PR #1, opened 2018-05-07
Benchmarks for SQLite write performance

scheiber-si:


Karsten1987:
I'll go ahead and merge as is. So we have it in the ros2 organization. 
Thanks @scheiber-si for putting in this effort.

scheiber-si:
It has been a team effort. Thanks to @nerfa-si, @lippmanm-si and @steinfus-si.

PR #12, opened 2018-07-30
initial command line interface

Karsten1987:
Provides entry points for record, play, info


Karsten1987:
CI (only on Linux, testing mainly for linters):
[Build Status](https://ci.ros2.org/job/ci_linux/4979/)

I would ask for early (style) feedback, because I see this as the base interface construct for the ros2 bag verb.

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
```
mikaelarguedas: Could this file just be empty and the dependency on rclpy removed from the package.xml ?

dirk-thomas:
Requesting feedback on the verbs and arguments is good. Until this patch contain any actual functionality I would suggest *not* to merge it yet. Simply because users building from source might be confused about the presence of the command but not doing anything.

wjwwood:
Well, they already added empty packages (see this as similar skeleton work). I don't think this repository is part of the default package list, so that mitigates it as well. I was trying to accommodate @Karsten1987's expressed desired to merge more frequent and smaller pr's (I think he mentioned that at some point).

@Karsten1987 you could add a disclaimer to all the commands for now, or merge into a working branch until you're ready for something on master.

PR #21, opened 2018-08-20
Adapt new interface

Martin-Idel-SI:
This PR adapts the demo to the new storage API.

This should now build and test cleanly on all supported platforms. The adaptation of the new interface is as lightweight as possible - in particular, we don't save the serialized bag message for now but only its string part. The true saving will be added with a later pull request.

Note:
I had to change the storage factory: Using template methods, it is not really a good idea if the implementation is not contained in the header file. It is not visible from the outside what kind of template specializations exist and indeed, this results in missing symbols on Windows. 
For now, I just exposed everything (i.e. we don't have a true pimpl here anymore), but we should reconsider the usage here. 
In fact, exposing the implementation, `librosbag2_storage` could be made a header only library for now. I introduced a dummy function which we will (in some way or other) need later on to avoid rewriting the `CMakeLists.txt`.

Also: It might be beneficial to introduce a visibility header for librosbag2. Otherwise the library cannot be linked against on Windows.


rosbag2_storage/src/impl/storage_factory_impl.hpp
```diff
@@ -66,7 +66,8 @@ get_interface_instance(
 
   std::shared_ptr<InterfaceT> instance = nullptr;
   try {
-    instance = class_loader->createSharedInstance(storage_id);
+    auto unmanaged_instance = class_loader->createUnmanagedInstance(storage_id);
```
Karsten1987: what's the difference here? Isn't this the same change in the end?

Issue #25, opened 2018-08-24
rosbag_write_integration_test is flaky


Martin-Idel-SI:
Currently, the rosbag_write_integration_test is flaky.

Valgrind and the debugger both suggest that cleanup of `shared_ptr<SerializedBagMessage>` does not work properly. However, the code never calls `reset` explicitly, nor is the memory in question released in any other way we can see.

Martin-Idel-SI:
We have heavily improved the tests. Sadly, the `rosbag2_write_integration_test` (and the second write `rosbag2_write_all_integration_test`) are still flaky:

Linux seems fine now, but we get the error `Expression: vector erase iterator outside range` following stack trace on Windows, which seems unrelated to our code:

```
 	msvcp140d.dll!00007ffecd72fc06()	Unknown
 	fastrtpsd-1.6.dll!std::vector<std::pair<eprosima::fastrtps::Publisher * __ptr64,eprosima::fastrtps::PublisherImpl * __ptr64>,std::allocator<std::pair<eprosima::fastrtps::Publisher * __ptr64,eprosima::fastrtps::PublisherImpl * __ptr64> > >::erase(std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<std::pair<eprosima::fastrtps::Publisher *,eprosima::fastrtps::PublisherImpl *> > > > _Where) Line 1475	C++
 	fastrtpsd-1.6.dll!eprosima::fastrtps::ParticipantImpl::removePublisher(eprosima::fastrtps::Publisher * pub) Line 84	C++
 	fastrtpsd-1.6.dll!eprosima::fastrtps::Domain::removePublisher(eprosima::fastrtps::Publisher * pub) Line 93	C++
 	rmw_fastrtps_shared_cpp.dll!rmw_fastrtps_shared_cpp::__rmw_destroy_publisher(const char * identifier, rmw_node_t * node, rmw_publisher_t * publisher) Line 67	C++
 	rmw_fastrtps_cpp.dll!rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher) Line 189	C++
 	rmw_implementation.dll!rmw_destroy_publisher(rmw_node_t * v2, rmw_publisher_t * v1) Line 245	C++
 	rcl.dll!rcl_publisher_fini(rcl_publisher_t * publisher, rcl_node_t * node) Line 210	C
 	rclcpp.dll!rclcpp::PublisherBase::~PublisherBase() Line 90	C++
[...]

PR #26, opened 2018-08-28
Allow an arbitrary topic to be recorded

Martin-Idel-SI:
This is based on https://github.com/ros2/rosbag2/pull/24

This PR adds the ability to record an arbitrary topic of an arbitrary ROS 2 type. To do this, the following changes are made:
- Extend the SQLite default storage plugin to contain a table of topics with type info
- Subscribe to and publish raw messages of arbitrary known topics, where the type info is retrieved from the corresponding type-info package (using the ideas presented in https://github.com/ros2/rosbag2/pull/20)
- To subscribe to messages without type info at compile time, we added a "GenericPublisher". Once this is possible with `rclcpp` this part can be deleted
- The `demo_record` takes the name of the topic to be recorded as a parameter. If the topic does not exist, rosbag record exits early.
- For now, the topic must be available at start up of `rosbag_record`. We have a small timeout to allow discovery of nodes and topics.

We make the following assumptions regarding types:
- When calling `node->get_topic_names_and_types()` all ROS 2 topics will be of the form `<package-name>/<message-type>` (e.g. `std_msgs/String`), while neither the package name nor the message type contain any further slashes.
- When a topic has multiple types, it is not a ROS 2 topic and won't be processed by rosbag

Known issues:
- The flaky test https://github.com/ros2/rosbag2/issues/25 is not resolved yet

### DIFF
--- a/rosbag2/CMakeLists.txt
+++ b/rosbag2/CMakeLists.txt
@@ -30,7 +30,7 @@ add_library(
   librosbag2
   SHARED
   src/rosbag2/rosbag2.cpp
-  src/rosbag2/demo_helpers.hpp)
+  src/rosbag2/typesupport_helpers.cpp)
 
 ament_target_dependencies(librosbag2 rclcpp rcutils std_msgs rcl Poco ament_index_cpp rosbag2_storage)
 target_include_directories(librosbag2
@@ -78,6 +78,15 @@ if(BUILD_TESTING)
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET rosbag2_read_integration_test)
     target_link_libraries(rosbag2_read_integration_test librosbag2)
+  endif()
+
+  ament_add_gmock(rosbag2_typesupport_helpers_test
+    test/rosbag2/rosbag2_typesupport_helpers_test.cpp
+    src/rosbag2/typesupport_helpers.cpp
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+  if(TARGET rosbag2_typesupport_helpers_test)
+    ament_target_dependencies(rosbag2_typesupport_helpers_test rcl Poco ament_index_cpp
+      rosidl_generator_c)
   endif()
 endif()
 
--- a/rosbag2/CMakeLists.txt
+++ b/rosbag2/CMakeLists.txt
@@ -70,6 +70,7 @@ if(BUILD_TESTING)
   ament_add_gmock(rosbag2_write_integration_test
     test/rosbag2/rosbag2_write_integration_test.cpp
     test/rosbag2/rosbag2_test_fixture.hpp
+    test/rosbag2/test_helpers.hpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET rosbag2_write_integration_test)
     target_link_libraries(rosbag2_write_integration_test librosbag2)
@@ -78,6 +79,7 @@ if(BUILD_TESTING)
   ament_add_gmock(rosbag2_read_integration_test
     test/rosbag2/rosbag2_read_integration_test.cpp
     test/rosbag2/rosbag2_test_fixture.hpp
+    test/rosbag2/test_helpers.hpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET rosbag2_read_integration_test)
     target_link_libraries(rosbag2_read_integration_test librosbag2)
@@ -94,6 +96,7 @@ if(BUILD_TESTING)
 
   ament_add_gmock(rosbag2_integration_test
     test/rosbag2/rosbag2_integration_test.cpp
+    test/rosbag2/test_helpers.hpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET rosbag2_integration_test)
     target_link_libraries(rosbag2_integration_test librosbag2)
@@ -105,6 +108,7 @@ if(BUILD_TESTING)
     src/rosbag2/raw_subscription.cpp
     src/rosbag2/raw_publisher.cpp
     src/rosbag2/rosbag2_node.cpp
+    test/rosbag2/test_helpers.hpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET rosbag2_rosbag_node_test)
     ament_target_dependencies(rosbag2_rosbag_node_test rclcpp Poco ament_index_cpp std_msgs)
--- a/rosbag2/CMakeLists.txt
+++ b/rosbag2/CMakeLists.txt
@@ -32,6 +32,7 @@ add_library(
   src/rosbag2/rosbag2.cpp
   src/rosbag2/typesupport_helpers.cpp
   src/rosbag2/raw_publisher.cpp
+  src/rosbag2/raw_subscription.cpp
   src/rosbag2/rosbag2_node.cpp)
 
 ament_target_dependencies(librosbag2 rclcpp rcutils std_msgs rcl Poco ament_index_cpp rosbag2_storage)
--- a/rosbag2/CMakeLists.txt
+++ b/rosbag2/CMakeLists.txt
@@ -30,7 +30,9 @@ add_library(
   librosbag2
   SHARED
   src/rosbag2/rosbag2.cpp
-  src/rosbag2/typesupport_helpers.cpp)
+  src/rosbag2/typesupport_helpers.cpp
+  src/rosbag2/rosbag2_publisher.cpp
+  src/rosbag2/rosbag2_node.cpp)
 
 ament_target_dependencies(librosbag2 rclcpp rcutils std_msgs rcl Poco ament_index_cpp rosbag2_storage)
 target_include_directories(librosbag2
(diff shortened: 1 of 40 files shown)


rosbag2/test/rosbag2/test_helpers.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+#include "../../src/rosbag2/typesupport_helpers.hpp"
+
+namespace test_helpers
+{
+
+std::shared_ptr<rmw_serialized_message_t> serialize_string_message(std::string message)
```
Karsten1987: why not `const std::string & message`?

rosbag2_storage/src/impl/storage_factory_impl.hpp
```diff
@@ -54,13 +54,17 @@ std::shared_ptr<InterfaceT>
...
+  const std::string & uri,
+  bool last_attempt)
 {
   const auto & registered_classes = class_loader->getDeclaredClasses();
   auto class_exists = std::find(registered_classes.begin(), registered_classes.end(), storage_id);
   if (class_exists == registered_classes.end()) {
-    RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME,
-      "Requested storage id %s does not exist", storage_id.c_str());
+    // This prevents unnecessary error messages when loading a ReadWrite storage as ReadOnly storage
+    if (last_attempt) {
```
Karsten1987: what's the exact purpose of this? When is decided that this is a last attempt? Can you elaborate on this?

rosbag2_storage/test/test_plugin.hpp
```diff
@@ -32,13 +33,15 @@ class TestPlugin : public rosbag2_storage::storage_interfaces::ReadWriteInterfac
 
   rosbag2_storage::BagInfo info() override;
 
-  bool has_next() const override;
+  void create_topic(const std::string & name, const std::string & type_id) override;
```
Karsten1987: `topic_id` here is the topic type? Or which id is meant here? If so, I would recommend having it renamed to `type_type` in order to keep the vocabulary in sync with upstream

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -40,39 +44,14 @@ SqliteWrapper::~SqliteWrapper()
   sqlite3_close(db_ptr);
 }
 
-void SqliteWrapper::execute_query(const std::string & query)
+SqliteStatement SqliteWrapper::prepare_statement(std::string query)
```
Karsten1987: why the change from `const std::string &` ?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper_integration_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include <gmock/gmock.h>
+
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "rcutils/types.h"
+
+#include "../../../src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.hpp"
```
Karsten1987: I know it's a test, but given that long include path, is there any use to having this wrapper file public?

Martin-Idel-SI: It will be made public in #31, because at that point we need it in other tests as well. But until then, I think it's cleaner to do it like this.

rosbag2_storage/src/impl/storage_factory_impl.hpp
Martin-Idel-SI: Sure: Before this change, whenever you used `rosbag_play`, you'd get an error message "Requested storage id sqlite3 does not exist" because there is no ReadOnlyStorage `sqlite3` from pluginlib's perspective. Obviously, since there is a ReadWriteStorage `sqlite3`, we still open the storage and everything works correctly, but the error message is highly irritating.

rosbag2/src/rosbag2/rosbag2.cpp
```diff
@@ -27,59 +29,136 @@
...
 namespace rosbag2
 {
 
 const char * ROS_PACKAGE_NAME = "rosbag2";
 
+std::string Rosbag2::get_topic_type(
+  const std::string & topic_name, const std::shared_ptr<rclcpp::Node> & node)
+{
+  // TODO(Martin-Idel-SI): This is a short sleep to allow the node some time to discover the topic
+  // This should be replaced by an auto-discovery system in the future
```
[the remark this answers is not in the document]
Martin-Idel-SI: Here are my problems with this approach: 
- How do you plan to support subscribing to new topics which are not available at startup time? If I say `rosbag2 record -all` I would eventually expect also new topics that are created after startup to be recorded. In that case, this will have to be done in our C++ loops, so I don't think it's a good idea to do this in the python code.
- I thought that the python code would really only be a python wrapper. But if you get topics and types in python, then you'd have further functionality there. I think it would be cleaner to have all functionality in C++

rosbag2/src/rosbag2/rosbag2.cpp
```diff
@@ -27,59 +29,136 @@
...
-    auto subscription = node->create_subscription<std_msgs::msg::String>(
+    RCUTILS_LOG_INFO_NAMED(ROS_PACKAGE_NAME, "Waiting for messages...");
+
+    auto node = std::make_shared<Rosbag2Node>("rosbag2");
+
+    std::string type = get_topic_type(topic_name, node);
+
+    if (type.empty()) {
+      RCUTILS_LOG_INFO_NAMED(ROS_PACKAGE_NAME, "Topic could not be found. Abort.");
+      return;
```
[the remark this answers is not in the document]
anhosi: +1 for the `throw`

rosbag2_storage/test/test_plugin.hpp
Martin-Idel-SI: Changed to `topic_name` and `type`

rosbag2/src/rosbag2/rosbag2.cpp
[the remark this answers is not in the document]
Martin-Idel-SI: Done

rosbag2/test/rosbag2/test_helpers.hpp
Martin-Idel-SI: Done

rosbag2_storage/test/test_plugin.hpp
Martin-Idel-SI: I haven't seen `type_type` anywhere but changed the parameters to `name` and `type`.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
Martin-Idel-SI: No change, this is just git messing up. But changed to const ref.

rosbag2/src/rosbag2/rosbag2.cpp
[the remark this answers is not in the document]
Karsten1987: I see your point. I believe we could have an overload for record (for a later time):
* `record(const std::string & topic_name, ...)`
* `record(const std::string & topic_name, const std::string & topic_type)`

The reason I am bringing up the python interface is, because it's easier to load the type information from python directly.

The thing with recording new topics is tricky though. This has to be somehow specified how often we look for new topics then. I believe in the current ROS 1 implementation, only the topics available at startup time are recorded. Otherwise I would expect asking for all available topics after every `spin`.

rosbag2/src/rosbag2/rosbag2_node.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <memory>
+#include <string>
+
+#include "typesupport_helpers.hpp"
+
+namespace rosbag2
+{
+
+Rosbag2Node::Rosbag2Node(const std::string & node_name)
```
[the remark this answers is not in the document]
Martin-Idel-SI: Please see the other comment about the generic subscription.

rosbag2/CMakeLists.txt
```diff
@@ -20,13 +20,22 @@ find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(rosbag2_storage REQUIRED)
+find_package(ament_index_cpp REQUIRED)
```
Karsten1987: nitpick: Alphabetical order

rosbag2/CMakeLists.txt
```diff
@@ -20,13 +20,22 @@ find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(rosbag2_storage REQUIRED)
```
Karsten1987: nitpick: Alphabetical order

rosbag2/test/rosbag2/test_helpers.hpp
```diff
@@ -0,0 +1,80 @@
...
+std::shared_ptr<rmw_serialized_message_t> serialize_string_message(const std::string & message)
+{
+  auto test_message = std::make_shared<std_msgs::msg::String>();
+  test_message->data = message;
+
+  auto rcutils_allocator = rcutils_get_default_allocator();
+  auto initial_capacity = 8u + static_cast<size_t>(test_message->data.size());
+  auto msg = new rmw_serialized_message_t;
+  *msg = rcutils_get_zero_initialized_char_array();
+  auto ret = rcutils_char_array_init(msg, initial_capacity, &rcutils_allocator);
```
Karsten1987: the allocator pointer will dangle here.

PR #30, opened 2018-08-31
Record all topics

Martin-Idel-SI:
This PR is based on #27 

The goal of this PR is to allow to subscribe to all available topics. In addition, there are a few technical changes:
- We publish serialized topics directly as shared pointers
- We introduced logging macros for simple usage and to make sure that the correct package is given. The macros internally use the rcutils logging macros, so nothing has really changed on the surface
- We improve and clarify why and how we filter topics to remove non-ROS topics that cannot be written


rosbag2/include/rosbag2/rosbag2.hpp
```diff
@@ -37,19 +37,31 @@ class Rosbag2Node;
 class Rosbag2
 {
 public:
+  /**
+   * Records topics to a bagfile. Subscription happens at startup time, hence the topics must
+   * exist when "record" is called.
+   *
+   * @param file_name Name of the bagfile to write
```
Karsten1987: I believe we are using the notion of `\param` for doxygen

rosbag2/src/rosbag2/generic_subscription.cpp
```diff
@@ -98,9 +100,8 @@ GenericSubscription::borrow_serialized_message(size_t capacity)
         auto fini_return = rmw_serialized_message_fini(msg);
         delete msg;
         if (fini_return != RCL_RET_OK) {
-          RCUTILS_LOG_ERROR_NAMED(
-            "rosbag2",
-            "failed to destroy serialized message: %s", rcl_get_error_string_safe());
+          ROSBAG2_LOG_ERROR_STREAM(
```
Karsten1987: there is still a problem with the allocator IMO.
So assume we create a serialized message using this subscription, keep it in memory and destroy the subscription before destroying the message.
The message now has an invalid allocator.

rosbag2/src/rosbag2/rosbag2_node.cpp
```diff
@@ -51,11 +56,83 @@ std::shared_ptr<GenericSubscription> Rosbag2Node::create_generic_subscription(
...
   }
 
   return subscription;
 }
+std::map<std::string, std::string> Rosbag2Node::get_topics_with_types(
+  const std::vector<std::string> & topic_names)
+{
+  std::vector<std::string> sanitized_topic_names;
+  for (const auto & topic_name : topic_names) {
+    sanitized_topic_names.push_back(topic_name[0] != '/' ? "/" + topic_name : topic_name);
```
Karsten1987: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/expand_topic_name.h

This function does exactly what you're looking for :)

rosbag2/src/rosbag2/rosbag2_node.cpp
```diff
@@ -51,11 +56,83 @@ std::shared_ptr<GenericSubscription> Rosbag2Node::create_generic_subscription(
...
+  std::map<std::string, std::vector<std::string>> filtered_topics_and_types;
+  for (const auto & topic_and_type : topics_and_types) {
+    if (std::find(sanitized_topic_names.begin(), sanitized_topic_names.end(),
+      topic_and_type.first) != sanitized_topic_names.end())
+    {
+      filtered_topics_and_types.insert(topic_and_type);
+    }
+  }
+
+  return reduce_multiple_types_to_one(filter_topics_with_wrong_types(filtered_topics_and_types));
```
Karsten1987: do you really need to do this? This copies the complete map. Maybe we could use the first element of the type just by convention.

rosbag2/src/rosbag2/rosbag2_node.cpp
```diff
@@ -51,11 +56,83 @@ std::shared_ptr<GenericSubscription> Rosbag2Node::create_generic_subscription(
...
+}
+
+std::map<std::string, std::string>
+Rosbag2Node::get_all_topics_with_types()
+{
+  // TODO(Martin-Idel-SI): This is a short sleep to allow the node some time to discover the topic
+  // This should be replaced by an auto-discovery system in the future
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  return reduce_multiple_types_to_one(
+    filter_topics_with_wrong_types(this->get_topic_names_and_types()));
```
Karsten1987: the topic types returned from here are correct.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
```diff
@@ -0,0 +1,136 @@
...
+  {
+    messages_stored_counter_.insert({topic_name, 0});
+    publisher_futures_.push_back(std::async(
+        std::launch::async, [this, message, topic_name, number_expected_messages]() {
+          auto publisher = publisher_node_->create_publisher<std_msgs::msg::String>(topic_name);
+
+          while (rclcpp::ok() && messages_stored_counter_[topic_name] < number_expected_messages) {
+            publisher->publish(message->serialized_data);
+            // We need to wait a bit, in order for record to process the messages
+            std::this_thread::sleep_for(50ms);
```
Karsten1987: isn't this potentially dangerous to assume? What if the publisher rate is higher than this?

rosbag2_storage/include/rosbag2_storage/logging.hpp
```diff
@@ -0,0 +1,61 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
```
Karsten1987: Isn't this a copy of the first file? Why is this necessary?

rosbag2_storage/src/impl/storage_factory_impl.hpp
```diff
@@ -69,17 +66,17 @@ get_interface_instance(
     auto unmanaged_instance = class_loader->createUnmanagedInstance(storage_id);
     instance = std::shared_ptr<InterfaceT>(unmanaged_instance);
   } catch (const std::runtime_error & ex) {
-    RCUTILS_LOG_ERROR_NAMED(ROS_PACKAGE_NAME,
-      "unable to load instance of read write interface: %s", ex.what());
+    ROSBAG2_STORAGE_LOG_ERROR_STREAM(
```
Karsten1987: I don't really see a reason for this.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/logging.hpp
```diff
@@ -0,0 +1,61 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
```
Karsten1987: that's the third copy of this file ?!

Martin-Idel-SI:
Regarding the logging: We basically copied that behaviour from rviz where we always had similar macros in `rviz_rendering`, `rviz_common` - they weren't invented by us. However, we felt it would be a nice addition as it solves two problems:

- The named macros need the package name. As far as I see, there is no nice way to make sure that the package name is uniform (there is no variable `ROS_PACKAGE_NAME` or the like).
- We don't have a stream variant of the macros. We like the stream variant because it simplifies working with the different variables (no `c_str()` or `std::to_string()` or similar patterns) and is easier to read than the usual variant.

At least the first problem can't be addressed in rcutils as far as I can see.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
Martin-Idel-SI: This is just for regular functionality testing (no performance or the like), so I think it's fine to assume. Or what would be your preferred rationale?

rosbag2/src/rosbag2/generic_subscription.cpp
Martin-Idel-SI: We had a close look and I believe it should be ok as is. The content of the allocator gets copied inside `rmw_serialized_message_init(...)` (see https://github.com/ros2/rcutils/blob/f04cc666dd06347050ff5c18b80079922e729d92/src/char_array.c#L50) so it should still be valid even after the pointer is destroyed.

We did a check by prematurely destroying the allocator (create `new rcutils_allocator_t`, use it to initialize the serialized message and then immediately destroy it) and it ran perfectly and valgrind was also happy.

rosbag2/include/rosbag2/rosbag2.hpp
Martin-Idel-SI: Changed.

rosbag2/src/rosbag2/rosbag2_node.cpp
Martin-Idel-SI: Since this is done at startup time, it's not really a big overhead. Nevertheless, if you think it's ok to just take the first argument by default, that's of course fine with me!

rosbag2/src/rosbag2/rosbag2_node.cpp
Martin-Idel-SI: Makes sense to use then. Thank you

rosbag2/src/rosbag2/rosbag2_node.cpp
Martin-Idel-SI: Ok, I deleted our own filtering code, then.

rosbag2/src/rosbag2/rosbag2_node.cpp
```diff
@@ -61,12 +62,69 @@ std::shared_ptr<GenericSubscription> Rosbag2Node::create_generic_subscription(
...
+           [](rcutils_string_map_t * map) {
+             rcl_ret_t cleanup = rcutils_string_map_fini(map);
+             delete map;
+             if (cleanup != RCL_RET_OK) {
+               ROSBAG2_LOG_ERROR("Failed to deallocate string map when expanding topic.");
+             }
+           });
+}
+
+std::string Rosbag2Node::expand_topic_name(std::string topic_name)
```
Karsten1987: should this function take a `const std::string &`

rosbag2/src/rosbag2/generic_subscription.cpp
Karsten1987: true. Please ignore my comment then.

rosbag2/src/rosbag2/rosbag2_node.cpp
Martin-Idel-SI: Absolut.

PR #31, opened 2018-08-31
Improve sqlite usage and test stability

anhosi:
This pr is based on #30.

Improved handling of sqlite:
- Open sqlite db according to `io_flag`, i.e. read-only or read-write. Only an existing db can be opened read-only.
- Use journal mode [WAL](https://www.sqlite.org/wal.html) (write ahead logging) which improves concurrency and performance.
- Use the [multi-thread threading mode](https://www.sqlite.org/threadsafe.html) that allows concurrent db access without relying on mutexes to protect most db operations. Db handles may not be shared across threads but each thread must use its own handle.
- Provides a convenient method for accessing the first row of a query result.

Other:
- Fixed threading issues in tests
- Allow direct linking against the sqlite storage plugin. This is needed by the integration tests so that assertions for the db content are possible. Also this enables reuse of the sqlite storage plugin in other plugins.


anhosi:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5123/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1906/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4249/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5106/)

anhosi:
New CI after fixing package dependencies:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5126/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1907/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4250/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5107/)

Martin-Idel-SI:
New CI after fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5132/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1913/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4260/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5114/)

Martin-Idel-SI:
And another CI after fixing the typesupport assertion (names are apparently different on aarch, so let's just assert rosidl substring):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1914/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4261/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5115/)

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
```diff
@@ -24,6 +24,8 @@
 #include "rclcpp/rclcpp.hpp"
 #include "rosbag2/rosbag2.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_storage_default_plugins/sqlite/sqlite_exception.hpp"
```
Karsten1987: I think this implies a strong dependency from `rosbag2` to `rosbag2_storage_default_plugins`. Ideally all these tests should perform equivally with all plugins.

We can introduce a new package for integration or system tests like these.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
```diff
@@ -91,20 +87,50 @@ class RosBag2IntegrationTestFixture : public Rosbag2TestFixture
...
+          while (rclcpp::ok() && count_stored_messages(db, topic_name) < number_expected_messages) {
+            publisher->publish(message->serialized_data.get());
+            // rate limiting
             std::this_thread::sleep_for(50ms);
           }
         }
     ));
   }
 
+  size_t count_stored_messages(rosbag2_storage_plugins::SqliteWrapper & db, std::string topic_name)
```
Karsten1987: `const std::string &`

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
```diff
@@ -91,20 +87,50 @@ class RosBag2IntegrationTestFixture : public Rosbag2TestFixture
...
+  {
+    // protect against concurrent writes (from recording) that may make the count query throw.
+    while (true) {
+      try {
+        return count_stored_messages_in_db(db, topic_name);
+      } catch (const rosbag2_storage_plugins::SqliteException & e) {
+        (void) e;
```
Karsten1987: shouldn't this break the while loop? Is there a way to have the loop conditioned otherwise than `while (true)`? This seems just dangerous.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
```diff
@@ -91,20 +87,50 @@ class RosBag2IntegrationTestFixture : public Rosbag2TestFixture
...
+            // rate limiting
             std::this_thread::sleep_for(50ms);
           }
         }
     ));
   }
 
+  size_t count_stored_messages(rosbag2_storage_plugins::SqliteWrapper & db, std::string topic_name)
+  {
+    // protect against concurrent writes (from recording) that may make the count query throw.
```
Karsten1987: if called concurrently, shouldn't this be locked then?

rosbag2_storage/package.xml
```diff
@@ -8,12 +8,18 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>pluginlib</depend>
-  <depend>rcutils</depend>
+  <build_depend>pluginlib</build_depend>
```
Karsten1987: `build_depend`, `exec_depend` and `test_depend` can be combined within `<depend>...</depend>`

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_exception.hpp
```diff
@@ -18,10 +18,17 @@
 #include <stdexcept>
 #include <string>
 
+#include "rosbag2_storage_default_plugins/visibility_control.hpp"
+
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4275)
```
Karsten1987: what's that warning? Can you maybe put a little comment on this to trace it later on?

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.hpp
```diff
@@ -25,12 +25,19 @@
 #include <vector>
 
 #include "rosbag2_storage/serialized_bag_message.hpp"
-#include "sqlite_exception.hpp"
+#include "rosbag2_storage_default_plugins/sqlite/sqlite_exception.hpp"
+#include "rosbag2_storage_default_plugins/visibility_control.hpp"
+
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
```
Karsten1987: same as above. Comment could be helpful here.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.hpp
```diff
@@ -25,12 +25,19 @@
...
 
 namespace rosbag2_storage_plugins
 {
 
-class SqliteStatementWrapper : public std::enable_shared_from_this<SqliteStatementWrapper>
+class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStatementWrapper
+  : public std::enable_shared_from_this<SqliteStatementWrapper>
 {
 public:
   SqliteStatementWrapper(sqlite3 * database, std::string query);
```
Karsten1987: const ref?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
```diff
@@ -51,8 +51,9 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
 std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset()
 {
   int return_code = sqlite3_step(statement_);
-  if (return_code != SQLITE_OK && return_code != SQLITE_DONE) {
-    throw SqliteException("Error processing SQLite statement.");
+  if (return_code != SQLITE_OK && return_code != SQLITE_DONE && return_code != SQLITE_ROW) {
```
Karsten1987: nitpick: Maybe for readability, these checks can be excluded in a little inline function `is_valid` or something. So ease up the condition.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -25,7 +25,8 @@
 
 #include "rosbag2_storage/serialized_bag_message.hpp"
 #include "../logging.hpp"
-#include "sqlite_statement_wrapper.hpp"
+#include "rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.hpp"
```
Karsten1987: nit: alphabetical order with `../logging.hpp`

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
anhosi: The test runner already provides a timeout so the `while (true)` while never run forever.
If a query fails to lock the sqlite db the wrapper throws (see also comment below). This loop simply retries the query until it succeeds (or the test wrapper kills and fails the test). It could be rewritten but this would only hide the intention of looping until the query succeeds.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
anhosi: We decided no for the following reason: the `rosbag2_storage_plugins::SqliteWrapper` is not designed to handle db locking more gracefully than throwing an exception as for the "record" and "play" usecases there will be no concurrent usage of the `SqliteWrapper`. Even for a later "convert" usage concurrent usage of the same db is (almost certain) not needed. So the concurrent use is only needed in tests the handling of db locks is put in the test fixture.
Retry is the most common strategy for dealing with failures of obtaining a db lock. Every other strategy would yield much more complicated code and the test runners timeout protects against infinite retrying.

rosbag2/test/rosbag2/rosbag2_write_integration_fixture.hpp
anhosi: In order to write integration tests for `rosbag2` we need a storage plugin. Which storage plugin is used for the tests is irrelevant and hidden inside the test fixtures.
Having `rosbag2_storage_default_plugins` as test dependency for `rosbag2` or having a separate test package depending on `rosbag2_storage_default_plugins` and `rosbag2` is for me practically equivalent. From a developers point of view I would prefer the test dependency as then the tests reside in the same package as the code they test.

When the `rosbag2_transport` package is introduced the tests will move there. If we choose to move the tests into a package of their own I suggest to do so after/ with the introduction of `rosbag2_transport`.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
anhosi: I agree.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
```diff
@@ -51,13 +51,18 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
 std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset()
 {
   int return_code = sqlite3_step(statement_);
-  if (return_code != SQLITE_OK && return_code != SQLITE_DONE && return_code != SQLITE_ROW) {
+  if (!isQueryOk(return_code)) {
```
Karsten1987: nitpick: ros2 convention is lower case and underscore

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
```diff
@@ -51,13 +51,18 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
...
     throw SqliteException("Error processing SQLite statement. Return code: " +
             std::to_string(return_code));
   }
   return reset();
 }
 
+bool SqliteStatementWrapper::isQueryOk(int return_code)
```
Karsten1987: could be inline or set in a lamdba

PR #32, opened 2018-09-10
Add correct timing behaviour for rosbag play

Martin-Idel-SI:
This PR adds correct timing behaviour for rosbag play

- Uses a shared queue (introduces vendor package for shared queues - header only library!)
- Use one thread to get messages from storage
- Use one thread to publish according to timestamp in storage
- Use a simple heuristic to make sure that the storage will not get starved and doesn't write too many messages to memory

In addition, a few technical aspects have been improved

- Only depend on `test_msgs` (not `std_msgs`) in rosbag2 tests
- Fix pluginlib classloader lifetime for rosbag play


anhosi:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5199/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1965/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5179/)

rosbag2/src/rosbag2/rosbag2.cpp
```diff
@@ -111,34 +109,15 @@ Rosbag2::create_subscription(
   return subscription;
 }
 
-void Rosbag2::play(const std::string & file_name)
+void Rosbag2::play(const std::string & file_name, Rosbag2PlayOptions options)
```
Karsten1987: const ref?

rosbag2/test/rosbag2/rosbag2_rosbag_node_test.cpp
```diff
@@ -62,8 +62,8 @@ class RosBag2NodeFixture : public Test
     auto subscription = node_->create_generic_subscription(topic_name, type,
         [this, &counter, &messages](std::shared_ptr<rmw_serialized_message_t> message) {
           auto string_message = memory_management_
-          .deserialize_message<std_msgs::msg::String>(message);
-          messages.push_back(string_message->data);
+          .deserialize_message<test_msgs::msg::Primitives>(message);
```
Karsten1987: that format looks kind of strange

rosbag2/test/rosbag2/rosbag2_test_fixture.hpp
```diff
@@ -114,11 +117,11 @@ class Rosbag2TestFixture : public Test
   void write_messages(
     const std::string & db_name,
     std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> messages,
-    std::map<std::string, std::string> topic_types)
+    std::map<std::string, std::string> topics_and_types)
```
Karsten1987: const ref? Same for the vector above

rosbag2/test/rosbag2/rosbag2_test_fixture.hpp
```diff
@@ -114,11 +117,11 @@ class Rosbag2TestFixture : public Test
...
   {
     rosbag2_storage::StorageFactory factory;
     auto storage = factory.open_read_write(db_name, "sqlite3");
-    for (auto topic_and_type : topic_types) {
+    for (auto topic_and_type : topics_and_types) {
```
Karsten1987: `for (const auto & )` ?

Karsten1987:
The tests seem to (at least) flaky on my OSX. 
I've run the gmock suite 10 times and `rosbag2_record_all_integration_test` constantly fails on my machine.

rosbag2/src/rosbag2/player.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void Player::wait_for_filled_queue(
+  const Rosbag2PlayOptions & options, const std::future<void> & storage_loading_future) const
+{
+  while (
+    message_queue_.size_approx() < options.queue_buffer_length_ &&
+    is_pending(storage_loading_future))
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
Karsten1987: I could see that this parameter should be exposed or configurable. Could we pack this into the `Rosbag2PlayOptions` struct?

rosbag2/include/rosbag2/rosbag2_play_options.hpp
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+#ifndef ROSBAG2__ROSBAG2_PLAY_OPTIONS_HPP_
+#define ROSBAG2__ROSBAG2_PLAY_OPTIONS_HPP_
+
+#include <cstdint>
+
+namespace rosbag2
+{
+struct Rosbag2PlayOptions
```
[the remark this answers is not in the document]
anhosi: Agreed, however I would prefer to do this later with a followup pr as we do not yet have the CLI integration.

rosbag2/src/rosbag2/player.cpp
```diff
@@ -0,0 +1,134 @@
...
+namespace rosbag2
+{
+
+Player::Player(std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage)
+: storage_(storage), node_(std::make_shared<Rosbag2Node>("rosbag2_node"))
+{}
+
+bool is_pending(const std::future<void> & future)
+{
+  return !(future.valid() && future.wait_for(std::chrono::seconds(0)) == std::future_status::ready);
```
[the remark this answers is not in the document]
anhosi: It will return immediately.

rosbag2/src/rosbag2/player.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void Player::play_messages_from_queue(std::future<void> storage_loading_future)
+{
+  auto start_time = std::chrono::high_resolution_clock::now();
+
+  while (message_queue_.size_approx() != 0 || is_pending(storage_loading_future)) {
+    ReplayableMessage message;
+    if (message_queue_.try_dequeue(message)) {
+      std::this_thread::sleep_until(start_time + message.time_since_start);
```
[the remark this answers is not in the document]
anhosi: It would be a "very short sleep" and return immediately. I am not sure whether the `sleep_until(now)` increases the chance of halting the execution of this thread by the OS scheduler.
This scheduling code is simple and straight forward but might not be optimal at high message rates. If we want / need a more elaborate scheduling scheme I would prefer to discuss this in a followup issue.

rosbag2/src/rosbag2/demo_play.cpp
```diff
@@ -15,15 +15,18 @@
...
-
+#include "rosbag2/rosbag2_play_options.hpp"
 #include "rosbag2/rosbag2.hpp"
 
 int main(int argc, const char ** argv)
 {
   rclcpp::init(argc, argv);
 
+  auto options = rosbag2::Rosbag2PlayOptions();
+  options.queue_buffer_length_ = 1000;
```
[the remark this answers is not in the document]
anhosi: Yes, it is the queue size up to which the storage is preloaded for smooth replaying.

rosbag2/src/rosbag2/player.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  ReplayableMessage message;
+  if (storage_->has_next()) {
+    message.message = storage_->read_next();
+    message.time_since_start = std::chrono::nanoseconds(0);
+    time_first_message = TimePoint(std::chrono::nanoseconds(message.message->time_stamp));
+    message_queue_.enqueue(message);
+  }
+
+  auto queue_lower_boundary = options.queue_buffer_length_ - options.queue_buffer_length_ / 10;
```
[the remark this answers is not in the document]
anhosi: * About being part of `Rosbag2PlayOptions`: I am not sure. It would be an expert user option only and we definitely would need to provide a useful default, e.g. 90%.
* There is actually no cast necessary as `/` is integer division in C++ which is what we want here.

rosbag2/src/rosbag2/player.cpp
```diff
@@ -0,0 +1,134 @@
...
+bool is_pending(const std::future<void> & future)
+{
+  return !(future.valid() && future.wait_for(std::chrono::seconds(0)) == std::future_status::ready);
+}
+
+void Player::play(Rosbag2PlayOptions options)
+{
+  prepare_publishers();
+
+  auto storage_loading_future = std::async(std::launch::async,
```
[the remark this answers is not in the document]
anhosi: Not necessarily as it is only used in a single thread. But I agree that the code handling the `storage_loading_future` needs to be refactored.

rosbag2/src/rosbag2/player.cpp
anhosi: Definitely possible. I am not yet sure how useful this parameter will be. I suggest to discuss this here (#34).

rosbag2/src/rosbag2/replayable_message.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <chrono>
+#include <memory>
+
+#include "rosbag2_storage/serialized_bag_message.hpp"
+
+namespace rosbag2
+{
+
+struct ReplayableMessage
```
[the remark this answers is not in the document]
anhosi: I disagree on that one. The `SerializedBagMessage` represents a message with meta data that can be stored in a bag file.
The `ReplayableMessage` represents a `SerializedBagMessage` enriched with play-specific meta data that is needed for playing. These are different things and thus should have different types.

anhosi:
* We also see this flaky integration test. I created a issue (https://github.com/ros2/rosbag2/issues/35) because we are not sure whether this is a rosbag2 problem.

anhosi:
New CI after working on review comments:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5206/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1970/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4324/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5184/)

rosbag2/src/rosbag2/player.cpp
```diff
@@ -41,29 +41,33 @@ Player::Player(std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInte
...
 }
 
-void Player::wait_for_filled_queue(
-  const Rosbag2PlayOptions & options, const std::future<void> & storage_loading_future) const
+void Player::wait_for_filled_queue(const Rosbag2PlayOptions & options) const
 {
   while (
-    message_queue_.size_approx() < options.queue_buffer_length_ &&
-    is_pending(storage_loading_future))
+    message_queue_.size_approx() < options.read_ahead_queue_size &&
```
Karsten1987: does this queue have to be initialized in some way? Or is it guaranteed that `size_approx()` returns 0 by default initialization?

rosbag2/src/rosbag2/player.cpp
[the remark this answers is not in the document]
Karsten1987: Yeah, my point is that `start_time + message.time_since_start` could be smaller than `now`, right?
But from what I read, the `sleep_until` command doesn't block on a time point in the past. So this should be alright.

rosbag2/src/rosbag2/rosbag2.cpp
```diff
@@ -78,18 +76,29 @@ void Rosbag2::record(const std::string & file_name, const std::vector<std::strin
...
+    rclcpp::spin(node_);
   }
   subscriptions_.clear();
 }
 
+
+void Rosbag2::record(const std::string & file_name)
+{
+  auto topics_and_types = node_->get_all_topics_with_types();
+  std::vector<std::string> topic_names(topics_and_types.size());
```
Karsten1987: Should this function throw when no topics are available?

rosbag2/src/rosbag2/rosbag2.cpp
```diff
@@ -48,12 +51,7 @@ void Rosbag2::record(const std::string & file_name, const std::vector<std::strin
...
   }
 
-  auto node = std::make_shared<Rosbag2Node>("rosbag2");
-
-  auto topics_and_types = topic_names.empty() ?
-    node->get_all_topics_with_types() :
-    node->get_topics_with_types(topic_names);
-
+  auto topics_and_types = node_->get_topics_with_types(topic_names);
   if (topics_and_types.empty()) {
```
Karsten1987: I think this check should go into the other record function

anhosi: As long as we do not have topic discovery after startup we should also report if we could not detect the topics that were explicitely specified. So I think is is an error for both `record` overloads and the location is fine.

anhosi:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5212/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1975/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4333/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5191/)

Karsten1987:
new round of CI:

[Build Status](https://ci.ros2.org/job/ci_linux/5217/)
[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1979/)
[Build Status](https://ci.ros2.org/job/ci_osx/4337/)
[Build Status](https://ci.ros2.org/job/ci_windows/5191/)

Karsten1987:
I'll go ahead and merge this. The flaky test is already ticketed in https://github.com/ros2/rosbag2/issues/37

PR #33, opened 2018-09-17
Improve sqlite iterator interface

anhosi:
The `SqliteStatementWrapper::QueryResult::Iterator` represents the result set of a sqlite query that needs to be stepped. Using a copy of such a iterator would also step the statement and looping the original iterator would "miss" result rows. To avoid this surprising behavior this PR changes the iterator interface to disable copies.

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
@@ -51,14 +51,14 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
 std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset()
 {
   int return_code = sqlite3_step(statement_);
-  if (!isQueryOk(return_code)) {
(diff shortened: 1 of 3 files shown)


anhosi:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5198/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1964/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4318/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5178/)

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
```diff
@@ -51,14 +51,14 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
 std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset()
 {
   int return_code = sqlite3_step(statement_);
-  if (!isQueryOk(return_code)) {
+  if (!is_query_ok(return_code)) {
```
Karsten1987: just as a remark. I know it's only single threaded for now, but I believe operations like that can benefit from atomic compare_exchange.

PR #42, opened 2018-10-04
Display bag summary using `ros2 bag info`

greimela-si:
This PR adds a new ros2 bag CLI verb `info` which prints a summary of a bag content.

The printout looks like this:
```
$ ros2 bag info 2018_10_04-11_08_21

Files:            2018_10_04-11_08_21.db3
Bag size:         44.4 KiB
Storage id:       sqlite3
Storage format:   cdr
Duration:         21.3s
Start:            Oct  4 2018 11:08:21.483 (1538644101.483)
End               Oct  4 2018 11:08:42.486 (1538644122.486)
Messages:         22
Topics with Type: /string_topic; std_msgs/String; 22 msgs
```

These infos are stored in a metadata file at the end of the record process to allow getting infos about a bag without inspecting its content.
This is necessary since the user might not have access to the storage backend used to record the bagfile.

Some minor changes have been added in this PR as well:
- Store rosbags using a folder, not a file. 
  This folder contains the `metadata.yaml` file as well as all files created by the storage backend.
- Add record options `--output` to state the output file and `--storage` to select the storage backend.
  Default: Use the current timestamp as bag name and sqlite3 as storage backend
- Stabilize integration and e2e tests.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5250/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2005/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4364/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5223/)


ros2bag/ros2bag/verb/info.py
```diff
@@ -12,18 +12,24 @@
...
     def add_arguments(self, parser, cli_name):  # noqa: D102
-        arg = parser.add_argument(
+        parser.add_argument(
             'bag_file', help='bag file to introspect')
 
     def main(self, *, args):  # noqa: D102
         bag_file = args.bag_file
-        print('calling ros2 bag info on', bag_file)
+        if not os.path.exists(bag_file):
+            sys.exit("Error: bag file '{}' does not exist!".format(bag_file))
```
Karsten1987: I don't think `sys.exit` is the right thing to do here. I think a simple return string will suffice and `ros2cli` will handle it. 
see here: https://github.com/ros2/ros2cli/blob/master/ros2msg/ros2msg/verb/package.py#L30-L33

rosbag2/include/rosbag2/rosbag2_factory.hpp
```diff
@@ -0,0 +1,56 @@
...
+   */
+  virtual std::shared_ptr<Writer> create_writer(const StorageOptions & options);
+
+  /**
+   * Open a bagfile to read all messages sequentially according to their timestamp.
+   *
+   * @param options Storage options including uri and storage identifier
+   * @return
+   */
+  virtual std::shared_ptr<SequentialReader> create_sequential_reader(
```
Karsten1987: I would like to propose this function to be templated. I still feel that the inconsistency between `writer` and `sequential_reader` disturbs me (I still wonder why there is no `sequential_writer` then). So the functions `create_writer` and `create_reader` could be templated with defaults to `Writer` and `SequentialReader`. 
Or I think better is to have `Writer` and `Reader` being the base class/interface and `SequentialReader` is inheriting from this. But I feel as a user I don't need to know which reader I am actually having.

rosbag2/src/rosbag2/info.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "rosbag2_storage/metadata_io.hpp"
+
+namespace rosbag2
+{
+
+Info::Info(std::shared_ptr<rosbag2_storage::Rosbag2StorageFactory> storage_factory)
```
Karsten1987: Shouldn't this be inverse? That the StorageFactory can be asked to get the info and not Info asking Storage?

rosbag2/src/rosbag2/rosbag2_factory.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+std::shared_ptr<SequentialReader> Rosbag2Factory::create_sequential_reader(
+  const StorageOptions & options)
+{
+  return std::make_shared<SequentialReaderImpl>(options);
+}
+
+std::shared_ptr<Info> Rosbag2Factory::create_info()
+{
+  return std::make_shared<Info>();
```
Karsten1987: Shouldn't info then get a StorageFactory?

rosbag2/src/rosbag2/writer_impl.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "rosbag2_storage/rosbag2_storage_factory.hpp"
+#include "rosbag2_storage/rosbag2_storage_factory_impl.hpp"
+#include "rosbag2/storage_options.hpp"
+#include "rosbag2/types.hpp"
+#include "rosbag2/writer.hpp"
+
+namespace rosbag2
+{
+
+class WriterImpl : public Writer
```
Karsten1987: Why does a writer need the pimpl idiom here? What's there to abstract/hide?

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
```diff
@@ -12,25 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
-#define ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
+#ifndef ROSBAG2_STORAGE__METADATA_IO_HPP_
```
Karsten1987: why that change? What's happening here? Why was it necessary to rename that file?

rosbag2_storage/include/rosbag2_storage/rosbag2_storage_factory.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "rosbag2_storage/metadata_io.hpp"
+#include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
+#include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
+#include "rosbag2_storage/visibility_control.hpp"
+
+namespace rosbag2_storage
+{
+
+/// Factory to create instances of various storage interfaces
+class ROSBAG2_STORAGE_PUBLIC Rosbag2StorageFactory
```
Karsten1987: why do I need this when we have the `rosbag2_storage/storage_factory.hpp` ?

ros2bag/ros2bag/verb/info.py
greimela-si: Done, thanks

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
greimela-si: This file has not been renamed.
That's just git seing things that are not there...

greimela-si:
This PR has been splitted into #43, #44 and #45.

rosbag2/include/rosbag2/rosbag2_factory.hpp
greimela-si: We removed the factories and the pimpls in the smaller PR for now, since they have to be reworked after adding the C interface anyway.

rosbag2/src/rosbag2/info.cpp
greimela-si: See https://github.com/ros2/rosbag2/pull/42#discussion_r224424391

rosbag2/src/rosbag2/writer_impl.hpp
greimela-si: See https://github.com/ros2/rosbag2/pull/42#discussion_r224424391

rosbag2_storage/include/rosbag2_storage/rosbag2_storage_factory.hpp
greimela-si: See https://github.com/ros2/rosbag2/pull/42#discussion_r224424391

rosbag2/src/rosbag2/rosbag2_factory.cpp
greimela-si: See https://github.com/ros2/rosbag2/pull/42#discussion_r224424391

PR #45, opened 2018-10-11
 Display bag summary using `ros2 bag info` (reduced size)

greimela-si:
This PR adds a new ros2 bag CLI verb `info` which prints a summary of a bag content.

The printout looks like this:
```
$ ros2 bag info 2018_10_04-11_08_21

Files:            2018_10_04-11_08_21.db3
Bag size:         44.4 KiB
Storage id:       sqlite3
Storage format:   cdr
Duration:         21.3s
Start:            Oct  4 2018 11:08:21.483 (1538644101.483)
End               Oct  4 2018 11:08:42.486 (1538644122.486)
Messages:         22
Topics with Type: /string_topic; std_msgs/String; 22 msgs
```

These infos are stored in a metadata file at the end of the record process to allow getting infos about a bag without inspecting its content.
This is necessary since the user might not have access to the storage backend used to record the bagfile.
As a result, `rviz_yaml_cpp_vendor` is required as a dependency.

This PR is based on the changes introduced in #43 and contains these changes.
Therefore, this PR should be merged after #43.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2039/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5264/)

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -43,6 +43,12 @@ class RecordVerb(VerbExtension):
             '-s', '--storage', default='sqlite3',
             help='storage identifier to be used, defaults to "sqlite3"')
 
+    def create_bag_directory(self, uri):
+        try:
+            os.makedirs(uri)
+        except:
+            return "Error: Could not create bag folder '{}'.".format(uri)
+
     def main(self, *, args):  # noqa: D102
         if args.all and args.topics:
             return 'Invalid choice: Can not specify topics and -a at the same time.'
@@ -52,10 +58,7 @@ class RecordVerb(VerbExtension):
         if os.path.isdir(uri):
             return "Error: Output folder '{}' already exists.".format(uri)
 
-        try:
-            os.makedirs(uri)
-        except:
-            return "Error: Could not create bag folder '{}'.".format(uri)
+        self.create_bag_directory(uri)
 
         if args.all:
             rosbag2_transport_py.record(uri=uri, storage_id=args.storage, all=True)
@@ -63,3 +66,6 @@ class RecordVerb(VerbExtension):
             rosbag2_transport_py.record(uri=uri, storage_id=args.storage, topics=args.topics)
         else:
             self._subparser.print_help()
+
+        if os.path.isdir(uri) and not os.listdir(uri):
+            os.rmdir(uri)
--- a/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
@@ -37,7 +37,7 @@ struct BagMetadata
   int version = 1;  // upgrade this number when changing the content of the struct
   size_t bag_size = 0;  // Will not be serialized
   std::string storage_identifier;
-  std::string storage_format;
+  std::string serialization_format;
   std::vector<std::string> relative_file_paths;
   std::chrono::nanoseconds duration;
   std::chrono::time_point<std::chrono::high_resolution_clock> starting_time;
--- a/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
@@ -1,0 +1,50 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2_STORAGE__BAG_METADATA_HPP_
+#define ROSBAG2_STORAGE__BAG_METADATA_HPP_
+
+#include <chrono>
+#include <map>
+#include <string>
+#include <vector>
+#include <utility>
+
+#include "rosbag2_storage/topic_with_type.hpp"
+
+namespace rosbag2_storage
+{
+
+struct TopicMetadata
+{
+  TopicWithType topic_with_type;
+  size_t message_count;
+};
+
+struct BagMetadata
+{
+  int version = 1;
+  size_t bag_size = 0;  // Will not be serialized
+  std::string storage_identifier;
+  std::string storage_format;
+  std::vector<std::string> relative_file_paths;
+  std::chrono::nanoseconds duration;
+  std::chrono::time_point<std::chrono::high_resolution_clock> starting_time;
+  size_t message_count;
+  std::vector<TopicMetadata> topics_with_message_count;
+};
+
+}  // namespace rosbag2_storage
+
+#endif  // ROSBAG2_STORAGE__BAG_METADATA_HPP_
--- a/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
@@ -34,7 +34,7 @@ struct TopicMetadata
 
 struct BagMetadata
 {
-  int version = 1;
+  int version = 1;  // upgrade this number when changing the content of the struct
   size_t bag_size = 0;  // Will not be serialized
   std::string storage_identifier;
   std::string storage_format;
--- a/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
@@ -29,8 +29,10 @@ class MetadataIo
 public:
   static constexpr const char * const metadata_filename = "metadata.yaml";
 
-  ROSBAG2_STORAGE_PUBLIC void write_metadata(const std::string & uri, BagMetadata metadata);
-  ROSBAG2_STORAGE_PUBLIC BagMetadata read_metadata(const std::string & uri);
+  ROSBAG2_STORAGE_PUBLIC
+  void write_metadata(const std::string & uri, const BagMetadata & metadata);
+  ROSBAG2_STORAGE_PUBLIC
+  BagMetadata read_metadata(const std::string & uri);
 
 private:
   std::string get_metadata_file_name(const std::string & uri);
--- a/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
@@ -1,4 +1,4 @@
-// Copyright 2018 Open Source Robotics Foundation, Inc.
+// Copyright 2018, Bosch Software Innovations GmbH.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,25 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
-#define ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
+#ifndef ROSBAG2_STORAGE__METADATA_IO_HPP_
+#define ROSBAG2_STORAGE__METADATA_IO_HPP_
 
-#include "rosbag2_storage/bag_info.hpp"
+#include <string>
+
+#include "rosbag2_storage/bag_metadata.hpp"
+#include "rosbag2_storage/topic_with_type.hpp"
 #include "rosbag2_storage/visibility_control.hpp"
 
 namespace rosbag2_storage
 {
-namespace storage_interfaces
-{
 
-class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
+class MetadataIo
 {
 public:
-  virtual ~BaseInfoInterface() = default;
-  virtual BagInfo info() = 0;
+  static constexpr const char * const metadata_filename = "metadata.yaml";
+
+  ROSBAG2_STORAGE_PUBLIC void write_metadata(const std::string & uri, BagMetadata metadata);
+  ROSBAG2_STORAGE_PUBLIC BagMetadata read_metadata(const std::string & uri);
+
+private:
+  std::string get_metadata_file_name(const std::string & uri);
 };
 
-}  // namespace storage_interfaces
 }  // namespace rosbag2_storage
 
-#endif  // ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
+#endif  // ROSBAG2_STORAGE__METADATA_IO_HPP_
--- a/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
@@ -18,6 +18,7 @@
 #include <string>
 
 #include "rosbag2_storage/bag_metadata.hpp"
+#include "rosbag2_storage/filesystem_helper.hpp"
 #include "rosbag2_storage/topic_with_type.hpp"
 #include "rosbag2_storage/visibility_control.hpp"
 
@@ -31,8 +32,12 @@ public:
 
   ROSBAG2_STORAGE_PUBLIC
   void write_metadata(const std::string & uri, const BagMetadata & metadata);
+
   ROSBAG2_STORAGE_PUBLIC
   BagMetadata read_metadata(const std::string & uri);
+
+  ROSBAG2_STORAGE_PUBLIC
+  bool metadata_file_exists(const std::string & uri);
 
 private:
   std::string get_metadata_file_name(const std::string & uri);
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
@@ -17,6 +17,7 @@
 
 #include <string>
 
+#include "rosbag2_storage/storage_interfaces/base_info_interface.hpp"
 #include "rosbag2_storage/storage_interfaces/base_io_interface.hpp"
 #include "rosbag2_storage/storage_interfaces/base_read_interface.hpp"
 #include "rosbag2_storage/visibility_control.hpp"
@@ -27,10 +28,11 @@ namespace storage_interfaces
 {
 
 class ROSBAG2_STORAGE_PUBLIC ReadOnlyInterface
-  : public BaseIOInterface, public BaseReadInterface
+  : public BaseInfoInterface, public BaseIOInterface, public BaseReadInterface
 {
 public:
   virtual ~ReadOnlyInterface() = default;
+
   void open(const std::string & uri, IOFlag io_flag = IOFlag::READ_ONLY) override = 0;
 };
 
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
@@ -17,7 +17,6 @@
 
 #include <string>
 
-#include "rosbag2_storage/storage_interfaces/base_info_interface.hpp"
 #include "rosbag2_storage/storage_interfaces/base_io_interface.hpp"
 #include "rosbag2_storage/storage_interfaces/base_read_interface.hpp"
 #include "rosbag2_storage/visibility_control.hpp"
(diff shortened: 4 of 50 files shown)


Martin-Idel-SI:
New CI after rebase and fixes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5294/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2048/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4412/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5275/)

Martin-Idel-SI:
New CI after rebase:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5300/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2054/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4417/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5280/)

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+struct TopicMetadata
+{
+  TopicWithType topic_with_type;
+  size_t message_count;
+};
+
+struct BagMetadata
+{
+  int version = 1;
```
Karsten1987: What is version 1 meaning here? How is the user supposed to change this version?

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
```diff
@@ -12,25 +12,30 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
-#define ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
+#ifndef ROSBAG2_STORAGE__METADATA_IO_HPP_
```
Karsten1987: Why this change? What's the problem with `base_info_interface`? Remember that these interfaces were designed to be composable for many specific interfaces.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
```diff
@@ -12,25 +12,30 @@
...
-namespace storage_interfaces
-{
 
-class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
+class MetadataIo
 {
 public:
-  virtual ~BaseInfoInterface() = default;
-  virtual BagInfo info() = 0;
+  static constexpr const char * const metadata_filename = "metadata.yaml";
```
Karsten1987: is the user ever able to change this name? What's happening if a user changes this file to another name?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return true;
+  }
+};
+
+}  // namespace YAML
+
+namespace rosbag2_storage
+{
+
+void MetadataIo::write_metadata(const std::string & uri, BagMetadata metadata)
```
Karsten1987:
```suggestion
void MetadataIo::write_metadata(const std::string & uri, const BagMetadata & metadata)
```

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -175,9 +185,15 @@ void SqliteStorage::fill_topics_and_types()
...
 
-std::string SqliteStorage::get_database_name(const std::string & uri)
+std::unique_ptr<rosbag2_storage::BagMetadata> SqliteStorage::load_metadata(const std::string & uri)
 {
-  return rosbag2_storage::FilesystemHelper::get_folder_name(uri) + ".db3";
+  try {
+    rosbag2_storage::MetadataIo metadata_io;
+    return std::make_unique<rosbag2_storage::BagMetadata>(metadata_io.read_metadata(uri));
+  } catch (std::exception & e) {
+    (void) e;
```
Karsten1987: Shouldn't there at least be some kind of logging?

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
```diff
@@ -28,7 +27,7 @@ namespace storage_interfaces
 {
 
 class ROSBAG2_STORAGE_PUBLIC ReadOnlyInterface
-  : public BaseInfoInterface, public BaseIOInterface, public BaseReadInterface
```
Karsten1987: I don't understand this change, or maybe just don't agree with it.

When I open a bag file read only, I shall still have the opportunity to read the meta data info, right? How would I do this with this current change?

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
Martin-Idel-SI: The version number is supposed to allow later changes to the metadata format (e.g. if we also want to save info about services, actions, etc at some later point). Having a version field simplifies handling old bagfiles then.
Therefore, the user must not change this version.

Martin-Idel-SI: The version number is supposed to allow later changes to the metadata format (e.g. if we also want to save info about services, actions, etc at some later point). Having a version field simplifies handling old bagfiles then.
Therefore, the user must not change this version.

Martin-Idel-SI: The version number is supposed to allow later changes to the metadata format (e.g. if we also want to save info about services, actions, etc at some later point). Having a version field simplifies handling old bagfiles then.
Therefore, the user must not change this version.

Martin-Idel-SI: The version number is supposed to allow later changes to the metadata format (e.g. if we also want to save info about services, actions, etc at some later point). Having a version field simplifies handling old bagfiles then.
Therefore, the user must not change this version.

Martin-Idel-SI: The version number is supposed to allow later changes to the metadata format (e.g. if we also want to save info about services, actions, etc at some later point). Having a version field simplifies handling old bagfiles then.
Therefore, the user must not change this version.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Martin-Idel-SI: At the moment, changing this name is not supported. As we discussed, this would require the user to also be able to give a path not just to the directory but to the metadata file itself. The existence of the metadata.yaml file is an implementation detail and i won't exist for old bagfiles, for instance.

Exposing this to the user by allowing to specify the path to the yaml file directly might (as discussed) be a feature, but I'd suggest we open a ticket for now.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
botteroa-si: Ok, good point.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Martin-Idel-SI: At the moment, changing this name is not supported. As we discussed, this would require the user to also be able to give a path not just to the directory but to the metadata file itself. The existence of the metadata.yaml file is an implementation detail and i won't exist for old bagfiles, for instance.

Exposing this to the user by allowing to specify the path to the yaml file directly might (as discussed) be a feature, but I'd suggest we open a ticket for now.

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
Martin-Idel-SI: This version field was introduced to allow easy transition to new formats if necessary (e.g. we want to add information about service calls, parameters, etc.). Having a version field then simplifies backwards compatibility with old yaml files.

Therefore, it should never be changed by the user.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Martin-Idel-SI: Not at the moment. As discussed, with the current changes the user enters a path to a directory containing this file. For simple use cases, this should be sufficient: Treat the directory as bagfile and hide implementation details (e.g. the metadata file, database, etc.) from the user. 

For more complicated use cases, if you want to modify the folder structure, you can already do that by also changing the content metadata file, but until it is possible to specify the path to a yaml file containing the information, it is impossible to rename the metadata file.

We suggest to open a new issue to allow the user to enter a path to the yaml file directly as this is a separate feature. At that point, it would be simple to support renaming the metadata file.

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
Martin-Idel-SI: This version field was introduced to allow easy transition to new formats if necessary (e.g. we want to add information about service calls, parameters, etc.). Having a version field then simplifies backwards compatibility with old yaml files.

Therefore, it should never be changed by the user.

Martin-Idel-SI: This version field was introduced to allow easy transition to new formats if necessary (e.g. we want to add information about service calls, parameters, etc.). Having a version field then simplifies backwards compatibility with old yaml files.

Therefore, it should never be changed by the user.

Martin-Idel-SI: This version field was introduced to allow easy transition to new formats if necessary (e.g. we want to add information about service calls, parameters, etc.). Having a version field then simplifies backwards compatibility with old yaml files.

Therefore, it should never be changed by the user.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Martin-Idel-SI: At the moment, it is only possible to hand the name of the directory to `ros2 bag record` (as discussed). For simple use-cases, I think this is sufficient: We think of the folder as bagfile and hide the implementation details (database, metadata yaml file) from the user.

For more complicated use-cases it is already possible to restructure the folder completely by adapting the yaml file. As long as it's impossible to hand the path to the yaml file directly to ros2 bag, it's impossible to rename the yaml file.

We suggest to open an issue to allow specifying the path to a yaml file instead of the path to a folder containing a yaml file. With this change, it is possible to support renaming the metadata yaml file.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
Martin-Idel-SI: Here is our rationale:

- Reading the metadata yaml file should always be possible, even without opening the storage (e.g. when the storage is not known in advance) - and indeed, the storage should not be opened if it is not necessary.
- The logic to read the yaml file is the same for every plugin hence a storage plugin author does not need to implement it
- Therefore, it seems logical to expose this functionality outside of the storage_interface hierarchy.
Note that the information we get from the storage plugin itself is necessarily incomplete - it must be enriched with additional information from the filesystem.

It is of course possible to expose a non-abstract method inside the storage interface which just calls the the outside method. I'm just not sure I see the benefit.

In other words, with the current implementation, you can open your storage read-only (or read-write) and you can (in a separate call) query the information from the metadata file without having a storage handle.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
botteroa-si: Good point.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
botteroa-si: You're right, we'll do so.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Martin-Idel-SI: See other comment referring to this change.

botteroa-si:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5306/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2058/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4421/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5284/)

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
Karsten1987: That assumption only holds when bag files are written with rosbag2. When, however, no yaml file is specified for let's say ros1 bags, then how do I get this type of info?

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
Karsten1987: where is this field currently used in the code? Can we extract that info somehow from incoming messages?

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
Karsten1987: https://github.com/ros2/rosbag2/issues/49

Karsten1987:
The problem with the folder still persists:

```
✔  11:45:06  ~/workspace/osrf/rosbag2_ws
 ➭ ros2 bag record -a -s my_plugin
DISCLAIMER
ros2 bag is currently under development and not ready to use yet
[ERROR] [rosbag2_storage]: Could not load/open plugin with storage id 'my_plugin'.
[ERROR] [rosbag2_transport]: Failed to record: No storage could be initialized. Abort
```

Even tho the plugin couldn't be found, and thus the execution is aborted, the folder gets created.

rosbag2_storage/include/rosbag2_storage/metadata_io.hpp
```diff
@@ -12,25 +12,32 @@
...
+
+  ROSBAG2_STORAGE_PUBLIC
+  void write_metadata(const std::string & uri, const BagMetadata & metadata);
```
Karsten1987:
```suggestion
  void write_metadata(const std::string & uri, const BagMetadata & metadata);
  
```

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
```diff
@@ -28,7 +27,7 @@ namespace storage_interfaces
...
+  : public BaseIOInterface, public BaseReadInterface
 {
 public:
   virtual ~ReadOnlyInterface() = default;
```
Karsten1987:
```suggestion
   virtual ~ReadOnlyInterface() = default;
  
```

Karsten1987: The same change should be done in `read_write_interface.hpp` and `base_io_interface.hpp`

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
Karsten1987: Also, from an interface point of view. If I get a `ReadWriteInterface`, I would like to be able to call `get_info()` or `get_metadata()` on that instance. I don't see how I would do that right now, because with this change I have to rely on the plugin that this function is available.

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
anhosi: It is not used yet and its sole purpose is to distinguish the yaml format in case of changes. It must be independent of the content so a md5 sum (or something similar) is not useful.
An example use would be the case of a format change (after the first release). For that case rosbag can distinguish the to formats and provide an automatic conversion (or at least read also the old format correctly).

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
Martin-Idel-SI: I see your first point. That's a part where we do indeed need a function to read the information from the storage directly (for read write, this is possible).

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
Martin-Idel-SI: I hopefully found all occurences

Martin-Idel-SI: The change above is outdated. The interface is back in, so the only line changing in that file is the addition of a newline.

Karsten1987:
```
 ➭ ros2 bag record -a
DISCLAIMER
ros2 bag is currently under development and not ready to use yet
[ERROR] [rosbag2_storage]: Failed to load metadata: Exception on parsing info file: bad file
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -52,14 +58,14 @@ def main(self, *, args):  # noqa: D102
...
+        self.create_bag_directory(uri)
 
         if args.all:
             rosbag2_transport_py.record(uri=uri, storage_id=args.storage, all=True)
         elif args.topics and len(args.topics) > 0:
             rosbag2_transport_py.record(uri=uri, storage_id=args.storage, topics=args.topics)
         else:
             self._subparser.print_help()
+
+        if os.path.isdir(uri) and not os.listdir(uri):
```
Karsten1987: It seems to be there is a design flaw here. I believe there should be a function (exposed) which can check whether the specified plugin is correct. That is, before calling straight up `record`, I think we should be able to initialize and setup the recorder at first and then call record. Having the recorder setup first allows for error checking.

Then second, the code for removing the folder only works right now because record has to be cancelled by `ctrl-c` (because of the spin inside). I believe this should be handled by a signint handler, to cancel the record deliberately.

anhosi: IMHO the "design flaw" is that the directory creation happens already in the python layer before instead of having the storage taking care of that. The reason for this is that cross platform file handling is hard to do.
Creating a separate `initialize` method would only solidify this workaround and would still leave some scenarios uncovered that cannot be detected beforehand.

anhosi: Regarding the second point I agree that we will need to do our own signal handling for that and then we also need to change this "failure detection". I would prefer to do this with another pull request.

rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
```diff
@@ -73,11 +76,20 @@ class Rosbag2Transport
   ROSBAG2_TRANSPORT_PUBLIC
   void play(const StorageOptions & storage_options, const PlayOptions & play_options);
 
+  /**
+   * Print the bag info contained in the metadata yaml file.
+   *
+   * \param uri path to the metadata yaml file.
+   */
+  ROSBAG2_TRANSPORT_PUBLIC
+  void print_bag_info(const std::string & uri);
```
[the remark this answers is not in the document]
anhosi: This will not be done separately -> #51.

Karsten1987:
has to be rebased.

Martin-Idel-SI:
Done. Add new CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2101/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4465/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5335/)

rosbag2_transport/src/rosbag2_transport/rosbag2_transport.cpp
```diff
@@ -99,30 +99,37 @@ void Rosbag2Transport::play(
...
-  info_stream << "Start:                " << formatter->format_time_point(start_time) << std::endl;
-  info_stream << "End                   " << formatter->format_time_point(end_time) << std::endl;
-  info_stream << "Messages:             " << metadata.message_count << std::endl;
-  info_stream << "Topics with Type:     ";
-  formatter->format_topics_with_type(
-    metadata.topics_with_message_count, info_stream, indentation_spaces);
-
-  std::cout << info_stream.str() << std::endl;
+  try {
+    auto metadata = info_->read_metadata(uri);
```
Karsten1987: wouldn't the exception come out here? I believe it makes more sense to catch the exception right after this method.

Martin-Idel-SI:
Done. New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5414/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2121/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4488/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5355/)

PR #48, opened 2018-10-18
Implement converter plugin for CDR format and add converter plugins package

botteroa-si:
This PR is based on #47.

This PR implements a converter plugin to convert ROS 2 messages to `CDR` format and back, using `fastrtps` as `rmw` implementation.
The plugin is contained in the new package `rosbag2_converter_default_plugins`.

This new package also contains helper methods to extract `rmw` functions needed by the converters. It might be a good idea to expose these helpers, to benefit plugins developers. In order to do that, though, we need visibility control to be added to the package.


Martin-Idel-SI:
Rebased + CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5385/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2102/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4466/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5336/)

botteroa-si:
@Karsten1987, we don't really understand why all the CI jobs failed. It seems that the problem is not related with our code, though. Maybe you have a better understanding of what went wrong (we also tried to restart the jobs, but the result was unchanged).

dirk-thomas:
Based on the job parameters and error message are you maybe trying to run tests for a package which hasn't been built?

botteroa-si:
@dirk-thomas, thank you very much. You're right, the parameters we gave are in the case of this branch wrong, we didn't think about that.

botteroa-si:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5389/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2106/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5341/)

rosbag2_converter_default_plugins/plugin_description.xml
```diff
@@ -0,0 +1,9 @@
+<library path="rosbag2_converter_default_plugins">
+<class
+        name="cdr_converter"
+        type="CdrConverterPlugin"
+        base_class_type="rosbag2::FormatConverterInterface"
+>
+  <description>This is a converter plugin for cdr format.</description>
+</class>
+</library>
```
Karsten1987:
```suggestion
</library>

```

Not sure if this helps, but github _complains_ that there is no new line for this file.

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -0,0 +1,61 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cdr_converter.hpp"
+
+#include <memory>
+#include <string>
+
+#include "rcutils/strdup.h"
+#include "rmw/rmw.h"
```
Karsten1987: why is this including "rmw/rmw.h"? In case of a different middleware, this is there is no CDR support available?
Shouldn't this directly link against `fastcdr` or similar?

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "../logging.hpp"
+
+namespace rosbag2_converter_default_plugins
+{
+
+void CdrConverter::deserialize(
```
Karsten1987: I'd prefer a similar argument order than in the `rmw_deserialize` function:
https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h#L247-L253

the `ros_message` is the last argument, so being consistent with the place of the input/output arguments.

Karsten1987:
Just in case I missed something here. But this PR should convert ROS 2 messages to CDR and CDR messages to ROS 2 messages, right?
I don't think it's correct to use the `rmw_serialize` / `rmw_deserialize` functions here. If other, non-CDR midlewares are present, the `rmw_serialize` function will not convert it to CDR, but to the middleware dependent format.

Martin-Idel-SI:
My understanding is that it will work - but since there is no other rmw with a serialization format, yet, it's hard to verify (I don't like to implement a `test_rmw` as that seems like a lot of work). The plugin is explicitly linked against `rmw_fastrtps_cpp` (`lddtree` or dependency walker confirm this) and this should result in the `rmw_fastrtps_cpp`-Version being called regardless of which `rmw` is used to run `rosbag2_transport`.

But I agree that one has to be careful. That's why we suggested using Poco to grab the real `rmw_serialize` and `rmw_deserialize` functions directly from the corresponding library before. Here, we tried to follow your suggestion to link against fastrtps directly. 

The third way would be to reimplement `rmw_serialize` within this function. In other word, we'd have to reimplement this file here: https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp/src/rmw_serialize.cpp . I'm of split opinion here: This would duplicate at least 70 lines of code (maybe more if we have to implement some of the functions, too - I haven't checked) and if those lines are touched ever again, they will diverge. On the other hand, avoiding Poco AND having a very clear linking process that definitely works might be worth it.
@Karsten1987 Since you implemented the logic in https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_cpp, maybe you could copy it here if that's what you'd rather like to see?

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
Martin-Idel-SI: Makes perfect sense, will do.

rosbag2_converter_default_plugins/plugin_description.xml
Martin-Idel-SI: I guess it makes sense to fix that - I expect the xml linter to find this problem later on once it'll be enabled.

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
Martin-Idel-SI: See comment below.

dirk-thomas:
> ... since there is no other rmw with a serialization format ...

That is not true. There are a couple of non-DDS RMW implementations using different serialization protocols: see https://discourse.ros.org/t/non-dds-based-rmw-implementation/5890

> The plugin is explicitly linked against `rmw_fastrtps_cpp`

The problem is that you can't rely on FastRTPS even being present. It is absolutely viable to build ROS 2 without FastRTPS - either only with Connext or OpenSplice - or only with OPC UA or DPS.

Martin-Idel-SI:
@dirk-thomas Thank you for the link to the other rmw implementation - that means we can at least to some manual testing :+1: 

Regarding the other point, would that mean that we'd have to reimplement our own serialization routines or is there some part that would be okay to link against (statically if we must)?

dirk-thomas:
> would that mean that we'd have to reimplement our own serialization routines or is there some part that would be okay to link against (statically if we must)?

I am not sure how you map between rmw implementations and serialization formats. My point was aiming to clarify that you can't rely on a specific rmw impl. to always be present.

It is possible build ROS 2 with a single rmw impl. and it should then be possible to record and playback without requiring any extra dependencies. That is probably already the case? I was just wondering about it when you mentioned to always link against FastRTPS.

botteroa-si:
@dirk-thomas 
> It is possible build ROS 2 with a single rmw impl. and it should then be possible to record and playback without requiring any extra dependencies. That is probably already the case? 

Yes, currently `ros2 bag record` saves messages in the `rmw` format of the incoming messages, and the resulting bag file can be played back on any system which uses the same `rmw` implementation. 

With the changes introduced by PRs #56 and #57, it will be possible to specify the format to save the messages in, and to play back bags whose messages are in a different `rmw` format than the one in use. For this to be possible, it is necessary to be able to convert messages between different formats. In order to achieve this, the idea is to have converter plugins which can convert a `ROS2` message into the format used by a specific `rmw` implementation and back.
The object of this PR is a converter plugin for the `CDR` format.

Karsten1987:
new CI:
[Build Status](https://ci.ros2.org/job/ci_linux/5504/)
[Build Status](https://ci.ros2.org/job/ci_linux-aarch64/2172/)
[Build Status](https://ci.ros2.org/job/ci_osx/4544/)
waiting on windows buildfarm slot

Issue #16, opened 2018-07-30
Convert CDR to ROS 2


Karsten1987:
Implement a convert plugin to convert CDR to ROS 2 and vice versa.

PR #57, opened 2018-10-31
Use converters when recording a bag file

botteroa-si:
This PR is based on #56. It includes:
- the addition of the option `-f, --serialization-format` for the CLI verb `record`
- the `serialization-format` option defaults to the `rmw` format of the incoming messages
- if a conversion is needed, the appropriate converter plugin is used during the record process, and if such plugin does not exists, the record process fails gracefully
- now each topic has its own `rmw` serialization format (for the moment, though, a bag file can be played only if all topics have the same format. Since it is not yet possible to edit a bag file, this is not a problem. The possibility to play a bag with different serialization formats will be added in a followup PR)


anhosi:
Fixes #16

anhosi:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5645/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2262/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4670/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5541/)

Martin-Idel-SI:
This PR is now based on #64 - after this PR is merged, the number of files changed will shrink to 63.

The changeset still comprises a lot of files since we change the behaviour of the storage plugin: Instead of writing one topic type per storage storage, now every topic has its own type. This needs to be changed in the sqlite3 backend all the way to the printing statements.

rosbag2/include/rosbag2/writer.hpp
```diff
@@ -45,17 +48,26 @@ class ROSBAG2_PUBLIC Writer
...
   explicit
   Writer(
-    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> factory =
-    std::make_unique<rosbag2_storage::StorageFactory>());
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
```
Karsten1987:
```suggestion
    std::make_unique<rosbag2_storage::MetadataIo>());
    
```

rosbag2/src/rosbag2/converter.cpp
```diff
@@ -28,16 +28,18 @@
 namespace rosbag2
 {
 
+struct ConverterTypeSupport
```
Karsten1987: I believe this deserves some docs here. What's the difference between `type_support` and `introspection_type_support`? Does it make sense to rename the `type_support` into something more specific?

rosbag2/include/rosbag2/converter.hpp
```diff
@@ -60,11 +62,13 @@ class ROSBAG2_PUBLIC Converter
...
   convert(std::shared_ptr<const SerializedBagMessage> message);
 
+  void add_topic(const std::string & topic, const std::string & type);
+
 private:
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::unique_ptr<SerializationFormatConverterInterface> input_converter_;
   std::unique_ptr<SerializationFormatConverterInterface> output_converter_;
-  std::map<std::string, std::string> topics_and_types_;
+  std::map<std::string, ConverterTypeSupport> topics_and_types_;
```
Karsten1987: does this have to be an ordered map? What about `std::unordered_map`?

rosbag2/src/rosbag2/converter.cpp
```diff
@@ -58,9 +60,8 @@ Converter::~Converter()
 std::shared_ptr<SerializedBagMessage> Converter::convert(
   std::shared_ptr<const rosbag2::SerializedBagMessage> message)
 {
-  auto ts = get_typesupport(topics_and_types_[message->topic_name], "rosidl_typesupport_cpp");
-  auto introspection_ts =
-    get_typesupport(topics_and_types_[message->topic_name], "rosidl_typesupport_introspection_cpp");
+  auto ts = topics_and_types_.at(message->topic_name).type_support;
```
Karsten1987: `at` might throw. Are we positive that the map is always filled at this point?

rosbag2/src/rosbag2/converter.cpp
```diff
@@ -72,4 +73,14 @@ std::shared_ptr<SerializedBagMessage> Converter::convert(
   return output_message;
 }
 
+void Converter::add_topic(const std::string & topic, const std::string & type)
+{
+  ConverterTypeSupport type_support;
+  type_support.type_support = get_typesupport(type, "rosidl_typesupport_cpp");
```
Karsten1987: I think I'd prefer having this renamed into something like `rmw_type_support` or something.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,13 +43,29 @@ SequentialReader::open(const StorageOptions & options, const std::string & rmw_s
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-  auto storage_serialization_format = storage_->get_metadata().serialization_format;
+  auto topics = storage_->get_metadata().topics_with_message_count;
+  if (topics.empty()) {
+    return;
+  }
+
+  // Currently a bag file can only be played if all topics have the same serialization format.
```
Karsten1987: I thought we just introduced a mapping from topic to serialization format?

rosbag2/CMakeLists.txt
```diff
@@ -119,11 +121,36 @@ if(BUILD_TESTING)
...
-    test/rosbag2/types/test_ros2_message.cpp)
+    test/rosbag2/types/test_ros2_message.cpp
+    src/rosbag2/typesupport_helpers.cpp
+    src/rosbag2/types/ros2_message.cpp)
   if(TARGET test_ros2_message)
-    target_link_libraries(test_ros2_message rosbag2)
-    ament_target_dependencies(test_ros2_message test_msgs)
+    target_compile_definitions(test_ros2_message PRIVATE "ROSBAG2_BUILDING_DLL")
+    if(NOT DISABLE_SANITIZERS)
+      target_compile_options(test_ros2_message PUBLIC $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
```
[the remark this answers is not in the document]
anhosi: It did not work on the Mac of @botteroa-si.

rosbag2/src/rosbag2/converter.cpp
anhosi: I like your suggestion below `rmw_type_support`.

rosbag2/src/rosbag2/converter.cpp
anhosi: Yes we are. `add_topic` must be (and is) called for every topic before a message of such a topic can be converted. Failing to do so would be a implementation error so a exception would be ok.

rosbag2/include/rosbag2/converter.hpp
anhosi: Correct a `unordered_map` is sufficient. This is true for almost all our usages of `std::map`. As this refactoring will touch a lot of files I did it as separate pr #65.

rosbag2/src/rosbag2/converter.cpp
Karsten1987: but it will fetch the type support from the middleware, doesn't it?

anhosi: As discussed we will rename it in a follow-up pr-

rosbag2_transport/include/rosbag2_transport/record_options.hpp
```diff
@@ -25,6 +25,7 @@ struct RecordOptions
 public:
   bool all;
   std::vector<std::string> topics;
+  std::string rmw_serialization_format;
```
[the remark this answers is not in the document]
anhosi: I would do this only when this feature is being implemented. I am not yet sure how this should look like (or even how the cli usage should be). E.g. a map might also be an alternative.

rosbag2/CMakeLists.txt
```diff
@@ -119,11 +121,36 @@ if(BUILD_TESTING)
...
+      target_link_libraries(test_ros2_message $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
```
dirk-thomas, line 133: Just for the record: in the Jenkins CI jobs which are using Docker the leak sanitizer was never used for the test. See ros-infrastructure/ros_buildfarm#832.

PR #63, opened 2018-11-26
Auto discovery of new topics

Martin-Idel-SI:
Using `ros2 bag record`: Poll regularly for new topics if not all topics have been found yet or `ros2 bag record` is used with the `--all` option to subscribe to all topics.

The polling interval is currently set to 100ms and not exposed to the user, but this can easily be amended if necessary.

CI: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5648/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2265/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4672/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5544/)

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -48,10 +48,12 @@ class RecordVerb(VerbExtension):
              ' rmw currently in use')
         parser.add_argument(
             '--no-discovery', action='store_true',
-            help='disables topic auto discovery during recording')
+            help='disables topic auto discovery during recording: only topics present at '
+             'startup will be recorded')
         parser.add_argument(
             '-p', '--polling-interval', default=100,
-            help='time in ms to wait between querying available topics for recording'
+            help='time in ms to wait between querying available topics for recording. It has no '
+             'effect if --no-discovery is enabled.'
         )
 
     def create_bag_directory(self, uri):
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -46,6 +46,9 @@ class RecordVerb(VerbExtension):
             '-f', '--serialization-format', default='',
             help='rmw serialization format in which the messages are saved, defaults to the'
              ' rmw currently in use')
+        parser.add_argument(
+            '--no-discovery', action='store_true',
+            help='disables topic auto discovery during recording')
 
     def create_bag_directory(self, uri):
         try:
@@ -69,12 +72,14 @@ class RecordVerb(VerbExtension):
                 uri=uri,
                 storage_id=args.storage,
                 serialization_format=args.serialization_format,
-                all=True)
+                all=True,
+                no_discovery=args.no_discovery)
         elif args.topics and len(args.topics) > 0:
             rosbag2_transport_py.record(
                 uri=uri,
                 storage_id=args.storage,
                 serialization_format=args.serialization_format,
+                no_discovery=args.no_discovery,
                 topics=args.topics)
         else:
             self._subparser.print_help()
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -49,6 +49,10 @@ class RecordVerb(VerbExtension):
         parser.add_argument(
             '--no-discovery', action='store_true',
             help='disables topic auto discovery during recording')
+        parser.add_argument(
+            '-p', '--polling-interval', default=100,
+            help='time in ms to wait between querying available topics for recording'
+        )
 
     def create_bag_directory(self, uri):
         try:
@@ -73,13 +77,15 @@ class RecordVerb(VerbExtension):
                 storage_id=args.storage,
                 serialization_format=args.serialization_format,
                 all=True,
-                no_discovery=args.no_discovery)
+                no_discovery=args.no_discovery,
+                polling_interval=args.polling_interval)
         elif args.topics and len(args.topics) > 0:
             rosbag2_transport_py.record(
                 uri=uri,
                 storage_id=args.storage,
                 serialization_format=args.serialization_format,
                 no_discovery=args.no_discovery,
+                polling_interval=args.polling_interval,
                 topics=args.topics)
         else:
             self._subparser.print_help()
--- a/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
@@ -41,9 +41,6 @@ public:
     auto publisher_node = std::make_shared<rclcpp::Node>("publisher" + std::to_string(counter++));
     auto publisher = publisher_node->create_publisher<T>(topic_name);
 
-    // We need to publish one message to set up the topic for discovery
-    publisher->publish(message);
-
     publishers_.push_back([publisher, topic_name, message, expected_messages](
         CountFunction count_stored_messages) {
         if (expected_messages != 0) {
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
@@ -25,14 +25,15 @@ TEST_F(RecordFixture, record_end_to_end_test) {
   auto message = get_messages_primitives()[0];
   message->string_value = "test";
   size_t expected_test_messages = 3;
-  pub_man_.add_publisher("/test_topic", message, expected_test_messages);
 
   auto wrong_message = get_messages_primitives()[0];
   wrong_message->string_value = "wrong_content";
-  pub_man_.add_publisher("/wrong_topic", wrong_message);
 
   auto process_handle = start_execution("ros2 bag record --output " + bag_path_ + " /test_topic");
   wait_for_db();
+
+  pub_man_.add_publisher("/test_topic", message, expected_test_messages);
+  pub_man_.add_publisher("/wrong_topic", wrong_message);
 
   rosbag2_storage_plugins::SqliteWrapper
     db(database_path_, rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY);
--- a/rosbag2_transport/include/rosbag2_transport/record_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/record_options.hpp
@@ -27,7 +27,7 @@ public:
   bool all;
   std::vector<std::string> topics;
   std::string rmw_serialization_format;
-  std::chrono::milliseconds topic_polling_frequency;
+  std::chrono::milliseconds topic_polling_interval;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/include/rosbag2_transport/record_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/record_options.hpp
@@ -15,6 +15,7 @@
 #ifndef ROSBAG2_TRANSPORT__RECORD_OPTIONS_HPP_
 #define ROSBAG2_TRANSPORT__RECORD_OPTIONS_HPP_
 
+#include <chrono>
 #include <string>
 #include <vector>
 
@@ -26,6 +27,7 @@ public:
   bool all;
   std::vector<std::string> topics;
   std::string rmw_serialization_format;
+  std::chrono::milliseconds topic_polling_frequency;
 };
 
 }  // namespace rosbag2_transport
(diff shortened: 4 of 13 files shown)


anhosi:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5706/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2305/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4714/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5618/)

anhosi:
The macOS slave seems to have a connection problem.

anhosi:
Trying macOS again: [Build Status](https://ci.ros2.org/job/ci_osx/4716/)

anhosi:
The macOS build fails because of Clang warnings in non rosbag2 code.

anhosi:
We added the option `--no-discovery` that allows to disable the auto discovery. If this option is used the resulting bag may be empty if the initial (and only) subscription fails. We consider this option "advanced" and did not implement checks whether all requested topics could be subscribed as this hard to do consistently.

New round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5741/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2325/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4743/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5641/)

ros2bag/ros2bag/verb/record.py
```diff
@@ -46,6 +46,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '-f', '--serialization-format', default='',
             help='rmw serialization format in which the messages are saved, defaults to the'
              ' rmw currently in use')
+        parser.add_argument(
+            '--no-discovery', action='store_true',
```
Karsten1987: I guess we could leave the argument as is for now given that the description of it mentions that this only applies to during runtime.

rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
```diff
@@ -41,9 +41,6 @@ class PublisherManager
     auto publisher_node = std::make_shared<rclcpp::Node>("publisher" + std::to_string(counter++));
     auto publisher = publisher_node->create_publisher<T>(topic_name);
 
-    // We need to publish one message to set up the topic for discovery
-    publisher->publish(message);
```
Karsten1987: is this just legacy or why is this code removed?

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -25,15 +25,16 @@ TEST_F(RecordFixture, record_end_to_end_test) {
...
-  pub_man_.add_publisher("/test_topic", message, expected_test_messages);
 
   auto wrong_message = get_messages_primitives()[0];
   wrong_message->string_value = "wrong_content";
-  pub_man_.add_publisher("/wrong_topic", wrong_message);
 
   auto process_handle = start_execution("ros2 bag record --output " + bag_path_ + " /test_topic");
   wait_for_db();
 
+  pub_man_.add_publisher("/test_topic", message, expected_test_messages);
```
Karsten1987: I believe we should also have a test for the `--no-discovery` to make sure we don't have any race conditions here.

rosbag2_transport/include/rosbag2_transport/record_options.hpp
```diff
@@ -24,8 +25,10 @@ struct RecordOptions
 {
 public:
   bool all;
+  bool is_discovery_disabled;
   std::vector<std::string> topics;
   std::string rmw_serialization_format;
+  std::chrono::milliseconds topic_polling_interval;
```
Karsten1987: would it make sense to expose this to the user as a parameter?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
   return subscription;
 }
 
+std::future<void> Recorder::launch_topics_discovery(
+  std::chrono::milliseconds topic_polling_interval,
```
Karsten1987: i think it makes sense to actually use ROS time (rclcpp time). so we can react on system clocks vs steady clock vs sim time.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
   return subscription;
 }
 
+std::future<void> Recorder::launch_topics_discovery(
```
Karsten1987: why do we need this function? Isn't this function just a wrapper around the lambda and its call?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
...
+        if (is_every_topic_subscribed(topics_to_record)) {
+          return;
+        }
+
+        subscribe_all_missing_topics(all_topics_and_types_to_subscribe);
+        std::this_thread::sleep_for(topic_polling_interval);
+      } while (rclcpp::ok() && !is_discovery_disabled);
+    };
+
+  return std::async(std::launch::async, subscribe_to_topics);
```
Karsten1987: i am not quite sure I understand why this have to be an async call?

rosbag2_transport/src/rosbag2_transport/recorder.hpp
```diff
@@ -42,10 +47,21 @@ class Recorder
...
+  void subscribe_all_missing_topics(
+    const std::unordered_map<std::string, std::string> & all_topics_and_types);
+  void subscribe_topic(const rosbag2::TopicMetadata & topic);
+  bool is_every_topic_subscribed(const std::vector<std::string> & topics_to_record) const;
+  void record_messages() const;
 
   std::shared_ptr<rosbag2::Writer> writer_;
   std::shared_ptr<Rosbag2Node> node_;
   std::vector<std::shared_ptr<GenericSubscription>> subscriptions_;
+  std::vector<std::string> subscribed_topics_;
```
Karsten1987: i believe an `unordered_set<std::string>` makes more sense here, because you really just need a quick lookup whether it's subscribed or not.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
...
 
+std::future<void> Recorder::launch_topics_discovery(
+  std::chrono::milliseconds topic_polling_interval,
+  bool is_discovery_disabled,
+  const std::vector<std::string> & topics_to_record)
+{
+  auto subscribe_to_topics =
+    [this, topics_to_record, topic_polling_interval, is_discovery_disabled] {
+      do {
+        auto all_topics_and_types_to_subscribe = topics_to_record.empty() ?
```
Karsten1987: I think this is pretty complicated for what it is.
wouldn't it be easier to have this function called once as a blocking call to start the recording and then asynchronously call `node_->get_topics_with_types` and subscribe to the not-yet-subscribed once? I think this makes it easier to follow what's happening here.

By extracting the new topics, the call to `is_every_topic_subscribed` is then not necessary anymore, because we know which topics to subscribe.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
...
+          node_->get_all_topics_with_types() :
+          node_->get_topics_with_types(topics_to_record);
+
+        if (is_every_topic_subscribed(topics_to_record)) {
+          return;
+        }
+
+        subscribe_all_missing_topics(all_topics_and_types_to_subscribe);
+        std::this_thread::sleep_for(topic_polling_interval);
+      } while (rclcpp::ok() && !is_discovery_disabled);
```
Karsten1987: given my comment before, this condition can then be executed once on the caller side to only start the async call when discovery is enabled.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -83,4 +75,65 @@ Recorder::create_subscription(
...
+  const std::vector<std::string> & topics_to_record) const
+{
+  return !topics_to_record.empty() && subscriptions_.size() == topics_to_record.size();
+}
+
+void Recorder::subscribe_all_missing_topics(
+  const std::unordered_map<std::string, std::string> & all_topics_and_types)
+{
+  for (const auto & topic_with_type : all_topics_and_types) {
+    bool already_subscribed = find(
```
Karsten1987: `std::unordered_set<std::string>` looks to me like the better data structure as it has O(1) lookup time and no search is necessary.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -48,6 +51,8 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
   storage_options.uri = std::string(uri);
   storage_options.storage_id = std::string(storage_id);
   record_options.all = all;
+  record_options.is_discovery_disabled = no_discovery;
+  record_options.topic_polling_interval = std::chrono::milliseconds(100);
```
Karsten1987: does it make sense to expose these 100 milliseconds ?

rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
anhosi: With auto discovery we no longer need to advertise the topic before starting rosbag in the test.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
anhosi: I think I do not understand what the possible advantage would be. The polling interval controls how often rosbag looks for new topics. I would expect this to happen in regularly in "rosbag"-time independent of what the ros time is "doing".

rosbag2_transport/include/rosbag2_transport/record_options.hpp
anhosi: This would have a similar expert-level as `--no-discovery`, so yes.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
anhosi: Yes, but it very much improves the readability of `record()`.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
anhosi: > By extracting the new topics, the call to `is_every_topic_subscribed` is then not necessary anymore, because we know which topics to subscribe.

I don't understand. In the `--all` case we do not know what topics to expect.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
anhosi: I changed it this way, readability does not really improve imho.

botteroa-si:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5805/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2373/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4800/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5702/)

botteroa-si:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5805/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2373/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4800/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5702/)

anhosi:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5809/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2376/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5707/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5839/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2403/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4830/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5734/)

PR #70, opened 2018-12-03
Split converters

Martin-Idel-SI:
This PR splits the converter interfaces into two interfaces: A serializer and a deserializer. This allows to have plugins which only support converting to OR from a custom serialization format (such as the plugin for old rosbags).

The changes also mimic the storage_interface architecture:
- As a plugin author, I can choose to write a plugin for serializers, deserializers or both at the same time (implementing only one interface)
- As a user in rosbag2, I only have to interact with the serializers and deserializers (if I can't cirectly use the converter class), so not much changes

The main change is the `serialization_format_converter_factory` and its impl file as well as the usage in `converter.cpp`.

In addition, the error message if a converter is missing was changed (see end to end test for an example).

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5733/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2319/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4736/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5635/)

### DIFF
--- a/rosbag2/include/rosbag2/converter_interfaces/serialization_format_converter_interface.hpp
+++ b/rosbag2/include/rosbag2/converter_interfaces/serialization_format_converter_interface.hpp
@@ -33,10 +33,14 @@
 namespace rosbag2
 {
 
+namespace converter_interfaces
+{
+
 class SerializationFormatConverterInterface
   : public SerializationFormatSerializerInterface, public SerializationFormatDeserializerInterface
 {};
 
+}  // namespace converter_interfaces
 }  // namespace rosbag2
 
 #endif  // ROSBAG2__CONVERTER_INTERFACES__SERIALIZATION_FORMAT_CONVERTER_INTERFACE_HPP_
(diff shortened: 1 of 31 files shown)


rosbag2/include/rosbag2/converter_interfaces/serialization_format_converter_interface.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * This is a convenience class for plugin developers. When developing a plugin to both write and
+ * read a specified serialization format, inherit from this class
+ */
+namespace rosbag2
+{
+
+namespace converter_interfaces
+{
+
+class SerializationFormatConverterInterface
```
Karsten1987: given that we have `interfaces` in the namespace, we could maybe rename `SerializationFormatConverterInterface` to only `SerializationFormatConverter`.

Martin-Idel-SI: Sounds good to me. Done.

PR #76, opened 2018-12-18
find openssl

Karsten1987:
@anhosi @botteroa-si 

I don't exactly know how this happened, but somehow this commit was lost. Please review.


botteroa-si:
@Karsten1987 yep, it seems that somehow it went lost and I also don't get why. I tried the commit again and it works as before. Therefore lgtm.

Issue #74, opened 2018-12-15
ros1_rosbag_storage_vendor missing dependencies.


nuclearsandwich:
The patches in this package suggest to me that it must always build the vendored project, in which case it must specify the dependencies of that project in order to build as a released package. The release job is [currently failing](http://build.ros2.org/job/Cbin_uB64__ros1_rosbag_storage_vendor__ubuntu_bionic_amd64__binary/1/) due to the absence of pluginlib in the build container. A grep through the patches suggests there are likely other build and runtime dependencies that should be specified.

```
❯ git grep find_package
CMakeLists.txt:find_package(ament_cmake REQUIRED)
resources/cmakeliststxt.diff:+find_package(ament_cmake REQUIRED)
resources/cmakeliststxt.diff:+find_package(pluginlib 2.0 REQUIRED)
resources/cmakeliststxt.diff: find_package(console_bridge REQUIRED)
resources/cmakeliststxt.diff:-find_package(catkin REQUIRED COMPONENTS cpp_common pluginlib roscpp_serialization roscpp_traits rostime roslz4)
resources/cmakeliststxt.diff: find_package(Boost REQUIRED COMPONENTS date_time filesystem program_options regex)
resources/cmakeliststxt.diff: find_package(BZip2 REQUIRED)
resources/cmakeliststxt.diff:+find_package(ros1_bridge REQUIRED)
resources/cmakeliststxt.diff:+find_package(PkgConfig)
resources/cmakeliststxt.diff:-    find_package(rostest)
resources/cmakeliststxt.diff:+#    find_package(ament_cmake_gmock)
resources/rosbag1_encryption_patch.diff:     find_package(rostest)
```

Karsten1987:
What's the best way of testing that all dependencies are available? When running CI on `ci.ros2.org` these builds are clean, also when running in a new docker container. Any best practice advice in order to avoid constant re-blooming?

PR #77, opened 2018-12-18
add ros1 rosbag_storage dependencies to vendor package

Karsten1987:
fixes ros2/rosbag2#74 ... hopefully :)


Martin-Idel-SI:
I don't see any problems with this change.

PR #82, opened 2018-12-20
Fix patch command osx

Karsten1987:
open for visibility

### DIFF
--- a/ros1_rosbag_storage_vendor/package.xml
+++ b/ros1_rosbag_storage_vendor/package.xml
@@ -8,6 +8,7 @@
   <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>git</buildtool_depend>
 
   <build_depend>boost</build_depend>
   <build_depend>bzip2</build_depend>


ros1_rosbag_storage_vendor/CMakeLists.txt
```diff
@@ -39,20 +39,26 @@ include(ExternalProject)
 # vendor package building against the wrong pluginlib, because the paths get switched.
 # Symptoms of wrong paths: The build of the rosbag for ROS 1 plugins fails with missing symbols
 # like ros::package::getPlugins(...) when linking the vendor package
+if (APPLE)
+  set(PATCH_CMD git apply)
```
wjwwood: Why is this necessary?

Karsten1987: When calling only patch on it, on my machine only the following files were patched:

```
patching file tools/rosbag_storage/CMakeLists.txt
patching file tools/rosbag_storage/encryptor_plugins.xml
patching file tools/rosbag_storage/include/rosbag/aes_encryptor.h
patching file tools/rosbag_storage/include/rosbag/encryptor.h
patching file tools/rosbag_storage/include/rosbag/gpgme_utils.h
patching file tools/rosbag_storage/include/rosbag/no_encryptor.h
patching file tools/rosbag_storage/src/aes_encryptor.cpp
patching file tools/rosbag_storage/src/gpgme_utils.cpp
patching file tools/rosbag_storage/src/encryptor.cpp
patching file tools/rosbag_storage/test/test_aes_encryptor.cpp
patching file tools/rosbag_storage/include/rosbag/bag.h
patching file tools/rosbag_storage/src/bag.cpp
patching file tools/rosbag_storage/src/bag.cpp
patching file tools/rosbag_storage/encryptor_plugins.xml
patching file tools/rosbag_storage/package.xml
```
Notably, the file `src/no_encryptor.cpp` did not get created and therefore my build does not compile. I don't really know what exactly the source of this problem is, but with `git apply` everything works fine.

So I can't really tell why the following lines are not being considered:
https://github.com/ros2/rosbag2/blob/master/ros1_rosbag_storage_vendor/resources/rosbag1_encryption_patch.diff#L543-L558

wjwwood: Hmm, ok, I was just curious. Sometimes I play with the `-p0` vs `-p1` and it helps.

PR #83, opened 2019-01-12
Update README.md

Karsten1987:
with the `*` the correct `ros2bag` package will be selected.


Karsten1987:
I'll merge given that the current information in the readme is definitely not working.

PR #86, opened 2019-02-06
enforce unique node names

Karsten1987:
fixes the unique node problem referred to in https://github.com/ros2/rcl/issues/375

the template function indeed creates a new `static int counter` per type and thus non-unique node names when different types are used.

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
@@ -33,14 +33,21 @@ using CountFunction = std::function<size_t(const std::string &)>;
 class PublisherManager
 {
 public:
+  ~PublisherManager()
+  {
+    publishers_.clear();
+    publisher_nodes_.clear();
+  }
+
   template<class T>
   void add_publisher(
     const std::string & topic_name, std::shared_ptr<T> message, size_t expected_messages = 0)
   {
-    static int counter = 0;
-    auto publisher_node = std::make_shared<rclcpp::Node>("publisher" + std::to_string(counter++));
+    auto node_name = std::string("publisher") + std::to_string(counter_++);
+    auto publisher_node = std::make_shared<rclcpp::Node>(node_name);
     auto publisher = publisher_node->create_publisher<T>(topic_name);
 
+    publisher_nodes_.push_back(publisher_node);
     publishers_.push_back([publisher, topic_name, message, expected_messages](
         CountFunction count_stored_messages) {
         if (expected_messages != 0) {
@@ -72,6 +79,8 @@ public:
   }
 
 private:
+  int counter_ = 1;
+  std::vector<std::shared_ptr<rclcpp::Node>> publisher_nodes_;
   std::vector<std::function<void(CountFunction)>> publishers_;
 };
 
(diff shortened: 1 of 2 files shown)


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6130/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2603/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5057/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5998/)

Karsten1987:
@jacobperron the current CI fails because of https://github.com/ament/ament_lint/issues/116.
how do i ensure cppcheck is correctly invoked?

jacobperron:
> @jacobperron the current CI fails because of [ament/ament_lint#116](https://github.com/ament/ament_lint/issues/116).
how do i ensure cppcheck is correctly invoked?

That's unfortunate, I ended up excluding passing include directories from external packages to `cppcheck` because of performance: 

https://github.com/ament/ament_lint/blob/4f9bfc00106496eeafa592bc5dfbc33b052d643d/ament_cmake_cppcheck/cmake/ament_cmake_cppcheck_lint_hook.cmake#L44-L47

One way around it would be to manually invoke `ament_cppcheck` and tell it about the header files containing the macros in `rclcpp` (rather than relying on `ament_lint_auto`). This means that you'd also have to invoke the other desired linters manually too. I think the `ament_cppcheck` command would look something like this:

```cmake
ament_cppcheck(INCLUDE_DIRS ${rclcpp_INCLUDE_DIRS})
```

I think this is another instance motivating ament/ament_lint#119, or for a feature enabling more control over the configurations for the linters invoked by ament.

dirk-thomas:
Our package template already demonstrates how individual linters part of the common linters can be skipped: https://github.com/ros2/ros2cli/blob/5dd01716edbb899e9ee270e56cf47a77b94997a2/ros2pkg/ros2pkg/resource/ament_cmake/CMakeLists.txt.em#L92-L94

Karsten1987:
> That's unfortunate, I ended up excluding passing include directories from external packages to `cppcheck` because of performance:

does that mean no _external_ package can use `ament_lint_auto`? Or asked differently, which packages can use it? How is external defined here?
if every package has to manually exclude `cppcheck` then automatic lookup for linters defeats kind of the purpose, no?

jacobperron:
> does that mean no _external_ package can use `ament_lint_auto`? Or asked differently, which packages can use it? How is external defined here?

To elaborate, to handle error you're seeing I added support in [ament/ament_lint#119](https://github.com/ament/ament_lint/issues/119) to pass a list of include directories to cppcheck so that it can resolve the "unknown macros". Unfortunately, a larger list of directories results in a very slow execution of cppcheck. So, when cppcheck is invoked automatically with `ament_lint_auto`, we limit the number of include directories to just those of the package being linted. An "external" package for the purpose of the automatic cppcheck, in this case, is anything that is not in `rosbag2_transport`.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6286/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2706/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5166/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6125/)

Karsten1987:
unrelated test errors/warnings. That means the cppcheck is correctly disabled as well as the node names are unique and don't produce any warnings.

PR #88, opened 2019-02-11
Dnae adas/sequential writing

DensoADAS:
Feature: Writing serialized ROS2 messages to a binary file instead of sqlite3 database.
 * speed improvement
 * allows to split big recordings
 * allows to write to multiple disks (for higher write performance)

Question: Should the identifier for the storage plugin be changed? (rosbag2 recorded with this system cannot be played by previous versions, but vice versa)

### DIFF
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -86,7 +86,7 @@ private:
     uint64_t size;
     uint32_t file_index;
   };
-#pragma pack(pop) //back to whatever the previous packing mode was
+#pragma pack(pop)  // back to whatever the previous packing mode was
 
   struct BinarySequentialFile
   {
@@ -94,7 +94,8 @@ private:
     std::fstream file_stream;
     std::mutex lock;
 
-    BinarySequentialFile(const std::string &_path) : path(_path)
+    explicit BinarySequentialFile(const std::string & _path)
+    : path(_path)
     {
     }
   };
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <fstream>
 
 #include "rcutils/types.h"
 #include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
@@ -85,6 +86,8 @@ private:
   ReadQueryResult::Iterator current_message_row_;
   std::unordered_map<std::string, int> topics_;
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
+  bool extra_file_;
+  std::fstream data_file_;
 };
 
 }  // namespace rosbag2_storage_plugins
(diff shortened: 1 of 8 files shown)


rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -19,6 +19,8 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <fstream>
```
Karsten1987: please order the includes in alphabetical order

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -82,6 +85,50 @@ void SqliteStorage::open(
...
+
+    // in write mode we add at least one serialized file
+    // (it would be possible to split the data into into multiple files)
+    if(!ro)
+    {
+      data_files_.push_back(
+        std::make_unique<BinarySequentialFile>(database_name_ + ".extra")
+      );
+    }
+    else if(metadata->relative_file_paths.size()>=1)
```
Karsten1987: spaces around the operators

Karsten1987: how does such a metafile now look like? What's the enforced format for the relative file paths ?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -96,8 +143,34 @@ void SqliteStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMe
...
+    MetaInformation data = {
+      static_cast<uint64_t>(file_handle->file_stream.tellg()),
+      static_cast<uint64_t>(message->serialized_data->buffer_length),
+      file_index
+    };
+
+    file_handle->file_stream.write(reinterpret_cast<const char *>(message->serialized_data->buffer),
+      message->serialized_data->buffer_length);
+
+    auto serialized_data = std::make_shared<rcutils_uint8_array_t>();
```
Karsten1987: I am having trouble to understand what's happening with this `serialized_data` instance.

Karsten1987:
I have a fundamental question for this PR.
What's the intended usage of this extra file? The way I see this every incoming message will be stored twice now. Once in the original `.db3` and additionally in the `db3.extra` file. 
I would actually recommend having this as a new and separate storage plugin with a new identifier. This further than allows to stay backwards compatible to existing rosbags but also makes the distinguishing between sqlite and native file clear.

DensoADAS:
Separated the plugin into "sqlite3_sequential". Main difference between both plugins are:
 a) in the original version the content of the messages are stored in the database itself (the CPU encodes the message; the database gets quite big for images or similar content)
 b) while in the sequential-version
    - the content of the message is stored in one (or more) binary files (which can be written without encoding)
    - in the database itself, only the offset (uint64) and the size of the message (uint64) and the file index (corresponding to the relative file paths in the metafile) are stored

DensoADAS:
* changes due to request
* rebased

Karsten1987:
@DensoADAS sorry for stalling this PR for so long. Do you mind rebasing this PR once more? Or is that no longer relevant? I remember there was some connection to the newly introduced intra process communication.

DensoADAS:
We're waiting for [rclpp PR #291 973](https://github.com/ros2/rclcpp/pull/973)
We prepared an updated version of rosbag2 compatible to these changes.

emersonknapp:
@DensoADAS are you still planning on implementing this feature? The code has diverged considerably since this was opened, so I imagine it would need a bit of rework. 

It also seems like this could be implemented in a separate package as a standalone plugin, rather than going into `rosbag2_storage_default_plugins`. I'm inclined to only install a single storage plugin by default, allowing other storage implementations to be optionally installed. If so, that would mean this should definitely be its own package, and I'm inclined to think, its own repository as well.

I'm going to close this for now, and if you would like to continue the conversation/development about including it in this repository, please feel free to reopen.

Issue #85, opened 2019-02-05
refactor rosbag2_tests to run without ros1 dependency


Karsten1987:
currently the `rosbag2_tests` package depends on the `rosbag2_bag_v2_plugin` and thus can't be built without this. This also means that the ros1-agnostic rosbag tests can't be built.

PR #90, opened 2019-02-12
removed dependency to ros1_bridge package

DensoADAS:
fixes ros2/rosbag2#85

removed dependency to ros1_bridge package:
 * checking if package is available
 * if not skipping (with warnings)
 * now rosbag2_tests builds on systems without ros1


Karsten1987:
extremely helpful patch. Running CI on it:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6171/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2632/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5086/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6027/)

I believe the cmake warning has to be turned into`STATUS` in order make CI green.

Karsten1987:
new round of CI, disabling `ros1_rosbag_storage_vendor` `rosbag2_bag_v2_plugins`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6172/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2633/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5087/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6028/)

that being said, I would really appreciate if we could use this PR to extend the same logic to the two excluded packages so that this complete repo can actually be built without any ros1 dependency.

Karsten1987:
Looking at the latest CI, this patch is not sufficient. Looks like we should patch the two packages in question as well in . order to run CI correctly.

rosbag2_tests/CMakeLists.txt
```diff
@@ -22,10 +22,12 @@ find_package(ament_index_cpp REQUIRED)
 
 find_package(PkgConfig)
 if(PKG_CONFIG_FOUND)
-  find_package(ros1_bridge REQUIRED)
-  include(${ros1_bridge_DIR}/find_ros1_interface_packages.cmake)
-  include(${ros1_bridge_DIR}/find_ros1_package.cmake)
-  find_ros1_package(roscpp)
+  find_package(ros1_bridge)
+  if(ros1_bridge_DIR)
```
dirk-thomas: This should check the `ros1_bridge_FOUND` variable instead.

rosbag2_tests/CMakeLists.txt
```diff
@@ -22,10 +22,12 @@ find_package(ament_index_cpp REQUIRED)
...
+    include(${ros1_bridge_DIR}/find_ros1_interface_packages.cmake)
+    include(${ros1_bridge_DIR}/find_ros1_package.cmake)
```
dirk-thomas: It would be better to update the `ros1_bridge` package to include these files when being found.

Karsten1987:
depends on https://github.com/ros2/ros1_bridge/pull/164

rosbag2_bag_v2_plugins/CMakeLists.txt
```diff
@@ -17,23 +17,17 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 
-find_package(PkgConfig)
-if(NOT PKG_CONFIG_FOUND)
-  message(WARNING "Failed to find PkgConfig, skipping...")
+# include bridge first to not create package if ros1 packages cannot be found
+find_package(ros1_bridge QUIET)
+if (NOT ros1_bridge_FOUND)
```
Karsten1987:
```suggestion
if(NOT ros1_bridge_FOUND)
```

PR #100, opened 2019-04-12
Fix issue with ros2bag record python frontend

sagniknitr:
Currently "ros2 bag record" command causes crashes when used with arguments. The PR fixes it.
@Karsten1987 review and get back to me.

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -48,7 +48,7 @@ class RecordVerb(VerbExtension):
             help='disables topic auto discovery during recording: only topics present at '
              'startup will be recorded')
         parser.add_argument(
-            '-p', '--polling-interval', default=100,
+            '-p', '--polling-interval', type=int, default=100,
             help='time in ms to wait between querying available topics for recording. It has no '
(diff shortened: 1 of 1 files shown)


ros2bag/ros2bag/verb/record.py
```diff
@@ -48,10 +48,12 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             help='disables topic auto discovery during recording: only topics present at '
              'startup will be recorded')
         parser.add_argument(
-            '-p', '--polling-interval', default=100,
+            '-p', '--polling-interval', type=int, default=100,
             help='time in ms to wait between querying available topics for recording. It has no '
              'effect if --no-discovery is enabled.'
         )
+        self._subparser = parser
```
Karsten1987: You don't need that line.

sagniknitr: @Karsten1987  but without it won't line number 92 throw error ? ( inside the else statement which calls             `self._subparser.print_help()` )

Karsten1987: you're absolutely right!

Karsten1987:
thanks a lot for the patch!

PR #105, opened 2019-04-30
Exclude topics

danieldube:
Changed the PlayOptions class to hold a topic filter function. This function tells the player, whether a topic should be included in the playback or not.

The python API now uses the topic filter to provide an exclude topic feature.


rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -124,6 +128,32 @@ rosbag2_transport_play(PyObject * Py_UNUSED(self), PyObject * args, PyObject * k
...
+
+        Py_DECREF(topic);
+      }
+      Py_DECREF(topic_exclude_iterator);
+    }
+
+    if (exclude_topics_list.empty() == false) {
+      auto topic_filter_function = [exclude_topics_list](const std::string& topic)
+      {
+        auto entry = exclude_topics_list.find(topic);
```
Karsten1987: Could this block just be replaced by 
```return (exclude_topics_list.find(topic) != exclude_topics_list.end());```

File: ros2bag/ros2bag/verb/play.py [hunk 2]
```diff
@@ -43,4 +46,5 @@ def main(self, *, args):  # noqa: D102
             uri=bag_file,
             storage_id=args.storage,
             node_prefix=NODE_NAME_PREFIX,
-            read_ahead_queue_size=args.read_ahead_queue_size)
+            read_ahead_queue_size=args.read_ahead_queue_size,
+            exclude_topics=args.exclude_topics)
```
Karsten1987: The check for unique topic names can be done in python. I guess that's a bit easier and avoids the need of a `std::set`. 
Something like `exclude_topics = list(set(args.exclude_topics))`

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -124,6 +128,32 @@ rosbag2_transport_play(PyObject * Py_UNUSED(self), PyObject * args, PyObject * k
   play_options.node_prefix = std::string(node_prefix);
   play_options.read_ahead_queue_size = read_ahead_queue_size;
 
+  std::set<std::string> exclude_topics_list;
```
Karsten1987: Given the command that the topic names could be filtered for uniqueness on the python side, I believe it makes sense to parse the topics as a vector and reuse the code for converting the PyObject to a vector from the `record` function. I believe this conversion can be excluded in a separate function.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -24,8 +24,11 @@ namespace rosbag2_transport
 struct PlayOptions
 {
 public:
+  using TopicFilter = std::function<bool(const std::string& topic)>;
```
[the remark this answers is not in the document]
danieldube: Thanks for the fast review. #1 of the change requests should is done. The rest will follow over the next days.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
danieldube: Refactored the PyObject parser code into a generic parse_python_list() function.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -24,8 +24,11 @@ namespace rosbag2_transport
...
+
   size_t read_ahead_queue_size;
   std::string node_prefix = "";
+  TopicFilter topic_filter = [](const std::string&){return false;};
```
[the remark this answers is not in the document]
danieldube: Done.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -16,6 +16,7 @@
 #include <chrono>
 #include <string>
 #include <vector>
+#include <set>
```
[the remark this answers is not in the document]
danieldube: Done.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
danieldube: I would argue, that the current implementation is more readable and therefore more expressive than putting everything into one line. However, if you like to change this, I'll do it.

ros2bag/ros2bag/verb/play.py
danieldube: For me a set would be the natural choice of container. I guess at this point it doesn't make a difference, but the idea is to have a O(log n) complexity for excluding topic names. 

Using a set on python side, wouldn't guarantee, that the API isn't misused from Python side in the future. Using a sorted vector and binary search for the topic lookup bloats the code. @Karsten1987, are you sure you want to have a std::vector in favor of a std::set?

danieldube:
Currently the Jenkins can not build the new test case for the exclude_topic_filter. The error message is 
```/tmp/ws/src/rosbag2/rosbag2_transport/test/rosbag2_transport/test_play.cpp:111:61: error: ‘Primitives’ is not a member of ‘test_msgs::msg’
   auto primitive_message = std::make_shared<test_msgs::msg::Primitives>();``` (http://build.ros2.org/job/Dpr__rosbag2__ubuntu_bionic_amd64/8/consoleText )

I can not reproduce it on my local system. The class `test_msgs::msg::Primitives` is well defined in the header file `primitives__struct.hpp`. The same object is used in the test case `recorded_messages_are_played_for_all_topics` defined in the same source file. The only thing I recognized is, that the line number 111 stated by Jenkins is wrong. Actually the rejected code is in line 112.

I'm out of ideas. @Karsten1987, do you have a clue where I should start looking for this problem?

**Comment by Karsten1987** [comment]
The `test_msgs` in Dashing were recently renamed. There is no longer a `test_msgs::msg::Primitives` message. That type was renamed to `BasicTypes`. 
We recently had to change all these test messages ourselves: 
https://github.com/ros2/rosbag2/pull/111/files

In any case, thanks for iterating over this PR. I'll review your PR as soon as possible.

**Comment by Karsten1987** [comment]
Looking at this PR again, I believe you actually have to rebase your branch on top of the latest master.

**Comment by danieldube** [comment]
Thanks, this was exactly what I was missing. The current rosbag2 master is not compiling using the latest ROS2 stable release anymore. Since I can not build the ROS2 master on my system, I'm trying to fix the build problems in blind mode. Let's see what Jenkins is saying. ;)

File: rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp [hunk 7]
```diff
@@ -22,6 +23,23 @@
 #include "rosbag2_transport/storage_options.hpp"
 #include "rmw/rmw.h"
 
+
+template<class InserterIterator>
```
Karsten1987: nice 👍

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
Karsten1987: I thought to believe that it's easier for the compiler to inline and optimize this as there is no if and thus no branch prediction and stuff might happen. But it's solely premature optimization thinking :)

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -104,3 +104,56 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
     Each(Pointee(Field(&test_msgs::msg::Arrays::float32_values,
     ElementsAre(40.0f, 2.0f, 0.0f)))));
 }
+
```
Karsten1987:
```suggestion
```

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -24,8 +24,11 @@ namespace rosbag2_transport
 struct PlayOptions
 {
 public:
+  using TopicFilter = std::function<bool (const std::string & topic)>;
+
   size_t read_ahead_queue_size;
   std::string node_prefix = "";
+  TopicFilter exclude_topic_filter = [](const std::string &) {return false;};
```
Karsten1987: When I am thinking about this, I believe it's a bit overkill to provide a lambda/function for this. Why can't a simple `std::vector<std::string>` not suffice? Especially when using this as an user facing option. I believe a list of vectors is easier and we can internally apply a generic lambda.

Karsten1987:
@danieldube Is there any update from your side on this PR?

danieldube:
Nope. Didn't have time yet to get back to this.

Karsten1987:
@danieldube just a friendly ping for an update

danieldube:
I will not be able to do the necessary changes right now. I will reopen the request, when there is some progress. @Karsten1987, thanks for the review.

PR #116, opened 2019-05-08
ignore cast function type warning

Karsten1987:
This fixes the warning shown in CentOS. According to https://docs.python.org/3/c-api/structures.html#c.PyMethodDef this cast is correct and necessary. Thus suppress the warning for this package.

CentOS CI: https://ci.ros2.org/job/ci_linux-centos/7/
Warning is unrelated

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
@@ -155,6 +155,9 @@ rosbag2_transport_info(PyObject * Py_UNUSED(self), PyObject * args, PyObject * k
 }
 
 /// Define the public methods of this module
+#if !defined(_WIN32)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wcast-function-type"
 static PyMethodDef rosbag2_transport_methods[] = {
   {
     "record", reinterpret_cast<PyCFunction>(rosbag2_transport_record), METH_VARARGS | METH_KEYWORDS,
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Can't this be done with a more narrow scope?

Karsten1987:
narrowed done to the specific lines of code. Using the diagnostics pragma to ignore the warnings.
new CentOS CI here:
[Build Status](https://ci.ros2.org/job/ci_linux-centos/8/)

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -170,6 +173,8 @@ static PyMethodDef rosbag2_transport_methods[] = {
   },
   {nullptr, nullptr, 0, nullptr}  /* sentinel */
 };
+# pragma GCC diagnostic pop
+#endif
```
dirk-thomas: The inner block should still be used on non-Windows platforms, no? Currently the `ifdef` wraps all of it - you probably want to `endif` after line 160 and `ifdef` again before line 176.

Karsten1987: absolutely! Corrected in https://github.com/ros2/rosbag2/pull/116/commits/928a4eeea7874107accd1026ee1dd86601acd049

dirk-thomas:
Have you tried building this on Linux? It seems it fails in the latest release: http://build.ros2.org/view/Dbin_uB64/job/Dbin_uB64__rosbag2_transport__ubuntu_bionic_amd64__binary/1/

PR #124, opened 2019-06-04
increase concurrentqueue download timeout

mikaelarguedas:
It currently fails to build on my machine due to slow download speed for this particular archive


mikaelarguedas:
looks like the fluke on github content was resolved and I could build this multiple times successfully in the last days.

Closing

Issue #127, opened 2019-06-08
rosbag2 fails to record messages generated directly from IDL


davidhodo:
## Bug report
- Operating System:
  - MacOS 10.14.5 (Mojave)
- Installation type:
  - Binaries
- Version or commit hash:
  - Dashing
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
ROS2 bag fails to subscribe to messages generated directly from .idl files.  An example package that creates a simple string message using idl can be found at: [ros2_idl_demo](https://github.com/davidhodo/ros2_idl_demo).  Building the message (idl_msgs) and sample publisher (idl_pub) packages and then running `ros2 run idl_pub idl_pub` and 'ros2 bag record -a' or 'ros2 bag record /chatter' fails to record the message published from idl_pub.

#### Expected behavior

```
davidhodo@Davids-MacBook-Pro:idl_test_ws∫ ros2 bag record /chatter
[INFO] [rosbag2_storage]: Opened database 'rosbag2_2019_06_08-12_29_58'.
[INFO] [rosbag2_transport]: Listening for topics...
[INFO] [rosbag2_transport]: Subscribed to topic '/chatter'
[INFO] [rosbag2_transport]: All requested topics are subscribed. Stopping discovery...
```

#### Actual behavior

```
davidhodo@Davids-MacBook-Pro:idl_test_ws∫ ros2 bag record /chatter
[INFO] [rosbag2_storage]: Opened database 'rosbag2_2019_06_08-12_17_47'.
[INFO] [rosbag2_transport]: Listening for topics...
```

#### Additional information
Could this be related to the message type / package name changes when using idl directly as described in this [issue](https://github.com/ros2/ros2cli/issues/277)? 

[UPDATE]
In typesupport_helpers.cpp:99 there is a hardcoded "\_\_msg\_\_" in the generation of the symbol name.  Changing this to "\_\_idl\_\_" allows the idl message to be recorded at the expense of no longer working for standard messages.  This value should be set based on the package type.

davidhodo:
Resolved by PR #128

PR #128, opened 2019-06-09
Add support for parsing middle module name from type

davidhodo:
Allows support for message types generated from both msg and idl files.  Attempts to solve #127.  I'm making the assumption that all type names are of the form package_name/middle_module (i.e. idl or msg)/type.

### DIFF
--- a/rosbag2/include/rosbag2/typesupport_helpers.hpp
+++ b/rosbag2/include/rosbag2/typesupport_helpers.hpp
@@ -29,7 +29,7 @@ const rosidl_message_type_support_t *
 get_typesupport(const std::string & type, const std::string & typesupport_identifier);
 
 ROSBAG2_PUBLIC
-const std::pair<std::string, std::string> extract_type_and_package(const std::string & full_type);
+const std::tuple<std::string, std::string, std::string> extract_type_and_package(const std::string & full_type);
 
 }  // namespace rosbag2
 
--- a/rosbag2/include/rosbag2/typesupport_helpers.hpp
+++ b/rosbag2/include/rosbag2/typesupport_helpers.hpp
@@ -29,9 +29,13 @@ ROSBAG2_PUBLIC
 const rosidl_message_type_support_t *
 get_typesupport(const std::string & type, const std::string & typesupport_identifier);
 
+[[deprecated("use extract_type_identifier(const std::string & full_type) instead")]]
+ROSBAG2_PUBLIC
+const std::pair<std::string, std::string> extract_type_and_package(const std::string & full_type);
+
 ROSBAG2_PUBLIC
 const std::tuple<std::string, std::string, std::string>
-extract_type_and_package(const std::string & full_type);
+extract_type_identifier(const std::string & full_type);
 
 }  // namespace rosbag2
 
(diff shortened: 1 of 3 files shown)


Karsten1987:
Thanks for the patch. I think that probably definitely has to be addressed.

However, when running `colcon test --packages-select rosbag2` on your branch, I get a lot of test failures. Could you verify that the tests all run successfully on your end?

Karsten1987:
I just opened https://github.com/davidhodo/rosbag2/pull/1 to address the concerns of failing tests.

I am not sure however how easily we can backport this to dashing, because as is it's currently breaking API.

davidhodo:
Thanks for those fixes.  It works fine for me with your changes.  I merged the pull request.  Understood on the API change.  It would be nice to get it in the next patch release since we're working with some outsides groups that also need the fixes.  I'll see if I can figure out a way to do it without changing the API, but I'm not sure it's possible.

Karsten1987:
Running a first round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7335/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3500/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6011/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7156/)


I think in order to remain API compatible we should maintain the function `extract_type_and_package` as is, but deprecate this function. Then add a new function e.g. `extract_type_identifier` (we can come up with a better function name), which extracts the type correctly with a middle module. We can then further decide that the first function calls the new functions and simply discards the middle module in order to avoid code duplication.
Does this make sense?

In order to bring this to a patch release, we should ideally merge this PR today as we ship the first patch release tomorrow.

davidhodo:
It does.  I'll take a stab at it shortly.

davidhodo:
See what you think of those changes.  On OSX I had to build with colcon build --cmake-args -DCMAKE_CXX_FLAGS="-Wno-error=deprecated-declarations" to prevent the deprecated functions calls in the tests from showing up as errors.  I assume we should keep tests in there for the legacy versions until they are removed.

Karsten1987:
This is great. I like it. I took the liberty to push on your branch to a small patch which adds a pragma to suppress the deprecation warning in order to trigger another CI build.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7337/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3502/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6014/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7158/)

davidhodo:
Not a problem.  Thanks for adding that.  I wasn't sure the best way to fix it.  Also, I noticed you added a default value of msg if there are only two parts.  Do you know of a case where that is possible or are you just being cautious?  I'm trying to fix a [similar issue](https://github.com/ros2/ros2cli/pull/278) in ros2cli and am assuming that types always have 3 parts.

Karsten1987:
Not sure I completely understood. Whenever trying to access the typesupport only by specifying e.g. `test_msgs/BasicTypes` the middle element will be empty. In order to get the correct symbol from the library, it makes sense to default to `msg` then, because before dashing that was the case.

davidhodo:
Got it.  Thanks.

PR #130, opened 2019-06-11
[DO NOT MERGE] Use ROS isolated version of tests

pbaughman:
This is my suggested fix for rosbag2 tests failing when run in parallel with other tests.

This PR relies on
  * ~~https://github.com/ament/ament_cmake/pull/177~~ Merged
  * ~~https://github.com/ros2/ament_cmake_ros/pull/4~~ Merged

Using the 'ros_isolated' version of ament_add_gmock means that the tests are given a ROS_DOMAIN_ID before they're run so they won't see ROS traffic from other tests.

I've checked this by running the rosbag2 tests in a loop while I run
>colcon test --packages-up-to rcl rclcpp rclpy
In another terminal window.  Prior to this change, I would see test failures every time.

I didn't change all of the tests to use the ros_isolated version - only the ones that were failing because of interference.

### DIFF
--- a/rosbag2_tests/CMakeLists.txt
+++ b/rosbag2_tests/CMakeLists.txt
@@ -25,6 +25,7 @@ if(BUILD_TESTING)
   skip_ros1_tests_if_necessary()
 
   find_package(ament_cmake_gmock REQUIRED)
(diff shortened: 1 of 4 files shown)


pbaughman:
The upstream API changed a little bit after this PR was opened.  Closing until it's fixed

PR #113, opened 2019-05-07
Export pluginlib to downstream packages

esteve:
I'm not entirely sure if pluginlib should be exported, but with this change and https://github.com/ros/pluginlib/pull/154 I was able to crosscompile rosbag2 for aarch64.

PR #132, opened 2019-06-13
Export ament_index and class_loader dependencies

kurcha01-arm:
Fixes missing exported dependencies:
 - ament_index_cpp and class_loader for rosbag2
 - ament_index_cpp for rosbag2_storage
 - class_loader for rosbag2_storage_default_plugins

Signed-off-by: Kurtis Charnock <kurtis.charnock@arm.com>

### DIFF
--- a/rosbag2/CMakeLists.txt
+++ b/rosbag2/CMakeLists.txt
@@ -73,7 +73,11 @@ install(
 
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
-ament_export_dependencies(rosbag2_storage rosidl_typesupport_introspection_cpp)
+ament_export_dependencies(
+  ament_index_cpp
+  class_loader
+  rosbag2_storage
+  rosidl_typesupport_introspection_cpp)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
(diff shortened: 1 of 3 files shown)


kurcha01-arm:
Hello,

[ro2/cross_compile](https://github.com/ros2/cross_compile) is failing because the linker can't find these packages. find_package() finds them when building natively but using ld's --sysroot option with a ros2 install location outside the sysroot prevents them from being found. This also stops any workarounds with options like rpath, since the sysroot path gets prefixed to any rpath specified.

By adding the packages as dependencies, find_package doesn't need to search the filesystem, fixing the problem.

See also [rmw_fastrtps/pull/294](https://github.com/ros2/rmw_fastrtps/pull/294)

Karsten1987:
thanks for the patch. Also is this related to #113? @esteve Does this solve your problem here as well?

esteve:
@Karsten1987 I haven't tried this PR yet, but with https://github.com/ros/pluginlib/pull/154 and https://github.com/ros2/rosbag2/pull/113 I was able to cross compile rosbag2 for aarch64.

However, the changes in this PR are a bit too invasive IMHO, exporting only `pluginlib` was enough to support crosscompiling, not `class_loader`. Also, `ament_index_cpp` shouldn't be exported at all, since it's part of the build phase in this package, not for downstream packages.

kurcha01-arm:
Swapping this patch for #113 fixed the problem for me. IMO we should abandon this and go with @esteve 's

Thanks!

Karsten1987:
With #113 merged, I am going ahead and close this. Feel free to re-open if you concerns were not addressed.

PR #133, opened 2019-06-14
Use ROS isolated version of tests

pbaughman:
This is my suggested fix for rosbag2 tests failing when run in parallel with other tests.

I've checked this by running the rosbag2 tests in a loop while I run

colcon test --packages-up-to rcl rclcpp rclpy
In another terminal window. Prior to this change, I would see test failures every time.

I didn't change all of the tests to use the ros_isolated version - only the ones that were failing because of interference.


pbaughman:
`test_record` and `test_rosbag2_node` appear to have the same problem.  Should I just set all of these to use the isolated version out of an abundance of caution?

PR #135, opened 2019-06-25
Fix test failures on armhf

prajakta-gokhale:
Fix rosbag2_transport unit test failures on armhf.

Related to armhf tier-2 support fixes: https://github.com/ros2/ros2/issues/721
Fixes unit test failure reported [here](https://ci.ros2.org/job/ci_linux-armhf/16/testReport/rosbag2_transport/FormatterTestFixture/format_file_size_returns_correct_format/).

Signed-off-by: Prajakta Gokhale <prajaktg@amazon.com>

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/formatter.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/formatter.cpp
@@ -90,7 +90,7 @@ std::string Formatter::format_time_point(
          " (" + formatted_duration["time_in_sec"] + ")";
 }
 
-std::string Formatter::format_file_size(size_t file_size)
+std::string Formatter::format_file_size(uint64_t file_size)
 {
   double size = static_cast<double>(file_size);
   static const char * units[] = {"B", "KiB", "MiB", "GiB", "TiB"};
--- a/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
@@ -33,13 +33,13 @@ struct TopicInformation
 
 struct BagMetadata
 {
-  int version = 1;  // upgrade this number when changing the content of the struct
-  size_t bag_size = 0;  // Will not be serialized
+  int version = 2;  // upgrade this number when changing the content of the struct
+  uint64_t bag_size = 0;  // Will not be serialized
   std::string storage_identifier;
   std::vector<std::string> relative_file_paths;
   std::chrono::nanoseconds duration;
   std::chrono::time_point<std::chrono::high_resolution_clock> starting_time;
-  size_t message_count;
+  uint64_t message_count;
   std::vector<TopicInformation> topics_with_message_count;
 };
 
--- a/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
+++ b/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
@@ -72,7 +72,7 @@ struct convert<rosbag2_storage::TopicInformation>
   static bool decode(const Node & node, rosbag2_storage::TopicInformation & metadata)
   {
     metadata.topic_metadata = node["topic_metadata"].as<rosbag2_storage::TopicMetadata>();
-    metadata.message_count = node["message_count"].as<size_t>();
+    metadata.message_count = node["message_count"].as<uint64_t>();
     return true;
   }
 };
@@ -89,7 +89,7 @@ struct convert<std::chrono::nanoseconds>
 
   static bool decode(const Node & node, std::chrono::nanoseconds & time_in_ns)
   {
-    time_in_ns = std::chrono::nanoseconds(node["nanoseconds"].as<size_t>());
+    time_in_ns = std::chrono::nanoseconds(node["nanoseconds"].as<uint64_t>());
     return true;
   }
 };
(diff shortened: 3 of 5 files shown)


prajakta-gokhale:
Gist: https://gist.githubusercontent.com/prajakta-gokhale/211bac73533a4576537436f416baea2a/raw/d03c798754795f5e307bdf4d7922654491587bca/ros2.repos
Job: ci_launcher

clalancette:
> I wonder if we should just disable this test for 32 bit platform. It doesn't feel correct to say that these maxed out values are actually a tebi/pebibyte in the first place. Thoughts?

I agree with that part, it is very odd to say this.

But actually, this is problematic in other ways.  First of all, this should have nothing to do with address space size; this is formatting for a file size, and there is no reason you can't have a 1 TiB file on 32-bit armhf.

The real problem looks to be in the API.  It is using a `size_t` for the number passed into [format_file_size](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/formatter.hpp#L39), which is only 32-bit on armhf, so it can only represent up to 4GiB.  I think the correct fix here is to switch to a uint64_t for the format_file_size API, which means that all of this will work on both 32-bit and 64-bit platforms.

prajakta-gokhale:
> The real problem looks to be in the API. It is using a `size_t` for the number passed into [format_file_size](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/formatter.hpp#L39), which is only 32-bit on armhf, so it can only represent up to 4GiB. I think the correct fix here is to switch to a uint64_t for the format_file_size API, which means that all of this will work on both 32-bit and 64-bit platforms.

I didn't realize that, thanks for pointing it out! Will change it in the next commit.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7435/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3580/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6097/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7266/)
* Linux-armhf [Build Status](https://ci.ros2.org/job/ci_linux-armhf/18/)

prajakta-gokhale:
@clalancette please take a look agian, thanks!

rosbag2_transport/src/rosbag2_transport/formatter.cpp
```diff
@@ -90,7 +90,7 @@ std::string Formatter::format_time_point(
          " (" + formatted_duration["time_in_sec"] + ")";
 }
 
-std::string Formatter::format_file_size(size_t file_size)
+std::string Formatter::format_file_size(uint64_t file_size)
```
clalancette: This seems better to me.

However, we have a further problem higher in the stack.  The only caller of this method in real code is [formatter.cpp](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/formatter.cpp#L42), where it passes a `BagMetadata.bag_size` into the function.  [BagMetadata](https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L34) defines the bag_size as a `size_t`, which is also incorrect.  The real issue is that we are now breaking on-disk serialized code, but using a `size_t` in here was always a bug; it will be a different size on different platforms.  My suggestion would be to bump the version to "2", change the bag_size to a `uint64_t`, and also change the message_count to a `uint64_t`; then it will be more cross-platform.  However, that is a big change to the whole ecosystem, so I'd like to get a further opinion from @Karsten1987 before moving ahead with that.

prajakta-gokhale: Yes, I was looking at [metadata_io.cpp](https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp#L111) which is similarly serializing `metadata.yaml` test resource files incorrectly into `size_t` instead of `uint64_t`. That is causing test failures in [rosbag2_tests](https://ci.ros2.org/job/ci_linux-armhf/17/testReport/junit/rosbag2_tests/).

I have a similar fix for it that uses the correct variable type, but I'd like for us to decide on an approach before I open that PR.

prajakta-gokhale:
@Karsten1987 could you please review these changes and let us know what you think about changing member variable types from `size_t` to `uint64_t`?

Karsten1987:
I agree, having `size_t` here is indeed prone to error. Therefore, I agree with @clalancette for the proposed changes. I am still a bit undecided whether we have to bump the version when it's obviously a wrong behavior.

clalancette:
> I am still a bit undecided whether we have to bump the version when it's obviously a wrong behavior.

So for the `bag_size` itself, there is actually a comment there saying that it is not serialized.  If we just change that, we may not have to bump the version.  But the fact that `message_count` is also a size_t, and *is* serialized, means (to me) that we have to bump the version.  There probably aren't a ton of bag files out there now, but having a different size without bumping the version may lead to difficult-to-debug bugs.

prajakta-gokhale:
> There probably aren't a ton of bag files out there now, but having a different size without bumping the version may lead to difficult-to-debug bugs.

I agree with that reasoning. For the sake of fixing unit test failures in the nightlies, we could merge this PR as is and create a new one to change variable types (and bump up the version?)

clalancette:
> I agree with that reasoning. For the sake of fixing unit test failures in the nightlies, we could merge this PR as is and create a new one to change variable types (and bump up the version?)

I'd prefer not to do that; it will be too easy to forget about this problem once it is "hidden" with this PR merged.

@prajakta-gokhale Please make the changes to BagMetadata switching the size_t -> uint64_t, and bumping the version.  @Karsten1987 can then review, and if he still disagrees with the version bump it's just a one-line change.

prajakta-gokhale:
@Karsten1987 I have changed the variable types and bumped up `BagMetadata::version` in the latest commit. Please take a look.

emersonknapp:
Lets also check that the armhf tests are fixed :) [Build Status](https://ci.ros2.org/job/ci_linux-armhf/54/)

emersonknapp:
@Karsten1987 we're happy with this - if you are too let's get this merged - I think this is the final PR for `ci_linux-armhf` to be fully green.

Issue #136, opened 2019-06-25
Build fails if you don't have fastrtps


ivanpauno:
The building of `rosbag2_converter_default_plugins` is failing not quietly when you don't have `rmw_fastrtps_cpp`. I think that's not ok, you should be able to build ros2 without it (e.g.: only with opensplice or connext).

Failing CI builds with only opensplice (from [comment](https://github.com/ros2/rmw_opensplice/issues/264#issuecomment-505181300)):

* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=7421)](http://ci.ros2.org/job/ci_linux/7421/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=3572)](http://ci.ros2.org/job/ci_linux-aarch64/3572/)
* macOS [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=6089)](http://ci.ros2.org/job/ci_osx/6089/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=7244)](http://ci.ros2.org/job/ci_windows/7244/)

Further info:

The problem is here:
https://github.com/ros2/rosbag2/blob/2ad74a4d7a81bcb82e084d5d26c60415c644766e/rosbag2_converter_default_plugins/CMakeLists.txt#L53-L57

The build is only failing when `BUILD_TESTING=1`, but that's usual.

dirk-thomas:
@Karsten1987 can you please address this since it make building without FastRTPS more effort.

PR #137, opened 2019-06-25
disable plugins/tests which need rmw_fastrtps_cpp if unavailable

ivanpauno:
Fixes https://github.com/ros2/rosbag2/issues/136.

### DIFF
--- a/rosbag2_converter_default_plugins/CMakeLists.txt
+++ b/rosbag2_converter_default_plugins/CMakeLists.txt
@@ -24,44 +24,44 @@ find_package(rosbag2 REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rmw REQUIRED)
+find_package(rmw_fastrtps_cpp QUIET)
 
-add_library(${PROJECT_NAME} SHARED
-  src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp)
+if(rmw_fastrtps_cpp_FOUND)
+  add_library(${PROJECT_NAME} SHARED
+    src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp)
 
-ament_target_dependencies(${PROJECT_NAME}
-  ament_index_cpp
-  pluginlib
-  Poco
-  rosbag2
-  rcutils
-  rmw
-  rosidl_generator_cpp)
+  ament_target_dependencies(${PROJECT_NAME}
+    ament_index_cpp
+    pluginlib
+    Poco
+    rosbag2
+    rcutils
+    rmw
+    rosidl_generator_cpp)
 
-target_include_directories(${PROJECT_NAME}
-  PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>)
+  target_include_directories(${PROJECT_NAME}
+    PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>)
 
-pluginlib_export_plugin_description_file(rosbag2 plugin_description.xml)
+  pluginlib_export_plugin_description_file(rosbag2 plugin_description.xml)
 
-install(
-  TARGETS ${PROJECT_NAME}
-  ARCHIVE DESTINATION lib
-  LIBRARY DESTINATION lib
-  RUNTIME DESTINATION bin)
+  install(
+    TARGETS ${PROJECT_NAME}
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin)
 
-if(BUILD_TESTING)
-  find_package(ament_cmake_gmock REQUIRED)
-  find_package(ament_lint_auto REQUIRED)
-  find_package(rcutils REQUIRED)
-  find_package(rmw_fastrtps_cpp QUIET)
-  find_package(rosbag2 REQUIRED)
-  find_package(rosbag2_test_common REQUIRED)
-  find_package(test_msgs REQUIRED)
+  if(BUILD_TESTING)
+    find_package(ament_cmake_gmock REQUIRED)
+    find_package(ament_lint_auto REQUIRED)
+    find_package(rcutils REQUIRED)
+    find_package(rosbag2 REQUIRED)
+    find_package(rosbag2_test_common REQUIRED)
+    find_package(test_msgs REQUIRED)
 
-  ament_lint_auto_find_test_dependencies()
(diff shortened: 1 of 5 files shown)


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7427/)

ivanpauno:
I think this is ready to merge

Karsten1987:
I don't think that's the right approach.
We dynamically, but hardcoded, load `rmw_fastrtps_cpp` inside the code: https://github.com/ros2/rosbag2/blob/master/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp#L63

---
I believe we should always find `rmw_fastrtps_cpp` and don't compile the library at all if fastrtps is not available. With this patch, we currently build the library correctly, but it fails on runtime due to the missing libraries. That also explains the set of test failures (the segfaults) of the CI job you've linked.
When fast-rtps is not available, we should further disable all tests which rely on the converter plugins, as they are the only source requiring fastrtps.

ivanpauno:
> I don't think that's the right approach.
> We dynamically, but hardcoded, load `rmw_fastrtps_cpp` inside the code: https://github.com/ros2/rosbag2/blob/master/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp#L63
> 
> I believe we should always find `rmw_fastrtps_cpp` and don't compile the library at all if fastrtps is not available. With this patch, we currently build the library correctly, but it fails on runtime due to the missing libraries. That also explains the set of test failures (the segfaults) of the CI job you've linked.
> When fast-rtps is not available, we should further disable all tests which rely on the converter plugins, as they are the only source requiring fastrtps.

Ok, I will check what has to be disabled an update the PR.

ivanpauno:
@Karsten1987 In https://github.com/ros2/rosbag2/pull/137/commits/bb38b7a1b65771b9cf14f51bcb1590203bac36b1, I completely disabled the build of `rosbag2_converter_default_plugins` if you don't have `fastrtps`. I also disabled the three tests in `rosbag2_tests` which I think that they depend at runtime in the former package.

I also saw in the previous CI that `test_rosbag2_node` and `test_play` were failing, but I'm not sure if they also depend on the converter plugin, or if it was a random error related to using `opensplice`.
Can you check please?
Thanks!

ivanpauno:
@Karsten1987 Friendly ping

Karsten1987:
Sorry @ivanpauno I am currently pretty underwater and won't have time to deep dive into it for the next two weeks. 

It makes sense to me to completely disable the converter plugin tests when fast rtps is not available. I could further imagine (without having precisely looked at the code) that some end-to-end tests might have to be disabled or compiled differently when the converters are not present in the current setup.

ivanpauno:
> Sorry @ivanpauno I am currently pretty underwater and won't have time to deep dive into it for the next two weeks.
> 
> It makes sense to me to completely disable the converter plugin tests when fast rtps is not available. I could further imagine (without having precisely looked at the code) that some end-to-end tests might have to be disabled or compiled differently when the converters are not present in the current setup.

No problem, I will wait then. Thanks for the update!

rosbag2_converter_default_plugins/package.xml
```diff
@@ -14,8 +14,10 @@
   <depend>rosidl_generator_cpp</depend>
   <depend>rcutils</depend>
   <depend>rmw</depend>
+  <depend>rmw_fastrtps_cpp</depend>
   <depend>rosbag2</depend>
 
+
```
Karsten1987: superfluous new line

rosbag2_converter_default_plugins/CMakeLists.txt
```diff
@@ -24,55 +24,57 @@ find_package(rosbag2 REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rmw REQUIRED)
+find_package(rmw_fastrtps_cpp QUIET)
 
-add_library(${PROJECT_NAME} SHARED
-  src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp)
+if(rmw_fastrtps_cpp_FOUND)
```
Karsten1987: I would somehow inverse this logic and return `if(NOT rmw_fastrtps_cpp_FOUND)` with at least a notification that this library is not going to be compiled in full. This will also ease the diff a little bit as we don't have an extra layer of indentation.

rosbag2_converter_default_plugins/package.xml
dirk-thomas: Please revert unrelated (whitespace) changes.

dirk-thomas:
Please run CI for this change with a non-FastRTPS RMW impl. If the build doesn't fail (which would be a big step forward) we will get this merged in the current state.

dirk-thomas:
> I believe the right change here is to make the plugins work with whatever RMW implementation available and try to avoid a hardcoded version of rmw_fastrtps_cpp.

That might be a good goal but is outside of the scope of this PR.

ivanpauno:
I've addressed comments and rebased with master.

CI (only opensplice, up to rosbag2 related packages):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7747/)

ivanpauno:
Again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7753/)

ivanpauno:
> Again:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/7753/)

It builds correctly now (only opensplice). I solved the linter failures.
There are two other build failures, but they shouldn't happen if fastrtps is available (they don't appear locally, I have the three rmw implementations).

Double checking (fastrtps and connext, same packages as before):
* Linux [![Build Status]* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7756/)

ivanpauno:
> Double checking (fastrtps and connext, same packages as before):
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/7756/)

Again, because I chose opensplice and connext by error :man_facepalming: :
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7758/)

rosbag2_transport/package.xml
```diff
@@ -14,6 +14,7 @@
   <depend>rmw</depend>
   <depend>shared_queues_vendor</depend>
 
+  <test_depend>ament_index_cpp</test_depend>
```
Karsten1987: nitpick: Alphabetical order

dirk-thomas:
@Karsten1987 can you please make sure that this gets backported into Dashing in time for the upcoming patch release.

dirk-thomas:
> can you please make sure that this gets backported into Dashing in time for the upcoming patch release.

Once the fix has been merged to the `dashing` branch the following job should start to pass (or at least become unstable): http://build.ros2.org/view/Dci/job/Dci__nightly-cyclonedds_ubuntu_bionic_amd64/

PR #146, opened 2019-08-09
[sqlite3_vendor] install package.xml

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

Otherwise rosdep cannot find it and install it from the debs instead of using the local version: https://github.com/osrf/docker_images/issues/304


Karsten1987:
thanks for the patch.

PR #176, opened 2019-10-11
Splitting/Reader changes

piraka9011:
This contains the changes necessary to read from multiple bag files as part of the rosbag2 splitting feature.

These changes do not affect the previous functionality of ros2bag play. 

TODO:
* [ ] Unit tests


rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -79,6 +80,23 @@ class ROSBAG2_PUBLIC SequentialReader
...
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file();
+
+  /**
+   * Update the current database file to read from and remove the first (front)
+   * element in the list of files to read from.
+   *
+   * Expected usage:
+   * if (has_next_file()) get_next_file();
```
thomas-moulard: please document what will happen if has_next_file() is false and this function is called.

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -104,6 +122,10 @@ class ROSBAG2_PUBLIC SequentialReader
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  std::string storage_id_;
```
thomas-moulard: what does this field mean? this needs a comment

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -104,6 +122,10 @@ class ROSBAG2_PUBLIC SequentialReader
...
+  std::string file_uri_;
```
thomas-moulard: ditto for the other fields

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,11 +41,20 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
-  storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
+  file_uri_ = storage_options.uri;
```
thomas-moulard: instead of copying this one by one, can't we just keep a local copy of storage_options? This will scale better as we add more options.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,11 +41,20 @@ void
...
+  storage_id_ = storage_options.storage_id;
+  storage_ = storage_factory_->open_read_only(file_uri_, storage_id_);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-  auto topics = storage_->get_metadata().topics_with_message_count;
+
+  BagMetadata bag_metadata = storage_->get_metadata();
```
thomas-moulard:
```suggestion
  const auto bag_metadata = storage_->get_metadata();
```

1/ if it can be const, make it const
2/ prefer auto to using types directly: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-auto

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -79,6 +80,23 @@ class ROSBAG2_PUBLIC SequentialReader
...
+  virtual bool has_next_file();
+
+  /**
+   * Update the current database file to read from and remove the first (front)
+   * element in the list of files to read from.
+   *
+   * Expected usage:
+   * if (has_next_file()) get_next_file();
+   */
+  void get_next_file();
```
thomas-moulard: this should probably return some kind of status as it may fail

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,11 +41,20 @@ void
...
+  // Get all the files that need to be read from
+  file_paths_ = bag_metadata.relative_file_paths;
+  get_next_file();
```
thomas-moulard: this sounds unsafe, what if we have no files at all in the metadata? we should check whether or not this operation succeeded.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +80,29 @@ SequentialReader::open(
   }
 }
 
+void SequentialReader::get_next_file()
+{
+  current_file_ = file_paths_.front();
+  file_paths_.erase(file_paths_.begin());
```
thomas-moulard: Dropping information there seems unnecessary, why not just replacing current_file_ by current_file_index_?

Also, the reason why std::vector does not have pop_front() is because it is inefficient, std::dequeue is probably closer to what we'd like if we really want to erase here.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +80,29 @@ SequentialReader::open(
...
+}
+
+bool SequentialReader::has_next_file()
```
thomas-moulard:
```suggestion
bool SequentialReader::has_next_file() const
```

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -79,6 +80,23 @@ class ROSBAG2_PUBLIC SequentialReader
    */
   virtual bool has_next();
 
+    /**
+   * Ask whether there is another database file to read from the list of relative
+   * file paths.
+   *
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file();
```
thomas-moulard:
```suggestion
  virtual bool has_next_file() const;
```

thomas-moulard: why virtual?

piraka9011:
A new base `Reader` has been made to be extended with other features.
The base `Reader` class only reads from one file.
The `SequentialReader` class can read from multiple files.

The purpose of this refactor is to extend the `Reader` with extra features in the future such as compression.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * \return true if there is a file to read from. false otherwise.
+   */
+  bool open_read_only(const std::string & file, const std::string & storage_id);
+
+  rosbag2_storage::BagMetadata get_metadata();
+
+private:
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
```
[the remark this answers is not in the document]
piraka9011: @Karsten1987 might have more insight into this. 
If I understood this well, there can only be one storage object open at a time to avoid multiple objects reading from a database.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "rosbag2/types.hpp"
+#include "rosbag2/visibility_control.hpp"
+
+// This is necessary because of using stl types here. It is completely safe, because
+// a) the member is not accessible from the outside
+// b) there are no inline functions.
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
```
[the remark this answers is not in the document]
piraka9011: Artifact from the previous reader, can refactor as well.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "rosbag2_storage/storage_factory.hpp"
+#include "rosbag2_storage/storage_factory_interface.hpp"
+#include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
+#include "rosbag2/converter.hpp"
+#include "rosbag2/logging.hpp"
+#include "rosbag2/serialization_format_converter_factory.hpp"
+#include "rosbag2/serialization_format_converter_factory_interface.hpp"
+#include "rosbag2/storage_options.hpp"
+#include "rosbag2/types.hpp"
+#include "rosbag2/visibility_control.hpp"
```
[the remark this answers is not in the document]
piraka9011: logging is for the `ROSBAG2_LOGGING_INFO/_STREAM()` logging statements. Can x2 check to see which aren't being used.

rosbag2/src/rosbag2/reader.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  ROSBAG2_LOG_INFO_STREAM("Reader opened.");
+}
+
+bool Reader::has_next()
+{
+  if (storage_) {
+    return storage_->has_next();
+  }
+  throw std::runtime_error("(has_next) Bag is not open. Call open() before reading.");
```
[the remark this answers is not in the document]
piraka9011: Left from a debug... 😮 removing

rosbag2/src/rosbag2/reader.cpp
```diff
@@ -0,0 +1,123 @@
...
+      storage_serialization_format,
+      converter_options.output_serialization_format,
+      converter_factory_);
+    auto topics = storage_->get_all_topics_and_types();
+    for (const auto & topic_with_type : topics) {
+      converter_->add_topic(topic_with_type.name, topic_with_type.type);
+    }
+  }
+
+  ROSBAG2_LOG_INFO_STREAM("Reader opened.");
```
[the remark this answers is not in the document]
piraka9011: Will move these to the DEBUG level then.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,133 @@
...
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2
+{
+/**
+ * The Reader allows opening and reading messages of a bag. Messages will be read
+ * sequentially according to timestamp.
+ */
+class ROSBAG2_PUBLIC Reader
```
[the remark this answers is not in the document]
piraka9011: Some methods need to be overridden by various tests in the other rosbag2 packages so that they can be mocked.

rosbag2/include/rosbag2/reader.hpp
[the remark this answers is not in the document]
thomas-moulard: I had more time to read the code tonight and it seems that we should change the factories to return `std::unique_ptr`, but this is for another PR...

Rationale: https://stackoverflow.com/questions/37884728/does-c11-unique-ptr-and-shared-ptr-able-to-convert-to-each-others-type

> std::unique_ptr is the C++11 way to express exclusive ownership, but one of its most attractive features is that it easily and efficiently converts to a std::shared_ptr.
> 
> This is a key part of why std::unique_ptr is so well suited as a factory function return type. Factory functions can’t know whether callers will want to use exclusive ownership semantics for the object they return or whether shared ownership (i.e., std::shared_ptr) would be more appropriate. By returning a std::unique_ptr, factories provide callers with the most efficient smart pointer, but they don’t hinder callers from replacing it with its more flexible sibling.
> 
> std::shared_ptr to std::unique_ptr is not allowed. Once you’ve turned lifetime management of a resource over to a std::shared_ptr, there’s no changing your mind. Even if the reference count is one, you can’t reclaim ownership of the resource in order to, say, have a std::unique_ptr manage it.
> 
> Reference: Effective Modern C++. 42 SPECIFIC WAYS TO IMPROVE YOUR USE OF C++11 AND C++14. Scott Meyers.

tl;dr it is usually better to have factories return unique_ptr, if your user wants to convert to a shared_ptr, it can always happen transparently.

rosbag2/include/rosbag2/reader.hpp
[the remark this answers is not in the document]
thomas-moulard: Yep, my point exactly, you don't use this function in the header, so this should be moved to the source file.

The rationale is that if you include a header here, any other file including this header will now slow down due to the indirect inclusion. If you include headers in cpp files, you will only make the compilation of this particular file slower. Following this rule is important to make large software compile in a reasonable time frame. I read somewhere that on average, when compiling a single C++ file, 10% of the compliation time is parsing/evaluating STL headers (!).

rosbag2/src/rosbag2/reader.cpp
```diff
@@ -0,0 +1,123 @@
...
+{}
+
+Reader::~Reader()
+{
+  storage_.reset();  // Necessary to ensure that the storage is destroyed before the factory
+}
+
+void Reader::reset()
+{
+  storage_.reset();
```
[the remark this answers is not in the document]
piraka9011: Already taken care of in [destructor](https://github.com/ros2/rosbag2/blob/dc2b161f7f019d595edcac926f227a088ea175e4/rosbag2/src/rosbag2/converter.cpp#L58).

`reset()` is exposed here for the `SequentialReader`

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,133 @@
...
+  bool has_storage();
+
+  /**
+   * Open a new database file to read from.
+   *
+   * \return true if there is a file to read from. false otherwise.
+   */
+  bool open_read_only(const std::string & file, const std::string & storage_id);
+
+  rosbag2_storage::BagMetadata get_metadata();
```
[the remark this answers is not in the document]
piraka9011: No. Compiler might optimize it though.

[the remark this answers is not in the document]
piraka9011: Changing it avoid copying touches code in another package. Can be directly optimized in another PR.

rosbag2/src/rosbag2/reader.hpp
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Open a new database file to read from.
+   *
+   * \return true if there is a file to read from. false otherwise.
+   */
+  bool open_read_only(const std::string & file, const std::string & storage_id);
+
+  /**
+   * Gets a copy of the metadata file read from the storage factory.
+   * TODO(piraka9011): Do we want to optimize to avoid copying?
```
thomas-moulard: remove from doxygen doc as it's not describing the function

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -51,65 +35,80 @@ class ROSBAG2_PUBLIC SequentialReader
     std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
     std::make_unique<rosbag2_storage::StorageFactory>(),
     std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
-    std::make_shared<SerializationFormatConverterFactory>());
+    std::make_shared<SerializationFormatConverterFactory>()) - ;
```
thomas-moulard: remove?

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -51,65 +35,80 @@ class ROSBAG2_PUBLIC SequentialReader
...
+     * automatically closed on destruction.
+     *
+     * If the `output_serialization_format` within the `converter_options` is not the same as the
+     * format of the underlying stored data, a converter will be used to automatically convert the
+     * data to the specified output format.
+     * Throws if the converter plugin does not exist.
+     *
+     * \param storage_options Options to configure the storage
+     * \param converter_options Options for specifying the output data format
+     */
```
thomas-moulard: let's avoid comestic changes or let's move that to a separate pr

piraka9011: IDE autoformat.

thomas-moulard: You can use `git checkout -p` in this case.

piraka9011:
@Karsten1987 Full scope of reader changes for reading split bags.

Karsten1987:
@piraka9011 do you mind preparing https://github.com/ros2/rosbag2/pull/181 in a way that it's somewhat complete in the feature it's addressing? It's somehow difficult to review a PR by cross referencing another PR.

piraka9011:
> do you mind preparing #181 in a way that it's somewhat complete in the feature it's addressing? It's somehow difficult to review a PR by cross referencing another PR.

We can do that, I'll start working on adding the complete feature in that PR (#181). You can also use this as a reference for any changes we introduce to the `Reader`.

Karsten1987:
Reading through this I am not completely clear I get the benefit of introducing this `Reader` class. The way I imagine a `SequentialReader` to work with multiple bag files is actually relatively straight forward:
The SequentialReader should know about all individual bagfiles to open from the metadata. Then simply open the first one until it's reaching the end, close the connection, open up a next connection until all bag files are played. We can make it a contract within the metadata file that the filenames have to be ordered in a chronological order.

> Add a field to StorageOptions called file_offset, to indicate that the SequentialReader should read from file file_offset (from 2nd file, 3rd file, etc.) present in the metadata file.
Add a MultiFileSequentialReader, which provides the same interface, except that it contains the logic to actually iterate over all the file. No new interface is added, has_next_message will just automatically open the next db file as needed. This class should act as a decorator / pimpl and just forward the calls to the underlying SequentialReader unless we are at a file limit.

I don't see what's the difference between a MultiFileSequentialReader with exactly one file and a SequentialReader. That being said, I believe a SequentialReader should be able to iterate over multiple files transparently. 
I could image a MultiFileReader if the files specified in the metadata are not strictly chronologically ordered. For example, I could image joining two existing bag files which should be fused and replayed jointly.

thomas-moulard:
While this approach is possible, the objective of having a separate class is to simplify testing and keep SequentialReader focused on one thing (reading a file, vs. reading a directory), as per the Single-Responsibility principle. We also found that decoupling Reader from SequentialReader improves the testability of the multi-file logic. Instead of having to actually read things (or mock part of the class out), we can inject a MockSequentialReader in the reader.

piraka9011:
Closing in favor of #187

PR #181, opened 2019-10-17
Splitting: Introduce new Reader

piraka9011:
Create a new `Reader` class that has the previous functionality of
`SequentialReader`.
`Reader` will be used to compose other `Reader`s with different
features such as reading from multiple bags.

This is an effort to rework draft PR #176 into multiple smaller PRs.

### DIFF
--- a/rosbag2/include/rosbag2/reader.hpp
+++ b/rosbag2/include/rosbag2/reader.hpp
@@ -1,0 +1,162 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2__READER_HPP_
+#define ROSBAG2__READER_HPP_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "rosbag2_storage/storage_factory.hpp"
+#include "rosbag2_storage/storage_factory_interface.hpp"
+#include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
+#include "rosbag2/converter.hpp"
+#include "rosbag2/serialization_format_converter_factory.hpp"
+#include "rosbag2/serialization_format_converter_factory_interface.hpp"
+#include "rosbag2/storage_options.hpp"
+#include "rosbag2/types.hpp"
+#include "rosbag2/visibility_control.hpp"
+
+// This is necessary because of using stl types here. It is completely safe, because
+// a) the member is not accessible from the outside
+// b) there are no inline functions.
+// TODO(piraka9011): PIMPL for reader, exposing only sequential_reader API.
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2
+{
+/**
+ * The Reader allows opening and reading messages of a bag. Messages will be read sequentially
+ * according to the timestamp. This is not the same as a SequentialReader which reads both
+ * messages and multiple database files sequentially.
+ */
+  class ROSBAG2_PUBLIC Reader
+    {
+      public:
+      // TODO(piraka9011): Re-evaluate usage of shared_ptr in converter.
+      explicit Reader(
+      std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+      std::make_unique<rosbag2_storage::StorageFactory>(),
+      std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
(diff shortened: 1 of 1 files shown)


piraka9011:
@Karsten1987 requesting your review.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+{
+/**
+ * The Reader allows opening and reading messages of a bag. Messages will be read sequentially
+ * according to the timestamp. This is not the same as a SequentialReader which reads both
+ * messages and multiple database files sequentially.
+ */
+  class ROSBAG2_PUBLIC Reader
+    {
+      public:
+      // TODO(piraka9011): Re-evaluate usage of shared_ptr in converter.
```
dabonnie: File as an issue?

piraka9011: Per hallway discussion w/ @emersonknapp, these will be new user stories for our board.

Karsten1987:
No exactly sure I get the scope of this particular PR. Is that `.hpp` file served as an interface class? In which case, should it have pure virtual functions each individual reader (such as the `SequentialReader`) should implement? 
If not, shouldn't there be a `.cpp` file as well for it then?

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<SerializationFormatConverterFactory>());
+
+  virtual ~Reader();
+
+  /**
+   * Close the storage object. Calls the reset function of the storage implementation which
+   * should close database connections.
+   */
+  void reset();
```
Karsten1987: why is `reset()` not virtual? Can we ensure that a reset can be handled likewise for every potential reader implementation?

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<SerializationFormatConverterFactory>());
+
+  virtual ~Reader();
+
+  /**
+   * Close the storage object. Calls the reset function of the storage implementation which
+   * should close database connections.
```
Karsten1987: the reader instance should not know about the underlying storage type. So it could be a database, but also a ROS1 bagfile. just a little nitpick for the comment here.

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+   */
+  void reset();
+
+  /**
+   * Open a rosbag for reading messages sequentially (time-ordered). Throws if file could not be
```
Karsten1987: is that comment still valid? Are the messages always be read sequentially? If so, what's the difference to the `SequentialReader`?

rosbag2/include/rosbag2/reader.hpp
```diff
@@ -0,0 +1,162 @@
...
+  // TODO(piraka9011): Do we want to optimize to avoid copying?
+  rosbag2_storage::BagMetadata get_metadata() const;
+
+private:
+  /**
+   * Checks if all topics in the bagfile have the same RMW serialization format.
+   * Currently a bag file can only be played if all topics have the same serialization format.
+   *
+   * \param topics Vector of TopicInformation with metadata.
+   * \throws runtime_error if any topic has a different serialization format from the rest.
```
Karsten1987: I am not sure if I agree with that. It should definitely be possible for rosbags to store topics where the serialization format is not the same. So for example it's a valid use case to enhance a bag file with raw UDP captures from a sensor (i.e. velodyne) and store it with a separate serialization format. 
The requirement can be though that the serialization format doesn't change per topic.

piraka9011:
> No exactly sure I get the scope of this particular PR. 

The full scope can be found in draft PR #176.

> Is that `.hpp` file served as an interface class? In which case, should it have pure virtual functions each individual reader (such as the `SequentialReader`) should implement?

This PR is a small portion of the changes we are introducing to the `SequentialReader` in order to enable reading from multiple files. 

The idea is that we have a `Reader` that simply reads the next message in the database which is also used to compose other forms of readers (ex. a `SequentialReader` that will be used to read multiple storage files in sequence).

The goal is to also enable further extensions of the `Reader` for upcoming features such as compression.

Karsten1987:
If I interpret this correctly, a `SequentialReader` can be having multiple `Reader` instances to read from multiple bag files simultaneously? If so, why can't it just open multiple connection handles directly? Also, if `Reader` isn't an interface class, what's the difference between an instance of the `SequentialReader` with one internal `reader` compared to instantiating and using a `Reader` directly?

piraka9011:
> If I interpret this correctly, a `SequentialReader` can be having multiple `Reader` instances to read from multiple bag files simultaneously? 

The `SequentialReader` will be composed of a single `Reader` that is used to read messages from storage. The primary difference between the two is that the `SequentialReader` will have the ability to read from a split bag file in sequence.

> If `Reader` isn't an interface class, what's the difference between an instance of the `SequentialReader` with one internal `reader` compared to instantiating and using a `Reader` directly?

As before, `Reader` is for reading from a single file and a `SequentialReader` is for reading from multiple files sequentially. We're trying to avoid inheritance to minimize coupling and allow us to extend the functionality of reading from storage in the future.

PR #182, opened 2019-10-17
Add bagfile splitting support to storage_options and Writer

zmichaels11:
Refactored changes in PR #171 
This is part of an effort to rework PR #158 into multiple, smaller PRs.

### Changes
* Add `max_bagfile_size` to `storage_options`. This is used to indicate when a bagfile should be split. A default value of 0 indicates no splitting.
* Add function `should_split_bagfile` to `Writer`. This will be used by `Writer` to determine when a bagfile should be split.

### Issues
* ros-security/aws-roadmap#11

PR #158, opened 2019-09-12
Split recording bagfiles based on configurable size

prajakta-gokhale:
This change is to introduce a new CLI option in `ros2 bag record` to split a recording into multiple files based on the given `--bag-size` parameter. 

Related to #150 .

This does not include changes to README and tests for the features, those will be addressed in separate PRs.

PR #183, opened 2019-10-17
Add get_identifier to io-interfaces for support in bagfile splitting

zmichaels11:

### Changes
* Add `get_identifier` to `BaseIOInterface`. This will be used by `Writer` to generate the `BagMetadata` without requiring merging of the metadata returned by each sub-bagfile in the splitting process.

### Dependent PRs
* PR #182 

### Issues
* ros-security/aws-roadmap#11

### DIFF
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp
@@ -49,10 +49,10 @@ public:
   virtual uint64_t get_bagfile_size() const = 0;
 
   /**
-   * Returns the identifier for the plugin.
+   * Returns the identifier for the storage plugin.
    * \returns the identifier.
    */
-  virtual std::string get_identifier() const = 0;
+  virtual std::string get_storage_identifier() const = 0;
 };
 
 }  // namespace storage_interfaces
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp
@@ -47,6 +47,12 @@ public:
    * \returns the size of the bagfile in bytes.
    */
   virtual uint64_t get_bagfile_size() const = 0;
+
+  /**
+   * Returns the identifier for the plugin.
+   * \returns the identifier.
+   */
+  virtual std::string get_identifier() const = 0;
 };
 
 }  // namespace storage_interfaces
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -66,7 +66,7 @@ public:
 
   uint64_t get_bagfile_size() const override;
 
-  std::string get_identifier() const override;
+  std::string get_storage_identifier() const override;
 
 private:
   void initialize();
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -66,6 +66,8 @@ public:
 
   uint64_t get_bagfile_size() const override;
 
+  std::string get_identifier() const override;
+
 private:
   void initialize();
   void prepare_for_writing();
@@ -73,7 +75,7 @@ private:
   void fill_topics_and_types();
 
   std::unique_ptr<rosbag2_storage::BagMetadata> load_metadata(const std::string & uri);
-  bool database_exists(const std::string & uri);
+  static bool database_exists(const std::string & uri);
   bool is_read_only(const rosbag2_storage::storage_interfaces::IOFlag & io_flag) const;
 
   using ReadQueryResult = SqliteStatementWrapper::QueryResult<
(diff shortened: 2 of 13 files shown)


rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -66,14 +66,16 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
...
+
 private:
   void initialize();
   void prepare_for_writing();
   void prepare_for_reading();
   void fill_topics_and_types();
 
   std::unique_ptr<rosbag2_storage::BagMetadata> load_metadata(const std::string & uri);
-  bool database_exists(const std::string & uri);
+  static bool database_exists(const std::string & uri);
```
Karsten1987: I think this shouldn't have been a class member function in the first place, but that function should be a free function IMO.

zmichaels11: I was considering defining it as static in the cpp file, however I thought it made more sense to have it prototyped somewhere. If anyone has any other input, I'll keep it as is for now.

Karsten1987: it doesn't necessarily have to be a static function, can just be a free function. 
> I thought it made more sense to have it prototyped somewhere

What do you mean with this?

I just think that it makes not so much sense to have this function as a class member function.

zmichaels11: I wanted the definition before its use. I can define it as a free function inside `SqliteStorage.cpp`

zmichaels11: Just moved `database_exists` into a file-private free function inside `sqlite_storage.cpp`

zmichaels11:
@Karsten1987 this one's next.
Just rebased onto master after merge of PR #182.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -214,21 +223,20 @@ std::unique_ptr<rosbag2_storage::BagMetadata> SqliteStorage::load_metadata(const
   }
 }
 
-bool SqliteStorage::database_exists(const std::string & uri)
+bool SqliteStorage::is_read_only(const rosbag2_storage::storage_interfaces::IOFlag & io_flag) const
 {
-  std::ifstream database(uri);
-  return database.good();
+  return io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY;
 }
```
Karsten1987: How is that change related? Why introducing `is_read_only` here?

zmichaels11: That function already existed. I think the reason why the diff lists it as a change is because I moved `database_exists` to the top of the file since its now a free function.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp
```diff
@@ -47,6 +47,12 @@ class ROSBAG2_STORAGE_PUBLIC BaseIOInterface
    * \returns the size of the bagfile in bytes.
    */
   virtual uint64_t get_bagfile_size() const = 0;
+
+  /**
+   * Returns the identifier for the plugin.
+   * \returns the identifier.
+   */
+  virtual std::string get_identifier() const = 0;
```
Karsten1987: I'd propose to refine the function name. What do you think about `get_storage_identifier()`? I could see that we could have other things being identified in the future.

zmichaels11: The reason why I went with `get_identifier` vs `get_storage_identifier` was because I thought it was redundant since the call is on a storage object.

Karsten1987: yes, I understand that. My thinking though is that identifier or ID could be later used with, e.g. individual instances of the same storage identifier. I think there is no harm in the redundancy.

zmichaels11: The function actually is first defined in `BaseIOInterface`, should it instead be `get_io_identifier`? Or do you think `get_storage_identifier` would be a better choice?

Karsten1987: I don't think it'd be super clear to me what an `io` identifier is. Also, we used `storage identifier` also for the command line argument:
https://github.com/ros2/rosbag2/blob/master/ros2bag/ros2bag/verb/play.py#L30

zmichaels11: That's enough justification on my side to change it to `get_storage_identifier`.

zmichaels11:
@Karsten1987 applied suggestions and rebased onto master

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8505/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4389/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8414/)

PR #184, opened 2019-10-17
(API) Generate bagfile metadata in Writer

zmichaels11:
This is part of an effort to rework PR #158  into multiple, smaller PRs.

### Changes
* `Writer` can now create `BagMetadata`, this enables splitting of bagfiles without having to merge intermediate bagfile metadata.

### Dependent PRs
- [x] PR #182 
- [x] PR #183 

### Issues
* ros-security/aws-roadmap#11

### DIFF
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -109,7 +109,7 @@ private:
   uint64_t max_bagfile_size_;
 
   // Used to track topic -> message count
-  std::unordered_map<std::string, TopicInformation> topics_;
+  std::unordered_map<std::string, TopicInformation> topics_names_to_info_;
 
   rosbag2_storage::BagMetadata metadata_;
 
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -116,11 +116,11 @@ private:
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
 
-  // Record TopicInformation into metadata
-  void finalize_metadata();
-
   // Prepares the metadata by setting initial values.
   void init_metadata();
+
+  // Record TopicInformation into metadata
+  void finalize_metadata();
 };
 
 }  // namespace rosbag2
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -97,13 +97,6 @@ public:
    */
   virtual void write(std::shared_ptr<SerializedBagMessage> message);
 
-  /**
-   * Check if the current recording bagfile needs to be split and rolled over to new file.
-   *
-   * \return true if the bagfile should be split.
-   */
-  virtual bool should_split_bagfile() const;
-
 private:
   std::string uri_;
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -108,14 +108,16 @@ private:
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
 
+  // Used to track topic -> message count
+  std::unordered_map<std::string, TopicInformation> topics_;
+
+  rosbag2_storage::BagMetadata metadata_;
+
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
-  std::vector<std::string> relative_file_paths_;
-  uint64_t message_count_;
-  std::unordered_map<std::string, TopicInformation> topics_;
-  rcutils_time_point_value_t start_time_;
-  rcutils_time_point_value_t end_time_;
-  rosbag2_storage::BagMetadata generate_metadata_() const;
+
+  // Record TopicInformation into metadata
+  void finalize_metadata();
 };
 
 }  // namespace rosbag2
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -95,6 +95,13 @@ public:
    */
   virtual void write(std::shared_ptr<SerializedBagMessage> message);
 
+  /**
+   * Check if the current recording bagfile needs to be split and rolled over to new file.
+   *
+   * \return true if the bagfile should be split.
+   */
+  virtual bool should_split_bagfile() const;
+
 private:
   std::string uri_;
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -95,13 +95,6 @@ public:
    */
   virtual void write(std::shared_ptr<SerializedBagMessage> message);
 
-  /**
-   * Check if the current recording bagfile needs to be split and rolled over to new file.
-   *
-   * \return true if the bagfile should be split.
-   */
-  virtual bool should_split_bagfile() const;
-
 private:
   std::string uri_;
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -118,6 +118,9 @@ private:
 
   // Record TopicInformation into metadata
   void finalize_metadata();
+
+  // Prepares the metadata by setting initial values.
+  void init_metadata();
 };
 
 }  // namespace rosbag2
--- a/rosbag2/include/rosbag2/writer.hpp
+++ b/rosbag2/include/rosbag2/writer.hpp
@@ -17,6 +17,8 @@
 
 #include <memory>
 #include <string>
+#include <unordered_map>
+#include <vector>
 
 #include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
@@ -115,6 +117,12 @@ private:
 
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
+  std::vector<std::string> relative_file_paths_;
+  uint64_t message_count_;
+  std::unordered_map<std::string, TopicInformation> topics_;
+  rcutils_time_point_value_t start_time_;
+  rcutils_time_point_value_t end_time_;
+  rosbag2_storage::BagMetadata generate_metadata_() const;
 };
 
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -142,11 +142,9 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
 
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
-
   metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
 
   const auto duration = message_timestamp - metadata_.starting_time;
-
   metadata_.duration = std::max(metadata_.duration, duration);
 
   storage_->write(converter_ ? converter_->convert(message) : message);
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -14,6 +14,9 @@
 
 #include "rosbag2/writer.hpp"
 
+#include <rosbag2_storage/filesystem_helper.hpp>
+
+#include <algorithm>
 #include <memory>
 #include <stdexcept>
 #include <string>
@@ -34,7 +37,12 @@ Writer::Writer(
   storage_(nullptr),
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
-  max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT)
+  max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
+  relative_file_paths_({}),
+  message_count_(0),
+  topics_({}),
+  start_time_(INT64_MAX),
+  end_time_(INT64_MIN)
 {}
 
 Writer::~Writer()
@@ -63,7 +71,9 @@ void Writer::open(
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-  uri_ = storage_options.uri;
+
+  start_time_ = std::chrono::nanoseconds::max().count();
+  end_time_ = std::chrono::nanoseconds::min().count();
 }
 
 void Writer::create_topic(const TopicMetadata & topic_with_type)
@@ -76,6 +86,11 @@ void Writer::create_topic(const TopicMetadata & topic_with_type)
     converter_->add_topic(topic_with_type.name, topic_with_type.type);
   }
 
+  rosbag2_storage::TopicInformation info{};
+  info.topic_metadata = topic_with_type;
+
+  topics_.insert({topic_with_type.name, info});
+
   storage_->create_topic(topic_with_type);
 }
 
@@ -86,6 +101,7 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
   }
 
   storage_->remove_topic(topic_with_type);
+  topics_.erase(topic_with_type.name);
 }
 
 void Writer::write(std::shared_ptr<SerializedBagMessage> message)
@@ -93,6 +109,13 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
   if (!storage_) {
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
+
+  // Update the message count for the Topic.
+  ++topics_[message->topic_name].message_count;
+  ++message_count_;
+
+  start_time_ = std::min(start_time_, message->time_stamp);
+  end_time_ = std::max(end_time_, message->time_stamp);
 
   storage_->write(converter_ ? converter_->convert(message) : message);
 }
@@ -105,4 +128,34 @@ bool Writer::should_split_bagfile() const
     return storage_->get_bagfile_size() > max_bagfile_size_;
   }
 }
+
+rosbag2_storage::BagMetadata Writer::generate_metadata_() const
+{
+  auto time_of_first_message = std::chrono::nanoseconds(start_time_);
+  auto time_of_last_message = std::chrono::nanoseconds(end_time_);
+
+  rosbag2_storage::BagMetadata metadata{};
+
+  // Only populate metadata if storage exists
+  if (storage_) {
+    metadata.storage_identifier = storage_->get_identifier();
+
+    metadata.relative_file_paths = relative_file_paths_;
+    metadata.duration = time_of_last_message - time_of_first_message;
+    metadata.starting_time =
+      std::chrono::time_point<std::chrono::high_resolution_clock>(time_of_first_message);
+    metadata.message_count = message_count_;
+
+    metadata.bag_size = rosbag2_storage::FilesystemHelper::calculate_directory_size(uri_);
+
+    metadata.topics_with_message_count.reserve(topics_.size());
+
+    for (const auto & topic : topics_) {
+      metadata.topics_with_message_count.push_back(topic.second);
+    }
+  }
+
+  return metadata;
+}
+
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -143,7 +143,7 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
 
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
-  metadata_.starting_time = (std::min)(metadata_.starting_time, message_timestamp);
+  metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
 
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = (std::max)(metadata_.duration, duration);
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -38,17 +38,15 @@ Writer::Writer(
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
   max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
-  relative_file_paths_({}),
-  message_count_(0),
   topics_({}),
-  start_time_(INT64_MAX),
-  end_time_(INT64_MIN)
+  metadata_()
 {}
 
 Writer::~Writer()
 {
   if (!uri_.empty()) {
-    metadata_io_->write_metadata(uri_, storage_->get_metadata());
+    finalize_metadata();
+    metadata_io_->write_metadata(uri_, metadata_);
   }
 
   storage_.reset();  // Necessary to ensure that the storage is destroyed before the factory
@@ -72,11 +70,13 @@ void Writer::open(
     throw std::runtime_error("No storage could be initialized. Abort");
   }
 
-  relative_file_paths_.push_back(storage_->get_relative_path());
-
   uri_ = storage_options.uri;
-  start_time_ = std::chrono::nanoseconds::max().count();
-  end_time_ = std::chrono::nanoseconds::min().count();
+
+  metadata_ = rosbag2_storage::BagMetadata{};
+  metadata_.storage_identifier = storage_->get_storage_identifier();
+  metadata_.starting_time = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds::max());
+  metadata_.relative_file_paths.push_back(storage_->get_relative_path());
 }
 
 void Writer::create_topic(const TopicMetadata & topic_with_type)
@@ -89,10 +89,12 @@ void Writer::create_topic(const TopicMetadata & topic_with_type)
     converter_->add_topic(topic_with_type.name, topic_with_type.type);
   }
 
-  rosbag2_storage::TopicInformation info{};
-  info.topic_metadata = topic_with_type;
+  if (topics_.find(topic_with_type.name) == topics_.end()) {
+    rosbag2_storage::TopicInformation info{};
+    info.topic_metadata = topic_with_type;
 
-  topics_.insert({topic_with_type.name, info});
+    topics_.insert({topic_with_type.name, info});
+  }
 
   storage_->create_topic(topic_with_type);
 }
@@ -103,8 +105,13 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
     throw std::runtime_error("Bag is not open. Call open() before removing.");
   }
 
-  storage_->remove_topic(topic_with_type);
-  topics_.erase(topic_with_type.name);
+  auto it = topics_.find(topic_with_type.name);
+
+  if (it != topics_.end()) {
+    topics_.erase(it);
+
+    storage_->remove_topic(topic_with_type);
+  }
 }
 
 void Writer::write(std::shared_ptr<SerializedBagMessage> message)
@@ -115,10 +122,20 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
 
   // Update the message count for the Topic.
   ++topics_[message->topic_name].message_count;
-  ++message_count_;
+  ++metadata_.message_count;
 
-  start_time_ = std::min(start_time_, message->time_stamp);
-  end_time_ = std::max(end_time_, message->time_stamp);
+  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds(message->time_stamp));
+
+  if (metadata_.starting_time > message_timestamp) {
+    metadata_.starting_time = message_timestamp;
+  }
+
+  auto duration = message_timestamp - metadata_.starting_time;
+
+  if (duration > metadata_.duration) {
+    metadata_.duration = duration;
+  }
 
   storage_->write(converter_ ? converter_->convert(message) : message);
 }
@@ -132,37 +149,20 @@ bool Writer::should_split_bagfile() const
   }
 }
 
-rosbag2_storage::BagMetadata Writer::generate_metadata_() const
+void Writer::finalize_metadata()
 {
-  auto time_of_first_message = std::chrono::nanoseconds(start_time_);
-  auto time_of_last_message = std::chrono::nanoseconds(end_time_);
+  metadata_.bag_size = 0;
 
-  rosbag2_storage::BagMetadata metadata{};
-
-  // Only populate metadata if storage exists
-  if (storage_) {
-    metadata.storage_identifier = storage_->get_storage_identifier();
-
-    metadata.relative_file_paths = relative_file_paths_;
-    metadata.duration = time_of_last_message - time_of_first_message;
-    metadata.starting_time =
-      std::chrono::time_point<std::chrono::high_resolution_clock>(time_of_first_message);
-    metadata.message_count = message_count_;
-
-    metadata.bag_size = 0;
-
-    for (const auto& path : relative_file_paths_) {
-        metadata.bag_size += rosbag2_storage::FilesystemHelper::get_file_size(path);
-    }
-
-    metadata.topics_with_message_count.reserve(topics_.size());
-
-    for (const auto & topic : topics_) {
-      metadata.topics_with_message_count.push_back(topic.second);
-    }
+  for (const auto & path : metadata_.relative_file_paths) {
+    metadata_.bag_size += rosbag2_storage::FilesystemHelper::get_file_size(path);
   }
 
-  return metadata;
+  metadata_.topics_with_message_count.clear();
+  metadata_.topics_with_message_count.reserve(topics_.size());
+
+  for (const auto & topic : topics_) {
+    metadata_.topics_with_message_count.push_back(topic.second);
+  }
 }
 
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -114,6 +114,8 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
 
   if (topics_names_to_info_.erase(topic_with_type.name) > 0) {
     storage_->remove_topic(topic_with_type);
+  } else {
+    throw std::runtime_error("Tried to remove a non-existing topic");
   }
 }
 
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -59,7 +59,7 @@ void Writer::init_metadata()
   metadata_ = rosbag2_storage::BagMetadata{};
   metadata_.storage_identifier = storage_->get_storage_identifier();
   metadata_.starting_time = std::chrono::time_point<std::chrono::high_resolution_clock>(
-    std::chrono::nanoseconds::max());
+    (std::chrono::nanoseconds::max)());
   metadata_.relative_file_paths = {storage_->get_relative_path()};
 }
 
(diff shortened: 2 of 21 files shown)


rosbag2/src/rosbag2/writer.cpp
```diff
@@ -90,7 +111,52 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
+  if (storage_) {
+    metadata.storage_identifier = storage_->get_identifier();
+
+    metadata.relative_file_paths = relative_file_paths_;
+    metadata.duration = time_of_last_message - time_of_first_message;
+    metadata.starting_time =
+      std::chrono::time_point<std::chrono::high_resolution_clock>(time_of_first_message);
+    metadata.message_count = message_count_;
+
+    metadata.bag_size = rosbag2_storage::FilesystemHelper::calculate_directory_size(uri_);
```
Karsten1987: I understand that we used that function before, but I actually think now that we can't solely rely on the directory size. There could (even though shouldn't) be files in the same directory which are essentially not related to the payload of the bag file. 
I think it's preferable to iterate over the individual relative paths and ask for the filesize: https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/filesystem_helper.hpp#L132-L137

rosbag2/include/rosbag2/writer.hpp
```diff
@@ -95,13 +97,27 @@ class ROSBAG2_PUBLIC Writer
...
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage_;
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
+  uint64_t max_bagfile_size_;
+  std::vector<std::string> relative_file_paths_;
+  uint64_t message_count_;
+  std::unordered_map<std::string, TopicInformation> topics_;
+  rcutils_time_point_value_t start_time_;
+  rcutils_time_point_value_t end_time_;
```
Karsten1987: Thinking about this, could we not directly instantiate a BagMetaData struct here? Most of the things in the `generate_metadata_()` function could actually directly be offloaded.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -59,7 +71,10 @@ void Writer::open(
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
+
   uri_ = storage_options.uri;
+  start_time_ = std::chrono::nanoseconds::max().count();
+  end_time_ = std::chrono::nanoseconds::min().count();
```
Karsten1987: This could directly be set in a metadata struct.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -90,7 +111,52 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
 
+  // Update the message count for the Topic.
+  ++topics_[message->topic_name].message_count;
+  ++message_count_;
+
+  start_time_ = std::min(start_time_, message->time_stamp);
+  end_time_ = std::max(end_time_, message->time_stamp);
```
Karsten1987: the bagmetadata struct could be updated here directly.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -90,7 +111,52 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
+bool Writer::should_split_bagfile() const
+{
+  if (max_bagfile_size_ == rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT) {
+    return false;
+  } else {
+    return storage_->get_bagfile_size() > max_bagfile_size_;
+  }
+}
+
+rosbag2_storage::BagMetadata Writer::generate_metadata_() const
```
Karsten1987: I don't feel super strong about it, but I think most of the items (if not all) of the metadata information could be updated in place without the need of additional class members such as `start_time_`, .. `topics_`. The `generate_metadata` function would then actually just translate into a `get_metadata` function.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Makes sense, shouldn't be too hard since `Writer` already knows of all of the bagfiles.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: I pulled it out because I thought less characters per line would be more clean. After looking at it again, I think I should move both lines inside `if (storage_)` block.

rosbag2/include/rosbag2/writer.hpp
zmichaels11: The reason I didn't go with that design is because I used an `unordered_map` instead of a `vector` for topics. If I stored an internal `BagMetadata`, I'd have two indexing of topics.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: I spent some time attempting to embed `BagMetadata` into `Writer`. `BagMetadata` has `std::chrono::time_point<std::chrono::high_resolution_clock> starting_time` and `std::chrono::nanoseconds duration`. While it can be done, it requires converting between time formats at every message write. 

I found it cleaner to instead resole the time values when metadata is written out.

zmichaels11:
@Karsten1987 I merged #184 into this PR and rebased onto master. 
Please review this one.

rosbag2/include/rosbag2/writer.hpp
Karsten1987: Why would you need to index any of them? The metadata has all the index information you'd need. 
Agreed, the `std::vector<TopicInformation>` is not perfect for o(1) indexing, but that shouldn't be a big pain point IMO.

rosbag2/src/rosbag2/writer.cpp
Karsten1987: I can understand that. If it makes things easier, we can change/enhance the metadata information with a `end_time` if that makes things easier.

zmichaels11: I think it would be best to do that in a separate PR and iterate on what we currently have.

How about this:
* I build a local copy of `BagMetadata` in `Writer`
* I also store a `end_time` variable in `Writer` and translate that into `duration` on metadata write in the dtor
* once all bagsplitting is in, I open an issue to move `end_time` from `Writer` into `BagMetadata`

rosbag2/include/rosbag2/writer.hpp
zmichaels11: I think I know how to resolve this. See the [below comment](https://github.com/ros2/rosbag2/pull/184#discussion_r338214426) on what I think I should do right now. I'll resolve this conversation once I implement something.

zmichaels11:
@Karsten1987 ok I reworked it so that it can build as much of `BagMetadata` inline.
I replaced `generate_metadata()` with `finalize_metadata`. The only operations that occur within that function are summing up bagfile size and moving the `TopicInformation` from the `Writer` owned hashmap to the metadata's vector.

rosbag2_storage/test/rosbag2_storage/test_read_only_plugin.cpp
```diff
@@ -54,6 +54,12 @@ std::vector<rosbag2_storage::TopicMetadata> TestReadOnlyPlugin::get_all_topics_a
   return std::vector<rosbag2_storage::TopicMetadata>();
 }
 
+std::string TestReadOnlyPlugin::get_relative_path() const
+{
+  std::cout << "\nreturning relative path\n";
+  return "";
```
Karsten1987: does it make sense to return a dummy path in here to test for?

rosbag2_storage/test/rosbag2_storage/test_plugin.cpp
```diff
@@ -79,6 +79,12 @@ rosbag2_storage::BagMetadata TestPlugin::get_metadata()
   return rosbag2_storage::BagMetadata();
 }
 
+std::string TestPlugin::get_relative_path() const
+{
+  std::cout << "\returning relative path\n";
+  return "";
```
Karsten1987: does it make sense to return a dummy path in here to test for?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +120,23 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
 
+  // Update the message count for the Topic.
+  ++topics_[message->topic_name].message_count;
+  ++metadata_.message_count;
```
Karsten1987: not that it matters, but I believe instead of constantly incrementing this message count, we could accumulate the number of messages in the `finalize_metadata` function when iterating through all topics.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +120,23 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
+
+  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds(message->time_stamp));
+
+  if (metadata_.starting_time > message_timestamp) {
+    metadata_.starting_time = message_timestamp;
+  }
```
Karsten1987:
```
metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp)
```

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +120,23 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
+
+  auto duration = message_timestamp - metadata_.starting_time;
+
+  if (duration > metadata_.duration) {
+    metadata_.duration = duration;
+  }
```
Karsten1987:
```
metadata_.duration = std::max(metadata_.duration, duration)
```

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -105,4 +148,21 @@ bool Writer::should_split_bagfile() const
...
+  for (const auto & path : metadata_.relative_file_paths) {
+    metadata_.bag_size += rosbag2_storage::FilesystemHelper::get_file_size(path);
+  }
+
+  metadata_.topics_with_message_count.clear();
+  metadata_.topics_with_message_count.reserve(topics_.size());
+
+  for (const auto & topic : topics_) {
+    metadata_.topics_with_message_count.push_back(topic.second);
+  }
```
Karsten1987: really only a nitpick, but I believe the call to `std::transform` might also apply in the same manner in case we ever change the datatypes for `topics_` or `metadata_.topics_with_message_count`. 
```
std::transform(topics_.begin(), topics_.end(), std::back_inserter(metadata_.topics_with_message_count),
  [](auto pair) {return pair.second;});
```

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -76,6 +89,13 @@ void Writer::create_topic(const TopicMetadata & topic_with_type)
...
   }
 
+  if (topics_.find(topic_with_type.name) == topics_.end()) {
+    rosbag2_storage::TopicInformation info{};
+    info.topic_metadata = topic_with_type;
+
+    topics_.insert({topic_with_type.name, info});
+  }
+
   storage_->create_topic(topic_with_type);
```
Karsten1987: with the metadata handled in here, does it make sense to still call `storage_->create_topic(..)`?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -105,4 +148,21 @@ bool Writer::should_split_bagfile() const
     return storage_->get_bagfile_size() > max_bagfile_size_;
   }
 }
+
+void Writer::finalize_metadata()
+{
+  metadata_.bag_size = 0;
```
Karsten1987: should that maybe be done in the `open` function, when `metadata_` is being initialized?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -105,4 +148,21 @@ bool Writer::should_split_bagfile() const
...
+
+  for (const auto & path : metadata_.relative_file_paths) {
+    metadata_.bag_size += rosbag2_storage::FilesystemHelper::get_file_size(path);
+  }
+
+  metadata_.topics_with_message_count.clear();
```
Karsten1987: should that maybe be done in the open function, when metadata_ is being initialized? I mean, actually, I would expect we don't need that call.

rosbag2_storage/test/rosbag2_storage/test_read_only_plugin.cpp
zmichaels11: nice catch. It should so we can test it

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Oh we're not iterating over topics in `finalize_metadata`. Its iterating over `topics_`, which is an unordered map of topic name to `TopicInformation`.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
```diff
@@ -30,6 +32,13 @@ class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
   virtual ~BaseInfoInterface() = default;
 
   virtual BagMetadata get_metadata() = 0;
+
+  /**
+   * Retrieves the relative path to the backing of the storage plugin.
+   *
+   * \returns the relative path.
+   */
+  virtual std::string get_relative_path() const = 0;
```
Karsten1987: This is up for discussion, but when I think about it, I believe this function makes more sense on the `base_io_interface` then on the `info` one. 

We merged the other one already, but I think the `get_storage_identifier()` could actually be part of the info interface then.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Yeah, in the case of sqlite, it writes the topic name and serialization format into a table

rosbag2/src/rosbag2/writer.cpp
Karsten1987: Yes, we iterate through `topics_` which in fact holds the message count, right? So the sum of all `message_count` of each item in the `topics_` should result in the overall message count. I think that's a bit more efficient than incrementing the `metadata_.message_count` with every call (I mean you do exactly that already in the line above).

rosbag2/src/rosbag2/writer.cpp
Karsten1987: Sorry my bad. I wasn't precise enough. I was thinking to call `create_topic` only once per topic, essentially move this call up into the `if` block

rosbag2/src/rosbag2/writer.cpp
zmichaels11: This is just me being overly cautious in the event that future commits touch `metadata_`. `bag_size` is already set to 0 in `open` with the line: 
`metadata_ = rosbag2_storage::BagMetadata{};`.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Exact same reason as [the previous comment](https://github.com/ros2/rosbag2/pull/184#discussion_r338328148); this line should be a no-op. 

I'm just being explicit that `metadata_.topics_with_message_count` will only hold the values from `topics_`.

rosbag2/src/rosbag2/writer.cpp
Karsten1987: But what would you do in case it's not zero? The same with a call to `.clear`. What if there is already data in it (wrongly presumably)?
I guess what I am aiming at is, that if we do it (here), there should be a check for whatever reason it's not 0 and bring up an error message that there might be data lost/ignored. Otherwise, I think the call doesn't bring much.

zmichaels11: Actually I have a better reason why I did it this way:

I think the code is a little more clear that `metadata_.bag_size` is the sum of the size of all files since it gets set to 0 before the accumulating for-loop.

rosbag2_storage/test/rosbag2_storage/test_plugin.cpp
```diff
@@ -79,6 +79,12 @@ rosbag2_storage::BagMetadata TestPlugin::get_metadata()
   return rosbag2_storage::BagMetadata();
 }
 
+std::string TestPlugin::get_relative_path() const
+{
+  std::cout << "\returning relative path\n";
```
thomas-moulard:
```suggestion
  std::cout << "\\returning relative path\n";
```

rosbag2/test/rosbag2/mock_storage.hpp
```diff
@@ -39,6 +39,7 @@ class MockStorage : public rosbag2_storage::storage_interfaces::ReadWriteInterfa
   MOCK_METHOD0(get_metadata, rosbag2_storage::BagMetadata());
   MOCK_CONST_METHOD0(get_bagfile_size, uint64_t());
   MOCK_CONST_METHOD0(get_storage_identifier, std::string());
+  MOCK_CONST_METHOD0(get_relative_path, std::string());
```
thomas-moulard: nit: alphasort

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -63,7 +69,14 @@ void Writer::open(
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
+
   uri_ = storage_options.uri;
+
+  metadata_ = rosbag2_storage::BagMetadata{};
```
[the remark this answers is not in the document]
thomas-moulard: More like, we should assert or throw if the order ctor -> open -> close -> open -> close -> dtor is not respected.
If metadata_ is not already initialized, the call to open should fail.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -34,13 +37,16 @@ Writer::Writer(
...
+  topics_({}),
+  metadata_()
 {}
 
 Writer::~Writer()
 {
   if (!uri_.empty()) {
-    metadata_io_->write_metadata(uri_, storage_->get_metadata());
+    finalize_metadata();
+    metadata_io_->write_metadata(uri_, metadata_);
```
thomas-moulard: could we call close() here instead of having logic in the dtor?

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Oh I misinterpreted your question. Yeah I can do that.
I think it will be a wash when it comes to performance, but it does group together the operations if I do move it into `finalize_metadata`

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -63,7 +69,14 @@ void Writer::open(
...
     throw std::runtime_error("No storage could be initialized. Abort");
   }
+
   uri_ = storage_options.uri;
+
+  metadata_ = rosbag2_storage::BagMetadata{};
+  metadata_.storage_identifier = storage_->get_storage_identifier();
+  metadata_.starting_time = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds::max());
+  metadata_.relative_file_paths.push_back(storage_->get_relative_path());
```
thomas-moulard:
```suggestion
  metadata_.relative_file_paths = {storage_->get_relative_path()};
```

should work and would make it more explicit we are actually resetting the vector

rosbag2/src/rosbag2/writer.cpp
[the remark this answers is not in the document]
zmichaels11: yeah I agree with @thomas-moulard here, it should throw if `open` is called multiple times

rosbag2/src/rosbag2/writer.cpp
thomas-moulard: http://cpp.sh/7j2mn

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -76,6 +89,13 @@ void Writer::create_topic(const TopicMetadata & topic_with_type)
     converter_->add_topic(topic_with_type.name, topic_with_type.type);
   }
 
+  if (topics_.find(topic_with_type.name) == topics_.end()) {
+    rosbag2_storage::TopicInformation info{};
+    info.topic_metadata = topic_with_type;
+
+    topics_.insert({topic_with_type.name, info});
```
thomas-moulard: We should check `insert` return value and assert or throw if the topic already exist

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -85,7 +105,13 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
     throw std::runtime_error("Bag is not open. Call open() before removing.");
   }
 
-  storage_->remove_topic(topic_with_type);
+  auto it = topics_.find(topic_with_type.name);
+
+  if (it != topics_.end()) {
```
thomas-moulard: Why are we removing topics which do not exist?

zmichaels11: This was a bug on my end. Fixing

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -85,7 +105,13 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
...
+    topics_.erase(it);
+
+    storage_->remove_topic(topic_with_type);
```
thomas-moulard: Consider the following:

```c++
if (topics_.erase(topic_with_type.name) > 0) {
  storage_->remove_topic(topic_with_type);
}
```

rosbag2/include/rosbag2/writer.hpp
```diff
@@ -106,8 +108,16 @@ class ROSBAG2_PUBLIC Writer
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
 
+  // Used to track topic -> message count
+  std::unordered_map<std::string, TopicInformation> topics_;
```
thomas-moulard: `topics_name_to_info_` ?

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Actually hold on, `if (it != topics_.end()` is true when the topic exists.
If you meant the line `storage_->remove_topic(topic_with_type)` which happens inside the if-statement, then its because it needs to remove the topic from the internal storage. In the case of `sqlite3`, this involves updating a table.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +120,23 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
 
+  // Update the message count for the Topic.
+  ++topics_[message->topic_name].message_count;
```
thomas-moulard: At should be used here, as I believe we should not implicitly create entries to this map, without calling create_topic first.

```suggestion
  ++topics_.at(message->topic_name).message_count;
```

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Applying suggestion.

thomas-moulard: this is one logical unit, consider creating a private function `initialize_metadata` to isolate this block.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Applying suggestion.

rosbag2/include/rosbag2/writer.hpp
zmichaels11: I agree with more descriptive members; applying suggestion.

rosbag2/src/rosbag2/writer.cpp
thomas-moulard: What I meant in this case, is that we should have:

```c++
if (it == topics_.end()) {
 throw std::runtime_error("tried to delete a non-existing topic");
} else {
  topics_.erase(it);
  storage_->remove_topic(topic_with_type);
}

File: rosbag2/src/rosbag2/writer.cpp (hunk 20 above)
Comment (zmichaels11) [reply rosbag2/src/rosbag2/writer.cpp]: Applied suggestion

File: rosbag2/src/rosbag2/writer.cpp (hunk 22 above)
Comment (zmichaels11) [reply rosbag2/src/rosbag2/writer.cpp]: Oh yeah, I can do that.

Comment (zmichaels11) [reply rosbag2/src/rosbag2/writer.cpp]: Done!

**Comment by zmichaels11** [comment]
Opened Issue #195 regarding tests on FSM since the logic in `Writer`'s FSM was not changed in this PR.

File: rosbag2/src/rosbag2/writer.cpp [hunk 26]
```diff
@@ -114,6 +114,8 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
 
   if (topics_names_to_info_.erase(topic_with_type.name) > 0) {
     storage_->remove_topic(topic_with_type);
+  } else {
+    throw std::runtime_error("Tried to remove a non-existing topic");
```
thomas-moulard: it'd be preferable to have this contain the topic name.

```c++
std::stringstream ss;
ss << "failed to remove non-existing topic \"" << topic_with_type.name.c_str() << "\"";
throw std::runtime_error(ss.str());
```

Note: unless I directly create code suggestions you can merge, the code in my comments is usually not production-ready, it's usually just enough so the reader can understand the intent, hence this additional comment :)

Karsten1987:
> The dtor also is doing more than calling close, which makes me nervous (storage factory reset being put aside, which anyway is not necessary here IIUC).

We can introduce `close()` here, but I don't think it's a big deal to have that logic in the constructor.  The storage factory reset is not necessary, but we have to ensure that the storage instance is closed correctly before the factory, otherwise the plugin handles get lost.

rosbag2/include/rosbag2/writer.hpp
```diff
@@ -106,8 +108,19 @@ class ROSBAG2_PUBLIC Writer
...
+  rosbag2_storage::BagMetadata metadata_;
+
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
+
+  // Record TopicInformation into metadata
+  void finalize_metadata();
+
+  // Prepares the metadata by setting initial values.
+  void init_metadata();
```
Karsten1987: absolute nitpick, but I'd put `init_metadata` before `finalize_metadata`

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -14,6 +14,9 @@
 
 #include "rosbag2/writer.hpp"
 
+#include <rosbag2_storage/filesystem_helper.hpp>
```
Karsten1987: Isn't that being flagged by cpplint?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -34,19 +37,31 @@ Writer::Writer(
...
+    metadata_io_->write_metadata(uri_, metadata_);
   }
 
   storage_.reset();  // Necessary to ensure that the storage is destroyed before the factory
   storage_factory_.reset();
 }
 
+void Writer::init_metadata()
+{
+  metadata_ = rosbag2_storage::BagMetadata{};
```
Karsten1987: once again, isn't that line redundant to line 42 in the initializer list? Or is `init_metadata` called multiple times to reset the struct?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +137,18 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
 
+  // Update the message count for the Topic.
+  ++topics_names_to_info_.at(message->topic_name).message_count;
+
+  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds(message->time_stamp));
+
```
Karsten1987: nitpick: consider removing the line

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +137,18 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
+  metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
+
+  const auto duration = message_timestamp - metadata_.starting_time;
+
```
Karsten1987: nitpick: consider removing the line

rosbag2/src/rosbag2/writer.cpp
zmichaels11: cpplint seems to think its a c-header and wants it before the rest of the headers.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: it is redundant, but it also indicates that `metadata_` is zeroed out within `init_metadata()`

rosbag2/src/rosbag2/writer.cpp
zmichaels11: I think defining the local variables `message_timestamp` and `duration` improves the readability of this code block.
The alternative is something like:

```c++
metadata_.starting_time = std::min(
  metadata_.starting_time,
  std::chrono::time_point<std::chrono::high_resolution_clock>(
    std::chrono::nanoseconds(message->time_stamp)));

metadata_.duration = std::max(
  metadata_.duration,
  std::chrono::time_point<std::chrono::high_resolution_clock>(
    std::chrono::nanoseconds(message->time_stamp) - metadata_.starting_time);
```

rosbag2/src/rosbag2/writer.cpp
Karsten1987: right. That's why we normally include the package headers with `"` instead of `<>`

rosbag2/src/rosbag2/writer.cpp
Karsten1987: comment was meant for the empty lines.

zmichaels11:
Rebased and applied the last few suggestions.

@piraka9011 can we get some CI badges on this?

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
zmichaels11: Opened issue #197 regarding this.

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8545/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4416/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6965/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8455/)

zmichaels11:
I think I figured out why windows failed; forgot to include chrono in `Writer.cpp`

@piraka9011 can you retest Windows only?

piraka9011:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8456/)

piraka9011:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8459/)

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -34,19 +38,31 @@ Writer::Writer(
   storage_(nullptr),
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
-  max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT)
+  max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
+  topics_names_to_info_({}),
```
thomas-moulard:
```suggestion
  topics_names_to_info_(),
```

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +138,16 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
...
 
+  // Update the message count for the Topic.
+  ++topics_names_to_info_.at(message->topic_name).message_count;
+
+  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds(message->time_stamp));
+  metadata_.starting_time = (std::min)(metadata_.starting_time, message_timestamp);
+
+  const auto duration = message_timestamp - metadata_.starting_time;
+  metadata_.duration = (std::max)(metadata_.duration, duration);
```
thomas-moulard:
```suggestion
  metadata_.duration = std::max(metadata_.duration, duration);
```

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -94,6 +138,16 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
     throw std::runtime_error("Bag is not open. Call open() before writing.");
   }
 
+  // Update the message count for the Topic.
+  ++topics_names_to_info_.at(message->topic_name).message_count;
+
+  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
+    std::chrono::nanoseconds(message->time_stamp));
+  metadata_.starting_time = (std::min)(metadata_.starting_time, message_timestamp);
```
thomas-moulard:
```suggestion
  metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
```

rosbag2/src/rosbag2/writer.cpp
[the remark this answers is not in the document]
zmichaels11: This is related to issue #195 and should be resolved outside of this PR.

rosbag2/src/rosbag2/writer.cpp
zmichaels11: This is related to issue #195 and should be resolved outside of this PR.

zmichaels11:
@piraka9011 one last time, but with more unit tests

rosbag2_storage/test/rosbag2_storage/test_plugin.cpp
```diff
@@ -79,6 +80,12 @@ rosbag2_storage::BagMetadata TestPlugin::get_metadata()
   return rosbag2_storage::BagMetadata();
 }
 
+std::string TestPlugin::get_relative_path() const
+{
+  std::cout << "\nreturning relative path\n";
+  return plugin_constants::DUMMY_FILEPATH;
```
Karsten1987: I can't comment on the lines above, but consider shifting the return value `default_max_bagfile_size` into the `plugin_constants` header as well.

rosbag2_storage/test/rosbag2_storage/plugin_constants.hpp
```diff
@@ -0,0 +1,26 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Karsten1987: maybe a detail, but consider renaming this file to `test_plugin_constants.hpp` to stay in sync with `test_plugin.hpp` and to highlight that it's clearly used only in the test plugin.

zmichaels11:
* Renamed `plugin_constants` -> `test_constants`
* Moved `default_max_bagfile_size` into `test_constants`

Karsten1987:
Please run a new round of CI on this. Also make sure you build packages up to ros2bag and rosbag2_tests and test with packages-select-regex rosbag2

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/d15a12ab0ff8cd4c3f20c7c1f14180b8/raw/7cda0198d9bee189ba8f61a3da0f8801c1ee378e/ros2.repos
BUILD args: `--packages-up-to rosbag2 rosbag2_tests`
TEST args: `--packages-select rosbag2`
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8553/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4425/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6973/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8467/)

Karsten1987:
Given that this PR is touching more than one package, I think we should test all rosbag2 related packages. 
Therefore I proposed to set packages-select-regex rosbag2 as test args instead of just testing rosbag2 to avoid regression.

piraka9011:
Sorry I missed that. Here's the new build:

Gist: https://gist.githubusercontent.com/piraka9011/d15a12ab0ff8cd4c3f20c7c1f14180b8/raw/7cda0198d9bee189ba8f61a3da0f8801c1ee378e/ros2.repos
BUILD args: --packages-up-to rosbag2 rosbag2_tests
TEST args: --packages-select-regex rosbag2
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8555/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4428/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6975/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8470/)

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/21c04ad8e74d0f75a3f6bfae01264072/raw/7cda0198d9bee189ba8f61a3da0f8801c1ee378e/ros2.repos
BUILD args: --packages-up-to rosbag2 rosbag2_tests
TEST args: --packages-select-regex rosbag2
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8556/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4429/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6976/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8471/)

zmichaels11:
@Karsten1987 we're green!

PR #185, opened 2019-10-17
Rosbag splitting in Writer

zmichaels11:

### Changes
* `open` in `BaseIOInterface` now must point to the storage file
* `Writer` is responsible for writing and generating `BagMetadata`
* `Writer` rolls over to a new storage when the bagfile exceeds `max_bagfile_size` by opening a new storage.

### Dependent PRs
- [x] PR #182 
- [x] PR #183 
- [x] PR #184 

### Issues
* ros-security/aws-roadmap#11

### DIFF
--- a/rosbag2/src/rosbag2/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/sequential_reader.cpp
@@ -21,15 +21,19 @@
 #include <vector>
 
 #include "rosbag2/info.hpp"
+#include "rosbag2_storage/metadata_io.hpp"
 
 namespace rosbag2
 {
 
 SequentialReader::SequentialReader(
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
-  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory)
-: storage_factory_(std::move(storage_factory)), converter_factory_(std::move(converter_factory)),
-  converter_(nullptr)
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory,
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io)
+: storage_factory_(std::move(storage_factory)),
+  converter_factory_(std::move(converter_factory)),
+  converter_(nullptr),
+  metadata_io_(std::move(metadata_io))
 {}
 
 SequentialReader::~SequentialReader()
@@ -41,11 +45,12 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
+  metadata_ = metadata_io_->read_metadata(storage_options.uri);
   storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-  auto topics = storage_->get_metadata().topics_with_message_count;
+  auto topics = metadata_.topics_with_message_count;
   if (topics.empty()) {
     return;
   }
--- a/rosbag2/src/rosbag2/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/sequential_reader.cpp
@@ -21,6 +21,7 @@
 #include <vector>
 
 #include "rosbag2/info.hpp"
+#include "rosbag2/logging.hpp"
 #include "rosbag2_storage/metadata_io.hpp"
 
 namespace rosbag2
@@ -46,13 +47,18 @@ SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
   metadata_ = metadata_io_->read_metadata(storage_options.uri);
+  if (metadata_.relative_file_paths.empty()) {
+    ROSBAG2_LOG_WARN("No file paths were found in metadata.");
+    return;
+  }
   storage_ = storage_factory_->open_read_only(
-    metadata_.relative_file_paths[0], storage_options.storage_id);
+    metadata_.relative_file_paths.at(0), storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
   auto topics = metadata_.topics_with_message_count;
   if (topics.empty()) {
+    ROSBAG2_LOG_WARN("No topics were listed in metadata.");
     return;
   }
 
--- a/rosbag2/src/rosbag2/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/sequential_reader.cpp
@@ -46,7 +46,8 @@ SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
   metadata_ = metadata_io_->read_metadata(storage_options.uri);
-  storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
+  storage_ = storage_factory_->open_read_only(
+    metadata_.relative_file_paths[0], storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -39,6 +39,7 @@ Writer::Writer(
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
   max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
+  bagfile_counter_(0),
   topics_names_to_info_(),
   metadata_()
 {}
@@ -75,7 +76,9 @@ void Writer::open(
     converter_ = std::make_unique<Converter>(converter_options, converter_factory_);
   }
 
-  storage_ = storage_factory_->open_read_write(storage_options.uri, storage_options.storage_id);
+  auto bagfile_uri = next_bagfile_uri_();
+
+  storage_ = storage_factory_->open_read_write(bagfile_uri, storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
@@ -132,6 +135,24 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
   }
 }
 
+void Writer::split_bagfile_()
+{
+  auto bagfile_uri = next_bagfile_uri_();
+
+  storage_ = storage_factory_->open_read_write(bagfile_uri, metadata_.storage_identifier);
+
+  if (!storage_) {
+    throw std::runtime_error("No storage could be initialized. Abort");
+  }
+
+  metadata_.relative_file_paths.push_back(storage_->get_relative_path());
+
+  // Re-register all Topics since we rolled-over to a new bagfile.
+  for (const auto & topic : topics_names_to_info_) {
+    storage_->create_topic(topic.second.topic_metadata);
+  }
+}
+
 void Writer::write(std::shared_ptr<SerializedBagMessage> message)
 {
   if (!storage_) {
@@ -141,6 +162,9 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
   // Update the message count for the Topic.
   ++topics_names_to_info_.at(message->topic_name).message_count;
 
+  if (should_split_bagfile()) {
+    split_bagfile_();
+  }
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
   metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
@@ -176,6 +200,21 @@ void Writer::finalize_metadata()
     metadata_.topics_with_message_count.push_back(topic.second);
     metadata_.message_count += topic.second.message_count;
   }
+}
+
+std::string Writer::next_bagfile_uri_()
+{
+  std::stringstream db_name;
+  db_name << rosbag2_storage::FilesystemHelper::get_folder_name(uri_);
+
+  // Only append the counter after we have split.
+  if (bagfile_counter_ > 0) {
+    db_name << "_" << bagfile_counter_;
+  }
+
+  ++bagfile_counter_;
+
+  return rosbag2_storage::FilesystemHelper::concat({uri_, db_name.str()});
 }
 
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -14,8 +14,6 @@
 
 #include "rosbag2/writer.hpp"
 
-#include <rosbag2_storage/filesystem_helper.hpp>
-
 #include <algorithm>
 #include <chrono>
 #include <memory>
@@ -25,9 +23,29 @@
 
 #include "rosbag2/info.hpp"
 #include "rosbag2/storage_options.hpp"
+#include "rosbag2_storage/filesystem_helper.hpp"
 
 namespace rosbag2
 {
+
+namespace
+{
+std::string format_storage_uri(const std::string & base_folder, uint64_t storage_count)
+{
+  // Right now `base_folder_` is always just the folder name for where to install the bagfile.
+  // The name of the folder needs to be queried in case Writer is opened with a relative path.
+  std::stringstream storage_file_name;
+  storage_file_name << rosbag2_storage::FilesystemHelper::get_folder_name(base_folder);
+
+  // Only append the counter after we have split.
+  // This is so bagfiles have the old naming convention if splitting is disabled.
+  if (storage_count > 0) {
+    storage_file_name << "_" << storage_count;
+  }
+
+  return rosbag2_storage::FilesystemHelper::concat({base_folder, storage_file_name.str()});
+}
+}  // namespace
 
 Writer::Writer(
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
@@ -39,16 +57,15 @@ Writer::Writer(
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
   max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
-  bagfile_counter_(0),
   topics_names_to_info_(),
   metadata_()
 {}
 
 Writer::~Writer()
 {
-  if (!uri_.empty()) {
+  if (!base_folder_.empty()) {
     finalize_metadata();
-    metadata_io_->write_metadata(uri_, metadata_);
+    metadata_io_->write_metadata(base_folder_, metadata_);
   }
 
   storage_.reset();  // Necessary to ensure that the storage is destroyed before the factory
@@ -69,7 +86,7 @@ void Writer::open(
   const ConverterOptions & converter_options)
 {
   max_bagfile_size_ = storage_options.max_bagfile_size;
-  uri_ = storage_options.uri;
+  base_folder_ = storage_options.uri;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
@@ -77,9 +94,9 @@ void Writer::open(
     converter_ = std::make_unique<Converter>(converter_options, converter_factory_);
   }
 
-  auto bagfile_uri = next_bagfile_uri();
+  const auto storage_uri = format_storage_uri(base_folder_, 0);
 
-  storage_ = storage_factory_->open_read_write(bagfile_uri, storage_options.storage_id);
+  storage_ = storage_factory_->open_read_write(storage_uri, storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
@@ -136,13 +153,14 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
 
 void Writer::split_bagfile()
 {
-  auto bagfile_uri = next_bagfile_uri();
-
-  storage_ = storage_factory_->open_read_write(bagfile_uri, metadata_.storage_identifier);
+  const auto storage_uri = format_storage_uri(
+    base_folder_,
+    metadata_.relative_file_paths.size());
+  storage_ = storage_factory_->open_read_write(storage_uri, metadata_.storage_identifier);
 
   if (!storage_) {
     std::stringstream errmsg;
-    errmsg << "Failed to rollover bagfile to new file: \"" << bagfile_uri << "\"!";
+    errmsg << "Failed to rollover bagfile to new file: \"" << storage_uri << "\"!";
 
     throw std::runtime_error(errmsg.str());
   }
@@ -203,22 +221,6 @@ void Writer::finalize_metadata()
     metadata_.topics_with_message_count.push_back(topic.second);
     metadata_.message_count += topic.second.message_count;
   }
-}
-
-std::string Writer::next_bagfile_uri()
-{
-  std::stringstream db_name;
-  db_name << rosbag2_storage::FilesystemHelper::get_folder_name(uri_);
-
-  // Only append the counter after we have split.
-  // This is so bagfiles have the same naming conventions if splitting is disabled.
-  if (bagfile_counter_ > 0) {
-    db_name << "_" << bagfile_counter_;
-  }
-
-  ++bagfile_counter_;
-
-  return rosbag2_storage::FilesystemHelper::concat({uri_, db_name.str()});
 }
 
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -162,7 +162,7 @@ void Writer::split_bagfile()
 
   metadata_.relative_file_paths.push_back(storage_->get_relative_file_path());
 
-  // Re-register all Topics since we rolled-over to a new bagfile.
+  // Re-register all topics since we rolled-over to a new bagfile.
   for (const auto & topic : topics_names_to_info_) {
     storage_->create_topic(topic.second.topic_metadata);
   }
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -35,13 +35,8 @@ std::string format_storage_uri(const std::string & base_folder, uint64_t storage
   // Right now `base_folder_` is always just the folder name for where to install the bagfile.
   // The name of the folder needs to be queried in case Writer is opened with a relative path.
   std::stringstream storage_file_name;
-  storage_file_name << rosbag2_storage::FilesystemHelper::get_folder_name(base_folder);
-
-  // Only append the counter after we have split.
-  // This is so bagfiles have the old naming convention if splitting is disabled.
-  if (storage_count > 0) {
-    storage_file_name << "_" << storage_count;
-  }
+  storage_file_name << rosbag2_storage::FilesystemHelper::get_folder_name(base_folder) <<
+    "_" << storage_count;
 
   return rosbag2_storage::FilesystemHelper::concat({base_folder, storage_file_name.str()});
 }
@@ -78,7 +73,7 @@ void Writer::init_metadata()
   metadata_.storage_identifier = storage_->get_storage_identifier();
   metadata_.starting_time = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds::max());
-  metadata_.relative_file_paths = {storage_->get_relative_path()};
+  metadata_.relative_file_paths = {storage_->get_relative_file_path()};
 }
 
 void Writer::open(
@@ -165,7 +160,7 @@ void Writer::split_bagfile()
     throw std::runtime_error(errmsg.str());
   }
 
-  metadata_.relative_file_paths.push_back(storage_->get_relative_path());
+  metadata_.relative_file_paths.push_back(storage_->get_relative_file_path());
 
   // Re-register all Topics since we rolled-over to a new bagfile.
   for (const auto & topic : topics_names_to_info_) {
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -141,7 +141,10 @@ void Writer::split_bagfile()
   storage_ = storage_factory_->open_read_write(bagfile_uri, metadata_.storage_identifier);
 
   if (!storage_) {
-    throw std::runtime_error("No storage could be initialized. Abort");
+    std::stringstream errmsg;
+    errmsg << "Failed to rollover bagfile to new file: \"" << bagfile_uri << "\"!";
+
+    throw std::runtime_error(errmsg.str());
   }
 
   metadata_.relative_file_paths.push_back(storage_->get_relative_path());
@@ -164,6 +167,7 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
   if (should_split_bagfile()) {
     split_bagfile();
   }
+
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
   metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
@@ -207,6 +211,7 @@ std::string Writer::next_bagfile_uri()
   db_name << rosbag2_storage::FilesystemHelper::get_folder_name(uri_);
 
   // Only append the counter after we have split.
+  // This is so bagfiles have the same naming conventions if splitting is disabled.
   if (bagfile_counter_ > 0) {
     db_name << "_" << bagfile_counter_;
   }
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -69,6 +69,7 @@ void Writer::open(
   const ConverterOptions & converter_options)
 {
   max_bagfile_size_ = storage_options.max_bagfile_size;
+  uri_ = storage_options.uri;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
--- a/rosbag2/src/rosbag2/writer.cpp
+++ b/rosbag2/src/rosbag2/writer.cpp
@@ -77,14 +77,12 @@ void Writer::open(
     converter_ = std::make_unique<Converter>(converter_options, converter_factory_);
   }
 
-  auto bagfile_uri = next_bagfile_uri_();
+  auto bagfile_uri = next_bagfile_uri();
 
   storage_ = storage_factory_->open_read_write(bagfile_uri, storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-
-  uri_ = storage_options.uri;
 
   init_metadata();
 }
@@ -136,9 +134,9 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
   }
 }
 
-void Writer::split_bagfile_()
+void Writer::split_bagfile()
 {
-  auto bagfile_uri = next_bagfile_uri_();
+  auto bagfile_uri = next_bagfile_uri();
 
   storage_ = storage_factory_->open_read_write(bagfile_uri, metadata_.storage_identifier);
 
@@ -164,7 +162,7 @@ void Writer::write(std::shared_ptr<SerializedBagMessage> message)
   ++topics_names_to_info_.at(message->topic_name).message_count;
 
   if (should_split_bagfile()) {
-    split_bagfile_();
+    split_bagfile();
   }
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
@@ -203,7 +201,7 @@ void Writer::finalize_metadata()
   }
 }
 
-std::string Writer::next_bagfile_uri_()
+std::string Writer::next_bagfile_uri()
 {
   std::stringstream db_name;
   db_name << rosbag2_storage::FilesystemHelper::get_folder_name(uri_);
--- a/rosbag2/test/rosbag2/test_sequential_reader.cpp
+++ b/rosbag2/test/rosbag2/test_sequential_reader.cpp
@@ -21,8 +21,8 @@
 
 #include "rosbag2/sequential_reader.hpp"
 #include "rosbag2_storage/bag_metadata.hpp"
-#include "rosbag2_storage/topic_metadata.hpp"
 #include "rosbag2_storage/metadata_io.hpp"
+#include "rosbag2_storage/topic_metadata.hpp"
 
 #include "mock_converter.hpp"
 #include "mock_converter_factory.hpp"
@@ -72,8 +72,7 @@ public:
   std::string storage_serialization_format_;
 };
 
-TEST_F(SequentialReaderTest, read_next_uses_converters_to_convert_serialization_format)
-{
+TEST_F(SequentialReaderTest, read_next_uses_converters_to_convert_serialization_format) {
   std::string output_format = "rmw2_format";
 
   auto format1_converter = std::make_unique<StrictMock<MockConverter>>();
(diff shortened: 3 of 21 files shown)


zmichaels11:
Epr failed due to a slight change in db name - this PR removed the extension since `Writer` needs to specify the name and `Writer` can't know the file extension until after the storage opens. It also cannot assume `db3` in case the storage plugin does not back as `SQLite3`.

I can resolve this by either:
* blocking until Reader changes are synced up and rewriting all e2e tests 
* Append the extension in `open` and introduce a `get_relative_paths` to `io interfaces`

I'll probably do the second and introduce a new PR that should be merged in before this one.

zmichaels11:
@Karsten1987 this PR is ready for review.
Please don't merge until I verify that it doesn't break any of the work done for reading split bagfiles.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -235,14 +222,14 @@ std::string SqliteStorage::get_storage_identifier() const
 
 std::string SqliteStorage::get_relative_path() const
 {
-  return database_name_;
+  return relative_path_;
 }
 
 rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
```
Karsten1987: didn't we say that `get_metadata()` should not be implemented in the individual storage plugins any longer?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -278,7 +265,11 @@ rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
   metadata.starting_time =
     std::chrono::time_point<std::chrono::high_resolution_clock>(std::chrono::nanoseconds(min_time));
   metadata.duration = std::chrono::nanoseconds(max_time) - std::chrono::nanoseconds(min_time);
-  metadata.bag_size = rosbag2_storage::FilesystemHelper::calculate_directory_size(database_name_);
+  metadata.bag_size = 0;
+
+  for (const auto & relative_path : metadata.relative_file_paths) {
```
Karsten1987: I am confused. How can `metadata.relative_file_paths` ever contains multiple files in the current implementation?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
```diff
@@ -91,7 +91,9 @@ class StorageTestFixture : public TemporaryDirectoryFixture
     std::unique_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> writable_storage =
       std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
 
-    writable_storage->open(temporary_dir_path_);
+    auto db_file = rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_, "rosbag"});
```
Karsten1987: why that change? Why appending `rosbag` to the filename?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -126,76 +132,38 @@ TEST_F(StorageTestFixture, get_all_topics_and_types_returns_the_correct_vector)
   }));
 }
 
-TEST_F(StorageTestFixture, get_metadata_returns_correct_struct) {
```
Karsten1987: So if you decide to delete these tests, they have to be reintegrated in the rosbag2 API. I think it makes sense to modify this PR to also remove the `get_metadata()` function from the storage interfaces and handle the logic completely in the c++ API of rosbag2.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -140,6 +141,7 @@ class RecordFixture : public TemporaryDirectoryFixture
 
   std::string bag_path_;
   std::string database_path_;
+  std::string storage_path_;
```
Karsten1987: What's the difference between these three variables now? What's the difference between `database_path_` and `storage_path_`?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -178,4 +205,20 @@ void Writer::finalize_metadata()
   }
 }
 
+std::string Writer::next_bagfile_uri()
+{
+  std::stringstream db_name;
```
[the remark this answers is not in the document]
zmichaels11: Would `storage_name` be better?

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -178,4 +205,20 @@ void Writer::finalize_metadata()
...
+  db_name << rosbag2_storage::FilesystemHelper::get_folder_name(uri_);
+
+  // Only append the counter after we have split.
+  // This is so bagfiles have the same naming conventions if splitting is disabled.
```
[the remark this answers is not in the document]
zmichaels11: The plan was to try and not break anything that depends on non-split bagfiles.
So the first bagfile is named `<bagfile>` with the first split being `<bagfile>_1`. 

This should allow anything that expects the old naming convention to continue to work, even if bagsplitting is enabled. (although only the first segment would be read in that case)

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: Yes, I opened Issue #197 which includes removing `get_metadata()` from the storage plugins. I wanted to keep it separate from this PR since that's an API-breaking change.

rosbag2/src/rosbag2/writer.cpp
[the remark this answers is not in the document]
Karsten1987: What do you have in mind which could expect the old naming convention?
The reader side generally shouldn't care whether it's prefixed or just composed by random file paths.

rosbag2/src/rosbag2/writer.cpp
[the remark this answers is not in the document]
Karsten1987: yes, that sounds better. Or something with `storage_file_name`.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: If you decide to do that in a follow PR, then we can't delete the tests for it though.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
zmichaels11: `Writer` consumes `bag_path_`, `SqliteStorage` consumes `storage_path_`, and `SqliteWrapper` can consume `database_path_`

This is related to what I commented here: https://github.com/ros2/rosbag2/pull/185#issuecomment-543895687.

The e2e test in `rosbag2_tests` opens the database either through `Writer`, `SqliteStorage`, or `SqliteWrapper`. 
`SqliteWrapper` hasn't changed and expects the path to the db, which is `database_path_`
`SqliteStorage` now also expects the db as well, but without the extension since that may be plugin-specific. This is `storage_path_`
`Writer` expects the base path, which is `bag_path_`.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
zmichaels11: `SqliteStorage` is being invoked directly here, so `db_file` needs to be the path to the actual database storage. Before this change, it was expecting the base folder since storage also interacted with the metadata file.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: Ok, I'll block this PR on that Issue and open up a new PR.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -68,20 +69,21 @@ void Writer::open(
...
+  uri_ = storage_options.uri;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
   {
     converter_ = std::make_unique<Converter>(converter_options, converter_factory_);
   }
 
-  storage_ = storage_factory_->open_read_write(storage_options.uri, storage_options.storage_id);
+  auto bagfile_uri = next_bagfile_uri();
```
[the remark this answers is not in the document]
zmichaels11: Applied change.

rosbag2/src/rosbag2/writer.cpp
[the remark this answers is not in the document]
zmichaels11: Renamed to `storage_file_name`

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -178,4 +205,20 @@ void Writer::finalize_metadata()
   }
 }
 
+std::string Writer::next_bagfile_uri()
```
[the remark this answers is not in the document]
zmichaels11: * replaced `next_bagfile_uri` with free function `format_bagfile_uri(base_folder, storage_count)`
* Removed `bagfile_counter_` and used  `metadata_.relative_file_paths.size()` instead

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
zmichaels11: `get_metadata` will be removed in Issue #197. Restored test for this PR.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
Karsten1987: > SqliteStorage now also expects the db as well, but without the extension since that may be plugin-specific. This is storage_path_

I don't think I understand this. In the metadata.yaml, there should be a path to an existing file listed in `relative_file_paths`. This filepath should then be passed to a `Reader` which opens it through the appropriate plugin. The Sqlite storage plugin should thus be able to handle a file which ends in `.db3`.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -25,10 +23,30 @@
...
+{
+std::string format_storage_uri(const std::string & base_folder, uint64_t storage_count)
+{
+  // Right now `base_folder_` is always just the folder name for where to install the bagfile.
+  // The name of the folder needs to be queried in case Writer is opened with a relative path.
+  std::stringstream storage_file_name;
+  storage_file_name << rosbag2_storage::FilesystemHelper::get_folder_name(base_folder);
+
+  // Only append the counter after we have split.
+  // This is so bagfiles have the old naming convention if splitting is disabled.
```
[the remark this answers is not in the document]
zmichaels11: I was talking with @piraka9011 about this last week, part of the reason for supporting the old convention is so the codebase doesn't break if this merges in before the Reader changes.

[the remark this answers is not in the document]
Karsten1987: why would that be the case? The relative file path doesn't follow any naming conventions at the moment. The only requirement is that it points to a valid file path.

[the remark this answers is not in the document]
zmichaels11: I thought about this one a bit more and I think the only way it would break anything is if something is making assumptions on what the bagfiles are named instead of reading the metadata. Reading the bagfile relative path from the metadata should be the correct way to retrieve the bagfiles.

I'll change this not have the branch

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -48,43 +48,29 @@ namespace rosbag2_storage_plugins
...
-
-    database_name_ = metadata->relative_file_paths[0];
-  } else {
-    if (is_read_only(io_flag)) {
-      throw std::runtime_error("Failed to read from bag '" + uri + "': No metadata found.");
-    }
-
-    database_name_ = rosbag2_storage::FilesystemHelper::get_folder_name(uri) + ".db3";
-  }
+  relative_path_ = uri + ".db3";
```
Karsten1987: Looking at this again, I think this is not the correct approach. A call to `open()` should not compose a URI. 

In the existing implementation, we could do that because of loading the metadata at the same time. I think the write approach here is that we introduce a new function on the storage interfaces which might give you a plugin specific file extension or something which then allows you to create the individual uri in the Writer API, i.e. `format_storage_uri()`.

In both cases, read and write, the passed in `uri` should point to a valid sqlite database file name. In the case of  writing a new file, this database will be created as given in the uri.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -235,14 +221,14 @@ std::string SqliteStorage::get_storage_identifier() const
 
 std::string SqliteStorage::get_relative_path() const
 {
-  return database_name_;
+  return relative_path_;
```
Karsten1987: I thought we talked about this somewhere, but I think it makes sense to rename this function and variable name to `get_relative_file_path` to make clear that we talk about the file and not the directory or something else.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
Karsten1987: also, looking at this again, I don't understand why `rosbag` is essential here.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
```diff
@@ -113,8 +115,11 @@ class StorageTestFixture : public TemporaryDirectoryFixture
   {
     std::unique_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> readable_storage =
       std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
+
+    auto db_file = rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_, "rosbag"});
+
     readable_storage->open(
-      temporary_dir_path_, rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY);
+      db_file, rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY);
```
Karsten1987: with the changes introduced in this PR, I think this `db_file` should point to a existing `.db3` file

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: Moving the composition of the `uri` into a separate function will require changes to `StorageFactory` and `StorageFactoryImpl` since `Writer` does not know what the extension should be until after the storage interface has been opened.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: Renamed `get_relative_path` to `get_relative_file_path`

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: This change was done because `calculate_directory_size` was determined to not be the best way to calculate `bag_size` in another issue.

Since `metadata.relative_file_paths` is hardcoded now to `get_relative_file_paths()`, I am replacing this accumulating for-loop with `calculate_directory_size(get_relative_file_path())`

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: An option I have considered is to have `SqliteStorage::open` accept `uri` with and without the extension. In this case, `Writer` can continue to open the plugin via `StorageFactory` without knowing the extension and `Reader` can expect the storage plugin to open the bagfile with the relative file path obtained from the metadata.

@Karsten1987  would this suggestion work for you?

Karsten1987: I think that would work. I think we can assume that every time we pass in `IOFlag::READ_WRITE` a new file is created - opposed to `IOFlag::APPEND`. see https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp#L30-L31

In the case of `IOFlag::READ_WRITE` the `.db3` suffix will be added.

Karsten1987:
@zmichaels11 do you mind not force pushing after a round of reviews? That makes it easier for me to keep track of individual changes.

zmichaels11:
> @zmichaels11 do you mind not force pushing after a round of reviews? That makes it easier for me to keep track of individual changes.

sorry about that

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: Ok, I'll make the change to apply `.db3` to the `uri` when `READ_WRITE` is passed.

zmichaels11: I've applied this change.

`SqliteStorage::open(uri, READ_WRITE)` will append the extension, since `Writer` calls this indirectly through `StorageFactory`, before it knows what extension the `db` should be.

`SqliteStorage::open(uri, READ_ONLY)` expects uri to have the extension since `Reader` calls this by passing a `relative_file_path` read from metadata.

`APPEND` behaves similarly to `READ_ONLY`.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -132,6 +146,28 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
...
+  if (!storage_) {
+    std::stringstream errmsg;
+    errmsg << "Failed to rollover bagfile to new file: \"" << storage_uri << "\"!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  metadata_.relative_file_paths.push_back(storage_->get_relative_file_path());
+
+  // Re-register all Topics since we rolled-over to a new bagfile.
```
Karsten1987: nit: `topics`

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
```diff
@@ -38,7 +38,7 @@ class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
    *
    * \returns the relative path.
    */
-  virtual std::string get_relative_path() const = 0;
+  virtual std::string get_relative_file_path() const = 0;
```
Karsten1987: This function makes more sense in the `io` interface, what do you think?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -139,14 +150,16 @@ TEST_F(StorageTestFixture, get_metadata_returns_correct_struct) {
...
+  const auto db_filename =
+    rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_, "rosbag.db3"});
+
+  readable_storage->open(db_filename, rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY);
+  const auto metadata = readable_storage->get_metadata();
 
   EXPECT_THAT(metadata.storage_identifier, Eq("sqlite3"));
   EXPECT_THAT(metadata.relative_file_paths, ElementsAreArray({
-    rosbag2_storage::FilesystemHelper::get_folder_name(temporary_dir_path_) + ".db3"
+    db_filename
```
Karsten1987: this could fit into one line, right?

Karsten1987: same below.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -184,18 +199,58 @@ TEST_F(StorageTestFixture, get_metadata_returns_correct_struct_if_no_messages) {
 TEST_F(StorageTestFixture, remove_topics_and_types_returns_the_empty_vector) {
   std::unique_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> writable_storage =
     std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
-  writable_storage->open(temporary_dir_path_);
+
+  // extension is omitted since storage is created; io_flag = READ_WRITE
+  const auto read_write_filename = rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_,
```
Karsten1987: nit: Maybe break after the `=`?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -40,51 +40,68 @@ bool database_exists(const std::string & uri)
...
-
-    database_name_ = metadata->relative_file_paths[0];
-  } else {
-    if (is_read_only(io_flag)) {
-      throw std::runtime_error("Failed to read from bag '" + uri + "': No metadata found.");
-    }
-
-    database_name_ = rosbag2_storage::FilesystemHelper::get_folder_name(uri) + ".db3";
+  if (is_read_write(io_flag)) {
+    relative_path_ = uri + FILE_EXTENSION;
```
Karsten1987: what happens if that file already exists?

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
zmichaels11: I have an issue to move this #197 which I'm picking up after bagsplit merges.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: imo: it should throw an exception.

I looked into how its handled in `SqliteWrapper` and `APPEND` and `READ_WRITE` are handled the same.

Unless you can provide an alternative, I'm going to rewrite this section to `throw std::runtime_error` when `is_read_write(io_flag) && database_exists(relative_path_)`

zmichaels11: @Karsten1987 I can push the changes if you're fine with it throwing when file exists and `READ_ONLY`.

Karsten1987: sounds good to me.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
zmichaels11: Applied suggestion

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
zmichaels11: Applied suggestion

rosbag2/src/rosbag2/writer.cpp
zmichaels11: Applied suggestion

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -40,51 +40,73 @@ bool database_exists(const std::string & uri)
...
 
-  if (metadata) {
-    if (metadata->relative_file_paths.empty()) {
+    // READ_WRITE requires the DB to not exist.
+    if (database_exists(relative_path_)) {
       throw std::runtime_error(
-              "Failed to read from bag '" + uri + "': Missing database file path in metadata");
+              "Failed to create bag: File '" + relative_path_ + "' file already exists!");
```
Karsten1987: duplicate `file`

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -40,51 +40,73 @@ bool database_exists(const std::string & uri)
...
-  ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO_STREAM("Opened database '" << uri << "'.");
+  // Reset the read and write statements in case the database changed.
+  // These will be reinitialized lazily on the first read or write.
+  read_statement_ = nullptr;
+  write_statement_ = nullptr;
+
+  ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO_STREAM("Opened database '" <<
+    relative_path_ <<
+    "' for " <<
+    to_string(io_flag) << ".");
```
Karsten1987: That should fit into two lines, shouldn't it?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -95,7 +97,9 @@ TEST_F(StorageTestFixture, get_next_returns_messages_in_timestamp_order) {
   write_messages_to_sqlite(string_messages);
   std::unique_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> readable_storage =
     std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
-  readable_storage->open(temporary_dir_path_);
+
+  auto db_filename = rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_, "rosbag.db3"});
```
Karsten1987: nit: consistent line breaking, see above where line breaks happen after the `=`

Karsten1987:
so when testing this I get the following:

```
 ➭ ros2 bag record -a
[ERROR] [rosbag2_storage]: Could not open 'rosbag2_2019_11_12-15_48_47/rosbag2_2019_11_12-15_48_47_0' with 'sqlite3'. Error: Failed to setup storage. Error: Could not read-write open database. Error code: 14
[ERROR] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
[ERROR] [rosbag2_transport]: Failed to record: No storage could be initialized. Abort
```

zmichaels11:
are you using an up-to-date copy of the repo?
I get:
```
ros2 bag record -a
[INFO] [rosbag2_storage]: Opened database 'rosbag2_2019_11_12-16_09_42/rosbag2_2019_11_12-16_09_42_0.db3' for READ_WRITE.
[INFO] [rosbag2_transport]: Listening for topics...
[INFO] [rosbag2_transport]: Subscribed to topic '/rosout'
[INFO] [rosbag2_transport]: Subscribed to topic '/parameter_events'
[INFO] [rosbag2_transport]: Subscribed to topic '/test_topic'
^C[INFO] [rclcpp]: signal_handler(signal_value=2)
```

Karsten1987:
Have you played around with the [WAL](https://www.sqlite.org/wal.html#avoiding_excessively_large_wal_files)? From what I understand in the documentation and also see when running your branch, is that this files gets 4mb in size before it actually writes the data to the `.db3` file. 

I see that the `.db3` file has around 84k (I set the max bag file to 100k), when the 4mb are synced and I wonder how that is possible. That being said, I really would like to have that test to see if actually all messages are correctly stored in the bag. 84k vs 4mb seems kind of odd to me.

zmichaels11:
I'm looking into adding a test for validating bagsplitting in this PR.

Its going to be a little more intensive than just duplicating and modifying a test in `rosbag2 tests` that invokes `ros2 bag record` since enabling bagsplitting from `ros2 bag` is in a different PR.

I'm thinking its probably best to add a unit test for bagsplitting on a mocked storage since the operation of bagsplitting is owned by `Writer`.

Karsten1987:
> Its going to be a little more intensive than just duplicating and modifying a test in rosbag2 tests that invokes ros2 bag record since enabling bagsplitting from ros2 bag is in a different PR.

Not sure I completely understand this. The command line option doesn't really play a role in this new PR. I do believe we (just) need a system test in ros2_tests which basically just opens a Writer instance, manually adds a couple of messages and verifies that all messages are correctly stored across various bag files. That tests has to be successful for all provided storage implementations.
This can be achieved independent of the command line tools and even the transport layer.

> I'm thinking its probably best to add a unit test for bagsplitting on a mocked storage since the operation of bagsplitting is owned by Writer.

This makes sense to me for validating the splitting logic within the Writer itself.

zmichaels11:
Added a unit test that validates that Writer is splitting and the bagfiles increment correctly.

rosbag2/src/rosbag2/writer.cpp
```diff
@@ -132,6 +146,28 @@ void Writer::remove_topic(const TopicMetadata & topic_with_type)
   }
 }
 
+void Writer::split_bagfile()
+{
+  const auto storage_uri = format_storage_uri(
+    base_folder_,
+    metadata_.relative_file_paths.size());
```
piraka9011: One issue I came across with the decompress is that I had to add 1 to the size since the file wasn't pushed into the relative_file_paths prior to `format_storage_uri` being called. Not sure if that's addressed here.

zmichaels11: Are you basing off of an older version? metadata's `relative_file_paths` is initialized with the initial storage's `relative_file_path` which is formatted with `storage_count=0`.

The unit test I wrote should be validating that the relative file paths are incrementing from 0 to n

zmichaels11:
> Not sure I completely understand this. The command line option doesn't really play a role in this new PR. I do believe we (just) need a system test in ros2_tests which basically just opens a Writer instance, manually adds a couple of messages and verifies that all messages are correctly stored across various bag files. That tests has to be successful for all provided storage implementations.
> This can be achieved independent of the command line tools and even the transport layer.

I took a crack at that and I think its going to require enough effort to justify a new PR. Most of what you suggested works, except that there isn't a cross-platform way to create directory structures in c++. 

I think that we should wrap up this PR, port `filesystem::create_directory` from C++17 as another PR, and include the integration test with the work defined in https://github.com/ros-security/aws-roadmap/issues/10

I have added a unit test to `Writer` that [verifies splitting](https://github.com/aws-ros-dev/rosbag2/blob/1fba68b175870e4eb170dc315d5cb7755be203e3/rosbag2/test/rosbag2/test_writer.cpp#L163)

Karsten1987:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8645/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4488/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8564/)

The changes for the writer API look good to me now. However, I am unable to play a bag file with this PR. I presume it's because the changes for loading the metadata appropriately on the reader side are not part of this PR. 

```
 ➭ ros2 bag play rosbag2_2019_11_14-14_16_02
[ERROR] [rosbag2_storage]: Could not open 'rosbag2_2019_11_14-14_16_02' with 'sqlite3'. Error: Failed to setup storage. Error: Could not read-only open database. Error code: 10
[ERROR] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
[ERROR] [rosbag2_transport]: Failed to play: No storage could be initialized. Abort
```

Pointing me to the direction of 
```c++
  if (is_read_write(io_flag)) {
    relative_path_ = uri + FILE_EXTENSION;

    // READ_WRITE requires the DB to not exist.
    if (database_exists(relative_path_)) {
      throw std::runtime_error(
              "Failed to create bag: File '" + relative_path_ + "' already exists!");
    }
  } else {  // APPEND and READ_ONLY
    relative_path_ = uri;

    // APPEND and READ_ONLY require the DB to exist
    if (!database_exists(relative_path_)) {
      throw std::runtime_error(
              "Failed to read from bag: File '" + relative_path_ + "' does not exist!");
    }
  }
```
which lets me believe that the relative path is still set to the base folder path and the actual metadata is not being read. 

As a side note: This also means that `database_exists()` is not as accurate as it should be. It will check whether _a_ file exists, not if that file actually represents a valid database.

How do you guys want to proceed with this? I can't merge this PR as-is as this would break rosbag2 because of not being able to replay bag files.

zmichaels11:
I had @piraka9011 look at this since he has context on the Reader.
He's going to push a commit with the minimal number of changes required to get this functionality working.

zmichaels11:
@Karsten1987 I think we resolved your issues.
Anas pushed a commit that resolved the Reader issues.
I pushed a commit that has `SqliteWrapper` throw if the database is bad.

rosbag2/test/rosbag2/test_sequential_reader.cpp
```diff
@@ -22,9 +22,11 @@
 #include "rosbag2/sequential_reader.hpp"
 #include "rosbag2_storage/bag_metadata.hpp"
 #include "rosbag2_storage/topic_metadata.hpp"
+#include "rosbag2_storage/metadata_io.hpp"
```
Karsten1987: alpha order

rosbag2/test/rosbag2/test_sequential_reader.cpp
```diff
@@ -38,48 +40,48 @@ class SequentialReaderTest : public Test
...
   std::unique_ptr<rosbag2::SequentialReader> reader_;
+  std::unique_ptr<MockMetadataIo> metadata_io_;
+  rosbag2_storage::BagMetadata metadata_;
+  std::string storage_serialization_format_;
 };
 
-TEST_F(SequentialReaderTest, read_next_uses_converters_to_convert_serialization_format) {
-  std::string storage_serialization_format = "rmw1_format";
+TEST_F(SequentialReaderTest, read_next_uses_converters_to_convert_serialization_format)
+{
```
Karsten1987: why that change? I think `{` stays on the same line for GTEST

Karsten1987: same below.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -73,15 +67,15 @@ void SqliteStorage::open(
...
+    if (rosbag2_storage::FilesystemHelper::file_exists(relative_path_)) {
       throw std::runtime_error(
               "Failed to create bag: File '" + relative_path_ + "' already exists!");
     }
   } else {  // APPEND and READ_ONLY
     relative_path_ = uri;
 
     // APPEND and READ_ONLY require the DB to exist
-    if (!database_exists(relative_path_)) {
+    if (!rosbag2_storage::FilesystemHelper::file_exists(relative_path_)) {
```
Karsten1987: If I put in a valid file, which doesn't point to a `.db3` file, this check will still return true. Do you think it makes sense to check for a `.db3` suffix?

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,11 +45,12 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
+  metadata_ = metadata_io_->read_metadata(storage_options.uri);
   storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
```
Karsten1987: So how does this help? We still pass the same `uri` to the open call. How did you test this latest change?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
zmichaels11: I'm unsure on this one. Sqlite3 doesn't actually care what the extension is and Line 78 is enforcing that the database exists. The line I added to `SqliteWrapper` constructor throws when the database cannot be opened or if its invalid, so validity is being checked.

Do you think, from a design point of view, that `SqliteStorage` should only be allowed to open databases whose `uri` ends with `db3`?

rosbag2/src/rosbag2/sequential_reader.cpp
zmichaels11: I just noticed that `test_rosbag2_play_end_to_end` is disabled

zmichaels11: Ok I changed `open_read_only` to accept `metadata_.relative_file_paths[0]`

Steps to test:
1. run `ros2 bag record -a` in first terminal
2. run `ros2 run demo_nodes_py talker` in another terminal
3. ctrl+c second terminal
4. ctrl+c first terminal
5. run `ros2 run listener` in first terminal
6. run `ros2 bag play rosbag2_2019_11_14-16_19_43/`
7. kill all terminals

Output:
```
[INFO] [listener]: I heard: [Hello World: 0]
[INFO] [listener]: I heard: [Hello World: 1]
```

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: i guess that's okay then.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,11 +45,13 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
-  storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
+  metadata_ = metadata_io_->read_metadata(storage_options.uri);
+  storage_ = storage_factory_->open_read_only(
+    metadata_.relative_file_paths[0], storage_options.storage_id);
```
Karsten1987: this should at least check if `relative_file_paths` is non empty otherwise this will lead to a segfault.

Karsten1987:
Except that one outstanding comment, that looks good to me.

piraka9011:
@Karsten1987 added a check!

zmichaels11:
@piraka9011 I was actually about to do the same, except throw instead.
You need to add in `test_sequential_reader_cpp` 
```c++
metadata_.relative_file_paths = {"dummy path"};
```

piraka9011:
I followed the same convention as the next check for the topics in L#58. Updating tests.

Karsten1987:
An immediate return is fine, but I think in both cases we should log a warning saying that either no concrete files are found or no topics are listed.

piraka9011:
@zmichaels11 Updated tests
@Karsten1987 Added warning

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8646/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4489/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7053/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8565/)

PR #186, opened 2019-10-18
Add function to BaseIOInterface to retrieve backing file

zmichaels11:

### Changes
* Add `get_relative_path` to `BaseIOInterface`. This was introduced in response to https://github.com/ros2/rosbag2/pull/185#issuecomment-543895687 

### Dependent PRs
- [x] PR #182 
- [x] PR #183 

### Issues
* ros-security/aws-roadmap#11

This PR will require changes to PR #185

### DIFF
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
@@ -30,6 +30,13 @@ public:
   virtual ~BaseInfoInterface() = default;
 
   virtual BagMetadata get_metadata() = 0;
+
+  /**
+   * Retrieves the relative path to the backing of the storage plugin.
+   *
+   * \returns the relative path.
+   */
+  virtual std::string get_relative_path() const = 0;
 };
 
 }  // namespace storage_interfaces
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
@@ -16,6 +16,8 @@
 #ifndef ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
 #define ROSBAG2_STORAGE__STORAGE_INTERFACES__BASE_INFO_INTERFACE_HPP_
 
+#include <string>
+
 #include "rosbag2_storage/bag_metadata.hpp"
 #include "rosbag2_storage/visibility_control.hpp"
 
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -64,6 +64,8 @@ public:
 
   rosbag2_storage::BagMetadata get_metadata() override;
 
+  std::string get_relative_path() const override;
+
   uint64_t get_bagfile_size() const override;
 
   std::string get_storage_identifier() const override;
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -233,6 +233,11 @@ std::string SqliteStorage::get_storage_identifier() const
   return "sqlite3";
 }
 
+std::string SqliteStorage::get_identifier() const
+{
+  return "sqlite3";
+}
+
 rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
 {
   rosbag2_storage::BagMetadata metadata;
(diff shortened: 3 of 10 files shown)


zmichaels11:
@Karsten1987 this is the next PR

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
```diff
@@ -30,6 +32,13 @@ class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
   virtual ~BaseInfoInterface() = default;
 
   virtual BagMetadata get_metadata() = 0;
+
+  /**
+   * Retrieves the relative path to the backing of the storage plugin.
+   *
+   * \returns the relative path.
+   */
+  virtual std::string get_relative_path() const = 0;
```
Karsten1987: I don't think this function should be on the `BaseInfoInterface` class. This would somehow conflict with `get_metadata()` as I would expect the `get_metadata()` to have a fleshed out metadata with all relative paths set (in the reader case for example).

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -64,6 +64,8 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
 
   rosbag2_storage::BagMetadata get_metadata() override;
 
+  std::string get_relative_path() const override;
```
Karsten1987: Thinking about this, I believe `get_metadata()` should not be part of each storage interface as it is actually storage agnostic. That being said, `get_relative_path()` should IMO just be renamed to `get_current_file_path()` or similar. The metadata should then be solely handled in the writer/reader level. 
Otherwise, I feel that `get_relative_path()` is in conflict with `get_metadata()` as the latter should be returning a struct with all the relative paths set to it.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -233,6 +233,11 @@ std::string SqliteStorage::get_storage_identifier() const
   return "sqlite3";
 }
 
+std::string SqliteStorage::get_relative_path() const
```
Karsten1987: I'd rename this to `get_current_file_path()` or similar.

zmichaels11: Since `Writer` handles bagfile splitting, the storage plugin only knows about a single file path.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
zmichaels11: PR #184 adds `generate_metadata` to `Writer` and PR #185 replaces the calls to the storage plugin's `get_metadata` with `Writer`'s `generate_metadata`.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
zmichaels11: I agree here. This PR and #184 are moving ownership of metadata out of the storage plugin and into `Writer`. 

The reason it is `get_relative_path()` instead of `get_current_file_path` or `get_relative_paths` is because only one path is opened by the storage plugin when the Writer invokes `open_read_write`. In the event of bagfile splitting, `Writer` calls `open_read_write` again once the bagfile should split, but with the next relative file path.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
Karsten1987: Can we enhance this PR then to be somewhat feature complete? I think that makes reviewing the PR easier.

zmichaels11: I can merge this PR with #184 in that case. I was trying to keep the number of changes to a minimum.

Karsten1987: that would make sense to me. I appreciate splitting up changes into multiple PRs, however I'd like to have them clustered by features. So one feature could be to extract the metadata generation from the storage plugins into a writer/reader API.
In this particular case, please also make sure that this change has to be applicable to the `rosbag2_bag_v2` plugin, which might not have a `metadata.yaml` file: https://github.com/ros2/rosbag2_bag_v2

zmichaels11: Ok. I'll close this PR after I finish integrating these changes into #184

Karsten1987:
@zmichaels11 Can you tell me which of the three PRs you want me to review now? I am a bit confused whether it's #184 #185 or indeed this one as they are all somewhat related to the same thing as far as I can tell from this point.

zmichaels11:
@Karsten1987 sorry for the confusion.
I'm merging this PR into #184, so that will be the next one for you to review. I'll ping you again shortly after I finish testing.

zmichaels11:
Merged into #184

PR #187, opened 2019-10-18
Minimal multifile reader

piraka9011:
Introduce the minimal changes needed to read multiple storage files.
Part of an effort to rework PR #158 into smaller PRs.

@Karsten1987 this will be the first step for the reader/writer interface changes. Adding this will enable us to at least test the splitting feature end-to-end.

### Changes 

* Add `has_next_file()` and `get_next_file()` private methods to `SequentialReader`.
* Add private variables for storing a list of files to read from and current file being read.

### DIFF
--- a/rosbag2/include/rosbag2/sequential_reader.hpp
+++ b/rosbag2/include/rosbag2/sequential_reader.hpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 
+#include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
 #include "rosbag2_storage/storage_factory_interface.hpp"
 #include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
@@ -51,7 +52,9 @@ public:
     std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
     std::make_unique<rosbag2_storage::StorageFactory>(),
     std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
-    std::make_shared<SerializationFormatConverterFactory>());
+    std::make_shared<SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
 
   virtual ~SequentialReader();
 
@@ -69,7 +72,8 @@ public:
    * \param converter_options Options for specifying the output data format
    */
   virtual void open(
-    const StorageOptions & storage_options, const ConverterOptions & converter_options);
+    const StorageOptions & storage_options,
+    const ConverterOptions & converter_options);
 
   /**
    * Ask whether the underlying bagfile contains at least one more message.
@@ -126,12 +130,12 @@ public:
    */
   virtual std::string get_current_uri() const;
 
-
 private:
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
-  std::unique_ptr<rosbag2_storage::BagMetadata> metadata_;
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
+  rosbag2_storage::BagMetadata metadata_;
   std::unique_ptr<Converter> converter_;
   StorageOptions storage_options_{};
   std::vector<std::string> file_paths_{};  // List of database files.
--- a/rosbag2/include/rosbag2/sequential_reader.hpp
+++ b/rosbag2/include/rosbag2/sequential_reader.hpp
@@ -41,7 +41,7 @@ namespace rosbag2
 {
 /**
  * The SequentialReader allows opening and reading messages of a bag. Messages will be read
- * sequentially according to timestamp.
+ * sequentially according to timestamp. Functions are virtual so that they can be mocked in tests.
  */
 class ROSBAG2_PUBLIC SequentialReader
 {
@@ -99,7 +99,6 @@ public:
    */
   virtual std::vector<TopicMetadata> get_all_topics_and_types();
 
-private:
   /**
    * Ask whether there is another database file to read from the list of relative
    * file paths.
@@ -109,21 +108,34 @@ private:
   virtual bool has_next_file() const;
 
   /**
-   * Update the current database file to read from and remove the first (front)
-   * element in the list of files to read from.
+   * Increment the current file iterator to point to the next file in the list of relative file
+   * paths.
    *
    * Expected usage:
-   * if (has_next_file()) get_next_file();
+   * if (has_next_file()) load_next_file();
    */
-  std::string get_next_file();
+  virtual void load_next_file();
 
+  /**
+   * Return the relative file path pointed to by the current file iterator.
+   */
+  virtual std::string get_current_file() const;
+
+  /**
+   * Return the URI of the current file (i.e. no extensions).
+   */
+  virtual std::string get_current_uri() const;
+
+
+private:
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
+  std::unique_ptr<rosbag2_storage::BagMetadata> metadata_;
   std::unique_ptr<Converter> converter_;
   StorageOptions storage_options_{};
-  std::vector<std::string> file_paths_ {};  // List of database files.
-  std::vector<std::string>::iterator current_file_iterator_ {};  // Index of file to read from
+  std::vector<std::string> file_paths_{};  // List of database files.
+  std::vector<std::string>::iterator current_file_iterator_{};  // Index of file to read from
 };
 
 }  // namespace rosbag2
--- a/rosbag2/include/rosbag2/sequential_reader.hpp
+++ b/rosbag2/include/rosbag2/sequential_reader.hpp
@@ -100,10 +100,30 @@ public:
   virtual std::vector<TopicMetadata> get_all_topics_and_types();
 
 private:
+  /**
+   * Ask whether there is another database file to read from the list of relative
+   * file paths.
+   *
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file() const;
+
+  /**
+   * Update the current database file to read from and remove the first (front)
+   * element in the list of files to read from.
+   *
+   * Expected usage:
+   * if (has_next_file()) get_next_file();
+   */
+  std::string get_next_file();
+
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  StorageOptions storage_options_{};
+  std::vector<std::string> file_paths_ {};  // List of database files.
+  std::vector<std::string>::iterator current_file_iterator_ {};  // Index of file to read from
 };
 
 }  // namespace rosbag2
--- a/rosbag2/src/rosbag2/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/sequential_reader.cpp
@@ -53,6 +53,8 @@ SequentialReader::open(
     return;
   }
   file_paths_ = bag_metadata.relative_file_paths;
+  assert(!file_paths_.empty());
+  current_file_iterator_ = file_paths_.begin();
 
   // Currently a bag file can only be played if all topics have the same serialization format.
   auto storage_serialization_format = topics[0].topic_metadata.serialization_format;
@@ -77,19 +79,14 @@ SequentialReader::open(
 
 std::string SequentialReader::get_next_file()
 {
-  {
     assert(current_file_iterator_ != file_paths_.end());
     current_file_iterator_++;
     return *current_file_iterator_;
-  }
 }
 
 bool SequentialReader::has_next_file() const
 {
-  {
-    assert(!file_paths_.empty());
     return current_file_iterator_ + 1 != file_paths_.end();
-  }
 }
 
 bool SequentialReader::has_next()
--- a/rosbag2/src/rosbag2/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/sequential_reader.cpp
@@ -79,14 +79,14 @@ SequentialReader::open(
 
 std::string SequentialReader::get_next_file()
 {
-    assert(current_file_iterator_ != file_paths_.end());
-    current_file_iterator_++;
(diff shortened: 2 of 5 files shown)


rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -100,10 +100,30 @@ class ROSBAG2_PUBLIC SequentialReader
...
+   */
+  std::string get_next_file();
+
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  StorageOptions storage_options_;
+  std::vector<std::string> file_paths_ {};  // List of database files.
+  std::size_t current_file_offset_ {0};  // Index of file to read from file_paths_.
```
zmichaels11: I believe this has the same as https://github.com/ros2/rosbag2/pull/182#discussion_r336274369

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +75,34 @@ SequentialReader::open(
...
+}
+
 bool SequentialReader::has_next()
 {
   if (storage_) {
+    // If there's no new message, check if there's at least another file to read and update storage
+    // to read from there. Otherwise, check if there's another message.
+    if (!storage_->has_next()) {
+      if (has_next_file()) {
+        std::string current_file = get_next_file();
```
zmichaels11:
```c++
const auto current_file = get_next_file()
```

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,6 +41,7 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
+  storage_options_; = storage_options;
```
zmichaels11: ?

rosbag2/src/rosbag2/sequential_reader.cpp
zmichaels11: btw this one is a suggestion.

rosbag2/src/rosbag2/sequential_reader.cpp
piraka9011: :)

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -100,10 +100,30 @@ class ROSBAG2_PUBLIC SequentialReader
...
+   */
+  std::string get_next_file();
+
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  StorageOptions storage_options_{};
+  std::vector<std::string> file_paths_ {};  // List of database files.
+  std::size_t current_file_offset_ {0};  // Index of file to read from file_paths_.
```
prajakta-gokhale: Prefer using `uint64_t` instead of `size_t` wherever possible, to maintain cross-architecture compatibility.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -50,6 +52,9 @@ SequentialReader::open(
     return;
   }
 
+  const auto bag_metadata = storage_->get_metadata();
```
Karsten1987: could this be combined with line 50?
```
auto topics = storage_->get_metadata().topics_with_message_count;
```

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -100,10 +100,30 @@ class ROSBAG2_PUBLIC SequentialReader
...
+   */
+  std::string get_next_file();
+
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  StorageOptions storage_options_{};
+  std::vector<std::string> file_paths_ {};  // List of database files.
+  uint64_t current_file_offset_ {0};  // Index of file to read from file_paths_.
```
Karsten1987: could this maybe just be an iterator of the vector above?

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +76,34 @@ SequentialReader::open(
   }
 }
 
+std::string SequentialReader::get_next_file()
+{
+  {
+    assert(current_file_offset_ < file_paths_.size());
+    current_file_offset_++;
+    return file_paths_.at(current_file_offset_);
+  }
```
Karsten1987: iterator could make things easier here as `get_next_file()` essentially would only be dereferencing it

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +76,34 @@ SequentialReader::open(
...
+}
+
+bool SequentialReader::has_next_file() const
+{
+  {
+    assert(!file_paths_.empty());
+    return current_file_offset_ + 1 < file_paths_.size();
+  }
```
Karsten1987: this could potentially become a one-liner when using an iterator
```
return it_ != file_paths_.end()
```

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -50,6 +52,9 @@ SequentialReader::open(
     return;
   }
 
+  const auto bag_metadata = storage_->get_metadata();
+  file_paths_ = bag_metadata.relative_file_paths;
```
Karsten1987: do we need this? Could we not just use the bag_metadata and store this as a class member?

piraka9011:
> I think using an iterator here makes things a bit easier and less verbose.

Initial implementation was using an iterator. We had an internal discussion and decided on indexing, however, we can go with iterator as well.

rosbag2/src/rosbag2/sequential_reader.cpp
piraka9011: Just in the constructor. Simplified per suggestion above.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +76,34 @@ SequentialReader::open(
   }
 }
 
+std::string SequentialReader::get_next_file()
+{
+  {
+    assert(current_file_iterator_ != file_paths_.end());
```
zmichaels11: You can store the end iterator in `open` instead of `file_paths_` since you only ever use `file_paths_` if you are comparing to the end iterator.

Karsten1987:
@piraka9011 is that ready for review? If so, I'd very much like to have some basic test for the new functionality, that indeed all the file paths mentioned in the metadata is being loaded and correctly iterated over.

piraka9011:
@Karsten1987 I had unit tests in another PR #192. I can move those here instead.

Karsten1987:
moving the tests in here seems fine to me as it's part of the feature.

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8550/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4421/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6970/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8462/)

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -100,10 +100,30 @@ class ROSBAG2_PUBLIC SequentialReader
   virtual std::vector<TopicMetadata> get_all_topics_and_types();
 
 private:
+  /**
+   * Ask whether there is another database file to read from the list of relative
+   * file paths.
+   *
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file() const;
```
Karsten1987: couldn't that be part of the public interface as well?

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -100,10 +100,30 @@ class ROSBAG2_PUBLIC SequentialReader
...
+   */
+  std::string get_next_file();
+
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_;
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_;
   std::unique_ptr<Converter> converter_;
+  StorageOptions storage_options_{};
+  std::vector<std::string> file_paths_ {};  // List of database files.
+  std::vector<std::string>::iterator current_file_iterator_ {};  // Index of file to read from
```
Karsten1987: I'd recommend consistency in whether you want to apply spacing between the variable name and `{}`

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,14 +42,18 @@ void
...
 {
+  storage_options_ = storage_options;
   storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
   if (!storage_) {
     throw std::runtime_error("No storage could be initialized. Abort");
   }
-  auto topics = storage_->get_metadata().topics_with_message_count;
+  const auto bag_metadata = storage_->get_metadata();
+  auto topics = bag_metadata.topics_with_message_count;
   if (topics.empty()) {
```
Karsten1987: I know this is not part of your changes, but I would feel like there should be a warning or something saying that the bagfile is empty.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +76,29 @@ SequentialReader::open(
   }
 }
 
+std::string SequentialReader::get_next_file()
+{
+  assert(current_file_iterator_ != file_paths_.end());
+  current_file_iterator_++;
+  return *current_file_iterator_;
+}
```
Karsten1987: I am having a bit of a problem with that implementation, because it does do two things in fact. It increases the file iterator and returns the new current file.
I would recommend either changing the name to something like `void load_next_file()` or let `get_next_file() const` really only return the next path and the iterator has to be advanced outside of the scope of the function.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -71,9 +76,29 @@ SequentialReader::open(
...
+
 bool SequentialReader::has_next()
 {
   if (storage_) {
+    // If there's no new message, check if there's at least another file to read and update storage
+    // to read from there. Otherwise, check if there's another message.
+    if (!storage_->has_next()) {
+      if (has_next_file()) {
+        const auto current_file = get_next_file();
```
Karsten1987: this line is the motivation for my previous comments. At this point here it's not clear to me without having insight knowledge of the code, that the `current_file_iterator_` has been moved forward and thus `open_read_only` can easily be called without calling `open` on this class again.

rosbag2/test/rosbag2/test_multifile_reader.cpp
```diff
@@ -0,0 +1,99 @@
...
+  std::unique_ptr<rosbag2::SequentialReader> reader_;
+  std::string serialization_format {"rmw1_format"};
+};
+
+TEST_F(MultifileReaderTest, has_next_reads_next_file)
+{
+  EXPECT_CALL(*storage_, has_next()).Times(AtLeast(2)).WillRepeatedly(Return(false));
+  reader_->open(rosbag2::StorageOptions(), {serialization_format, serialization_format});
+  reader_->has_next();
+  reader_->read_next();
```
Karsten1987: I am not sure I understand this test. So by calling `has_next()` and `read_next()` you expect `has_next()` to be called twice and returns both times false? But wouldn't you want to test that `has_next_file()` is getting called and returns true for the first time and false for the second time?

piraka9011: The goal of this test was to mimic how this might be used when reading from an actual file.
This might be due to my limited knowledge of gmock, but would calling `has_next()` not call `has_next_file()` as well since that's what it's doing under the hood?
If not then a better sequence would be similar to what you suggested:
1. `has_next()`        -> True
2. `has_next_file()` -> True
3. `read_next()`      -> True
4. `has_next()`        -> False
5. `has_next_file()` -> False

Karsten1987: I like the sequence you proposed. Also because I think it's easier to read and understand later on.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -41,14 +42,18 @@ void
 SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
+  storage_options_ = storage_options;
   storage_ = storage_factory_->open_read_only(storage_options.uri, storage_options.storage_id);
```
piraka9011: This should be updated to reflect the new writer changes, where the URI should change from:
`<file_name>`
to
`<directory_name>/<file_name>`

piraka9011:
@Karsten1987 I've updated the tests to reflect the changes for:
* Using metadata_io to read the metadata since `storage::get_metadata` does not get all files.
* Testing for multifile reading functionality.

I will close this PR in favor of another one that uses the reader interface in [aws-ros-dev/rosbag2#4](https://github.com/aws-ros-dev/rosbag2/pull/4) but would like your feedback on this.

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -14,21 +14,28 @@
...
 
 namespace rosbag2
 {
 
 SequentialReader::SequentialReader(
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
-  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory)
-: storage_factory_(std::move(storage_factory)), converter_factory_(std::move(converter_factory)),
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory,
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io)
```
Karsten1987: From what I see here, the only purpose of the `metadata_io` is to load the actual metadata once. I think it makes sense to consider passing in the metadata directly instead of the `io`. That potentially means the io handle to the metadata.yaml file is held open for the lifetime of this sequential reader instance, but also means I have to mock the complete io instead of just the metadata struct for testing.

piraka9011: I worry that doing so may lead us back to the same problem we had before when we loaded the metadata from the storage plugin and we don't have up to date information.

While the reader should not be responsible for ever writing metadata, there may be a time where we need to reload the metadata (ex. streaming from bags that are also being written) and it would be helpful to have metadata_io read the updated file.

That, and this new constructor would mirror the writer nicely.

Karsten1987: makes sense to me.

Karsten1987:
@piraka9011 Feel free to close this PR then.

PR #191, opened 2019-10-22
Create new interface for SequentialReader

piraka9011:
This PR refactors the implementation of the `SequentialReader` to use opaque pointers (PIMPL Idiom).
We can now preserve ABI compatibility when making changes to the reader.

This PR is part a sequence of changes to introduce bag splitting to rosbag2.

This addresses the following [AWS Roadmap issue](https://github.com/ros-security/aws-roadmap/issues/97)

## Changes

1. Introduced new pointer to implementation `ReaderImpl`.
2. Moved implementation of `SequentialReader` to `src/impl/sequential_reader.xpp`.


rosbag2/src/rosbag2/impl/sequential_reader_impl.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "rosbag2/sequential_reader.hpp"
+
+namespace rosbag2
+{
+
+class ROSBAG2_LOCAL ReaderImpl
```
thomas-moulard: please include the header containing this macro here.

rosbag2/CMakeLists.txt
```diff
@@ -40,7 +40,8 @@ add_library(${PROJECT_NAME} SHARED
   src/rosbag2/serialization_format_converter_factory.cpp
   src/rosbag2/typesupport_helpers.cpp
   src/rosbag2/writer.cpp
-  src/rosbag2/types/introspection_message.cpp)
+  src/rosbag2/types/introspection_message.cpp
```
Karsten1987: alpha sorting

rosbag2/src/rosbag2/impl/sequential_reader_impl.hpp
```diff
@@ -0,0 +1,73 @@
...
+{
+public:
+  ReaderImpl(
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory);
+
+  ~ReaderImpl();
```
Karsten1987: space between the function declarations.

rosbag2/CMakeLists.txt
prajakta-gokhale: Done.

rosbag2/src/rosbag2/impl/sequential_reader_impl.hpp
prajakta-gokhale: Fixed.

rosbag2/src/rosbag2/impl/sequential_reader_impl.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace rosbag2
+{
+
+ReaderImpl::ReaderImpl(
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory)
+: storage_factory_(std::move(storage_factory)),
+  converter_factory_(std::move(converter_factory))
+{}
+
```
Karsten1987: duplicate newline

rosbag2/src/rosbag2/impl/sequential_reader_impl.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+private:
+  /**
+   * Checks if all topics in the bagfile have the same RMW serialization format.
+   * Currently a bag file can only be played if all topics have the same serialization format.
+   *
+   * \param topics Vector of TopicInformation with metadata.
+   * \throws runtime_error if any topic has a different serialization format from the rest.
+   */
+  virtual void check_topics_serialization_formats(const std::vector<TopicInformation> & topics);
```
Karsten1987: Is the reason you declare this private function as virtual that you'd want to apply the [non-virtual interface idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Non-Virtual_Interface) here?

prajakta-gokhale:
> I understand what you are doing here and I principally agree with an impl approach to save ABI compatibility. However, the problem I have with this is that I think we've discussed this in our meeting that a `SequentialReader` shall be composed by individual `Reader` instances which handle the storage.
> With that in mind, the `ReaderImpl` class should in my opinion be an interface class and the SequentialReader should take a pointer to such an interface implementation. With that, we'd stay ABI compatible as well as having the flexibility to pass in different types of `Readers` to the sequential reader.

If I understand correctly, one way of doing this that won't break ABI compatibility is:
* `ReaderImpl` interface (abstract class) 
* Subclasses of the interface like say `SingleFileReaderImpl`, `MultiFileReaderImpl`, `CompressionReaderImpl` etc. 
* Corresponding `SingleFileReader`, `MultiFileReader` classes that have pointers to their respective implementations

Does this align with what you are suggesting?

> I am okay with having it strictly `pimpl'ed, but then I would expect this class to be final with no virtual functions on it. I am still not completely clear on how you'd envision to incorporate future features such as compression or encryption with this design.

The way we were thinking about this is - all reader classes (compression, encryption etc.) will have an opaque pointer to the `ReaderImpl` class and have other non-virtual functions to provide specific reader capabilities. I agree that in this case the `ReaderImpl` class needs to be final with no virtual functions in it. The specialized reader classes will also need to be final.

rosbag2/include/rosbag2/impl/sequential_reader_impl.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "rosbag2_storage/storage_factory.hpp"
+#include "rosbag2_storage/storage_factory_interface.hpp"
+#include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
+
+#include "rosbag2/converter.hpp"
+#include "rosbag2/reader_interfaces/base_reader_interface.hpp"
+#include "rosbag2/serialization_format_converter_factory.hpp"
+#include "rosbag2/serialization_format_converter_factory_interface.hpp"
+#include "rosbag2/visibility_control.hpp"
+
```
Karsten1987: duplicate new line

rosbag2/src/rosbag2/sequential_reader.cpp
```diff
@@ -12,88 +13,50 @@
...
 
 namespace rosbag2
 {
 
 SequentialReader::SequentialReader(
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory)
-: storage_factory_(std::move(storage_factory)), converter_factory_(std::move(converter_factory)),
-  converter_(nullptr)
+: reader_impl_(std::make_unique<SequentialReaderImpl>(
```
Karsten1987: I feel like this should actually be an input parameter. The way I feel composition should work is that the constructor takes its specialization as parameters.
And with the `reader_impl_` being a pointer to an interface class, this can be changed dynamically. 

The way I see this:

```
SequentialReader::SequentialReader(
  ... // storage,
  ... // converter,
  std::unique_ptr<BaseReaderInterface> reader_impl)
: reader_impl_(std::move(reader_impl))
{}
```

This way a SequentialReader can easily be extended to take in an implementation of the `BaseReaderInterface` which allows e.g. compression or future enhancements. 
On top of that the SequentialReader class can be final.

In fact, with this pattern the `SequentialReader` could be renamed to solely `Reader` where the input parameter dictates its behavior. So another implementation of the `BaseReaderInterface` could be a non sequential reading.

rosbag2/include/rosbag2/sequential_reader.hpp
```diff
@@ -39,15 +39,17 @@
...
 namespace rosbag2
 {
+
+class BaseReaderInterface;
+
 /**
  * The SequentialReader allows opening and reading messages of a bag. Messages will be read
  * sequentially according to timestamp.
  */
 class ROSBAG2_PUBLIC SequentialReader
```
Karsten1987: could be `final` now.

rosbag2/src/rosbag2/impl/sequential_reader_impl.cpp
prajakta-gokhale: Fixed.

rosbag2/include/rosbag2/impl/sequential_reader_impl.hpp
prajakta-gokhale: Fixed.

Karsten1987:
@prajakta-gokhale Do you mind rebasing this and sign off the commits? The DCO bot is unhappy because of missing signed commits.

prajakta-gokhale:
> @prajakta-gokhale Do you mind rebasing this and sign off the commits? The DCO bot is unhappy because of missing signed commits.

Done. All checks pass now.

Karsten1987:
Thanks. I think it makes sense to apply the same pattern for the writer API, once #185 is merged. What do you think?

prajakta-gokhale:
I agree. We have an issue on our side https://github.com/ros-security/aws-roadmap/issues/98 to have the same structure in `Writer`.

Karsten1987:
with #185 merged, this PR has to be rebased. I propose to reflect the changes for the writer API in this PR as well to have only one PR to review and get the architectural changes in.

prajakta-gokhale:
If it's okay with you, I'd prefer creating a separate PR for `Writer` interfaces. It'll be easier to review and this PR can be merged after a rebase. What do you think?

Karsten1987:
that's okay for me. My only concern is the turnaround rate here. As mentioned previously offline, I would like to have the changes for splitting landed by tonight. Tuesday is rosdistro freeze.

Karsten1987:
@prajakta-gokhale looks like something went wrong while rebasing. I think you might have forgotten to commit `src/rosbag2/readers/sequential_reader.cpp`

prajakta-gokhale:
I forgot to commit some files I think. Reverting and rebasing again.

Karsten1987:
@prajakta-gokhale please make sure you rebase on the latest master and compile&test your changes locally. Also, I'd recommend just syncing the AWS master with your branch here.

Karsten1987:
closing this as #205 got merged.

PR #194, opened 2019-10-22
Fix flake8 errors and add missing lint tests

thomas-moulard:
Signed-off-by: Thomas Moulard <tmoulard@amazon.com>


Karsten1987:
thanks for the patch. That's certainly helpful.

PR #208, opened 2019-11-18
[Compression - 1] Interfaces for compression feature

piraka9011:
## Changes
* Added compression interfaces
* Added compression metadata
* Updated tests for new metadata version

### DIFF
--- a/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
+++ b/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
@@ -42,7 +42,7 @@ public:
    * @return A shared pointer to the bag message with compressed serialized_data.
    */
   virtual std::shared_ptr<SerializedBagMessage> compress_serialized_bag_message(
-    std::shared_ptr<SerializedBagMessage> & bag_message) = 0;
+    std::shared_ptr<SerializedBagMessage> bag_message) = 0;
 
   /**
    * Get the identifier of the compression algorithm. This is appended to the relative file path.
--- a/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
+++ b/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
@@ -1,0 +1,55 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+#define ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rosbag2/visibility_control.hpp"
+
+namespace rosbag2
+{
+class ROSBAG2_PUBLIC BaseCompressorInterface
+{
+public:
+  virtual ~BaseCompressorInterface() {}
+
+  /**
+   * Compress a file on disk.
+   *
+   * \param uri Input file to compress with file extension.
+   * \return The relative path to the compressed file with the compressed extension.
+   */
+  virtual std::string compress_uri(const std::string & uri) = 0;
+
+  /**
+   * Compress serialized_data in a serialized bag message.
+   *
+   * @param bag_message A serialized bag message.
+   * @return A shared pointer to the bag message with compressed serialized_data.
+   */
+  virtual std::shared_ptr<SerializedBagMessage> compress_serialized_bag_message(
+    std::shared_ptr<SerializedBagMessage> & bag_message) = 0;
+
+  /**
+   * Get the identifier of the compression algorithm. This is appended to the relative file path.
+   */
+  virtual std::string get_compression_identifier() const = 0;
+};
+
+}  // namespace rosbag2
+
+#endif  // ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
--- a/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
+++ b/rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
@@ -20,8 +20,29 @@
 
 #include "rosbag2/visibility_control.hpp"
 
+/**
+ * An interface for developers adding a new compression algorithm to rosbag2. These functions
+ * must be implemented to that a writer can properly compress a file or bag message.
+ * A corresponding decompressor must also be implemented.
+ *
+ * Example file compression usage:
+ *
+ * MyCompressor my_compressor();
+ * std::string current_uri = storage.get_relative_path();
+ * std::string compressed_path_uri = my_compressor.compress_uri(current_uri);
+ * relative_file_paths.push_back(compressed_path_uri);
+ *
+ * Example message compression usage:
+ *
+ * MyCompressor my_compressor();
+ * std::shared_ptr<SerializedBagMessage> bag_message = std::make_shared<SerializedBagMessage>();
+ * ...fill message
+ * std::shared_ptr<SerializedBagMessage> compressed_message =
+ *   my_compressor.compress_serialized_bag_message(bag_message);
+ */
 namespace rosbag2
 {
+
 class ROSBAG2_PUBLIC BaseCompressorInterface
 {
 public:
@@ -38,8 +59,8 @@ public:
   /**
    * Compress serialized_data in a serialized bag message.
    *
-   * @param bag_message A serialized bag message.
-   * @return A shared pointer to the bag message with compressed serialized_data.
+   * \param bag_message A serialized bag message.
+   * \return A shared pointer to the bag message with compressed serialized_data.
    */
   virtual std::shared_ptr<SerializedBagMessage> compress_serialized_bag_message(
     std::shared_ptr<SerializedBagMessage> bag_message) = 0;
--- a/rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
+++ b/rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
@@ -20,6 +20,25 @@
 
 #include "rosbag2/visibility_control.hpp"
 
+/**
+ * An interface for developers adding a new decompression algorithm to rosbag2. These functions
+ * must be implemented to that a reader can properly decompress a file or bag message.
+ * A corresponding compressor must also be implemented.
+ *
+ * Example file decompression usage:
+ *
+ * MyDecompressor my_decompressor();
+ * std::string current_uri = get_current_file();
+ * std::string compressed_path_uri = my_decompressor.decompress_uri(current_uri);
+ * storage = storage_factory.open_read_only(compressed_path_uri, storage_options.storage_id);
+ *
+ * Example message decompression usage:
+ *
+ * MyDecompressor my_decompressor();
+ * std::shared_ptr<SerializedBagMessage> bag_message = storage.read_next();
+ * std::shared_ptr<SerializedBagMessage> decompressed_message =
+ *   my_decompressor.decompress_serialized_bag_message(bag_message);
+ */
 namespace rosbag2
 {
 
@@ -31,16 +50,16 @@ public:
   /**
    * Decompress a file on disk.
    *
-   * @param uri Input file to decompress with file extension.
-   * @return The relative path to the decompressed file without the compressed extension.
+   * \param uri Input file to decompress with file extension.
+   * \return The relative path to the decompressed file without the compressed extension.
    */
   virtual std::string decompress_uri(const std::string & uri) = 0;
 
   /**
    * Decompress serialized_data in a serialized bag message.
    *
-   * @param bag_message A serialized bag message.
-   * @return A shared pointer to the bag message with decompressed serialized_data.
+   * \param bag_message A serialized bag message.
+   * \return A shared pointer to the bag message with decompressed serialized_data.
    */
   virtual std::shared_ptr<SerializedBagMessage> decompress_serialized_bag_message(
     std::shared_ptr<SerializedBagMessage> bag_message) = 0;
--- a/rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
+++ b/rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
@@ -1,0 +1,56 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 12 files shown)


rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
```diff
@@ -0,0 +1,55 @@
...
+   */
+  virtual std::string compress_uri(const std::string & uri) = 0;
+
+  /**
+   * Compress serialized_data in a serialized bag message.
+   *
+   * @param bag_message A serialized bag message.
+   * @return A shared pointer to the bag message with compressed serialized_data.
+   */
+  virtual std::shared_ptr<SerializedBagMessage> compress_serialized_bag_message(
```
zmichaels11: Is it intentional that `bag_messages` is passed by reference?
Elsewhere, `std::shared_ptr<SerializedBagMessage>` is passed by value. If we're going to go against that design choice, I think we should instead pass as a raw pointer.

piraka9011: Lets stick to convention.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -163,6 +163,10 @@ BagMetadata MetadataIo::read_metadata(const std::string & uri)
     YAML::Node yaml_file = YAML::LoadFile(get_metadata_file_name(uri));
     auto metadata = yaml_file["rosbag2_bagfile_information"].as<rosbag2_storage::BagMetadata>();
     metadata.bag_size = FilesystemHelper::calculate_directory_size(uri);
+    // Preserve backward compatibility with older metadata files
+    if (metadata.version == 3) {
```
dabonnie: As discussed let's check if the compression keys exist instead of versioning the file

rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
```diff
@@ -0,0 +1,56 @@
...
+#define ROSBAG2__COMPRESSION_INTERFACES__BASE_DECOMPRESSOR_INTERFACE_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rosbag2/visibility_control.hpp"
+
+namespace rosbag2
+{
+
```
dabonnie: Nit: interface documentation (purpose, how to use / implement, etc)

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+#define ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rosbag2/visibility_control.hpp"
+
+namespace rosbag2
+{
```
dabonnie: Nit: interface documentation (purpose, how to use / implement, etc)

piraka9011:
@Karsten1987 Could we get a review on this please.

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
```diff
@@ -0,0 +1,76 @@
...
+#ifndef ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+#define ROSBAG2__COMPRESSION_INTERFACES__BASE_COMPRESSOR_INTERFACE_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rosbag2/visibility_control.hpp"
+
+/**
+ * An interface for developers adding a new compression algorithm to rosbag2. These functions
```
Karsten1987: I would prefer to have one line per sentence. I know it's not consistent throughout the repo, but I think for new comments in the right approach.

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
```diff
@@ -0,0 +1,76 @@
...
+ * std::string compressed_path_uri = my_compressor.compress_uri(current_uri);
+ * relative_file_paths.push_back(compressed_path_uri);
+ *
+ * Example message compression usage:
+ *
+ * MyCompressor my_compressor();
+ * std::shared_ptr<SerializedBagMessage> bag_message = std::make_shared<SerializedBagMessage>();
+ * ...fill message
+ * std::shared_ptr<SerializedBagMessage> compressed_message =
+ *   my_compressor.compress_serialized_bag_message(bag_message);
```
Karsten1987: That compression implementation lives outside the `rosbag2` package, right? I mean the use case you describe here is applicable independently and outside the realm of rosbag.

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
Karsten1987: Shouldn't that bag message be passed in `const`? I think it'd be a bit strange if the original message was modified. 

Also, does it have to be a `SerializedBagMessage`? I could see that a serialized message would suffice and so this functionality could even live outside of rosbag2 completely.

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
piraka9011: Implementation of compression can be outside of rosbag2. However, we want to have a default one that can be used out of the box with the package.
We can remove the compression implementation out of the rosbag2 namespace.

Karsten1987: The namespace doesn't matter, really, but the dependencies kind of do. I would propose to shift the implementation into its own package (even if default) which could be designed similar to the `default_converter_plugins` or `default_storage_plugins`.

piraka9011: Sounds good, I can make that change.
Will make a `default_compression_plugin` package and start by adding the necessary compression features in the reader (another PR) so you can see how it works.

rosbag2/include/rosbag2/compression_interfaces/base_compressor_interface.hpp
Karsten1987: how was that resolved?

rosbag2/include/rosbag2/compression_interfaces/base_decompressor_interface.hpp
```diff
@@ -0,0 +1,75 @@
...
+  virtual std::string decompress_uri(const std::string & uri) = 0;
+
+  /**
+   * Decompress serialized_data in a serialized bag message.
+   *
+   * \param bag_message A serialized bag message.
+   * \return A shared pointer to the bag message with decompressed serialized_data.
+   */
+  virtual std::shared_ptr<SerializedBagMessage> decompress_serialized_bag_message(
+    std::shared_ptr<SerializedBagMessage> bag_message) = 0;
```
Karsten1987: same comment about constness.

piraka9011: I couldn't reply to your other comment, but basically we need access to the `bag_message->serialized_data` so we can actually compress it and then replace it in the serialized bag message.

Karsten1987: I don't understand this. Is the intention of this function to (de)compress data in place? If so, then I'd expect this function to have a `void` or `bool` return type. Also, this should then be documented that the input parameter will be changed. Also, if being changed, I don't think a `shared_ptr` is the right pointer type (raw pointer could be more applicable if so). 
If at one point the function makes a copy of the `bag_message` I think the input parameter should be const.

piraka9011: The goal of this function is to create a new `SerializedBagMessage` and deallocate the old one since we no longer need it.
We could do that in the function or outside. I prefer within the function since that way we don't put the responsibility on someone else.

On the other hand, we can keep it `const`, however, I don't see where the message would be used after it's compressed. 

Regarding using `SerializedBagMessage` specifically - We want to make it clear that compression is specifically for messages in `rosbag2`.  I don't think the goal of this compressor is to compress things outside of `rosbag2`.

piraka9011:
New changes:
* New `rosbag2_compression` package.
* Address const-ness of `bag_message`.

rosbag2_compression/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+option(DISABLE_SANITIZERS "disables the use of gcc saniztizers")
```
thomas-moulard:
```suggestion
option(DISABLE_SANITIZERS "Disable the use of gcc saniztizers")
```

rosbag2_compression/include/rosbag2_compression/base_compressor_interface.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * MyCompressor my_compressor();
+ * std::shared_ptr<SerializedBagMessage> bag_message = std::make_shared<SerializedBagMessage>();
+ * ...fill message
+ * std::shared_ptr<SerializedBagMessage> compressed_message =
+ *   my_compressor.compress_serialized_bag_message(bag_message);
+ */
+namespace rosbag2_compression
+{
+namespace compression_interfaces
+{
```
thomas-moulard: I don't see the point of having two levels of indirection there

rosbag2_compression/include/rosbag2_compression/base_compressor_interface.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#ifndef ROSBAG2_COMPRESSION__BASE_COMPRESSOR_INTERFACE_HPP_
+#define ROSBAG2_COMPRESSION__BASE_COMPRESSOR_INTERFACE_HPP_
+
+#include <memory>
+#include <string>
+
+#include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_compression/visibility_control.hpp"
+
```
thomas-moulard: alphasort everywhere

rosbag2_compression/include/rosbag2_compression/base_compressor_interface.hpp
```diff
@@ -0,0 +1,81 @@
...
+typedef rosbag2_storage::SerializedBagMessage SerializedBagMessage;
```
thomas-moulard: using...

rosbag2_compression/include/rosbag2_compression/base_compressor_interface.hpp
```diff
@@ -0,0 +1,81 @@
...
+  virtual std::string compress_uri(const std::string & uri) = 0;
+
+  /**
+   * Compress serialized_data in a serialized bag message.
+   *
+   * \param bag_message A serialized bag message.
+   * \return A shared pointer to the bag message with compressed serialized_data.
+   */
+  virtual std::shared_ptr<SerializedBagMessage> compress_serialized_bag_message(
+    const std::shared_ptr<SerializedBagMessage> bag_message) = 0;
```
thomas-moulard: Why not uptr?

rosbag2_compression/package.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
```
thomas-moulard: format=2 to align with the other packages?

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -33,14 +33,16 @@ struct TopicInformation
 
 struct BagMetadata
 {
-  int version = 2;  // upgrade this number when changing the content of the struct
+  int version = 3;  // Upgrade this number when changing the content of the struct
```
thomas-moulard: Let's split modifying the metadata into a separate PR

dabonnie: What if the metadata is required for functionality, i.e., pushing this PR without modifying metadata is non-functional code?

rosbag2_tests/resources/cdr_test/metadata.yaml
```diff
@@ -19,3 +19,5 @@ rosbag2_bagfile_information:
         type: test_msgs/Arrays
         serialization_format: cdr
       message_count: 4
+  compression_format: ""
```
dabonnie: Do you need a test resource that does not contain the compression metadata keys (test backwards compatibility)?

rosbag2_compression/include/rosbag2_compression/base_compressor_interface.hpp
piraka9011: SerializedBagMessage is usually a `shared_ptr` throughout the `rosbag2` packages. Converting it from shared to unique would involved changing a lot of the code elsewhere.

The uncompressed `bag_message` might also be used elsewhere.

piraka9011:
> Is there a real need to split compression and decompression in two different interfaces? It's technically not a problem since C++ allows multiple inheritance, but does it make sense to have an implementation available where only one of the two functionality is implemented?

Yes, when working on the PoC, we found it best to decouple the implementation. This allows us to perform specific actions for de/compression which can vary based on the library/algorithm we are using. 

We check which compression implementation should be used from the metadata and add checks when de/compressing to make sure we are using the right libraries. Most of the libraries themselves also have checks to make sure the compressed files are in the right format.

piraka9011:
@Karsten1987 Regarding the usage of `const` for `bag_message` compression - we think the best way forward is to have `compress_serialized_bag_message` compress the serialized data in place. That way we don't create a copy of a new `bag_message` (which could be very large). The new function would then be:

```c++
  /**
   * Compress serialized_data in a serialized bag message in place. 
   *
   * \param bag_message A serialized bag message whose serialized_data will be compressed.
   */
  virtual void compress_serialized_bag_message(
    std::shared_ptr<SerializedBagMessage> bag_message) = 0;
```

Karsten1987:
> This allows us to perform specific actions for de/compression which can vary based on the library/algorithm we are using.

Can you elaborate on these actions a little? I currently have a hard time grasping the benefits of allowing a library to only compress a bag message with a specific algorithm and don't provide an implementation for decompressing it.

> we think the best way forward is to have compress_serialized_bag_message compress the serialized data in place

I guess that's okay. I just want to highlight that you have to deal with concurrency issues then when still using a shared pointer. At a minimum you have to lock the instances while compressing/decompressing. Ideally, I think you might have to work with condition variables to notify each instances that the underlying data of their shared pointer has now fundamentally changed. 
That is to say, I don't think a shared pointer is the ideal pointer type for this. I think a unique pointer or a raw pointer makes more sense here.

piraka9011:
> Can you elaborate on these actions a little? I currently have a hard time grasping the benefits of allowing a library to only compress a bag message with a specific algorithm and don't provide an implementation for decompressing it.

There is a potential use case for using a certain compression format and then using another decompression library, but with a similar format to the initial compression format. Our team would also benefit from the current decoupling since:
1. We have an ask to collect statistics on de/compression times for playback.
2. Change compression libraries.

Moreover, it ensures the de/compressor has a single responsibility and leaves the door open for extending de/compressors down the line.

> At a minimum you have to lock the instances while compressing/decompressing....
> That is to say, I don't think a shared pointer is the ideal pointer type for this. I think a unique pointer or a raw pointer makes more sense here.

We had a similar discussion as well. The issue is that `SerializedBagMessage` is usually used throughout `rosbag2` as a `shared_ptr` and you can't convert a `shared_ptr` to `unique_ptr`. 
We can add locks if we find that the pointer to the `SerializedBagMessage` is not [unique](http://www.cplusplus.com/reference/memory/shared_ptr/unique/).

I'd like to hear any other suggestions from your end on this matter.

thomas-moulard:
I agree raw pointer is better.
Can we give a shot at implementing this with 2 different interfaces for now, unless you're really against it? Reader/Writer are split between two interfaces, and the same type of discussion apply there (do we really want to read, but not write or the opposite?). We can always follow-up and merge the interfaces down the road if we feel this is the wrong decision.

piraka9011:
Closing in favor of the various broken up PRs currently in progress.

Issue #197, opened 2019-10-24
Move get_storage_identifier and get_bagfile_size into BaseInfoInterface


zmichaels11:
As per https://github.com/ros2/rosbag2/pull/184#discussion_r338327221,

these API calls are probably better if defined in `BaseInfoInterface` rather than `BaseIOInterface`.

### Required Changes
- [x] Move `get_storage_identifier` from `BaseIOInterface` to `BaseInfoInterface`
- [x] Move `get_bagfile_size` from `BaseIOInterface` to `BaseInfoInterface`
- [x] Update `SqliteStorage`
- [x] ~Update `ros2/rosbag2_bag_v2`~ Not required

### Dependent PRs
- [x] #184 

### PRs
- [x] #209 - Ready for Review

Karsten1987:
@zmichaels11 are you still working on this?

zmichaels11:
Yes I am, I'll push my branch once I rebase onto master.

zmichaels11:
Addressing `Move get_storage_identifier from BaseInfoInterface to BaseIOInterface` and `Move get_bagfile_size from BaseInfoInterface to BaseIOInterface` in PR: https://github.com/ros2/rosbag2/pull/209

zmichaels11:
Removed requirement `remove get_metadata` due to bug resolved by #210.

As per https://github.com/ros2/rosbag2/pull/210#issuecomment-555275503, we should probably discuss on handling `get_metadata` since its required by bagfiles not written by `Writer`.

zmichaels11:
Remove `get_metadata` was moved into https://github.com/ros2/rosbag2/issues/213 due to being more complex and requiring planning.

zmichaels11:
Corrected the tasks in OP;
'get_storage_identifier` and `get_bagfile_size` now correctly state that they're moving INTO `BaseInfoInterface`

PR #210, opened 2019-11-19
load metadata from storage if no yaml file is found

Karsten1987:
This is necessary when loading legacy ros1 bag files, where no metadata.yaml file is present.
connected to https://github.com/ros2/rosbag2_bag_v2/pull/8
Signed-off-by: Knese Karsten <karsten@openrobotics.org>

### DIFF
--- a/rosbag2/src/rosbag2/readers/sequential_reader.cpp
+++ b/rosbag2/src/rosbag2/readers/sequential_reader.cpp
@@ -51,20 +51,37 @@ void SequentialReader::reset()
 void SequentialReader::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
-  metadata_ = metadata_io_->read_metadata(storage_options.uri);
-  if (metadata_.relative_file_paths.empty()) {
-    ROSBAG2_LOG_WARN("No file paths were found in metadata.");
-    return;
-  }
+  // If there is a metadata.yaml file present, load it.
+  // If not, let's ask the storage with the given URI for its metadata.
+  // This is necessary for non ROS2 bags (aka ROS1 legacy bags).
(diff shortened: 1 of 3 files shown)


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8667/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7076/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8587/)

CI with bag_v2_plugins:
[Build Status](https://ci.ros2.org/job/ci_packaging_linux/272/)

rosbag2/src/rosbag2/readers/sequential_reader.cpp
```diff
@@ -51,20 +51,37 @@ void SequentialReader::reset()
...
+      throw std::runtime_error("No storage could be initialized. Abort");
+    }
+  } else {
+    storage_ = storage_factory_->open_read_only(
+      storage_options.uri, storage_options.storage_id);
+    if (!storage_) {
+      throw std::runtime_error("No storage could be initialized. Abort");
+    }
+    metadata_ = storage_->get_metadata();
+    if (metadata_.relative_file_paths.empty()) {
```
piraka9011: I think this should be outside the if/else clause since we want to check the paths in both cases.

zmichaels11:
Just spoke with @piraka9011 
I think we should probably have a workshop session to discuss this and Issue #197 since this code path exists only for rosbags written by something that isn't `Writer`.

Karsten1987:
looking at the console output of the packaging job, all rosbag2 related tests passed. Merging.

PR #212, opened 2019-11-22
Validate bagfile splitting size input

prajakta-gokhale:
If `sqlite3` plugin is given a splitting size less than 84 kiB, it will still create bagfiles of ~84 kiB because of how often writes are flushed to disk. Splitting size smaller than this could cause unexpected behavior, so the input value needs to be validate and set correctly.

Related to https://github.com/ros-security/aws-roadmap/issues/6.

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -57,8 +57,9 @@ class RecordVerb(VerbExtension):
         )
         parser.add_argument(
             '-b', '--max-bag-size', type=int, default=0,
-            help='maximum size in bytes before the bagfile will be split.'
-              'Default it is zero, recording written in single bagfile and splitting is disabled.'
+            help='maximum size in bytes before the bagfile will be split. '
+                  'Default it is zero, recording written in single bagfile and splitting '
+                  'is disabled.'
         )
         self._subparser = parser
 
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -56,11 +56,9 @@ class RecordVerb(VerbExtension):
                   'is disabled.'
         )
         parser.add_argument(
-            '-b', '--bag-size', type=int, default=0,
-            help='provides a threshold for recording bagfile size in bytes, before the recording is split and '
-              'rolled over to a new bagfile. This value is only used as a guideline because actually enforcing a '
-              'maximum file size is difficult. The actual bagfile size may be a little over this value. '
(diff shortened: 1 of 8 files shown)


prajakta-gokhale:
@Karsten1987 please take a look.

ros2bag/ros2bag/verb/record.py
```diff
@@ -74,6 +86,8 @@ def main(self, *, args):  # noqa: D102
 
         self.create_bag_directory(uri)
 
+        self.validate_splitting_size(args)
```
Karsten1987: I don't see the purpose of doing this within the ros2bag package. In the end, every storage should validate that splitting option itself and so I think it makes more sense to have this implemented on a per storage level.
Saying I don't see the point to do this only for sqlite3 and simply ignore all other plugins.

prajakta-gokhale:
> I personally don't see the point of validating the max bag file size within the command line tool if only the default storage plugin will be checked and reacted on.
> I think this should be implemented per storage.

It doesn't have to validate only the default plugin here, we could add all of them if the values are known. The other option is to do it in the `Writer` where both the values are known.

I don't have a preference for one over another as long as we sanitize the input somewhere, because otherwise users can enter unsupported values and the splitting feature won't work.

Karsten1987:
> It doesn't have to validate only the default plugin here, we could add all of them if the values are known. The other option is to do it in the Writer where both the values are known.

That would mean that for every other storage plugin, even non-public ones, the ros2bag has to be modified. I think that's rather unfeasible.  I think we have to do that in the `Writer` class, but even have a change on the interfaces for each storage to validate the storage options or have `open` fail/return false if the given storage options don't fit.

> I don't have a preference for one over another as long as we sanitize the input somewhere, because otherwise users can enter unsupported values and the splitting feature won't work.

absolutely agree. I think it's just dangerous to silently apply a default value without warning the user. I think the command should just fail if the user applies an unsupported value.

prajakta-gokhale:
> I think we have to do that in the `Writer` class, but even have a change on the interfaces for each storage to validate the storage options or have `open` fail/return false if the given storage options don't fit.

That sounds reasonable to me. I will change this and add the check in `Writer` class with a corresponding new API in `storage_interfaces`.

> I think the command should just fail if the user applies an unsupported value.

With the above approach, we can have the `record` command fail when an unsupported argument is passed with the appropriate error response.

prajakta-gokhale:
Closing in favor of https://github.com/ros2/rosbag2/pull/226

PR #217, opened 2019-11-26
[Compression - 2] Utility headers

piraka9011:
Add logging and visibility macros.

Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

PR #216, opened 2019-11-26
[Compression - 1] Create new rosbag2 compression package

piraka9011:
Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

PR #223, opened 2019-12-02
Revert "[Compression - 3] Add compressor interface"

piraka9011:
Reverts ros2/rosbag2#218 because it wasn't squashed.

Signed-off-by: Anas Abou Allaban allabana@amazon.com


thomas-moulard:
I'll leave @Karsten1987 judge what needs to be done there, but we accidentally pushed 9 commits, then with it we go to 9 + 1 (revert) + 1 (new squashed commit) = 11 commits. Maybe we can just let it go for this time?

Karsten1987:
The same behavior happened already for #216 and #217 as both PRs previously weren't squashed. So in fact you've pushed 17 commits instead of only three needed if squashing, afaict. I am okay with it as it's mainly tackling your `rosbag2_compression` package. Which means you might want to clean up the changelog correctly when releasing. 
I would appreciate though to squash and merge future PRs.

piraka9011:
>  I would appreciate though to squash and merge future PRs.

Will do, thanks for your input!

thomas-moulard:
@Karsten1987 Is there any reason not to squash in this repo? You may want to just completely [disable other merge options in the repo settings directly](https://help.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests).

Karsten1987:
@thomas-moulard you should be able to re-write the history for this PR and manually squash commits into its top-level PRs.

PR #225, opened 2019-12-03
Transaction based sqlite3 inserts

sriramster:
Hi @Karsten1987  in continuation with the discussion we've had about using trasactions. 

This is a very high level implementation of how the code looks. Plaes have a look and let me know your thoughts.

https://gist.github.com/sriramster/2f9bdd620730855e6172c7136b56d6fc

Sriram

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
@@ -26,11 +26,11 @@ public:
   std::string uri;
   std::string storage_id;
 
-// The maximum size a bagfile can be, in bytes, before it is split.
-// A value of 0 indicates that bagfile splitting will not be used.
+  // The maximum size a bagfile can be, in bytes, before it is split.
+  // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_size;
 
-// The chunk_size value is used to control cache sizes for bulk writes.
+  // The chunk_size value is used to control cache sizes for bulk writes.
   size_t chunk_size;
 };
 
--- a/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
@@ -30,8 +30,10 @@ public:
   // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_size;
 
-  // The chunk_size value is used to control cache sizes for bulk writes.
-  size_t chunk_size;
+  // The chunk size indiciates how many messages should be hold in cache
+  // before these being written to disk.
+  // Defaults to 1, and effectively disables the caching.
+  size_t chunk_size = 1;
 };
 
 }  // namespace rosbag2_cpp
--- a/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
@@ -26,17 +26,11 @@ public:
   std::string uri;
   std::string storage_id;
 
-  /**
-   * The maximum size a bagfile can be, in bytes, before it is split.
-   * A value of 0 indicates that bagfile splitting will not be used.
-   */
+// The maximum size a bagfile can be, in bytes, before it is split.
+// A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_size;
-  /**
-   * The chunk_size value is used to control writes to sqlite db,
-   * Selecting a optimal chunk_size, optimizes the sqlite writes by
-   * enabling transaction's
-   */
 
+// The chunk_size value is used to control cache sizes for bulk writes.
   size_t chunk_size;
 };
 
--- a/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
@@ -30,10 +30,10 @@ public:
   // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_size;
 
-  // The chunk size indiciates how many messages should be hold in cache
+  // The cache size indiciates how many messages can maximally be hold in cache
   // before these being written to disk.
-  // Defaults to 1, and effectively disables the caching.
-  size_t chunk_size = 1;
+  // Defaults to 0, and effectively disables the caching.
+  uint64_t max_cache_size = 0;
 };
 
 }  // namespace rosbag2_cpp
--- a/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
@@ -31,6 +31,13 @@ public:
    * A value of 0 indicates that bagfile splitting will not be used.
    */
   uint64_t max_bagfile_size;
+  /**
+   * The chunk_size value is used to control writes to sqlite db,
+   * Selecting a optimal chunk_size, optimizes the sqlite writes by
+   * enabling transaction's
+   */
+
+  size_t chunk_size;
 };
 
 }  // namespace rosbag2_cpp
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -109,10 +109,11 @@ private:
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage_;
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
-  size_t chunk_size_;
 
-  // Intermediate caching to enable bulk writes using transaction on SQLITE3
-  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>>  cache_;
+  // Intermediate cache to write multiple messages into the storage.
+  // chunk size is the amount of messages to hold in storage before writing to disk.
+  size_t chunk_size_;
+  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
 
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -115,7 +115,7 @@ private:
 
   // Intermediate cache to write multiple messages into the storage.
   // `max_cache_size` is the amount of messages to hold in storage before writing to disk.
-  size_t max_cache_size_;
+  uint64_t max_cache_size_;
   std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
 
   // Used to track topic -> message count
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -109,6 +109,10 @@ private:
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage_;
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
+  size_t chunk_size_;
+
+  // Intermediate caching to enable bulk writes using transaction on SQLITE3
+  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>>  cache_;
 
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -110,13 +110,13 @@ private:
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
 
-  // Intermediate cache to write multiple messages into the storage.
-  // chunk size is the amount of messages to hold in storage before writing to disk.
-  size_t chunk_size_;
-  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
-
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
+
+  // Intermediate cache to write multiple messages into the storage.
+  // `max_cache_size` is the amount of messages to hold in storage before writing to disk.
+  size_t max_cache_size_;
+  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
 
   // Used to track topic -> message count
   std::unordered_map<std::string, rosbag2_storage::TopicInformation> topics_names_to_info_;
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -79,6 +79,7 @@ void SequentialWriter::open(
 {
   max_bagfile_size_ = storage_options.max_bagfile_size;
   base_folder_ = storage_options.uri;
+  chunk_size_ = storage_options.chunk_size;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
@@ -203,7 +204,11 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
-  storage_->write(converter_ ? converter_->convert(message) : message);
+  cache_.push_back(converter_ ? converter_->convert(message) : message);
+
+  if (cache_.size() >= chunk_size_) {
+    storage_->bulk_write(cache_);
+  }
 }
 
 bool SequentialWriter::should_split_bagfile() const
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -208,7 +208,7 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
 
   cache_.push_back(converter_ ? converter_->convert(message) : message);
   if (cache_.size() >= chunk_size_) {
-    storage_->bulk_write(cache_);
+    storage_->write(cache_);
   }
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -81,6 +81,8 @@ void SequentialWriter::open(
   base_folder_ = storage_options.uri;
   chunk_size_ = storage_options.chunk_size;
 
+  cache_.reserve(chunk_size_);
+
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
   {
@@ -205,7 +207,6 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   metadata_.duration = std::max(metadata_.duration, duration);
 
   cache_.push_back(converter_ ? converter_->convert(message) : message);
-
   if (cache_.size() >= chunk_size_) {
     storage_->bulk_write(cache_);
   }
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -206,6 +206,13 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
+  // No need to cache if there's only space for 1 message
+  // We'll call write directly in this case.
+  if (chunk_size_ <= 1) {
+    storage_->write(converter_ ? converter_->convert(message) : message);
+    return;
+  }
+
   cache_.push_back(converter_ ? converter_->convert(message) : message);
   if (cache_.size() >= chunk_size_) {
     storage_->write(cache_);
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -209,6 +209,9 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   cache_.push_back(converter_ ? converter_->convert(message) : message);
   if (cache_.size() >= chunk_size_) {
     storage_->write(cache_);
+    // reset cache
+    cache_.clear();
+    cache_.reserve(chunk_size_);
   }
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -77,11 +77,11 @@ void SequentialWriter::open(
   const StorageOptions & storage_options,
   const ConverterOptions & converter_options)
 {
-  max_bagfile_size_ = storage_options.max_bagfile_size;
   base_folder_ = storage_options.uri;
-  chunk_size_ = storage_options.chunk_size;
+  max_bagfile_size_ = storage_options.max_bagfile_size;
+  max_cache_size_ = storage_options.max_cache_size;
 
-  cache_.reserve(chunk_size_);
+  cache_.reserve(max_cache_size_);
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
@@ -206,19 +206,17 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
-  // No need to cache if there's only space for 1 message
-  // We'll call write directly in this case.
-  if (chunk_size_ <= 1) {
+  // if cache size is set to zero, we directly call write
+  if (max_cache_size_ == 0u) {
     storage_->write(converter_ ? converter_->convert(message) : message);
-    return;
-  }
-
-  cache_.push_back(converter_ ? converter_->convert(message) : message);
-  if (cache_.size() >= chunk_size_) {
-    storage_->write(cache_);
-    // reset cache
-    cache_.clear();
-    cache_.reserve(chunk_size_);
+  } else {
+    cache_.push_back(converter_ ? converter_->convert(message) : message);
+    if (cache_.size() >= max_cache_size_) {
+      storage_->write(cache_);
+      // reset cache
+      cache_.clear();
+      cache_.reserve(max_cache_size_);
+    }
   }
 }
 
--- a/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
@@ -258,12 +258,12 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
 
 TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
   const size_t counter = 1000;
-  const size_t chunk_size = 100;
+  const size_t max_cache_size = 100;
 
   EXPECT_CALL(
     *storage_,
     write(An<const std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> &>())).
-  Times(counter / chunk_size);
+  Times(counter / max_cache_size);
   EXPECT_CALL(
     *storage_,
     write(An<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>>())).Times(0);
@@ -278,7 +278,7 @@ TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
   message->topic_name = "test_topic";
 
   storage_options_.max_bagfile_size = 0;
-  storage_options_.chunk_size = chunk_size;
+  storage_options_.max_cache_size = max_cache_size;
 
   writer_->open(storage_options_, {rmw_format, rmw_format});
   writer_->create_topic({"test_topic", "test_msgs/BasicTypes", "", ""});
@@ -288,9 +288,9 @@ TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
   }
 }
 
-TEST_F(SequentialWriterTest, do_not_use_cache_if_chunk_size_is_one) {
+TEST_F(SequentialWriterTest, do_not_use_cache_if_cache_size_is_zero) {
   const size_t counter = 1000;
-  const size_t chunk_size = 1;
+  const size_t max_cache_size = 0;
 
   EXPECT_CALL(
     *storage_,
@@ -310,7 +310,7 @@ TEST_F(SequentialWriterTest, do_not_use_cache_if_chunk_size_is_one) {
   message->topic_name = "test_topic";
 
   storage_options_.max_bagfile_size = 0;
-  storage_options_.chunk_size = chunk_size;
+  storage_options_.max_cache_size = max_cache_size;
 
   writer_->open(storage_options_, {rmw_format, rmw_format});
   writer_->create_topic({"test_topic", "test_msgs/BasicTypes", "", ""});
--- a/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
@@ -227,6 +227,7 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
   message->topic_name = "test_topic";
 
   storage_options_.max_bagfile_size = max_bagfile_size;
+  storage_options_.chunk_size = 1;
 
   writer_->open(storage_options_, {rmw_format, rmw_format});
   writer_->create_topic({"test_topic", "test_msgs/BasicTypes", "", ""});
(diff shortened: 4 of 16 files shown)


rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -112,8 +145,22 @@ void SqliteStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMe
...
+#ifdef EN_TRANSACTION
+  if (!active_transaction_) {
+    activate_transaction();
+  }
+  // To work with a batched insert within a trasaction
+  no_of_inserts++;
+#endif
+
   write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
   write_statement_->execute_and_reset();
```
Karsten1987: with the transactions open, is this command staying the same?
Are the actual transactions orthogonal to the individual commits to the database?

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -88,6 +94,8 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   std::unordered_map<std::string, int> topics_;
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
+  std::atomic_bool active_transaction_;
+  uint64_t no_of_inserts;
```
Karsten1987: please follow the same style as in this file and prefix class member variables with `_`.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -70,6 +71,11 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
 
   std::string get_storage_identifier() const override;
 
+  // APIs to handle/control transactions
+  void activate_transaction();
+
+  void commit_transaction();
+
```
Karsten1987: Do these functions have to be public?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -101,6 +106,34 @@ void SqliteStorage::open(
     relative_path_ << "' for " << to_string(io_flag) << ".");
 }
 
+void SqliteStorage::activate_transaction()
+{
+  int rc = -1;
```
Karsten1987: This line can be shifted within the `if` block, right?

Karsten1987: Also for clarity, I would prefer having it initialized with `SQLITE_ERROR`. https://www.sqlite.org/rescode.html#error

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -101,6 +106,34 @@ void SqliteStorage::open(
...
+    rc = sqlite3_exec(database_->get_db_handle(), "BEGIN TRANSACTION;", NULL, 0, NULL);
+    active_transaction_.store(true, std::memory_order_relaxed);
+    if (rc != SQLITE_OK)
+      throw SqliteException("Failed to initialize transaction");
+  }
+}
+
+void SqliteStorage::commit_transaction()
+{
+  int rc = -1;
```
Karsten1987: same comment as above.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -101,6 +106,34 @@ void SqliteStorage::open(
     relative_path_ << "' for " << to_string(io_flag) << ".");
 }
 
+void SqliteStorage::activate_transaction()
+{
+  int rc = -1;
+
+  if (!active_transaction_) {
```
Karsten1987: I think it easier to change the logic here and return early:
```
if (active_transaction_) {
  return;
}
```

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
Karsten1987: Also, I think these two functions make more sense in the `SQLiteWrapper` class as they are really just implementation details and don't need any visibility in the storage implementation. 
That would also resolve the need for the `get_db_handle()` function.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -42,7 +42,8 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   : public rosbag2_storage::storage_interfaces::ReadWriteInterface
 {
 public:
-  SqliteStorage() = default;
+  SqliteStorage()
+       :no_of_inserts(0) {};
   ~SqliteStorage() override = default;
```
Karsten1987: Given that you specify a custom implementation of the destructor, the `default` keyword has to be removed here.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -101,6 +106,34 @@ void SqliteStorage::open(
...
+    if (rc != SQLITE_OK)
+      throw SqliteException("Failed to initialize transaction");
+  }
+}
+
+void SqliteStorage::commit_transaction()
+{
+  int rc = -1;
+
+  if (!active_transaction_) {
```
Karsten1987: I think this logic should be inverted, don't you think? This should be `if (active_transacton_) {`, right?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -101,6 +106,34 @@ void SqliteStorage::open(
...
+    rc = sqlite3_exec(database_->get_db_handle(), "END TRANSACTION;", NULL, 0, NULL);
```
Karsten1987: I believe this shouldn't be `END TRANSACTION` but rather `COMMIT`:
https://stackoverflow.com/questions/14806435/difference-between-end-transaction-and-commit-transaction

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -88,6 +94,8 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   std::unordered_map<std::string, int> topics_;
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
+  std::atomic_bool active_transaction_;
```
Karsten1987: this is an uninitialized variable. I propose `std::atomic_bool active_transaction_ {false};`

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
sriramster: Hi Karsten, the problem is the logic of keepign count of the number of inserts is within sqlite_storage.cpp, and to move that around looked like too much changes for me. But I'll still consider your comment.

sriramster:
Hi @Karsten1987  Thank you for the review and your time. I've updated some pieces based on your comments. 

Let me know if you've a test script, and I'd be more than happy to test and share the numbers with you.

sriramster:
Hi @Karsten1987  the branch compiles with master. Could you please have a look?

sriramster:
Hi @Karsten1987 Have you had time to look at the above patch set?

sriramster:
Hi Karsten, 

Just trying to be on your github notifications, for this pull request. Have you had time to review this? Do I need to rework anywhere?

Karsten1987:
I'll revisit it by the end of today. For the time being, could you rebase your PR on top of master? It looks like there are conflicting files.

sriramster:
@Karsten1987 done. Rebased.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <atomic>
```
Karsten1987: alpha sort

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -71,6 +73,10 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   std::string get_storage_identifier() const override;
 
   uint64_t get_minimum_split_file_size() const override;
+  // APIs to handle/control transactions
```
Karsten1987: new line between functions

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -24,6 +24,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <atomic>
```
Karsten1987: alpha sort

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -104,6 +111,38 @@ void SqliteStorage::open(
     relative_path_ << "' for " << to_string(io_flag) << ".");
 }
 
+void SqliteStorage::activate_transaction()
+{
+  if (active_transaction_) {
+    return;
+  } else {
```
Karsten1987: no need for the return branch here.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -104,6 +111,38 @@ void SqliteStorage::open(
     relative_path_ << "' for " << to_string(io_flag) << ".");
 }
 
+void SqliteStorage::activate_transaction()
+{
+  if (active_transaction_) {
```
Karsten1987: Should we warn the user here that an transition is already active?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -104,6 +111,38 @@ void SqliteStorage::open(
...
+      throw SqliteException("Failed to begin transaction");
+    }
+  }
+}
+
+void SqliteStorage::commit_transaction()
+{
+  if (!active_transaction_) {
+    return;
+  } else {
```
Karsten1987: no need for else branch

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -104,6 +111,38 @@ void SqliteStorage::open(
...
+    active_transaction_.store(true, std::memory_order_relaxed);
+    if (rc != SQLITE_OK) {
+      throw SqliteException("Failed to begin transaction");
+    }
+  }
+}
+
+void SqliteStorage::commit_transaction()
+{
+  if (!active_transaction_) {
```
Karsten1987: same comment as above about the warning

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.hpp
```diff
@@ -45,6 +45,8 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteWrapper
 
   operator bool();
 
+  DBPtr get_db_handle() const;
```
Karsten1987: why do we need this?

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -71,6 +73,10 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
...
+  void activate_transaction();
+
+  void commit_transaction();
```
Karsten1987: Why do these two functions have to be in the public interface? Ideally, the transactions should be opaque to the caller and should be handled solely by the sqlite plugin.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -115,8 +154,24 @@ void SqliteStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMe
...
+  // To work with a batched insert within a trasaction
+  no_of_inserts_++;
+#endif
+
   write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
   write_statement_->execute_and_reset();
+
+#ifdef EN_TRANSACTION
+  // Just for a design perspective we'll move this to a param from command line later
+  if (no_of_inserts_ > 10000) {
```
Karsten1987: Can we somehow shift this to a storage option item? I feel like we should do this right away while testing this PR to get a feeling about its usability and impact.

sriramster:
Hi @Karsten1987 

Fixed the points mentioned above, except for few things. 

* Warning user about active transaction. Do we need this? Since transactions are controlled by the plugin layer. And a user can not interact with a transaction, active or inactive. But it might be a point to debug if this why you intend to have it. I'll add it. 
* Adding the transaction number as a part of storage options, makes us change the open() API, from the base interfaces. Do you mind if I overload the function open() with a different control flow? The problem with this approach is, separation of reader and writer flow control APIs from class loading to sqlite changes. 

Please let me know your thoughts on this. I'll update the review accordingly 

Sorry for the delay, could'nt work over the weekends.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -42,8 +43,9 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   : public rosbag2_storage::storage_interfaces::ReadWriteInterface
 {
 public:
-  SqliteStorage() = default;
-  ~SqliteStorage() override = default;
+  SqliteStorage()
+  : no_of_inserts_(0) {}
```
Karsten1987: new line below.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -42,8 +43,9 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
...
+  ~SqliteStorage();
```
Karsten1987: why not declaring as `override`?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: @sriramster, do you mind answering to this?

sriramster:
Hi @Karsten1987  I've cleaned up a bit focussing your review comments. And working on the chunk_size/transaction_size as an option. Kind of stuck at point, just updating you. Have'nt given up yet here.

sriramster:
Hi @Karsten1987  I've pushed in the changes as suggested by you. Could you please have a look and let me know if they fit into the model.

Karsten1987:
@sriramster thanks a lot for not giving up on this ;-)

Up front: This PR has to be rebased once more as there're some merge conflicts.

I am seeing two potential ways on how to integrate a generic `chunk_size` option.

 * The SQLite plugin (in fact, every plugin) gets the chunk size passed in when being opened as `read_write`. It's then in the plugin's responsibility to deal appropriately with it. The `open_read_write()` function has therefore to be enhanced. I think this is what you've currently implemented in this PR.
 * We deal with chunk sizes exclusively in the `rosbag2_cpp` API. That is, we keep implement a cache in the c++ API which holds all incoming messages until it's succeeding the chunk size and only then asks the storage API to write it to disk. We therefore have to enhance the writer interface with an additional `write(std::vector<SerializedBagMessage>)` method which writes multiple messages at once. 

The first option would give more leeway to each individual plugin to come up with an optimal way of dealing with chunk sizes. The second one, however, eases the burden of each plugin as the logic has to be implemented only once and is therefore less error-prone.

I believe I am tending towards the second option here as I imagine that we could technically also leverage sqlite3 transaction when writing multiple messages at once. I further think that keeping the serialized bag messages in a vector within `rosbag2_cpp` does not introduce a big computation overhead. 

So essentially what I would imagine could work nicely:

```c++
// in rosbag2_cpp

rosbag2_cpp::Writer::write(
  std::shared_ptr<rosbag2_storage::SerializedBagMessage> message)
{
...
// There might exist an optimal data type for this,
// but at first a `std::vector` might do
cache.push_back(message);
if (cache.size() >= storage_options_.chunk_size) {
  // new signature of write in BaseWriterInterface to write multiple
  storage_->write(cache);
}
``` 

```c++
// in SqliteStorage plugin

void SqliteStorage::write(
  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage> & messages)
{
  start_sqlite_transaction();
  for (auto & serialized_bag_message : messages) {
     write(serialized_bag_message);  // call to normal write
  } 
  commit_transaction();
}
```

Does this make sense what I writing here? What do you think?

ros-discourse:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/1

sriramster:
@Karsten1987 I'm reworking this. Been a bit busy at work. Will update you at the earliest, sorry about the delay.

sriramster:
Hi @Karsten1987 I've reworked the patch set.

I'd designed it based on the first method we'd discussed above to make sure the control of transactios is closer to the sqlite plugins only. But in a way, I feel moving it to a higher layer would give the flexibility of working with transactions and plugins per-say. Not all the plugins might have transactions that is one reason why this was sqlite3 specific. Just stating thought process behind the idea. 

Pleae let me know if this patch set looks ok. 
Thanks for the time.

sriramster:
Hi  @Karsten1987 friendly reminder, have you had time to look into this?

sriramster:
Hi @Karsten1987 remind you again on this. :-)

rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
```diff
@@ -31,6 +31,13 @@ struct StorageOptions
    * A value of 0 indicates that bagfile splitting will not be used.
    */
   uint64_t max_bagfile_size;
+  /**
+   * The chunk_size value is used to control writes to sqlite db,
+   * Selecting a optimal chunk_size, optimizes the sqlite writes by
+   * enabling transaction's
+   */
```
Karsten1987: two things: 
* please adhere to the existing style in this file. Add a new line after `uint64_t max_bagfile_size;` and the comment as well as remove the extra line after the comment.
* keep in mind that the `chunk_size` is independent of the actual plugin being used as there are many of them. So please remove the notion of `sqlite` and `transaction`. Same below.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -81,6 +81,7 @@ void SequentialWriter::open(
 {
   max_bagfile_size_ = storage_options.max_bagfile_size;
   base_folder_ = storage_options.uri;
+  chunk_size_ = storage_options.chunk_size;
```
Karsten1987: the chunk size can also be used to `reserve` space in the `cache_` vector so a call to `push_back` doesn't reshuffle the storage.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -205,7 +206,11 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
-  storage_->write(converter_ ? converter_->convert(message) : message);
+  cache_.push_back(converter_ ? converter_->convert(message) : message);
+
```
Karsten1987: remove new line

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_write_interface.hpp
```diff
@@ -34,6 +34,8 @@ class ROSBAG2_STORAGE_PUBLIC BaseWriteInterface
   virtual ~BaseWriteInterface() = default;
 
   virtual void write(std::shared_ptr<const SerializedBagMessage> msg) = 0;
+  // Method for bulk writes based on transactions
```
Karsten1987: remove comment as this is implementation specific. Add a new line between the two functions though.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_write_interface.hpp
```diff
@@ -34,6 +34,8 @@ class ROSBAG2_STORAGE_PUBLIC BaseWriteInterface
...
+  virtual void bulk_write(std::vector<std::shared_ptr<const SerializedBagMessage>> msg) = 0;
```
Karsten1987: please pass the vector as `const &`.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -43,7 +44,7 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
 {
 public:
   SqliteStorage() = default;
-  ~SqliteStorage() override = default;
+  ~SqliteStorage() override;
```
Karsten1987: new line before destructor.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -56,6 +57,9 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
 
   void write(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message) override;
 
+  // API for bulk writes
```
Karsten1987: remove comment or write it as proper doxygen

Karsten1987:
@sriramster did you have a chance to look at the review? I would love to have this PR merged for the Foxy release.

sriramster:
Hi @Karsten1987  have updated the patch with review comments.

Karsten1987:
@sriramster please rebase your branch on top of `origin/master`. There are a bunch of changes in your PR which are already on master.

Similarly, please make sure to run `colcon build` and `colcon test` locally in your workspace. Your current branch doesn't compile. You can see the status of it in the GitHub action.

sriramster:
Hi @Karsten1987 

I do run colcon build and colcon test before pushing commits. I did find nothing suspicious, but I think the rebase was messed up. Anyways I've rebased, built it and ran the colcon test. Could you please have a look? 

Sriram

sriramster:
Hello @Karsten1987 

Have you had time to look into this? 

Thanks

Karsten1987:
@sriramster I've opened https://github.com/sriramster/rosbag2/pull/2 against your branch, which addresses my outstanding comments.

I still have test failures which your changes which may conflict with the existing tests for rosbag splitting. I'll have to take another look at these. In the meantime, I would recommend you rebase your branch and have a look at my PR (and eventually merge it) so we can push forward on this PR.

Karsten1987:
@sriramster I actually think it makes more sense at this point to simply synchronize your branch here with mine (`karsten_dev_2`) as that one is already rebased and has a bunch of commits on it which addresses the comments and adds additional tests.

```
git checkout sriram_dev_2
git reset --hard origin/karsten_dev_2
```
Let me know what you think. Alternatively, we can close this PR and re-open it from the other branch.

Karsten1987:
a first CI run (on `karsten_dev_2`):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9828/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5449/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7992/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9689/)

sriramster:
Hello @Karsten1987 

Whichever is easier, I've rebased synchronized sriram_dev_2 with karsten_dev_2, if you want me to create another branch and raise a new pull request I'll do it. 

I think the time we spent on this PR is what created the mess in my branch. Anyways, let me know. I'll be pushing sriram_dev_2 if in case we close the PR on the same branch. 

Sriram

Karsten1987:
thanks a bunch @sriramster for iterating with me over this. Your branch looks good and I think we can take it from there.

New round of CI from your branch:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5460/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8006/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9699/)

Karsten1987:
If I could get a review on this one from somebody else, that would be appreciated. I think this is good to go, but I am happy about another pair of eyes.

rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
```diff
@@ -26,11 +26,14 @@ struct StorageOptions
...
-   * A value of 0 indicates that bagfile splitting will not be used.
-   */
+  // The maximum size a bagfile can be, in bytes, before it is split.
+  // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_size;
+
+  // The chunk size indiciates how many messages should be hold in cache
+  // before these being written to disk.
+  // Defaults to 1, and effectively disables the caching.
+  size_t chunk_size = 1;
```
zmichaels11: Is there a reason why `size_t` is used instead of `uint64_t`?
Also what happens if `0` is used?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -79,6 +79,9 @@ void SequentialWriter::open(
 {
   max_bagfile_size_ = storage_options.max_bagfile_size;
   base_folder_ = storage_options.uri;
+  chunk_size_ = storage_options.chunk_size;
```
zmichaels11: I think this function should throw if `chunk_size_ == 0` is invalid.
Something like:
```c++
rcpputils::require_true(storage_options.chunk_size >= 1, "chunk_size must be at least 1.");
```
And then add `\throw std::invalid_argument` to the header documentation.

**This suggestion only applies if a value of 0 is invalid.**

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -203,7 +206,20 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
-  storage_->write(converter_ ? converter_->convert(message) : message);
+  // No need to cache if there's only space for 1 message
+  // We'll call write directly in this case.
+  if (chunk_size_ <= 1) {
+    storage_->write(converter_ ? converter_->convert(message) : message);
+    return;
+  }
```
zmichaels11: I'd prefer this to have the rest of the function written within an else block rather than having a return.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -121,6 +162,30 @@ void SqliteStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMe
...
+  }
+
+  activate_transaction();
+
+  for (auto & message : messages) {
+    auto topic_entry = topics_.find(message->topic_name);
+    if (topic_entry == end(topics_)) {
+      throw SqliteException(
+              "Topic '" + message->topic_name +
+              "' has not been created yet! Call 'create_topic' first.");
```
piraka9011, line 179:
```suggestion
      std::stringstream error_msg;
      error_msg << "Topic '" << message->topic_name << "' has not been created yet! Call 'create_topic' first.";
      throw SqliteException{error_msg.str()};
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Karsten1987: This is totally up for discussion. I think 0 is a valid entry as well. Maybe even a better indicator to disable the cache. Setting a cache size of 1 could then technically still use the internal cache.

Thoughts? Defaulting to zero and using the cache with everything except 0?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: This pattern is currently used throughout the file. I think we can open a follow-up style PR for this.

rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
Karsten1987: I know outside of this PR, but why can't the `max_bagfile_size` field above not be similarly defaulted to `0`?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Karsten1987: Ok, so I've opted for the one I've described. Only setting it to `0` will disable the cache. Every other value will use the cache.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Karsten1987: should be done.

Karsten1987:
new round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5482/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8026/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9722/)

Karsten1987:
In order to verify that the transactions are used correctly, I had to expose the `max_cache_size` through the `ros2cli` interface in https://github.com/ros2/rosbag2/pull/225/commits/bac35581e9d0c8af8b6f7ce9c071b04703223b54.
I tried to put that step actually in a separate PR, but it turned out that the sqlite3 statements were wrongly placed. That's corrected in 83042cc.

With the option being exposed in the command line interface, I'll have to enhance this PR with some end-to-end tests.

Karsten1987:
CI (with cache end to end test):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9867/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5483/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8027/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9723/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9868/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5484/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8028/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9724/)

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -113,6 +113,11 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
 
+  // Intermediate cache to write multiple messages into the storage.
+  // `max_cache_size` is the amount of messages to hold in storage before writing to disk.
+  size_t max_cache_size_;
```
zmichaels11, line 118: `max_cache_size` is a `uint64_t` in `StorageOptions`

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -253,3 +255,67 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
     EXPECT_EQ(expected_path, path);
   }
 }
+
+TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
+  const size_t counter = 1000;
+  const size_t max_cache_size = 100;
```
zmichaels11, line 261: use `uint64_t` to match `cache_size`

rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp
zmichaels11: IIRC not specifying a value for an int will default it to 0.
The code is more readable with it explicitly initialized to 0, so I think there should be a follow-up PR default initializing `max_bagfile_size` to 0.

sriramster:
@Karsten1987  Thanks for considering this. I see a lot of activity here, hope this gets merged soon.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
Karsten1987: good catch!

Karsten1987:
one more round:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9872/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5488/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8032/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9728/)

PR #228, opened 2019-12-04
Add workflow to run colcon build and colcon test

thomas-moulard:


thomas-moulard:
@Karsten1987 I'll double check tomorrow, but it seems this is working!

Karsten1987:
@thomas-moulard there seems to be quite some errors in the testing phase.
From what I can see, they are unrelated to the actual rosbag tests, but given that there are failures occurring, but the overall action returns as `successful` I am wondering if tests failures would be reported correctly. 
Do you have an example where the same code runs with a warning/error on test failures?

thomas-moulard:
We got CI failures in the past for sure, most likely what you're interpreting as an error is related to code coverage I believe (and hence is expected).

I opened a draft PR to demonstrate a failling build, it's this commit + another one breaking the build on purpose: https://github.com/ros2/rosbag2/pull/231

zmichaels11:
Rebased onto master and re-running GH Actions

PR #230, opened 2019-12-10
[Compression - 6] Add Zstd file decompression implementation

piraka9011:
## Changes

* Add `zstd_decompressor` header and source
* Add a new unit test for decompression
  * Compress a file then decompresses it using the same format. Make sure original size is the same.

## Dependencies
- [x] PR #220. 
- [x] Rebased after #220 is merged.

### DIFF
--- a/rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
@@ -29,6 +29,11 @@
 namespace rosbag2_compression
 {
 
+/**
+ * A BaseDecompressorInterface that is used to decompress bagfiles stored using ZStandard compression.
+ *
+ * ZstdDecompressor should only be initialized by Reader.
+ */
 class ROSBAG2_COMPRESSION_PUBLIC ZstdDecompressor : public BaseDecompressorInterface
 {
 public:
--- a/rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
@@ -1,0 +1,49 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2_COMPRESSION__ZSTD_DECOMPRESSOR_HPP_
+#define ROSBAG2_COMPRESSION__ZSTD_DECOMPRESSOR_HPP_
+
+#include <zstd.h>
+
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <string>
+
+#include "rosbag2_compression/base_decompressor_interface.hpp"
+#include "rosbag2_compression/visibility_control.hpp"
+
+namespace rosbag2_compression
+{
+
+class ROSBAG2_COMPRESSION_PUBLIC ZstdDecompressor : public BaseDecompressorInterface
+{
+public:
+  ZstdDecompressor() = default;
+
+  ~ZstdDecompressor() = default;
+
+  std::string decompress_uri(const std::string & uri) override;
+
+  void decompress_serialized_bag_message(
+    rosbag2_storage::SerializedBagMessage * bag_message) override;
+
+  std::string get_decompression_identifier() const override;
+};
+
+}  // namespace rosbag2_compression
+
+#endif  // ROSBAG2_COMPRESSION__ZSTD_DECOMPRESSOR_HPP_
--- a/rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
@@ -71,9 +71,10 @@ std::vector<uint8_t> get_input_buffer(const std::string & uri)
     throw std::runtime_error{errmsg.str()};
   }
 
-  // Allocate and read in
-  std::vector<uint8_t> compressed_buffer;
-  compressed_buffer.resize(compressed_buffer_length);
+  // Initialize compress_buffer with size = compressed_buffer_length.
+  // Uniform initialization cannot be used here since it will choose
+  // the initializer list constructor instead.
+  std::vector<uint8_t> compressed_buffer(compressed_buffer_length);
 
   const auto nRead = fread(
     compressed_buffer.data(), sizeof(decltype(compressed_buffer)::value_type),
@@ -207,8 +208,10 @@ std::string ZstdDecompressor::decompress_uri(const std::string & uri)
     ZSTD_getFrameContentSize(compressed_buffer.data(), compressed_buffer_length);
   check_frame_content(decompressed_buffer_length);
 
-  std::vector<uint8_t> decompressed_buffer;
-  decompressed_buffer.resize(decompressed_buffer_length);
+  // Initializes decompressed_buffer with size = decompressed_buffer_length.
+  // Uniform initialization cannot be used here since it will choose
+  // the initializer list constructor instead.
+  std::vector<uint8_t> decompressed_buffer(decompressed_buffer_length);
 
   const auto decompression_result = ZSTD_decompress(
     decompressed_buffer.data(), decompressed_buffer_length,
--- a/rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
@@ -13,33 +13,102 @@
 // limitations under the License.
 
 #include <chrono>
-#include <memory>
+#include <cstdio>
+#include <sstream>
 #include <string>
+#include <vector>
 
 #include "rcpputils/filesystem_helper.hpp"
 
-#include "logging.hpp"
 #include "rosbag2_compression/zstd_decompressor.hpp"
+
+#include "rosbag2_storage/filesystem_helper.hpp"
+
+#include "logging.hpp"
 
 namespace
 {
-std::unique_ptr<char[]> get_input_buffer(const std::string & uri, size_t & buffer_length)
+constexpr const char DECOMPRESSION_IDENTIFIER[] = "zstd";
+
+FILE * open_file(const std::string & uri, const std::string & read_mode)
 {
-  try {
-    std::ifstream infile{uri};
-    infile.exceptions(std::ifstream::failbit);
-    // Get size and allocate
-    infile.seekg(0, std::ios::end);
-    buffer_length = infile.tellg();
-    auto decompressed_buffer = std::make_unique<char[]>(buffer_length);
-    // Go back and read in contents
-    infile.seekg(0 /* off */, std::ios::beg);
-    infile.read(decompressed_buffer.get(), buffer_length);
-    return decompressed_buffer;
-  } catch (std::ios_base::failure & fail) {
-    ROSBAG2_COMPRESSION_LOG_ERROR_STREAM(
-      "Caught IO stream exception while reading file: " << fail.what());
-    throw fail;
+  FILE * fp{nullptr};
+#ifdef _WIN32
+  fopen_s(&fp, uri.c_str(), read_mode.c_str());
+#else
+  fp = std::fopen(uri.c_str(), read_mode.c_str());
+#endif
+  return fp;
+}
+
+std::vector<uint8_t> get_input_buffer(
+  const std::string & uri,
+  size_t & compressed_buffer_length)
+{
+  // Get the file size
+  compressed_buffer_length = rosbag2_storage::FilesystemHelper::get_file_size(uri);
+
+  // Read in buffer, handling accordingly
+  const auto file_pointer = open_file(uri.c_str(), "rb");
+  if (file_pointer == nullptr) {
+    std::stringstream errmsg;
+    errmsg << "Failed to open file: \"" << uri << "\" for binary reading!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  // Allocate and read in
+  std::vector<uint8_t> compressed_buffer;
+  compressed_buffer.reserve(compressed_buffer_length);
+  fread(compressed_buffer.data(), sizeof(uint8_t), compressed_buffer_length, file_pointer);
+
+  if (ferror(file_pointer)) {
+    fclose(file_pointer);
+
+    std::stringstream errmsg;
+    errmsg << "Unable to read binary data from file: \"" << uri << "\"!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  fclose(file_pointer);
+  return compressed_buffer;
+}
+
+void write_output_buffer(
+  const uint8_t * output_buffer,
+  const size_t output_buffer_length,
+  const std::string & uri)
+{
+  const auto file_pointer = open_file(uri.c_str(), "wb");
+  if (file_pointer == nullptr) {
+    std::stringstream errmsg;
+    errmsg << "Failed to open file: \"" << uri << "\" for binary writing!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  fwrite(output_buffer, sizeof(uint8_t), output_buffer_length, file_pointer);
+
+  if (ferror(file_pointer)) {
+    fclose(file_pointer);
+
+    std::stringstream errmsg;
+    errmsg << "Unable to write decompressed data to file: \"" << uri << "\"!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  fclose(file_pointer);
+}
+
+void throw_on_zstd_error(const size_t compression_result)
+{
+  if (ZSTD_isError(compression_result)) {
+    std::stringstream error;
+    error << "ZSTD decompression error: " << ZSTD_getErrorName(compression_result);
+
+    throw std::runtime_error(error.str());
   }
 }
 
@@ -47,30 +116,25 @@ void check_frame_content(const size_t frame_content)
 {
   if (frame_content == ZSTD_CONTENTSIZE_ERROR) {
     throw std::runtime_error("File not compressed with Zstd.");
-  }
-  if (frame_content == ZSTD_CONTENTSIZE_UNKNOWN) {
-    ROSBAG2_COMPRESSION_LOG_WARN("Unable to determine file size, considering upper bound.");
+  } else if (frame_content == ZSTD_CONTENTSIZE_UNKNOWN) {
+    throw std::runtime_error("Unable to determine file size");
   }
 }
 
-void check_decompression_result(const size_t decompression_result)
+void print_decompression_statistics(
+  const std::chrono::high_resolution_clock::time_point start,
+  const std::chrono::high_resolution_clock::time_point end,
+  const size_t decompressed_size,
+  const size_t compressed_size)
 {
-  if (ZSTD_isError(decompression_result)) {
-    std::stringstream error;
-    error << "ZSTD compression error: " << ZSTD_getErrorName(decompression_result);
-    throw std::runtime_error(error.str());
-  }
-  ROSBAG2_COMPRESSION_LOG_DEBUG("ZSTD decompressed file.");
-}
+  const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+  const auto decompression_ratio =
+    static_cast<double>(decompressed_size) / static_cast<double>(compressed_size);
 
-void print_statistics(
-  std::chrono::high_resolution_clock::time_point start,
-  std::chrono::high_resolution_clock::time_point end)
-{
-  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
   ROSBAG2_COMPRESSION_LOG_DEBUG_STREAM(
     "Decompression statistics:\n" <<
-      "Time: " << duration.count() << " microseconds");
+      "Time: " << duration.count() << " microseconds" <<
+      "Compression Ratio: " << decompression_ratio);
 }
 }  // namespace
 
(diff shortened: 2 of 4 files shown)


rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+#include "rcpputils/filesystem_helper.hpp"
+
+#include "rosbag2_compression/zstd_decompressor.hpp"
+
+#include "rosbag2_storage/filesystem_helper.hpp"
+
+#include "logging.hpp"
+
+namespace
```
piraka9011: Maybe move these to a common source file. But that's for another PR.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
```diff
@@ -81,3 +83,28 @@ TEST_F(CompressionHelperFixture, zstd_compress_file_uri)
...
+TEST_F(CompressionHelperFixture, zstd_decompress_file_uri)
+{
+  const auto uri = rosbag2_storage::FilesystemHelper::concat({temporary_dir_path_, "file1.txt"});
+  create_garbage_file(uri);
+  const auto initial_file_size = rosbag2_storage::FilesystemHelper::get_file_size(uri);
+
+  auto zstd_compressor = rosbag2_compression::ZstdCompressor();
+  const auto compressed_uri = zstd_compressor.compress_uri(uri);
+
+  ASSERT_EQ(0, std::remove(uri.c_str()));  // the test is invalid if the initial file is not deleted
```
piraka9011: nit: `The test...`

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: The helper functions for reading a file into a buffer and writing a file into a buffer cannot be pulled out since their implementations will change once streaming is supported.

zmichaels11:
Gist: https://gist.githubusercontent.com/zmichaels11/8caa09c53ea15fc9b854eafe8eccc589/raw/f8650ffdf23ddc9842120ed4479a0c8462cf9749/ros2.repos
BUILD args: --packages-up-to rosbag2_compression
TEST args: --packages-select rosbag2_compression
Job: ci_launcher

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8908/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4690/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7290/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8825/)

zmichaels11:
@Karsten1987 this PR is ready for review

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+    error << "ZSTD decompression error: " << ZSTD_getErrorName(compression_result);
+
+    throw std::runtime_error(error.str());
+  }
+}
+
+void check_frame_content(const size_t frame_content)
+{
+  if (frame_content == ZSTD_CONTENTSIZE_ERROR) {
+    throw std::runtime_error("File not compressed with Zstd.");
```
mm318: Is not being compressed with `Zstd` the only possibility that returns `ZSTD_CONTENTSIZE_ERROR`? Perhaps include "file could be corrupt", etc.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+  }
+}
+
+void print_decompression_statistics(
+  const std::chrono::high_resolution_clock::time_point start,
+  const std::chrono::high_resolution_clock::time_point end,
+  const size_t decompressed_size,
+  const size_t compressed_size)
+{
+  const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
```
mm318: Is microseconds the reasonable unit of measure? Will the rosbags typically be so small that they're compressable in the microsecond range?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+  if (file_pointer == nullptr) {
+    std::stringstream errmsg;
+    errmsg << "Failed to open file: \"" << uri << "\" for binary reading!";
+
+    throw std::runtime_error(errmsg.str());
+  }
+
+  // Allocate and read in
+  std::vector<uint8_t> compressed_buffer;
+  compressed_buffer.reserve(compressed_buffer_length);
```
mm318: You must `.resize()` before using `.data()`
```suggestion
  compressed_buffer.resize(compressed_buffer_length);
```

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+  fopen_s(&fp, uri.c_str(), read_mode.c_str());
+#else
+  fp = std::fopen(uri.c_str(), read_mode.c_str());
+#endif
+  return fp;
+}
+
+std::vector<uint8_t> get_input_buffer(
+  const std::string & uri,
+  size_t & compressed_buffer_length)
```
mm318: By using `.resize()`, you won't need to output `compressed_buffer_length`.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  size_t compressed_buffer_length{0};
+  const auto compressed_buffer = get_input_buffer(uri, compressed_buffer_length);
+
+  const auto decompressed_buffer_length =
+    ZSTD_getFrameContentSize(compressed_buffer.data(), compressed_buffer_length);
+  check_frame_content(decompressed_buffer_length);
+
+  std::vector<uint8_t> decompressed_buffer;
+  decompressed_buffer.reserve(decompressed_buffer_length);
```
mm318:
```suggestion
  decompressed_buffer.resize(decompressed_buffer_length);
```

rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <memory>
+#include <sstream>
+#include <string>
+
+#include "rosbag2_compression/base_decompressor_interface.hpp"
+#include "rosbag2_compression/visibility_control.hpp"
+
+namespace rosbag2_compression
+{
+
```
dabonnie: Since this is the only implemented decompressor I think it warrants documentation (https://google.github.io/styleguide/cppguide.html#Class_Comments).

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: `reserve` was chosen over `resize` due to `resize` default initializing each element.
I'll look into how the compiler chooses to handle this

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+#include "rosbag2_compression/zstd_decompressor.hpp"
+
+#include "rosbag2_storage/filesystem_helper.hpp"
+
+#include "logging.hpp"
+
+namespace
+{
+constexpr const char DECOMPRESSION_IDENTIFIER[] = "zstd";
```
dabonnie: This naming convention doesn't follow the [ROS2 style guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#id3).

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
```diff
@@ -81,3 +83,28 @@ TEST_F(CompressionHelperFixture, zstd_compress_file_uri)
...
+  auto zstd_decompressor = rosbag2_compression::ZstdDecompressor();
+  const auto decompressed_uri = zstd_decompressor.decompress_uri(compressed_uri);
+
+  const auto expected_decompressed_uri = uri;
+  const auto decompressed_file_size =
+    rosbag2_storage::FilesystemHelper::get_file_size(decompressed_uri);
+
+  EXPECT_NE(compressed_uri, uri);
+  EXPECT_NE(decompressed_uri, compressed_uri);
+  EXPECT_EQ(uri, expected_decompressed_uri);
```
mm318: Can you add a test comparing the contents of the files at `uri` and `decompressed_uri`?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+
```
dabonnie: Please document the anonymous functions for the future maintainers of the code, especially since they can throw. Including purpose, in the case of printing metrics below, would also be handy.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
mm318: Okay, I understand the performance concern, but using `.data()` after `.reserve()` is not valid: https://stackoverflow.com/questions/38972443/use-stdvectordata-after-reserve

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
```diff
@@ -81,3 +83,28 @@ TEST_F(CompressionHelperFixture, zstd_compress_file_uri)
   EXPECT_GT(compressed_file_size, 0u);
   EXPECT_TRUE(rosbag2_storage::FilesystemHelper::file_exists(compressed_uri));
 }
+
+TEST_F(CompressionHelperFixture, zstd_decompress_file_uri)
```
dabonnie: I don't see a test to check for bad input, i.e., the compressed file is invalid.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Would this test just verify that the file contents are not the same?

mm318: You take a file, compress it, then decompress it, and it will end up different? Why is that?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,201 @@
...
+  const std::chrono::high_resolution_clock::time_point start,
+  const std::chrono::high_resolution_clock::time_point end,
+  const size_t decompressed_size,
+  const size_t compressed_size)
+{
+  const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+  const auto decompression_ratio =
+    static_cast<double>(decompressed_size) / static_cast<double>(compressed_size);
+
+  ROSBAG2_COMPRESSION_LOG_DEBUG_STREAM(
```
dabonnie: Consider printing in JSON or another easily machine parseable format.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
mm318: Actually, forget `.resize()`, just construct the `vector` with the fill constructor (type (4) shown [here](https://en.cppreference.com/w/cpp/container/vector/vector)).

Please don't be concerned with the penalty of zero-initialization of the elements. That's the cost of using C++ over using C. You can get around it by writing your own allocator that does default-initialization instead of zero-/value-initialization for `uint8_t` and other POD types, but definitely please don't go down that route.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Oh I misread the your comment. I'll see what I can do for that.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Used `milliseconds` and converted to seconds by dividing by 1000.0. I think this should be more human readable.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Changed error message to just state that an error occurred on `ZSTD_CONTENTSIZE_ERROR` since ZStandard doesn't elaborate further why it failed.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Went with`resize`. 
I tried to invoke type (4) via uniform initialization, however the compiler kept selecting the initializer list constructor.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: See: https://github.com/ros2/rosbag2/pull/230#discussion_r362152097 for resolution

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: See: https://github.com/ros2/rosbag2/pull/230#discussion_r362152097 for resolution

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Added unit test that checks decompressed file contents against initial.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Added unit test that checks for failure when a bad uri is passed.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
mm318: Hmm, that's strange. Have you also tried constructing using brackets `()` instead of braces `{}`?

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
```diff
@@ -81,3 +99,68 @@ TEST_F(CompressionHelperFixture, zstd_compress_file_uri)
...
+  auto decompressor = rosbag2_compression::ZstdDecompressor{};
+  const auto decompressed_uri = decompressor.decompress_uri(compressed_uri);
+
+  const auto initial_data = read_file(uri);
+  const auto initial_file_size = rosbag2_storage::FilesystemHelper::get_file_size(uri);
+  const auto decompressed_data = read_file(decompressed_uri);
+  const auto decompressed_file_size =
+    rosbag2_storage::FilesystemHelper::get_file_size(decompressed_uri);
+
+  EXPECT_EQ(initial_data.size() * sizeof(char), initial_file_size);
```
mm318: Nit: It's a little confusing why it's sometimes `sizeof(char)` and sometimes `sizeof(uint8_t)`. Try something like `sizeof(decltype(initial_data)::value_type)` instead.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: That does work, however I think its kind of confusing that old style initialization has to be used.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Its `sizeof(char)` because `ifstream` can only read in `char` and on some systems `char` is not 1 byte wide. This is only done in the tests for comparing file contents.

I've changed the code to use `sizeof(decltype(vec)::value_type)` when possible since its more robust.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
mm318: You'll have to come to accept that C++ is nowhere near being a consistent language :)

zmichaels11: I changed it to use the type (4) constructor and added comments stating that uniform initialization cannot be used for initializing a `std::vector` with size.

rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
```diff
@@ -0,0 +1,54 @@
...
+{
+public:
+  ZstdDecompressor() = default;
+
+  ~ZstdDecompressor() = default;
+
+  std::string decompress_uri(const std::string & uri) override;
+
+  void decompress_serialized_bag_message(
+    rosbag2_storage::SerializedBagMessage * bag_message) override;
```
mm318: Why is this being passed by pointer? Will the `SerializedBagMessage` be modified in-place?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Writes the output buffer to the specified file path.
+ * \param output_buffer is the data to write.
+ * \param output_buffer_length is the number of bytes to write.
+ * \param uri is the relative file path to the output storage.
+ */
+void write_output_buffer(
+  const uint8_t * output_buffer,
+  const size_t output_buffer_length,
```
mm318: I think it's still better at this point for `write_output_buffer()` to have a more C++-like interface, eg. pass in a `const vector<uint8_t> &`.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
```diff
@@ -81,3 +101,68 @@ TEST_F(CompressionHelperFixture, zstd_compress_file_uri)
...
+  EXPECT_EQ(decompressed_data.size() * sizeof(char), decompressed_file_size);
```
mm318: Can
```
sizeof(decltype(initial_data)::value_type)
sizeof(decltype(decompressed_data)::value_type)
```
be used here?

Btw, `sizeof(char) == 1` always (http://www.cs.technion.ac.il/users/yechiel/c++-faq/sizeof-char.html), even if `CHAR_BITS > 8` (https://stackoverflow.com/questions/21080114/if-char-bit-8-what-is-the-size-of-other-types).

rosbag2_compression/include/rosbag2_compression/zstd_decompressor.hpp
zmichaels11: That was a design choice done in the PoC. I believe it may have been chosen due to `SerializedBagMessage` being wrapped in a `shared_ptr` upstream.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Done.
This might revert in the future when we implement streaming, however we don't need it now so passing `vector` by const ref is better.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
zmichaels11: Interesting, one of the reasons why `vector<uint8_t>` is used over `vector<char>` is to handle systems when "`char` is not a byte".

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,239 @@
...
+    throw std::runtime_error{"Unable to determine file size due to error."};
+  } else if (frame_content == ZSTD_CONTENTSIZE_UNKNOWN) {
+    throw std::runtime_error{"Unable to determine file size."};
+  }
+}
+
+/**
+ * Prints decompression statistics to the debug log stream.
+ * The log statement is formatted as JSON.
+ *
```
dabonnie: Nit: include an example of a printed line in the doc string.

rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp
mm318: If I understand correctly, `char` is always a byte, and if a byte is more than 8 bits on a particular system then `uint8_t` will not be available on that system (your `vector<uint8_t>` won't compile). Let's not worry about that right now.

zmichaels11: Yeah, that's all for a different PR.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Added the example to the documentation.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+  const size_t compressed_size)
+{
+  const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+  const auto decompression_ratio =
+    static_cast<double>(decompressed_size) / static_cast<double>(compressed_size);
+
+  ROSBAG2_COMPRESSION_LOG_DEBUG_STREAM(
+    "\"Decompression statistics\" : {" <<
+      "\"Time\" : " << (duration.count() / 1000.0) <<
+      ", \"Compression Ratio\" : " << decompression_ratio <<
```
piraka9011: `Decompression Ratio` and not Compression? Does it make sense to keep this though?

piraka9011: `Decompression Ratio`? We're not compressing.

zmichaels11: Fixed

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -190,7 +190,7 @@ void print_decompression_statistics(
   ROSBAG2_COMPRESSION_LOG_DEBUG_STREAM(
     "\"Decompression statistics\" : {" <<
       "\"Time\" : " << (duration.count() / 1000.0) <<
-      ", \"Compression Ratio\" : " << decompression_ratio <<
+      ", \"Deompression Ratio\" : " << decompression_ratio <<
```
dabonnie: Typo ;-)

zmichaels11:
@Karsten1987 friendly ping requesting review

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+  const auto compressed_buffer_length = rosbag2_storage::FilesystemHelper::get_file_size(uri);
+
+  // Read in buffer, handling accordingly
+  const auto file_pointer = open_file(uri.c_str(), "rb");
+  if (file_pointer == nullptr) {
+    std::stringstream errmsg;
+    errmsg << "Failed to open file: \"" << uri << "\" for binary reading!";
+
+    throw std::runtime_error{errmsg.str()};
+  }
```
Karsten1987: Could this check be done before getting the size of the `uri`?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * Read a file from the supplied uri into a vector.
+ *
+ * \param uri is the path to the file.
+ * \return the contents of the buffer as a vector.
+ */
+std::vector<uint8_t> get_input_buffer(const std::string & uri)
+{
+  // Get the file size
+  const auto compressed_buffer_length = rosbag2_storage::FilesystemHelper::get_file_size(uri);
```
Karsten1987: I think we should react here in case the returned filesize is 0 (aka the file is non-existing)

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    throw std::runtime_error{errmsg.str()};
+  }
+
+  // Initialize compress_buffer with size = compressed_buffer_length.
+  // Uniform initialization cannot be used here since it will choose
+  // the initializer list constructor instead.
+  std::vector<uint8_t> compressed_buffer(compressed_buffer_length);
+
+  const auto nRead = fread(
```
Karsten1987: nit: generally we don't have camelcase variable names in ROS2.

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+    compressed_buffer.data(), sizeof(decltype(compressed_buffer)::value_type),
+    compressed_buffer_length, file_pointer);
+
+  if (nRead != compressed_buffer_length) {
+    ROSBAG2_COMPRESSION_LOG_ERROR_STREAM("Bytes read !(" <<
+      nRead << ") != compressed_buffer_length (" << compressed_buffer_length <<
+      ")!");
+    // An error indicator is set by fread, so the following check will throw.
```
Karsten1987: This comment actually refers to line 90, right?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+  const auto file_pointer = open_file(uri.c_str(), "wb");
+  if (file_pointer == nullptr) {
+    std::stringstream errmsg;
+    errmsg << "Failed to open file: \"" << uri << "\" for binary writing!";
+
+    throw std::runtime_error{errmsg.str()};
+  }
+
+  const auto nWrite = fwrite(
+    output_buffer.data(), sizeof(uint8_t),
```
Karsten1987: what if output buffer is empty?

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
```diff
@@ -0,0 +1,242 @@
...
+    error << "ZSTD decompression error: " << ZSTD_getErrorName(compression_result);
+
+    throw std::runtime_error{error.str()};
+  }
+}
+
+/**
+ * Checks if the frame content is valid.
+ */
+void check_frame_content(const size_t frame_content)
```
Karsten1987: given that we have `throw_on_zstd_error` I think this function could be renamed accordingly as it really does only throwing on error.

zmichaels11:
> With compression and decompression being available with this, I would like to see a test which checks the validity of the compression. That is, a file being compressed and decompressed is byte equivalent with the initial file content.

There's a unit test for this: https://github.com/ros2/rosbag2/blob/9e18969f9d0d9b55d80a79124d12fa79b342e25a/rosbag2_compression/test/rosbag2_compression/test_zstd_compressor.cpp#L130

rosbag2_compression/src/rosbag2_compression/zstd_decompressor.cpp
zmichaels11: Yeah, the error log message on L84 exists to provide further information on an error from `fread`.

The comment on L87 is indicating that the code should throw at this state and that the check on L90 does the actual throw since `fread` may have set the error indicator for other reasons besides `nRead != compressed_buffer_length`.

zmichaels11:
@Karsten1987 I'm opening a PR to address your comments: https://github.com/ros2/rosbag2/pull/245

PR #231, opened 2019-12-12
[DO NOT MERGE] Demonstrate CI failure

thomas-moulard:


.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
+name: Test rosbag2
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
Karsten1987: shouldn't this point to `tmoulard/gh-action-ci-build-failure`

Karsten1987:
Interestingly enough, the linters didn't respond anything weird, but I see that the actual build check failed! So that seems like it works well.

.github/workflows/test.yml
thomas-moulard: This just describes what triggers a build: a push on master (for the post-merge check) or a PR. For PRs, whatever branch it is merged on will be used.

PR #241, opened 2019-12-28
Make rosbag2 a metapackage

### LINKED ISSUE #174: need for metapackage
I think there's the need for a sort of a metapackage. This package can then be used conveniently for triggering CI but also be part of the ros2-variants. 

I am thinking about `rosbag2_common_extensions`. Any idea for names?

Karsten1987:
fixes #174 

As discussed in the top level ticket, this PR renames the existing `rosbag2` package to `rosbag2_cpp` and introduces `rosbag2` as a simple metapackage with only listing runtime dependencies.
The renaming also entails changing the namespace accordingly, which makes that PR pretty brutal.


Karsten1987:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4716/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7318/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8855/)

Karsten1987:
windows again after fixing visibility macro:
[Build Status](https://ci.ros2.org/job/ci_windows/8856/)

Karsten1987:
And a final round of windows CI:
[Build Status](https://ci.ros2.org/job/ci_windows/8857/)

Putting this up for review.

Karsten1987:
The namespace should always reflect the package name otherwise this would lead to confusion IMO. Also discussed here: https://github.com/ros2/rosbag2/issues/174#issuecomment-543027661

As for the types, I agree this could have been separately, but really, this PR is already changing so many files and namespaces, that removing the aliases for the types doesn't make a huge impact either.

rosbag2_converter_default_plugins/test/rosbag2_converter_default_plugins/cdr/test_cdr_converter.cpp
```diff
@@ -17,16 +17,20 @@
...
+#include "rosbag2_cpp/typesupport_helpers.hpp"
+#include "rosbag2_cpp/types.hpp"
+
 #include "rosbag2_test_common/memory_management.hpp"
+
 #include "test_msgs/message_fixtures.hpp"
 
-using rosbag2::converter_interfaces::SerializationFormatConverter;
+#include "../../../src/rosbag2_converter_default_plugins/cdr/cdr_converter.hpp"
+#include "../../../src/rosbag2_converter_default_plugins/logging.hpp"
```
piraka9011: This is probably for another PR, but shouldn't these be found via the `CMakeLists.txt`?

piraka9011:
LGTM. It seems all the packages have been renamed correctly.

zmichaels11:
@Karsten1987 I think you should get more people's input before merging due to how big this is.
Besides that, LGTM

rosbag2_converter_default_plugins/test/rosbag2_converter_default_plugins/cdr/test_cdr_converter.cpp
wjwwood: This is ok since the package's structure is known to the tests, but it can be fragile. You could create a relative path in CMake and pass it as a compiler definition instead. At least then it wouldn't be hardcoded in so many places. But the use of a relative path here is in principle safe.

zmichaels11:
@Karsten1987 the build is probably going to fail due to a new unit test I added in the compression metadata PR that was just merged. I can add the fix if you want.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8965/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4735/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7337/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8893/)

piraka9011:
Running another windows build:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8896/)

Karsten1987:
windows: [Build Status](https://ci.ros2.org/job/ci_windows/8898/)

zmichaels11:
@Karsten1987 I think we're good for merge on green CI.

Issue #123, opened 2019-05-22
Performance Issues


julled:
I wanted to test the performance of the current implementation, if its capable of writing multiple Raw cameradatastreams to my SSD. 

Testusecase:
* 1xCamera
* Node to simulate datastream 1280x1040x3 Byte @ 30fps
    * rosidl error when i wanted to create a native image
    * Created data by creating a string with the corresponding size
* using sqlite plugin 
* Ubuntu 18.04 
* ROS2 packages / rosbag via apt installed

Result:
* 4,6 seconds recording: only 32 of ~138 images got saved

Did i miss something, or is rosbag2 not capable yet to save that amount of data?

Martin-Idel:
The SQLite plugin is not optimized as of now - rosbag2 still lacks a good performance testing suite to check for bottlenecks. Since no transactions are used as of today in the database backend, I would expect performance to be really bad for many small messages and better for larger messages at lower numbers.

Sending around 100 MB of data per second could be too much already on some systems.

Karsten1987:
Hmm, looks like we have to dig a bit deeper into performance optimizations for SQLite3. The assumption that it's more suited for larger data seems to be in conflict with this ros-answers question: https://answers.ros.org/question/328814/ros2-bag-issue/

@Martin-Idel Do you know of any good starting point for diving into these performance topics?

Martin-Idel:
I wanted to come back to this issue, because I find it interesting. I think the first step would be to create a good performance test suite to be sure that any changes actually help.

My starting point would probably be https://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite - as you can see there are quite some points to get started. From what I know the actual performance tuning will probably start with:
- Use transactions (this will definitely help, and it might help a lot)
- Use batching within transactions (larger changes required here, but this will help for small messages)
- Look into journal modes (might also require more user settings)

MarcoPedretti:
Hi,

I am facing a similar problem with the following configuration:
- 2xImage msg @~11FPS 1928x1208x4
- 1xPointCloud2 msg @~11FPS 900pts
- sqlite plugin
- Ubuntu 18.04
- USB3.0 External SSD (tested up to 480MB/s)

Here some results:
- **ROS2 nodes + rosbag2 from ROS2** 
  - topic_img_1 - 4.707FPS
  - topic_img_2 - 4.709FPS
  - topic_pts - 4.689FPS
[...]

PR #244, opened 2020-01-03
Simple framework for performance testing

Martin-Idel:
In view of #123 I said that we should have a framework for performance testing. Otherwise there is no good way to show that some change improves on the performance.

This PR adds a package for performance testing. It is extensible and tests can be easily altered to try different scenarios although much of this could be improved.

## Current features:

- Only build the package on demand. To build tests:
```
colcon build --cmake-args -DBUILD_PERFORMANCE=True
```
- Launch file based testing. To run the (currently only) test, do:
```
ros2 launch rosbag2_performance_testing large_message_launch.launch.py
```
- The `large_message_launch` test writes 60 seconds of two large images and one large point cloud (total write about 50 MB per second. That's about one tenth of where it should get more interesting and 1/100th of where rosbag might hopefully one day be, but my own system seems to be limited at a couple of 100 MB/s)
- You'll get a warning beforehand showing how much space is needed (this needs to be changed manually if you change the test) as well as a summary after the test completed showing if the test passes and if not which messages were lost
- New tests can be added by writing a new launch file (+ publishers if necessary), copying a bit of code and writing a yaml file with the expected outcome.

### The test will gather the following information:
- average and peak CPU-load (as measured via Python over 1s every 5s)
- average and peak memory load (as measured via Python every 5s)
- comparing the number of messages in the bagfile after write with an expectation (failing if less than 95% of any topic have been recorded)

### Thing I don't really know how to solve:
- Launch always needs to escalate to SIGTERM to finish resulting in ugly error messages. This might be related to the fact I explicitly call `ros2 bag` in a script with Popen, but I don't quite understand it.
- The output of the test could probably be improved

### Things that could be improved, but I think it's not necessary for now:
- Use parameters to make rates of publishers configurable
- Require test author to specify only the size of each message and calculate the size of a bagfile from there
- Factor out launch file code to launch the test script so that it doesn't have to be copied for new tests and then reorder tests

There is probably much more and maybe a fundamental flaw I overlooked, so any feedback is welcome!

## First findings
Note that the test already rarely passes on my machine (similar to what is reported in #123. Usually, one of the two images gets lost nearly all of the time, so I guess that the `write` takes too long and the concurrent write of the other topic just gets lost. If I change the test to just subscribe to one image topic but with the double rate (i.e. same I/O-load), the test usually passes. If I double the frequency (i.e. load > 100 MB) it fails again.

So what I would suggest is to first look into buffering and sorting out the concurrent writes to the storage. Maybe this will already stabilize the scenario here. Once we have buffering, we could also think about transactions writing a collection of messages all at once. 
I hope to look into it, but I can probably not propose a mergeable PR anytime soon.


piraka9011:
@Martin-Idel Thanks a lot for this! We were just discussing how we could measure `rosbag2` performance using ApexAI's [`performance_test`](https://gitlab.com/ApexAI/performance_test/) package.

I think we can leverage a lot of their tools as opposed to rewriting some nodes.
That and using the [`ros-tooling/file-talker`](https://github.com/ros-tooling/file_talker/) would be good.

Thoughts on using those instead of an internal package?

piraka9011:
> The output of the test could probably be improved

See Python's [`check_output()`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) where you can capture what is printed to `stdout`.

> Things that could be improved...

Most of this is taken care of by ApexAI's tools.

> First findings...

Looks like some optimizations need to be made to the SQLite storage plugin. That is for another investigation though. Performance testing would be the best way to start.

Martin-Idel:
@piraka9011 Interesting - I hadn't seen the issues on the aws roadmap! I had seen the performance-test framework a year ago, but back then, it couldn't do what we needed and I'm still not sure how to get what I'm doing here, which is quite simple. The file-talker seems interesting and should get rid of a lot of code.

It is very clear that the SQLite-plugin needs optimization (see also #123 ). I wrote or reviewed most of the original code and it was never optimized - there just wasn't time; we tried to make it simple and extensible and only develop optimized features once performance testing was done. I'm interested in performance optimizations, so I decided to just get started on working on performance testing because as you say, this needs to be done first.

Regarding the "findings", I just added them to show that this would already improve the status quo to get more insight into the performance of rosbag2.

piraka9011:
We can start by creating a launch file (or script) that:
a. Starts the ApexAI performance tests
b. Starts `ros2 bag record -a`.

And then postprocess the results as you did in `performance_test.py` but only:
a. Plot the results using the performance scripts
b. Perform a comparison on the number of messages

Let me know if that's sounds like a good path forward.

Karsten1987:
@Martin-Idel while testing, have you verified that all messages actually arrive correctly and in order? That rosbag loses data is one thing, but before that we should verify that the subscription actually receives all of the sent data to get a qualified statement of how many messages get lost only on rosbag's side and not during DDS/network traffic.

Martin-Idel:
@Karsten1987 Very good point - and that's something I'm thinking about. You are definitely right, some parts could come from the underlying dds. I tried (not in this PR) to count the messages and the timestamps in the subscription code (that's why it's here) and sometimes I seemed to have gotten most messages and sometimes I didn't, so one component of what I see is definitely DDS-related. However, it's also not completely clear because the queue could just have been full and I just dropped some messages and it would be a rosbag2 performance issue. 
Do you get warnings when the queue is full?

So maybe we should have several layers of performance tests? 
- We could completely evade rosbag2_transport first and just optimize the write to the sqlite backend using the rosbag2_cpp API. These tests are definitely easier to interpret and maybe it's enough to begin with?
- We could then try to do the full roundtrip as I describe here

My time is sparse, so if anybody else wants to continue the work, just let me know and I'll stop, otherwise I hope to steadily continue doing this over the next few weeks as I find it interesting.

I would probably start with a simple test suite using the file_talker and the rosbag2_cpp API and see where that gets us. That should be extremely simple to set up.

ros-discourse:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/1

Karsten1987:
@Martin-Idel thanks again for kicking things off here. I do believe we can close this PR in favor of https://github.com/ros2/rosbag2/pull/442 which seems to be inspired by your PR and goes along with what you've mentioned in #123 

I'll go ahead and close it. Please feel free to re-open it in case you'll see a need for it or have further insights.

PR #246, opened 2020-01-07
[Compression - 8] Enable reader to read from compressed files/messages

piraka9011:
This PR allows a Reader to decompress a file or serialized bag message by checking the bag file metadata.
It currently supports the Zstd decompressor only.
Creating a decompressor factory will be part of a later change.

## Changes

* Reader can now read from a compressed file or message.
* Validate metadata keys for compression, supporting older metadata versions.
* Unit test to check that compression occurs when specified in metadata.
* Unit test to check that compression options can be converted b/w string/enum.

Part of [aws-roadmap/#68](https://github.com/ros-security/aws-roadmap/issues/68).

PR #221, opened 2019-11-27
[Compression - 7] Add compression metadata

piraka9011:
## Changes
* Bump metadata version (2 -> 3)
* Adds fields:
  * `compression_format`
  * `compression_mode`
* Update tests to account for new metadata info
* **metadata version 2 is still supported**
  * Add unit test to verify metadata with version 2 can still be read.

PR #250, opened 2020-01-08
[Compression - 9] Enable writer to compress files/messages

piraka9011:
This PR adds compression options to the `rosbag2_cpp` `SequentialWriter` to support compression by file and message.
Compression options are used in the `rosbag2_transport` package to propagate the compression CLI flags.  

## Changes

* Add a function to initialize a compressor.
* Add a compression options struct.
* Change signature of writer's `open()` method to support compression options.
* Unit tests for compression.

## Dependencies

* [x] #246 
* [x] #249

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+=======
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+>>>>>>> Enable reader to read from decompressed files & messages
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
@@ -26,7 +26,7 @@ namespace rosbag2_cpp
  * Modes are used to specify whether to compress by individual serialized bag messages or by file.
  * rosbag2_cpp defaults to NONE.
  */
-enum class ROSBAG2_CPP_PUBLIC CompressionMode : uint32_t
+enum class ROSBAG2_CPP_PUBLIC CompressionMode: uint32_t
 {
   NONE = 0,
   FILE,
@@ -40,7 +40,8 @@ enum class ROSBAG2_CPP_PUBLIC CompressionMode : uint32_t
  * \param compression_mode A case insensitive string that is either "FILE" or "MESSAGE".
  * \return CompressionMode NONE if compression_mode is invalid. FILE or MESSAGE otherwise.
  */
-ROSBAG2_CPP_PUBLIC CompressionMode compression_mode_from_string(const std::string & compression_mode);
+ROSBAG2_CPP_PUBLIC CompressionMode compression_mode_from_string(
+  const std::string & compression_mode);
 
 /**
  * Converts a rosbag2_cpp::CompressionMode enum into a string.
@@ -49,6 +50,15 @@ ROSBAG2_CPP_PUBLIC CompressionMode compression_mode_from_string(const std::strin
  * \return The corresponding mode as a string.
  */
 ROSBAG2_CPP_PUBLIC std::string compression_mode_to_string(CompressionMode compression_mode);
+
+/**
+ * Compression options used in the writer which are passed down from the CLI in rosbag2_transport.
+ */
+struct CompressionOptions
+{
+  std::string compression_format;
+  CompressionMode compression_mode;
+};
 
 }  // namespace rosbag2_cpp
 #endif  // ROSBAG2_CPP__COMPRESSION_OPTIONS_HPP_
--- a/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-=======
-// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
->>>>>>> Enable reader to read from decompressed files & messages
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -149,6 +149,8 @@ private:
   // Used in invoking compression
   rosbag2_cpp::CompressionMode compression_mode_;
 
+  bool should_compress_last_file_;
+
   // Closes the current backed storage and opens the next bagfile.
   void split_bagfile();
 
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -126,8 +126,6 @@ private:
   // Used in invoking compression
   rosbag2_cpp::CompressionMode compression_mode_;
 
-  bool should_compress_last_file_;
-
   // Closes the current backed storage and opens the next bagfile.
   void split_bagfile();
 
@@ -140,13 +138,6 @@ private:
   // Record TopicInformation into metadata
   void finalize_metadata();
 
-  /**
-   * Check to make sure a valid bagfile split size is passed.
-   *
-   * \throws runtime_error If the size is too small (plugin specific).
-   */
-  void check_bagfile_size(const StorageOptions & storage_options);
-
 protected:
   /**
    * Initialize the compressor.
@@ -156,14 +147,9 @@ protected:
   virtual void init_compression(const CompressionOptions & compression_options);
 
   /**
-   * Checks if the compression by file option is specified and a compressor exists.
-   *
-   * If the above conditions are satisfied, compresses the most recent file and updates the
-   * metadata file paths.
-   *
-   * \return True if compression occurred, false otherwise.
+   * Compress the most recent file and update the metadata file path.
    */
-  virtual bool compress_file_and_update_metadata();
+  virtual void compress_last_file();
 
   /**
    * Checks if the compression by message option is specified and a compressor exists.
@@ -173,7 +159,7 @@ protected:
    * \param message The message to compress.
    * \return True if compression occurred, false otherwise.
    */
-  virtual bool compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
+  virtual void compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
 };
 
 }  // namespace writers
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -126,6 +126,8 @@ private:
   // Used in invoking compression
   rosbag2_cpp::CompressionMode compression_mode_;
 
+  bool should_compress_last_file_;
+
   // Closes the current backed storage and opens the next bagfile.
   void split_bagfile();
 
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -20,12 +20,14 @@
 #include <unordered_map>
 #include <vector>
 
+#include "rosbag2_cpp/compression_options.hpp"
 #include "rosbag2_cpp/converter.hpp"
 #include "rosbag2_cpp/serialization_format_converter_factory.hpp"
 #include "rosbag2_cpp/storage_options.hpp"
 #include "rosbag2_cpp/writer_interfaces/base_writer_interface.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
+#include "rosbag2_compression/base_compressor_interface.hpp"
 #include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
 #include "rosbag2_storage/storage_factory_interface.hpp"
@@ -71,7 +73,9 @@ public:
    * \param converter_options options to define in which format incoming messages are stored
    **/
   void open(
-    const StorageOptions & storage_options, const ConverterOptions & converter_options) override;
+    const StorageOptions & storage_options,
+    const ConverterOptions & converter_options,
+    const CompressionOptions & compression_options) override;
 
   void reset() override;
 
@@ -109,6 +113,7 @@ private:
   std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage_;
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
+  std::unique_ptr<rosbag2_compression::BaseCompressorInterface> compressor_;
 
   // Used in bagfile splitting; specifies the best-effort maximum sub-section of a bagfile in bytes.
   uint64_t max_bagfile_size_;
@@ -118,17 +123,54 @@ private:
 
   rosbag2_storage::BagMetadata metadata_;
 
+  // Used in invoking compression
+  rosbag2_cpp::CompressionMode compression_mode_;
+
   // Closes the current backed storage and opens the next bagfile.
   void split_bagfile();
 
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
 
+  /**
+   * Checks if the compression by file option is specified and a compressor exists.
+   *
+   * If the above conditions are satisfied, compresses the most recent file and updates the
+   * metadata file paths.
+   *
+   * \return True if compression occurred, false otherwise.
+   */
+  bool compress_file_and_update_metadata();
+
+  /**
+   * Checks if the compression by message option is specified and a compressor exists.
+   *
+   * If the above conditions are satisfied, compresses the serialized bag message.
+   *
+   * \param message The message to compress.
+   * \return True if compression occurred, false otherwise.
+   */
+  bool compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
+
   // Prepares the metadata by setting initial values.
-  void init_metadata();
+  void init_metadata(const CompressionOptions & compression_options);
 
   // Record TopicInformation into metadata
   void finalize_metadata();
+
+  /**
+   * Check to make sure a valid bagfile split size is passed.
+   *
+   * \throws runtime_error If the size is too small (plugin specific).
+   */
+  void check_bagfile_size(const StorageOptions & storage_options);
+
+  /**
+   * Initialize the compressor.
+   *
+   * \throws runtime_error If the compression implementation does not exist.
+   */
+  virtual void init_compression(const CompressionOptions & compression_options);
 };
 
 }  // namespace writers
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -106,6 +106,29 @@ public:
    */
   void write(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message) override;
 
+protected:
+  /**
+   * Initialize the compressor.
+   *
+   * \throws runtime_error If the compression implementation does not exist.
+   */
+  virtual void init_compression(const CompressionOptions & compression_options);
+
+  /**
+   * Compress the most recent file and update the metadata file path.
+   */
+  virtual void compress_last_file();
+
+  /**
+   * Checks if the compression by message option is specified and a compressor exists.
+   *
+   * If the above conditions are satisfied, compresses the serialized bag message.
+   *
+   * \param message The message to compress.
+   * \return True if compression occurred, false otherwise.
+   */
+  virtual void compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
+
 private:
   std::string base_folder_;
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_;
@@ -133,33 +156,10 @@ private:
   bool should_split_bagfile() const;
 
   // Prepares the metadata by setting initial values.
-  void init_metadata(const CompressionOptions & compression_options);
+  void init_metadata();
 
   // Record TopicInformation into metadata
   void finalize_metadata();
-
-protected:
-  /**
-   * Initialize the compressor.
-   *
-   * \throws runtime_error If the compression implementation does not exist.
-   */
-  virtual void init_compression(const CompressionOptions & compression_options);
-
-  /**
-   * Compress the most recent file and update the metadata file path.
-   */
-  virtual void compress_last_file();
-
-  /**
-   * Checks if the compression by message option is specified and a compressor exists.
-   *
-   * If the above conditions are satisfied, compresses the serialized bag message.
-   *
-   * \param message The message to compress.
-   * \return True if compression occurred, false otherwise.
-   */
-  virtual void compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
 };
 
 }  // namespace writers
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -132,26 +132,6 @@ private:
   // Checks if the current recording bagfile needs to be split and rolled over to a new file.
   bool should_split_bagfile() const;
 
-  /**
-   * Checks if the compression by file option is specified and a compressor exists.
-   *
-   * If the above conditions are satisfied, compresses the most recent file and updates the
-   * metadata file paths.
-   *
-   * \return True if compression occurred, false otherwise.
-   */
-  bool compress_file_and_update_metadata();
-
-  /**
-   * Checks if the compression by message option is specified and a compressor exists.
-   *
-   * If the above conditions are satisfied, compresses the serialized bag message.
-   *
-   * \param message The message to compress.
-   * \return True if compression occurred, false otherwise.
-   */
-  bool compress_message(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message);
-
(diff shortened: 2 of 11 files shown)


rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
```diff
@@ -50,5 +51,14 @@ ROSBAG2_CPP_PUBLIC CompressionMode compression_mode_from_string(const std::strin
  */
 ROSBAG2_CPP_PUBLIC std::string compression_mode_to_string(CompressionMode compression_mode);
 
+/**
+ * Compression options used in the writer which are passed down from the CLI in rosbag2_transport.
+ */
+struct CompressionOptions
+{
+  std::string compression_format;
+  CompressionMode compression_mode;
```
mm318: I think these can be better named. Is `compression_format` referring to the compression algorithm and `compression_mode` referring to the type of content that's being compressed?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -55,31 +57,63 @@ SequentialWriter::SequentialWriter(
...
   reset();
 }
 
-void SequentialWriter::init_metadata()
+void SequentialWriter::check_bagfile_size(const StorageOptions & storage_options)
+{
+  if (storage_options.max_bagfile_size != 0 &&
+    storage_options.max_bagfile_size < storage_->get_minimum_split_file_size())
+  {
+    throw std::runtime_error(
```
mm318: Probably more of a `std::invalid_argument` exception.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -88,25 +122,20 @@ void SequentialWriter::open(
...
-  if (max_bagfile_size_ != 0 &&
-    max_bagfile_size_ < storage_->get_minimum_split_file_size())
-  {
-    throw std::runtime_error(
-            "Invalid bag splitting size given. Please provide a different value.");
-  }
-
-  init_metadata();
+  check_bagfile_size(storage_options);
+  init_compression(compression_options);
```
mm318: Just a note. Not sure what can be done at this point since it seems to be baked into the design of this class, but this is quite a bit of `void` functions that throw.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -162,17 +191,32 @@ void SequentialWriter::remove_topic(const rosbag2_storage::TopicMetadata & topic
...
+bool SequentialWriter::compress_file_and_update_metadata()
+{
+  if (compressor_ && compression_mode_ == CompressionMode::FILE) {
+    // Get the uri of the last file and pop the uri
+    const auto uncompressed_uri = metadata_.relative_file_paths.back();
+    metadata_.relative_file_paths.pop_back();
+
+    // Compress the last file and push the new uri
+    const auto compressed_uri = compressor_->compress_uri(uncompressed_uri);
+    metadata_.relative_file_paths.push_back(compressed_uri);
```
mm318: Can you modify this file path in place instead of popping and pushing?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -162,17 +191,32 @@ void SequentialWriter::remove_topic(const rosbag2_storage::TopicMetadata & topic
   }
 }
 
+bool SequentialWriter::compress_file_and_update_metadata()
+{
+  if (compressor_ && compression_mode_ == CompressionMode::FILE) {
+    // Get the uri of the last file and pop the uri
+    const auto uncompressed_uri = metadata_.relative_file_paths.back();
```
mm318: Not sure how you guarantee `metadata_.relative_file_paths.back()` isn't already an URI to a compressed file. For example, `split_bagfile()` and `reset()` are called right after the other.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -204,6 +259,7 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
   const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
+  compress_message(message);
```
mm318: Do you really gain anything by compressing twice, once at the message level and once at the file level? There must be something in information theory that says this is not very effective.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
piraka9011: Maybe a helper function to check if the URI is already compressed (The extension matches the compression identifier)?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
piraka9011: The compression option passed to the reader only allows you to compress either by file or message. It can't be both.

mm318: Oh, never mind. The `CompressionMode` can only be `FILE` or `MESSAGE`.

rosbag2_cpp/include/rosbag2_cpp/compression_options.hpp
piraka9011: Yes, these were the naming conventions back when we updated the bag metadata in #221.

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -44,7 +44,15 @@ Writer::~Writer()
 void Writer::open(
   const StorageOptions & storage_options, const ConverterOptions & converter_options)
 {
-  writer_impl_->open(storage_options, converter_options);
+  writer_impl_->open(storage_options, converter_options, {});
```
zmichaels11: Nit:
I think it would be cleaner if the last parameter was passed explicitly. So:
```c++
open(storage_options, converter_options, CompressionOptions{});
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -55,31 +57,63 @@ SequentialWriter::SequentialWriter(
   storage_(nullptr),
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
+  compressor_(nullptr),
```
zmichaels11: If you're adding this, you should initialize it using braces.
Ignore this comment otherwise

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -55,31 +57,63 @@ SequentialWriter::SequentialWriter(
...
+
+void SequentialWriter::init_compression(const CompressionOptions & compression_options)
+{
+  if (compression_options.compression_mode != rosbag2_cpp::CompressionMode::NONE) {
+    if (compression_options.compression_format == "zstd") {
+      compressor_ = std::make_unique<rosbag2_compression::ZstdCompressor>();
+    } else {
+      std::stringstream err;
+      err << "Unsupported compression format " << compression_options.compression_format;
+      throw std::runtime_error(err.str());
```
zmichaels11:
```suggestion
     throw std::runtime_error{err.str()};
```

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -249,3 +249,123 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
...
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io)
+  : SequentialWriter(std::move(storage_factory), converter_factory, std::move(metadata_io)) {}
+
+  ~FakeSequentialWriter() override = default;
+
+  void init_compression(const rosbag2_cpp::CompressionOptions & compression_options) override
+  {
+    std::cout << "init_compression called with " << compression_options.compression_format <<
+      " compression format for " <<
+      rosbag2_cpp::compression_mode_to_string(compression_options.compression_mode) << std::endl;
```
zmichaels11: Replace with logger

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -88,25 +122,20 @@ void SequentialWriter::open(
...
-  init_metadata();
+  check_bagfile_size(storage_options);
+  init_compression(compression_options);
+  init_metadata(compression_options);
 }
 
 void SequentialWriter::reset()
 {
   if (!base_folder_.empty()) {
+    compress_file_and_update_metadata();
```
zmichaels11: This function returns a bool status; you're not checking it

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
mm318: It'd be more clear if you did the compression mode check outside of `compress_message()` and `compress_file_and_update_metadata()`. You can leave the `compressor_ != nullptr` check inside those functions though.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
mm318: The happy case seems to be: `split_bagfile()` is called, the next rosbag starts being recorded a new element is pushed back on `metadata_.relative_file_paths`, then recording is stopped and `reset()` is called.

The bad case is: `split_bagfile()` is called, then recording is stopped and `reset()` is called. This still feels like a pretty normal situation.

It's probably better to have a flag that tracks whether `metadata_.relative_file_paths.back()` has been compressed already or not, which gets set when a new rosbag starts and a when the current rosbag ends (I can't tell if this needs to be also threadsafe or not). Just checking the extension doesn't scale when you have more compression algorithms / identifiers available.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -118,17 +123,55 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
...
+
+  /**
+   * Checks if the compression by file option is specified and a compressor exists.
+   *
+   * If the above conditions are satisfied, compresses the most recent file and updates the
+   * metadata file paths.
+   *
+   * \return True if compression occurred, false otherwise.
+   */
+  virtual bool compress_file_and_update_metadata();
```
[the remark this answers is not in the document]
mm318: Given the `if` condition check that's in there right now, returning `false` is definitely not an exception, it's a norm.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
piraka9011: My thoughts are that each of these are a different step when calling `open()` that make sure the different components are setup correctly and throw when they fail to otherwise.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
piraka9011: I added a flag `should_compress_last_file_` but maybe all we need to do is check if we are splitting or not (since we know this is a situation that only occurs when splitting).
This just means checking for `max_bagfile_size_ == 0`.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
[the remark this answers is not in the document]
piraka9011: I made it return a `bool` since we don't check if we should be compressing or not outside of this function. So if calling the function results in a compression we return true, false otherwise.

We could have a similar structure to bag file splitting where we have a `should_compress_file/message()` and check that before calling the compressor. I don't know if that's better from a design standpoint.

And in any case, `compress_uri()` throws when it's unable to compress.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -74,7 +75,7 @@ void SequentialWriter::check_bagfile_size(const StorageOptions & storage_options
   if (storage_options.max_bagfile_size != 0 &&
     storage_options.max_bagfile_size < storage_->get_minimum_split_file_size())
   {
-    throw std::runtime_error(
+    throw std::invalid_argument(
```
zmichaels11:
```c++
std::invalid_argument{...}
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "rosbag2_cpp/writers/sequential_writer.hpp"
 
+#include <assert.h>
```
zmichaels11: Include `cassert` instead http://www.cplusplus.com/reference/cassert/

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -46,6 +49,7 @@ std::string format_storage_uri(const std::string & base_folder, uint64_t storage
 }
 }  // namespace
 
+// TODO(piraka9011) Initialize defaults in header file instead.
```
zmichaels11: Also add a TODO for replacing old-style constructors with uniform initialization

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -55,31 +59,63 @@ SequentialWriter::SequentialWriter(
...
-void SequentialWriter::init_metadata()
+void SequentialWriter::check_bagfile_size(const StorageOptions & storage_options)
+{
+  if (storage_options.max_bagfile_size != 0 &&
+    storage_options.max_bagfile_size < storage_->get_minimum_split_file_size())
+  {
+    throw std::invalid_argument{
+            "Invalid bag splitting size given. Please provide a different value."};
+  }
+}
```
zmichaels11: Is this relevant to this PR?
I think this code block could be extracted and done in a separate PR.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
mm318: I mean the issue is more with `SequentialWriter::open()` being a `void` function, so the only way to propagate failure from your sub-functions is to throw at some point.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -56,31 +58,53 @@ SequentialWriter::SequentialWriter(
...
-void SequentialWriter::init_metadata()
+void SequentialWriter::init_compression(const CompressionOptions & compression_options)
+{
+  if (compression_options.compression_mode != rosbag2_cpp::CompressionMode::NONE) {
+    if (compression_options.compression_format == "zstd") {
+      compressor_ = std::make_unique<rosbag2_compression::ZstdCompressor>();
+    } else {
+      std::stringstream err;
+      err << "Unsupported compression format " << compression_options.compression_format;
+      throw std::runtime_error{err.str()};
```
dabonnie:
```suggestion
      throw std::invalid_argument{err.str()};
```

To be consistent with the changes below.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -56,31 +58,53 @@ SequentialWriter::SequentialWriter(
...
+  compression_mode_{CompressionMode::NONE} {}
+
 
 SequentialWriter::~SequentialWriter()
 {
   reset();
 }
 
-void SequentialWriter::init_metadata()
+void SequentialWriter::init_compression(const CompressionOptions & compression_options)
```
dabonnie: I don't understand why this method isn't part of init_metadata: it's a bit confusing in ```SequentialWriter::open``` why you're calling two init functions taking the same input. I'd recommend the following:

```
SequentialWriter::open(...) 
{
    ....
    init_metadata(compression_options);
    ....
}

SequentialWriter::init_metadata(const CompressionOptions & compression_options);
{
    ....
    init_metadata(...);
    ....
}

SequentialWriter:: init_compression(const CompressionOptions & compression_options);
{
    ....
}

```

I'd also rename the method ```init_compression``` to ```init_compressor```.

If you don't want to combine methods, consider initializing your metadata first before creating the compressor_ resource.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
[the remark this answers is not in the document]
mm318: Okay, I realized throwing here is concerning, because the `~SequentialWriter()` destructor calls `reset()` which calls `compress_last_file()` which calls `compress_uri()`, which you're saying can potentially throw. This effectively means that the `~SequentialWriter()` destructor can throw.

Destructors that throw open up another whole can of worms https://isocpp.org/wiki/faq/exceptions#dtors-shouldnt-throw. There needs to be a try-catch somewhere, probably best in `~SequentialWriter()`.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
mm318: > However, `split_bagfile()` always adds a new file to the end so eventually, it should always be compressed on `reset()`.

Sorry, I missed the `metadata_.relative_file_paths.push_back()` in `split_bagfile()`, so I thought it was happening elsewhere.

However, I also noticed in `split_bagfile()` that you're potentially throwing between `compress_last_file()` and `metadata_.relative_file_paths.push_back()`, so you can still potentially end up with `metadata_.relative_file_paths.back()` being a file that's already compressed.

I'm not sure how such a thrown exception will be handled. How do you guarantee that the next time you enter `split_bagfile()`, `metadata_.relative_file_paths.back()` is not already compressed?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -89,25 +113,28 @@ void SequentialWriter::open(
...
 
-  init_metadata();
+  init_compression(compression_options);
+  init_metadata(compression_options);
 }
 
 void SequentialWriter::reset()
 {
   if (!base_folder_.empty()) {
+    if (compressor_ && compression_mode_ == CompressionMode::FILE) {
```
mm318: Nitpick: You didn't check for `compressor_ != nullptr` on [line 202](https://github.com/ros2/rosbag2/pull/250/files#diff-07603c1ee210fdbba6c3090284fc556dR202), why be inconsistent here? I think just don't have that check. If `compressor_ == nullptr` when the `compression_mode_ == FILE`, just let the assertion in `compress_last_file()` fire.

piraka9011: If the sequential writer fails to open because of bad options (ex. unable to open storage or bad splitting options) then `reset()` will be called without `compressor_` being initialized and the program will fail.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
piraka9011: > However, I also noticed in `split_bagfile()` that you're potentially throwing between `compress_last_file()` and `metadata_.relative_file_paths.push_back()`, so you can still potentially end up with `metadata_.relative_file_paths.back()` being a file that's already compressed.

That's correct, it seems a `bool` flag for that small section that could throw is required to make sure we don't double compress.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -20,12 +20,14 @@
...
 #include <vector>
 
+#include "rosbag2_cpp/compression_options.hpp"
 #include "rosbag2_cpp/converter.hpp"
 #include "rosbag2_cpp/serialization_format_converter_factory.hpp"
 #include "rosbag2_cpp/storage_options.hpp"
 #include "rosbag2_cpp/writer_interfaces/base_writer_interface.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
+#include "rosbag2_compression/base_compressor_interface.hpp"
```
Karsten1987: might want to consider a new line after this to group the includes on a per-package basis.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -102,13 +106,37 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
    */
   void write(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message) override;
 
+protected:
+  /**
+   * Initialize the compressor.
+   *
+   * \throws runtime_error If the compression implementation does not exist.
+   */
+  virtual void init_compression(const CompressionOptions & compression_options);
```
Karsten1987: what happens if the bag is already opened with a set of specific compression options and then this function is called with a different, potentially conflicting, set of it?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -15,15 +15,16 @@
...
 #include <chrono>
 #include <memory>
 #include <stdexcept>
 #include <string>
 #include <utility>
 
 #include "rcpputils/filesystem_helper.hpp"
-
 #include "rcutils/filesystem.h"
+#include "rosbag2_compression/zstd_compressor.hpp"
```
Karsten1987: new line above to group per package.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -56,16 +58,34 @@ SequentialWriter::SequentialWriter(
...
 SequentialWriter::~SequentialWriter()
 {
   reset();
 }
 
+void SequentialWriter::init_compression(const CompressionOptions & compression_options)
+{
+  if (compression_options.compression_mode != rosbag2_cpp::CompressionMode::NONE) {
+    if (compression_options.compression_format == "zstd") {
+      compressor_ = std::make_unique<rosbag2_compression::ZstdCompressor>();
```
Karsten1987: I am having a problem with this line as it's imposing a hard dependency on `Zstd` and all it's dependencies. 
`rosbag2_cpp` should potentially be compiled without any compression algorithm

piraka9011: We have [this issue](https://github.com/ros-security/aws-roadmap/issues/154) on our backlog next to make compression a plugin instead so that its not hardcoded.

Do you have any other suggestions in the meantime on how to avoid the hard dependency?
Maybe something along the lines of:

```cpp
auto zstd_compressor = std::make_unique<rosbag2_compression::ZstdCompressor>();
compressor_ = std::make_unique<rosbag2_compression::BaseCompressorInterface>(std::move(zstd_compressor));
```

If not, I suggest we work with the current implementation and then after introducing the plugin changes to `rosbag2_compression` use a `CompressorFactory` to make the writer more robust for different compression implementations.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
piraka9011: This function is only called in `open()` and is a protected method (for unit testing) so we don't expect to call it anywhere else.
We could simply move the functionality into `open()` only but that might make the function too large.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Karsten1987: Not sure I get your proposition right. We can either use a factory for loading it as a plugin dynamically or as discussed in a previous PR, inherit from the `SequentialWriter` or the `BaseReaderInterface` and make it live in the compression package. It can still be passed in by the caller side when instantiating the `final` Reader class. Example here: https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/player.cpp#L44-L46

The moment we include a compression header in the rosbag2_cpp package we introduce the dependency. That being said, the current PR also lacks the dependency of `rosbag2_compression` in its `package.xml`.

piraka9011: If I understand correctly, you prefer to isolate compression from `rosbag2_cpp`.
So instead there would be a `CompressionWriter` in `rosbag2_compression` that would implement compression by message or file which can be used in the other `rosbag2` packages.
However, that means losing the ability to split and compress simultaneously since we're overriding `write()` of the `SequentialWriter`.
Moreover, I don't see the issue of introducing compression as a dependency since we'd like to have that feature available for those using both the API and the CLI.

Karsten1987: Introducing the compression functionality is not the problem. So things like `compression_options` and general compression related structs might very well be living in this package. However, hard coding `ztd` is giving people no change to not compile zstd and its vendor package etc. This might not suit very well to every use case. 
Just the sqlite3 it is possible to compile rosbag2 without any notion of sqlite3.

piraka9011: > However, hard coding ztd is giving people no change to not compile zstd and its vendor package etc. 

I agree, which is why we want to work on making compression a plugin and work like `storage` and `storage_factory` does as opposed to the current implementation which is only `Zstd`.

However, since this is the first iteration, we want to make sure the whole compression pipeline works with a single compression format before extending it to other formats though a plugin.

Karsten1987: > P.S. Accidentally edited your comment, my bad.

That's one way to steer the direction of aa conversation ;-)

> However, since this is the first iteration, we want to make sure the whole compression pipeline works with a single compression format before extending it to other formats though a plugin.

That's fine. I think having a compression writer/reader implementation within `rosbag2_compression` is then the right way to go. It gives you the freedom to test your compression pipeline without introducing a dependency to `zstd` in `rosbag2_cpp`.

piraka9011: > I think having a compression writer/reader implementation within `rosbag2_compression` is then the right way to go.

What about splitting then? Basically implementing a `SequentialCompressionWriter`?
If that's the case we can consider going down that route instead for now.

zmichaels11:
The work done in this PR was addressed in:
* https://github.com/ros2/rosbag2/pull/263 (in review)
* https://github.com/ros2/rosbag2/pull/260 (merged)
* https://github.com/ros2/rosbag2/pull/258 (merged)

PR #257, opened 2020-01-15
Move compression artifacts from rosbag2_cpp to rosbag2_compression

piraka9011:
This removes the dependency of both `zstd` and `rosbag2_compression` from `rosbag2_cpp`.

This is part of a multi-PR effort to:
* Create a reader that can decompress files/messages in `rosbag2_compression`.
* Create a writer that can compress files/messages in `rosbag2_compression`.
* Select the appropriate reader/writer to use in `rosbag2_transport` based on the CLI.
* Add CLI options for compression in `ros2bag`.

Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

Issue #213, opened 2019-11-22
Resolve get_metadata in io-interfaces


zmichaels11:
Right now, `get_metadata` is defined in `BaseInfoInterface` and is only required in reading bagfiles that do not have a metadata file.

Issue #197 initially suggested removal of this function, however removal will cause breakage in `rosbag2_bag_v2`

I think this function should be removed since it exists only for compatibility and is required to be implemented by all plugins. Therefore some other mechanism should also be implemented to retain compatibility.

wjwwood:
@zmichaels11 I'll assign you for our triage process, feel free to reassign later.

zmichaels11:
> @zmichaels11 I'll assign you for our triage process, feel free to reassign later.

I'll create the required tickets to pull it into our sprint backlog.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/1

Karsten1987:
For bookmarking, I think also `get_all_topics_and_types` should be deprecated in the storage interface.

Taken from https://github.com/ros2/rosbag2/pull/371#issuecomment-613146317
> The BaseReaderInterface provides get_all_topics_and_types, but this calls into the storage, which implements this as a database read

MichaelOrlov:
- Closing as wouldn't be done. We need `get_metadata()` and use it from multiple places, and we are not going to remove it.

PR #258, opened 2020-01-15
Add a SequentialCompressionReader

piraka9011:
Implements a `SequentialCompressionReader` that uses `zstd` to both decompress files and read them sequentially.

Depends on #257.

### DIFF
--- a/rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
@@ -1,0 +1,153 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2_COMPRESSION__SEQUENTIAL_COMPRESSION_READER_HPP_
+#define ROSBAG2_COMPRESSION__SEQUENTIAL_COMPRESSION_READER_HPP_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "rosbag2_compression/base_decompressor_interface.hpp"
+#include "rosbag2_compression/compression_options.hpp"
+
+#include "rosbag2_cpp/converter.hpp"
+#include "rosbag2_cpp/reader_interfaces/base_reader_interface.hpp"
+#include "rosbag2_cpp/serialization_format_converter_factory.hpp"
+#include "rosbag2_cpp/serialization_format_converter_factory_interface.hpp"
+
+#include "rosbag2_storage/metadata_io.hpp"
+#include "rosbag2_storage/storage_factory.hpp"
+#include "rosbag2_storage/storage_factory_interface.hpp"
+#include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
+
+#include "visibility_control.hpp"
+#include "zstd_decompressor.hpp"
+
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2_compression
+{
+
+class ROSBAG2_COMPRESSION_PUBLIC SequentialCompressionReader
+  : public ::rosbag2_cpp::reader_interfaces::BaseReaderInterface
+{
+public:
+  explicit SequentialCompressionReader(
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<rosbag2_cpp::SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<rosbag2_cpp::SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
+
+  virtual ~SequentialCompressionReader();
+
+  void open(
+    const rosbag2_cpp::StorageOptions & storage_options,
+    const rosbag2_cpp::ConverterOptions & converter_options) override;
+
+  void reset() override;
+
+  bool has_next() override;
+
+  std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() override;
+
+  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() override;
+
+  /**
+   * Ask whether there is another database file to read from the list of relative
+   * file paths.
+   *
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file() const;
+
+  /**
+   * Return the relative file path pointed to by the current file iterator.
+   */
+  virtual std::string get_current_file() const;
+
+  /**
+   * Return the URI of the current file (i.e. no extensions).
+   */
+  virtual std::string get_current_uri() const;
+
+protected:
+  /**
+   * Increment the current file iterator to point to the next file in the list of relative file
+   * paths.
+   *
+   * Expected usage:
+   * if (has_next_file()) load_next_file();
+   */
+  virtual void load_next_file();
+
+  /**
+   * Initializes the decompressor if a compression mode is specified in the metadata.
+   *
+   * \throws std::invalid_argument If compression format doesn't exist.
+   */
+  virtual void setup_compression();
+
+  /**
+   * Set the file path currently pointed to by the iterator.
+   *
+   * \param file Relative path to the file.
+   */
+  virtual void set_current_file(const std::string & file);
+
+private:
+  /**
+   * Checks if all topics in the bagfile have the same RMW serialization format.
+   * Currently a bag file can only be played if all topics have the same serialization format.
+   *
+   * \param topics Vector of TopicInformation with metadata.
+   * \throws runtime_error if any topic has a different serialization format from the rest.
+   */
+  virtual void check_topics_serialization_formats(
+    const std::vector<rosbag2_storage::TopicInformation> & topics);
+
+  /**
+   * Checks if the serialization format of the converter factory is the same as that of the storage
+   * factory.
+   * If not, changes the serialization format of the converter factory to use the serialization
+   * format of the storage factory.
+   *
+   * \param converter_serialization_format
+   * \param storage_serialization_format
+   */
+  virtual void check_converter_serialization_format(
+    const std::string & converter_serialization_format,
+    const std::string & storage_serialization_format);
+
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_{};
+  std::shared_ptr<rosbag2_cpp::SerializationFormatConverterFactoryInterface> converter_factory_{};
+  std::shared_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> storage_{nullptr};
+  std::unique_ptr<rosbag2_cpp::Converter> converter_{nullptr};
+  std::unique_ptr<rosbag2_compression::BaseDecompressorInterface> decompressor_{nullptr};
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_{};
+  rosbag2_storage::BagMetadata metadata_{};
+  std::vector<std::string> file_paths_{};  // List of database files.
+  std::vector<std::string>::iterator current_file_iterator_{};  // Index of file to read from
+  rosbag2_compression::CompressionMode compression_mode_{
+    rosbag2_compression::CompressionMode::NONE};
+};
+
+}  // namespace rosbag2_compression
+
+#endif  // ROSBAG2_COMPRESSION__SEQUENTIAL_COMPRESSION_READER_HPP_
--- a/rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
@@ -102,7 +102,7 @@ protected:
    *
    * \throws std::invalid_argument If compression format doesn't exist.
    */
-  virtual void setup_compression();
+  virtual void setup_decompression();
 
   /**
    * Set the file path currently pointed to by the iterator.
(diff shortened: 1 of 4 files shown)


rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+#ifndef ROSBAG2_COMPRESSION__SEQUENTIAL_COMPRESSION_READER_HPP_
+#define ROSBAG2_COMPRESSION__SEQUENTIAL_COMPRESSION_READER_HPP_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "rosbag2_compression/compression_options.hpp"
+#include "rosbag2_compression/base_decompressor_interface.hpp"
```
zmichaels11: Alphasort

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+void SequentialCompressionReader::check_topics_serialization_formats(
+  const std::vector<rosbag2_storage::TopicInformation> & topics)
+{
+  auto storage_serialization_format = topics[0].topic_metadata.serialization_format;
+  for (const auto & topic : topics) {
+    if (topic.topic_metadata.serialization_format != storage_serialization_format) {
+      throw std::runtime_error(
+              "Topics with different rwm serialization format have been found. "
+              "All topics must have the same serialization format.");
```
zmichaels11: Nit: brace initialize

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_reader.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+class SequentialCompressionReaderTest : public Test
+{
+public:
+  SequentialCompressionReaderTest()
+  : storage_factory_(std::make_unique<StrictMock<MockStorageFactory>>()),
+    storage_(std::make_shared<NiceMock<MockStorage>>()),
+    converter_factory_(std::make_shared<StrictMock<MockConverterFactory>>()),
+    metadata_io_(std::make_unique<NiceMock<MockMetadataIo>>()),
+    storage_serialization_format_("rmw1_format")
```
zmichaels11: brace-initialize

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_reader.cpp
```diff
@@ -0,0 +1,103 @@
...
+  {
+    topic_with_type_ = rosbag2_storage::TopicMetadata{
+      "topic", "test_msgs/BasicTypes", storage_serialization_format_};
+    auto topics_and_types = std::vector<rosbag2_storage::TopicMetadata>{topic_with_type_};
+    auto message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
+    message->topic_name = topic_with_type_.name;
+
+    EXPECT_CALL(*storage_, get_all_topics_and_types())
+    .Times(AtMost(1)).WillRepeatedly(Return(topics_and_types));
+    EXPECT_CALL(*storage_, read_next()).WillRepeatedly(Return(message));
```
zmichaels11: I think you actually want `ON_CALL`.
iirc: `EXPECT_CALL` is more useful if you are checking if the function is being called

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() override;
+
+  /**
+   * Ask whether there is another database file to read from the list of relative
+   * file paths.
+   *
+   * \return true if there are still files to read in the list
+   */
+  virtual bool has_next_file() const;
```
emersonknapp: I know this is only the second time we're writing this pattern, but the sequential file URI logic is a decent amount of copy-pasted code. It may be worth a refactor to delegate functionality to a common sequential file handler.

piraka9011: I agree that it's a lot of similar code.
The problem with the sequential reader is that some of the functionality is private and not exposed in its public API. More over that's for another PR.

If we find ourselves doing this again though, then yes [a refactor is in order](https://refactoring.guru/refactoring/when).

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9072/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4820/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7427/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9005/)

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
```diff
@@ -0,0 +1,153 @@
...
+   * format of the storage factory.
+   *
+   * \param converter_serialization_format
+   * \param storage_serialization_format
+   */
+  virtual void check_converter_serialization_format(
+    const std::string & converter_serialization_format,
+    const std::string & storage_serialization_format);
+
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_{};
```
mm318: Nit: some `std::unique_ptr`s are default constructed (initialized with `{}`) and some are initialized to `nullptr`. Actually, both are unnecessary as C++ classes like smart pointers are well-behaved, unlike POD types like raw pointers.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+  reset();
+}
+
+void SequentialCompressionReader::reset()
+{
+  storage_.reset();
+}
+
+void SequentialCompressionReader::setup_compression()
```
mm318: Should this be `setup_decompression()`?

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
zmichaels11: Splitting `SequentialReader` may be a step in resolving Issue https://github.com/ros2/rosbag2/issues/213 since `SequentialCompressionReader` requires a metadata file. 

I think there will be refactoring in the future where `SequentialReader` may split again into a specialized form for ROS1 bagfile support. If that happens, I think it would be more beneficial to refactor then.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+std::string SequentialCompressionReader::get_current_file() const
+{
+  return *current_file_iterator_;
+}
+
+void SequentialCompressionReader::set_current_file(const std::string & file)
+{
+  *current_file_iterator_ = file;
```
mm318: I think it's much better to keep an index into `file_paths_` than to keep an iterator. An iterator is just as bad as a raw pointer, subject to invalidation. While it's easy to compare the iterator to `file_paths_.end()`, it's hard to check if an iterator is indeed somewhere between `file_paths_.begin()` and `file_paths_.end()`.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    storage_ = storage_factory_->open_read_only(
+      *current_file_iterator_, metadata_.storage_identifier);
+    if (!storage_) {
+      throw std::runtime_error{"No storage could be initialized. Abort"};
+    }
+  } else {
+    throw std::runtime_error{"Compression is not supported for legacy bag files."};
+  }
+  auto topics = metadata_.topics_with_message_count;
```
thomas-moulard:
```suggestion
  const auto topics = metadata_.topics_with_message_count;
```

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+  if (storage_) {
+    return storage_->get_all_topics_and_types();
+  }
+  throw std::runtime_error{"Bag is not open. Call open() before reading."};
+}
+
+bool SequentialCompressionReader::has_next_file() const
+{
+  return current_file_iterator_ + 1 != file_paths_.end();
```
thomas-moulard: This will crash is current_file_it_ is already end()

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+void SequentialCompressionReader::load_next_file()
+{
+  assert(current_file_iterator_ != file_paths_.end());
+  current_file_iterator_++;
```
thomas-moulard:
```suggestion
  ++current_file_iterator_;
```

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
mm318: Shouldn't this be a reference?
```suggestion
  auto & topics = metadata_.topics_with_message_count;
```

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+  if (decompressor_ && compression_mode_ == rosbag2_compression::CompressionMode::FILE) {
```
thomas-moulard: Can decompressor be null in normal operation? If it should never be null, we should not check for that and let the node crash instead of silently losing data.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+    auto message = storage_->read_next();
+    if (decompressor_ && compression_mode_ == rosbag2_compression::CompressionMode::MESSAGE) {
+      decompressor_->decompress_serialized_bag_message(message.get());
+    }
+    return converter_ ? converter_->convert(message) : message;
+  }
+  throw std::runtime_error{"Bag is not open. Call open() before reading."};
+}
+
+std::vector<rosbag2_storage::TopicMetadata> SequentialCompressionReader::get_all_topics_and_types()
```
mm318: Is this data generated on-the-fly? Why isn't it returned as a const reference?

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+std::string SequentialCompressionReader::get_current_uri() const
+{
+  auto current_file = get_current_file();
+  auto current_uri = rcpputils::fs::remove_extension(current_file);
+  return current_uri.string();
+}
+
+std::string SequentialCompressionReader::get_current_file() const
```
mm318: This can be returned as const reference.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+void SequentialCompressionReader::set_current_file(const std::string & file)
+{
+  *current_file_iterator_ = file;
+}
+
+void SequentialCompressionReader::check_topics_serialization_formats(
+  const std::vector<rosbag2_storage::TopicInformation> & topics)
+{
+  auto storage_serialization_format = topics[0].topic_metadata.serialization_format;
```
mm318: Not sure what data type this is, but if it's expensive to copy like a `std::string`, it should be assigned by reference (e.g.: `auto & storage_seriaization_format = ...`).

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
piraka9011: No, it will always throw in `setup_decompression` so yes this is a redundant check.

zmichaels11: I'm making a change to `setup_compression()` to eagerly throw in the one case where `decompressor_` is not set since it doesn't make sense for `SequentialCompressonReader` to be used on uncompressed bagfiles. I'm then replacing the `nullptr` checks on `decompressor_` with asserts

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
zmichaels11: Moving in favor of `{}` so all fields in `SequentialCompressionReader` are initialized similarly.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
zmichaels11: It might be, its forwarded from `BaseReaderInterface::get_all_topics_and_types` which is returned by value.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
zmichaels11: Its safer if it isn't since its queried from an iterator who's lifespan is determined by the currently loaded metadata.

zmichaels11:
@ros2/aws-oncall please rerun CI

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9073/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4822/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7428/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9011/)

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
mm318: My suggestion is to make them all consistent, with either no initializer or with empty brace initializer (`{}`).

zmichaels11:
All CI is green.
@Karsten1987 any last minute comments before I go ahead and merge?

rosbag2_compression/include/rosbag2_compression/sequential_compression_reader.hpp
Karsten1987: Why can't this class inherit from the `SequentialReader`?

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+#include "rosbag2_compression/sequential_compression_reader.hpp"
+
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "rcpputils/filesystem_helper.hpp"
```
Karsten1987: new lines between packages

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -0,0 +1,210 @@
...
+    }
+    return storage_->has_next();
+  }
+  throw std::runtime_error{"Bag is not open. Call open() before reading."};
+}
+
+std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialCompressionReader::read_next()
+{
+  if (storage_) {
+    assert(decompressor_);
```
Karsten1987: asserts won't be triggered when being compiled in release mode. Why not throwing here as well when `decompressor_` is null just like the storage?

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_reader.cpp
```diff
@@ -0,0 +1,102 @@
...
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "rosbag2_compression/sequential_compression_reader.hpp"
```
Karsten1987: new line

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
zmichaels11: `assert` was chosen since `decompressor_` cannot be null; `SequentialCompressionReader` should have already thrown an exception.

I can change it to throw an exception

piraka9011:
> Why can't this class inherit from the SequentialReader?

There's some private methods that need to be updated and/or refactored to support compression.
That's something we can address when we work on [aws-roadmap#154](https://github.com/ros-security/aws-roadmap/issues/154)

PR #261, opened 2020-01-18
switch to not deprecated API

dirk-thomas:
Related to ros2/ros2cli#436.

### DIFF
--- a/ros2bag/ros2bag/command/bag.py
+++ b/ros2bag/ros2bag/command/bag.py
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


zmichaels11:
This is causing an error in build & test: `Failed to load entry point 'bag': cannot import name 'add_subparsers_on_demand'`

dirk-thomas:
@zmichaels11 This needs to be tested against the branch of the referenced PR in order to pass.

dirk-thomas:
@zmichaels11 With the referenced PR being merged can you please retrigger the failing action.

dirk-thomas:
Technically this one will need a minimum version dependency on the next version of `ros2cli`.

PR #283, opened 2020-02-04
Wait for metadata to be written to disk

piraka9011:
Adds a method to the rosbag2 record E2E test fixture to wait until the metadata has been written to disk.
This resolves test flakiness on systems that are slow to flush all file system data to disk.

### DIFF
--- a/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
+++ b/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
@@ -89,7 +89,7 @@ public:
 
   void wait_for_metadata(int timeout_in_sec = 5)
   {
-    const auto metadata_path = rcpputils::fs::path{root_bag_path_} / "metadata.yaml";
+    const auto metadata_path = root_bag_path_ / "metadata.yaml";
     const auto start_time = std::chrono::steady_clock::now();
 
     while (std::chrono::steady_clock::now() - start_time < std::chrono::seconds(timeout_in_sec) &&
--- a/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
+++ b/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
@@ -101,7 +101,8 @@ public:
       std::this_thread::sleep_for(50ms);
     }
     // Final check for metadata if we timeout. Fail otherwise
-    ASSERT_EQ(metadata_path.exists(), true) << "Could not find metadata.yaml";
+    ASSERT_EQ(metadata_path.exists(), true) << "Could not find metadata file: \"" <<
+      metadata_path.string() << "\"";
   }
 
   void wait_for_db()
--- a/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
+++ b/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
@@ -87,6 +87,19 @@ public:
     return root_bag_path_ / (get_bag_file_name(split_index) + ".db3");
   }
 
+  void wait_for_metadata()
+  {
+    // Waits until metadata is written to disk since some platforms have slower IO sync times
+    // If the metadata does not exist, the test will timeout and fail
+    const auto metadata_path = rcpputils::fs::path{root_bag_path_} / "metadata.yaml";
+    while (true) {
+      if (metadata_path.exists()) {
+        return;
+      }
+      std::this_thread::sleep_for(50ms);
+    }
+  }
+
   void wait_for_db()
   {
     const auto database_path = get_bag_file_path(0);
--- a/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
+++ b/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
@@ -87,17 +87,18 @@ public:
     return root_bag_path_ / (get_bag_file_name(split_index) + ".db3");
   }
 
-  void wait_for_metadata()
+  void wait_for_metadata(int timeout_in_sec = 5)
   {
-    // Waits until metadata is written to disk since some platforms have slower IO sync times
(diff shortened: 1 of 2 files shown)


rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -69,6 +69,19 @@ class RecordFixture : public TemporaryDirectoryFixture
...
+  {
+    // Waits until metadata is written to disk since some platforms have slower IO sync times
+    // If the metadata does not exist, the test will timeout and fail
+    const auto metadata_path = rcpputils::fs::path{root_bag_path_} / "metadata.yaml";
+    while (true) {
+      if (metadata_path.exists()) {
+        return;
+      }
+      std::this_thread::sleep_for(50ms);
+    }
```
thomas-moulard: This will block forever if the file never gets written.

piraka9011: Is it better to have the test timeout with `colcon test` or specifying our own timeout?
This was left as an infinite loop since we expect the `colcon test` to fail on timeout.

dabonnie: Nit: specify in the docstring that you expect the timeout to happen as a result of colcon test.

zmichaels11: You could easily insert a parameter `max_checks` and rewrite `while (true)` with `for (int i = 0; i < max_checks; ++i)` 
Elsewhere have `constexpr const int kDefaultMaxChecks` assigned to 60s / 50ms.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -69,6 +69,20 @@ class RecordFixture : public TemporaryDirectoryFixture
     rclcpp::shutdown();
   }
 
+  void wait_for_metadata(int timeout = 5)
```
thomas-moulard:
```suggestion
  void wait_for_metadata(int timeout_in_sec = 5)
```

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -69,6 +69,20 @@ class RecordFixture : public TemporaryDirectoryFixture
...
+  {
+    const auto metadata_path = rcpputils::fs::path{root_bag_path_} / "metadata.yaml";
+    const auto start_time = std::chrono::steady_clock::now();
+    while (std::chrono::steady_clock::now() - start_time < std::chrono::seconds(timeout)) {
```
thomas-moulard: `... && rclcpp::ok()`

This way, this will actually unblock the thread when CTRL-C is triggered instead of forcing the process to be killed.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
zmichaels11: Maybe use the appropriate chrono type?

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -69,6 +69,22 @@ class RecordFixture : public TemporaryDirectoryFixture
...
+    while (std::chrono::steady_clock::now() - start_time < std::chrono::seconds(timeout_in_sec) &&
+      rclcpp::ok())
+    {
+      if (metadata_path.exists()) {
+        return;
+      }
+      std::this_thread::sleep_for(50ms);
+    }
+    // Final check for metadata if we timeout. Fail otherwise
+    ASSERT_EQ(metadata_path.exists(), true) << "Could not find metadata.yaml";
```
zmichaels11:
```suggestion
    ASSERT_EQ(metadata_path.exists(), true) << "Could not find metadata file: \"" <<
      metadata_path.string() << "\"";
```

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -69,6 +69,23 @@ class RecordFixture : public TemporaryDirectoryFixture
     rclcpp::shutdown();
   }
 
+  void wait_for_metadata(int timeout_in_sec = 5)
+  {
+    const auto metadata_path = rcpputils::fs::path{root_bag_path_} / "metadata.yaml";
```
zmichaels11: After rebase, this should be:
```suggestion
    const auto metadata_path = root_bag_path_ / "metadata.yaml";
```

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -87,6 +87,24 @@ class RecordFixture : public TemporaryDirectoryFixture
     return root_bag_path_ / (get_bag_file_name(split_index) + ".db3");
   }
 
+  void wait_for_metadata(int timeout_in_sec = 5)
```
thomas-moulard: It seems that `timeout_in_sec` is never used, so we don't need this parameter, we can just set `5` as a constant.
https://martinfowler.com/bliki/Yagni.html

piraka9011: Would it make sense to define it as a constant somewhere or just use `5` directly?

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/7edb4ff402ddd766a5f3fa38b6e5dd0e/raw/d8e79b898564d7307f95703d889dbf0156af8250/ros2.repos
BUILD args: --packages-up-to rosbag2_tests
TEST args: --packages-select rosbag2_tests
Job: ci_launcher

PR #282, opened 2020-02-04
Check for magic number on decompress_uri

zmichaels11:
### Changes
* Check for ZStandard magic number after reading in data inside `ZstdDecompressor::decompress_uri`

### Issues
* This  should fix ZStandard compression on armhf Linux builds

Signed-off-by: Zachary Michaels <zmichaels11@gmail.com>

PR #285, opened 2020-02-04
Change validation functions to accept output type of ZSTD_getFrameContentSize

zmichaels11:
Addresses the same issue to #282.
The root cause of the problem was an implicit conversion from `unsigned long long` to `size_t` (same on x86_64 platforms)

### Changes
* Accept `unsigned long long` directly. This is the return type of ZSTD_getFrameContentSize

Signed-off-by: Zachary Michaels <zmichaels11@gmail.com>

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
@@ -34,6 +34,7 @@ constexpr const int kDefaultZstdCompressionLevel = 1;
 
 // String constant used to identify ZstdCompressor.
 constexpr const char kCompressionIdentifier[] = "zstd";
+// Used as a parameter type in a function that accepts the output of ZSTD_compress.
 using ZstdCompressReturnType = decltype(ZSTD_compress(nullptr, 0, nullptr, 0, 0));
 
 /**
--- a/rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
@@ -137,7 +137,7 @@ void write_output_buffer(
 /**
  * Checks compression_result and throws a runtime_error if there was a ZSTD error.
  */
-void throw_on_zstd_error(const size_t compression_result)
+void throw_on_zstd_error(const unsigned long long compression_result)  // NOLINT(runtime/int)
 {
   if (ZSTD_isError(compression_result)) {
     std::stringstream error;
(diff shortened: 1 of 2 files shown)


zmichaels11:
please run this CI job
Gist: https://gist.githubusercontent.com/zmichaels11/96d24f9df801fdf5cba371185dfc0af1/raw/a7f0a443519a12a55fca1a4ed05fa7168e085a43/ros2.repos
BUILD args: --packages-up-to rosbag2_compression
TEST args: --packages-select rosbag2_compression
Job: ci_launche

zmichaels11:
* Linux armhf [Build Status](https://ci.ros2.org/job/ci_linux-armhf/76/)

zmichaels11:
> Could we avoid the NOLINT here?

The lint warning is from using `unsigned long long`. I opted to use NOLINT in favor of using a sized-type because its the return value of `ZSTD_getFrameContentSize` which is directly passed into the functions.

zmichaels11:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9244/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4930/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7551/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9163/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container
* Linux-armhf [Build Status](https://ci.ros2.org/job/ci_linux-armhf/77/)

zmichaels11:
> Nice call with the decltype!

I think you missed a hallway discussion with Thomas; paraphrasing:
`decltype` should be avoided since its deep magic. This current commit is a compromise from a previous (unpublished) iteration and uses type aliasing to comment via code that decltype yields the return type of the function.

rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
```diff
@@ -34,6 +34,8 @@ constexpr const int kDefaultZstdCompressionLevel = 1;
 
 // String constant used to identify ZstdCompressor.
 constexpr const char kCompressionIdentifier[] = "zstd";
+// Used as a parameter type in a function that accepts the output of ZSTD_compress.
+using ZstdCompressReturnType = decltype(ZSTD_compress(nullptr, 0, nullptr, 0, 0));
```
mm318: Maybe use https://en.cppreference.com/w/cpp/types/result_of instead of making a function call.

zmichaels11: decltype actually doesn't make a function call! (I'll update with the document stating this)

mm318: Never mind. `decltype` doesn't invoke anything, but `result_of` is better in this situation because you won't need to write the dummy values to "pass" into the `ZSTD_compress()` function.

zmichaels11: Yeah, I agree, unfortunately I merged this before you commented.

Issue #299, opened 2020-02-18
Solution proposed to implementing sim clock feature in ros2 bag play


scoopySHI:
### ros2 bag play with sim time (current time) as time stamper in header 
As mentioned in issue #99, there is no proper way to play the rosbag file with sim colock (current time) as the time stamper. Therefore the rosbag file cannot be properly handled with some packages like robot_localization.
@kikass13 and i managed to find a solution to implementing this feature. We figured out the exact position of "Header" in Fastrtps serialized data and `memcpy` the current sim time into it as the new time stamper.
This feature is triggered with` -c `as argument in ros2 bag play. Otherwise ros2bag play performs as default. <br/>
Example:

> ros2 bag play -c yourbagfiles.db3

Fork + branch:
[https://github.com/scoopySHI/rosbag2/tree/time_fix](https://github.com/scoopySHI/rosbag2/tree/time_fix) 
<br/>
**Test instances:**
Originally it was tested with sensor_msgs/Imu and sensor_msgs/NavSatFix. In order to test the feature more comprehensively, multiple custom ros messages are also used to generate the testing rosbag files. The header is not required to be always at the first place inside the message. These custom messages are meaningless but cover thoroughly the types of messages including string, wide string, array etc. There can also be custom messages in front of the header.
Rosbags and custom messages tested:
[https://github.com/scoopySHI/msg_ws](https://github.com/scoopySHI/msg_ws)
<br/>

We would like to hear any feedback plus test results of this feature. @Karsten1987  @dirk-thomas  @ivanpauno

kikass13:
:+1:

emersonknapp:
Closing this issue as there is already #99 tracking the feature, and this approach was decided against

Issue #99, opened 2019-04-11
play - publish /clock topic


AndreasAZiegler:
# Description

Feature request: Provide the `--clock` option so that rosbag2 publishes to the `/clock` topic alongside its messages.

# Implementation Suggestion

Using `starting_time` of the bag and the stored (received) timestamp with every message, we can start a publisher for the /clock topic on playback and run it at a fixed rate, publishing messages when the next clock message is past their publishing time. 

This requires no introspection or modification of the message contents themselves.

Note that `player.cpp` is already using message timestamps to sleep in between publishing, to approximate the original rates. In this loop (`play_messages_until_queue_empty`) we could potentially insert the clock publisher there.

Another option would be to put the clock publisher on a timer, and have it publish the last-published message time.

# Original Ticket

From ros 1 I'm used to the ```--clock``` option in order that the topics have time stamps as If the system would run right now. I couldn't find this option in the documentation of rosbag2. How can I activate simulated time with rosbag2?

Karsten1987:
Sorry for being so late on replying here. The concept of simulating time for rosbag2 is not yet realized. Therefore, there is currently no `--clock` option. 

I am going ahead and close this issue. If there is a follow up to this topic, please feel free to re-open it and continue the conversation.

IanTheEngineer:
Follow-on - @Karsten1987 & @jacobperron -  playing back a `/clock` from a bagfile is very much a core feature in the ROS1 `rosbag`. Otherwise, no node requiring time can utilize the collected bagfile data during playback. These nodes will use the system clock instead, which will conflict with the recorded data's header timestamps. Specifially, `tf` and nodes that subscribe to it are useless without `/clock`.

From a practical standpoint, I am running a high frequency system-clock-publishing-node publishing `/clock` onboard our robots simply to be recorded for bagfile purposes. This is pretty much the only workaround to debug data collected amcl, cartographer, perception nodes, etc.

scoopySHI:
Can't agree more @IanTheEngineer. With --clock publishing msg in current time should be a core feature in ros2bag since the nodes are using current system time. For instance, the robot_localization pkg cannot output correct prediction with "ros2 bag play" bagfile data. 
[...]

PR #300, opened 2020-02-18
feature: replace rosbag time stamper with sim time(current time)

scoopySHI:
The feature is mentioned in issue #299.


rosbag2_transport/CMakeLists.txt
```diff
@@ -26,8 +26,10 @@ find_package(rcl REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rosbag2 REQUIRED)
+find_package(sensor_msgs REQUIRED)
```
kikass13: remove this

rosbag2_transport/CMakeLists.txt
```diff
@@ -49,6 +51,8 @@ ament_target_dependencies(${PROJECT_NAME}
   rosidl_generator_cpp
   rmw
   shared_queues_vendor
+  sensor_msgs
```
kikass13: remove this

rosbag2_transport/CMakeLists.txt
kikass13: remove this

rosbag2_transport/CMakeLists.txt
kikass13: remove this

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
   } while (!is_storage_completely_loaded() && rclcpp::ok());
 }
 
+//Dynamic Alignment as Fastrtps
+unsigned long alignment(unsigned long data_size, unsigned long last_data_size, unsigned long current_position)
+{
+  return data_size > last_data_size ? (data_size - current_position % data_size) & (data_size-1):0;
```
kikass13: can we do this vendor specific thing here?

how can we use the dds vendor implementation instead of using copied code in here

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+}
+
+//deal with string
+void Player::deal_with_string(const uint8_t *dds_buffer)
```
kikass13: can we refactor those together? *1

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+
+  unsigned long one_offset = alignment(string_header, last_data_size, current_position);
+  current_position = current_position + one_offset;
+  memcpy(&length, (dds_buffer + current_position + 4), string_header);
+  last_data_size = sizeof(char);
+  current_position += (string_header + length);
+}
+
+//deal with wstring
+void Player::deal_with_wstring(const uint8_t *dds_buffer)
```
kikass13: can we refactor those together? *1

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+
+  unsigned long one_offset = alignment(string_header, last_data_size, current_position);
+  current_position = current_position + one_offset;
+  memcpy(&length, (dds_buffer + current_position + 4), string_header);
+  last_data_size = sizeof(uint32_t);
+  current_position += (string_header + length * sizeof(uint32_t));
+}
+
+//find out the real position of header in fastrtps serialized data
+void Player::calculate_position_with_align(const uint8_t * dds_buffer_ptr, const rosidl_typesupport_introspection_cpp::MessageMember *message_member, unsigned long stop_index)
```
kikass13: write comment with source from here, where did you get this from

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+        is_wstring = true;
+        break;
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
+        sub_members = static_cast<const rosidl_typesupport_introspection_cpp::MessageMembers *>(message_member[i].members_->data);
+        calculate_position_with_align(dds_buffer_ptr, sub_members->members_, sub_members->member_count_);
+        is_ros_msg_type = true;
+        break;
+      }
+    //standard element
+    if(!is_string && !is_wstring && !is_ros_msg_type && !message_member[i].is_array_)
```
kikass13: these checks seem kind of redundant, try cascading the clauses
```
 if(!is_ros_msg_type && !message_member[i].is_array_)
      if(!is_string && !is_wstring ...)
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+
+}
+
 void Player::play_messages_until_queue_empty()
 {
   ReplayableMessage message;
   while (message_queue_.try_dequeue(message) && rclcpp::ok()) {
     std::this_thread::sleep_until(start_time_ + message.time_since_start);
+
+    if (topics_ts_map_.find(message.message->topic_name) != topics_ts_map_.end() ) {
```
kikass13: use 
```
auto it = topics_ts_map_.find(message.message->topic_name)
 if (it != topics_ts_map_.end() ) {
...
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+      builtin_interfaces::msg::Time ros_time_to_set;
+
+      std::chrono::time_point<std::chrono::high_resolution_clock> time_to_be_set = start_time_ + message.time_since_start;
+
+      auto offset_index = topics_ts_map_[message.message->topic_name].stop_index;
```
kikass13: use
``` 
it->stop_index;
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+      const rosidl_typesupport_introspection_cpp::MessageMember * msg_ptr = topics_ts_map_[message.message->topic_name].msg_member_ptr;
```
kikass13: use iterator for that aswell

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+      ros_time_to_set.nanosec = static_cast<uint32_t>(round(time_to_be_set.time_since_epoch().count() - ros_time_to_set.sec*1e9));
+
+      //memcpy
+      uint8_t * buffer_temp = message.message->serialized_data->buffer;
+      dds_buffer_ptr = message.message->serialized_data->buffer;
+      calculate_position_with_align(dds_buffer_ptr, msg_ptr, offset_index);
+      size_t header_time_sec_size = sizeof (int32_t);
+      unsigned long last_offset = alignment(header_time_sec_size, last_data_size, current_position);
+      current_position += last_offset;
+      buffer_temp = buffer_temp + current_position + 4; //plus dds header
```
kikass13: can this +4 dds header be looked up in any way?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,222 @@ void Player::play_messages_from_queue()
...
+
+      memcpy(buffer_temp, &ros_time_to_set, sizeof(builtin_interfaces::msg::Time));
+
+      current_position = 0;
+      last_data_size = ULONG_MAX;
+    }
+
+//    if(options.clock_type == "current")
```
kikass13: remove this?

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -56,7 +71,12 @@ class Player
   void wait_for_filled_queue(const PlayOptions & options) const;
   void play_messages_from_queue();
   void play_messages_until_queue_empty();
+  void prepare_clock(ReplayableMessage & message);
```
kikass13: remove this?

.gitignore
```diff
@@ -3,5 +3,7 @@
 cmake-build-debug/
 cmake-build-release/
 build/
+install/
```
emersonknapp: this seems outside the scope of this feature

ros2bag/CHANGELOG.rst
```diff
@@ -2,8 +2,8 @@
 Changelog for package ros2bag
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Forthcoming
```
emersonknapp: It seems like CHANGELOG is outside the scope of this feature, shouldn't we be generating when we create the next release? Same for the other CHANGELOGs

ros2bag/package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>ros2bag</name>
-  <version>0.2.3</version>
+  <version>0.2.4</version>
```
emersonknapp: Shouldn't version bump be reserved for the release process? Same for the other package.xmls

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -26,6 +26,7 @@ struct PlayOptions
 public:
   size_t read_ahead_queue_size;
   std::string node_prefix = "";
+  std::string clock_type = "past";
```
emersonknapp: What values are accepted here? I am thinking this should probably be an enum.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -30,6 +30,7 @@
 #include "rosbag2_transport/logging.hpp"
 #include "rosbag2_node.hpp"
 #include "replayable_message.hpp"
+#include "sensor_msgs/msg/imu.hpp"
```
emersonknapp: remove?

kikass13: :+1

Karsten1987:
this PR has to be rebased on top of master. This should also make the diff way easier to read.

ros2bag/ros2bag/verb/play.py
```diff
@@ -27,6 +27,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '-s', '--storage', default='sqlite3',
             help='storage identifier to be used, defaults to "sqlite3"')
+        parser.add_argument(
+            '-c', '--clock', default='past',
```
Karsten1987: what does `past` mean here? Is there a way to describe this more accurately? Or at least the `help` text should reflect this. Default is set to `past` and the help text mentions `current time`.

rosbag2_transport/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ find_package(rcutils REQUIRED)
 find_package(rosbag2 REQUIRED)
 find_package(rmw REQUIRED)
 find_package(shared_queues_vendor REQUIRED)
+find_package(rosidl_typesupport_introspection_cpp)
```
Karsten1987: alphabetical sort. Same below.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
Karsten1987: Could this be part of the actual clock types being available in ros2? http://docs.ros2.org/eloquent/api/rcl/time_8h.html#a5c734f508ce06aec7974af10ad09d071

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -56,6 +57,11 @@ void Player::play(const PlayOptions & options)
 {
   prepare_publishers();
 
+  if(options.clock_type == "current")
```
Karsten1987: as @emersonknapp mentioned before I don't think that a string compare is the right thing here. At least an `enum` or if not the actual `rcl_clock_type` makes more sense to me here.

rosbag2_transport/src/rosbag2_transport/player.cpp
Karsten1987: How is this alignment related to time? Just trying to keep the set of changes semantically separated.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -126,17 +132,208 @@ void Player::play_messages_from_queue()
...
+      current_position += (string_header + length);
+    }
+  else {
+      last_data_size = sizeof(uint32_t);
+      current_position += (string_header + length * sizeof(uint32_t));
+  }
+}
+
+//find out the real position of header in fastrtps serialized data
+void Player::calculate_position_with_align(const uint8_t * dds_buffer_ptr, const rosidl_typesupport_introspection_cpp::MessageMember *message_member, unsigned long stop_index)
```
Karsten1987: I think this change should actually land in `rosidl_typesupport_introspection_cpp` rather than in rosbag2.
We've worked on a similar change for the serialized bag message, which resulted in here: https://github.com/ros2/rosidl/pull/416

So I am very hesitant to bring back this logic

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -20,12 +20,22 @@
 #include <memory>
 #include <queue>
 #include <string>
+#include <cmath>
+#include <cstring>
+#include <climits>
```
Karsten1987: alphasort

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -67,8 +85,16 @@ class Player
   mutable std::future<void> storage_loading_future_;
   std::shared_ptr<Rosbag2Node> rosbag2_transport_;
   std::unordered_map<std::string, std::shared_ptr<GenericPublisher>> publishers_;
+  //std::unordered_map<std::string, unsigned> topics_ts_map_;
```
Karsten1987: can this be deleted then?

ros2bag/package.xml
kikass13: yeah git suggested the signoff and we did it without caution. this will probably get reverted tomorrow (it's 10pm here ^-^)

rosbag2_transport/src/rosbag2_transport/player.cpp
kikass13: in our version of this feature, we are changing the timestamp inside the serialized message directly, that way we achieve a "simulated" time while publishing messages via the player class. 
This is related to issue #299

.gitignore
kikass13: correct, should be removed

rosbag2_transport/include/rosbag2_transport/play_options.hpp
kikass13: this was @scoopySHI 's way of getting the simulated clock argument into the core player class. I do agree that this should be done in a better way.

rosbag2_transport/src/rosbag2_transport/player.cpp
kikass13: I can confirm that this is probably not the right place for a lot of this logic. the point is that we don't really know if this PR is even necessary. There is currently no way (as far as we know) of achieving what #299 needs without compromising the relatively clean rosbag functionality. So this is just the first shot we have for this problem so any suggestion is helpful :)

rosbag2_transport/include/rosbag2_transport/play_options.hpp
scoopySHI: Maybe the name clock_type can lead to misunderstand. I'd like to replace it with "time in Header": it suggests which time is in the time stamper in Header (current system time or the time while recording). It has actually nothing to do with clock types in ROS2.

ros2bag/ros2bag/verb/play.py
scoopySHI: "past" means the time in time stamper in Header is the time when recording happened. "current time" means the time in time stamper in Header is current system time. Some packages like robot_localization deals with current system time, therefore we want to replace the time stamper as current time so that the rosbag can be directly played and as data input in those packages. @Karsten1987

rosbag2_transport/src/rosbag2_transport/player.cpp
scoopySHI: Because the offset in typesupport does not provides actual offset in fastrtps serialized data. (It provides the offset in machine memory, using the alignment of machine) In order to figure out the actual offset of "Header", we must calculate the offset according to Fastrtps alignment rule.

rosbag2_transport/src/rosbag2_transport/player.cpp
kikass13: @Karsten1987  the ros2/rosidl#416 you mentioned does not solve the problem. We do not know the type at hand (we cannot static cast something unknown into something unknown) ... 
``` std_msgs__msg__String msg = *static_cast<std_msgs__msg__String *>(message_memory); ```
in your solution does not help when dealing with fully unknown type deserialization. This could only work if we ```switch(type_string) case 1, case 2, case 3``` all possible message types - which is not possible due to the existence of custom messages (which are not known by default).

This works well for subscribers (where I give the expected Template type of the deserialization) ... but it should not work in places where i deserialize something random and look for fields.

Am I missing something here? Thanks :)

File: rosbag2_transport/CMakeLists.txt [hunk 25]
```diff
@@ -25,6 +25,7 @@ find_package(ament_cmake_ros REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
+find_package(rosidl_typesupport_introspection_cpp)
```
zmichaels11: Is `REQUIRED` not required here?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -132,17 +137,203 @@ void Player::play_messages_from_queue()
...
+        break;
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_CHAR:
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_INT8:
+        data_size = sizeof (char);
+        break;
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_FLOAT:
+        data_size = sizeof (float);
+        break;
+      case ::rosidl_typesupport_introspection_cpp::ROS_TYPE_FLOAT64:
+        data_size = sizeof (double);
```
zmichaels11: `sizeof(double)` may be sizes different than 64bits. Is this intentional?
I would have expected it to be 8 considering the switch case is `FLOAT64`

rosbag2_transport/CMakeLists.txt
kikass13: :+1:

rosbag2_transport/src/rosbag2_transport/player.cpp
kikass13: https://github.com/ros2/rmw_fastrtps/blob/9438cca2a6a21b2436684607fea8a78624363f80/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L525

:+1: 
Is an example for that:

we should replicate what was done here @scoopySHI

kikass13: https://github.com/ros2/rmw_fastrtps/blob/9438cca2a6a21b2436684607fea8a78624363f80/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L525

:+1: We should replicate what was done here

kikass13:
Thank you guys for the first feedback, mostly regarding style and broader implementation details.
@scoopySHI  and myself are not really sure about our approach regarding the solution to #299  and #99. We are not experts in the DDS specifics and if you guys have some nice ideas we are open for them :)

To summarice what @scoopySHI said in #299, we are searching for a fast way to get a "simulated" time in ros bag 2 similia to the feature ros1 had, so we have implemented a method of changing the timestamp of the current ros message to the current time without deserializing it. The solution follows these steps:
* initialize rosbag player
  * remember start time of first bag message and start time of current system clock
  * type introspect each messae in the bagfile and lookup type-supports via dynamic symbol lookup `get_symbol(type_str)` 
  * use each type-support to find header a header field in the message 
  * this information will be kept inside a lookup hashtable `unordered_map`
* play rosbag messages by dequeuing serialized messages
  * before publish, check each message for  it's type
  * to do that, we need to do a lookup into our hashtable (header fields message types are already in there)
  * calculate the header field offset in memory using the same dynamic alignment functions that the de-serialization uses.
  * create our stamp message and set it to the `stamp = current time + (current_message_stamp_in_bagfile - first_message_time_in_bagfile - )`, that way we can replicate the time difference of the past and propagate it into the future
  * memcopy our stamp into the offset calculated 
  * publish the message as normal

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -132,17 +137,203 @@ void Player::play_messages_from_queue()
...
+      //memcpy
+      uint8_t * buffer_temp = message.message->serialized_data->buffer;
+      dds_buffer_ptr_ = message.message->serialized_data->buffer;
+      calculate_position_with_align(dds_buffer_ptr_, msg_ptr, offset_index);
+      size_t header_time_sec_size = sizeof (int32_t);
+      unsigned long last_offset = alignment(header_time_sec_size, last_data_size_, current_position_);
+      current_position_ += last_offset;
+      buffer_temp = buffer_temp + current_position_ + 4;
+
+      memcpy(buffer_temp, &ros_time_to_set, sizeof(builtin_interfaces::msg::Time));
```
zmichaels11: Nit: Use `memcpy_s` I believe Windows will complain about not using it.

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -58,8 +73,14 @@ class Player
   void play_messages_from_queue();
   void play_messages_until_queue_empty();
   void prepare_publishers();
+  void prepare_topic_ts_map();
+  void calculate_position_with_align(const uint8_t * dds_buffer, const rosidl_typesupport_introspection_cpp::MessageMember *message_member, unsigned long stop_index);
+  void deal_with_string(const uint8_t * dds_buffer, bool is_wstring);
 
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
+  size_t last_data_size_ = ULONG_MAX;
+  unsigned long current_position_ = 0;
```
zmichaels11: Nit: I believe `current_position_` should be `std::ptrdiff_t` since it is used in pointer math.

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -58,8 +73,14 @@ class Player
   void play_messages_from_queue();
   void play_messages_until_queue_empty();
   void prepare_publishers();
+  void prepare_topic_ts_map();
+  void calculate_position_with_align(const uint8_t * dds_buffer, const rosidl_typesupport_introspection_cpp::MessageMember *message_member, unsigned long stop_index);
+  void deal_with_string(const uint8_t * dds_buffer, bool is_wstring);
 
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
+  size_t last_data_size_ = ULONG_MAX;
```
zmichaels11: A more `C++` way of doing this would be to use `numeric_limits::max()`

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -132,17 +137,203 @@ void Player::play_messages_from_queue()
...
+    if(!is_string && !is_wstring && !is_ros_msg_type && !message_member[i].is_array_)
+      {
+        one_offset = alignment(data_size, last_data_size_, current_position_);
+        current_position_ += (one_offset + data_size);
+        last_data_size_ = data_size;
+      }
+    //standard array
+    else if(!is_string && !is_wstring && !is_ros_msg_type && message_member[i].is_array_)
+      {
+        for (uint j = 0;j < message_member[i].array_size_; j++) {
```
zmichaels11: The iterating index should be the same type as `array_size_`

emersonknapp:
As a high level question, I want to summarize what we're seeing here -

* When playing back messages, if they have a Header, re-stamp the Header timestamp with the current ROS time

If I'm not mistaken, rosbag1 never had this feature, the feature it did expose is what's asked for in  https://github.com/ros2/rosbag2/issues/99. This involves publishing an interpolated `/clock` topic so that other nodes in the system can `use_sim_time` to run along.

Do we actually want to implement this re-stamping feature, or would we be better off aiming at a solution for #99 instead?

kikass13:
to comment on what @emersonknapp said: Yes you absolutely correct with your analysis. The relevance of this feature is ... questionable... for us at best. 

As far as I can tell, there is no implementation for an underlying /clock topic in Nodes,  but I could be wrong with that. As I said in my comment before, @scoopySHI and myself are not really sure about this approach either.

We implemented it currently because we are doing integration tests with the current robot_localisation (RL) in eloquent and issue #99 literally destroyed our ability to debug the RL, which is, in it's current state, so broken that we cannot use it properly. 

So your question about re-implementing the clock topic feature for simulating time (in the Base rclcpp:Node as well as here  in rosbag) is valid and needs to be addressed by some guys other than us because, we (myself and @scoopySHI) are not really an authority when it comes to ros2 guidelines of how to do stuff. 

This solution though has one benefit in relation to the old /clock solution, It is stanralone and does therfore not involve or compromise nodes initialization behavior , the system clock or any other implementation. Nonetheless could our solution be seen as "dirty as fuck", which is not something I would personally disagree with :D

emersonknapp:
The `use_sim_time` parameter works in ros2 on Nodes, which makes them use the `/clock` topic. The only change to make that work would be here in rosbag2 to publish the clock.

scoopySHI:
> The `use_sim_time` parameter works in ros2 on Nodes, which makes them use the `/clock` topic. The only change to make that work would be here in rosbag2 to publish the clock.

Thanks for the hint, i will test it with our stack and then try to implement the `/clock` publisher inside ros2bag.  For now i will close this PR since i think right now it is not the best way to solve the problem referring to issue #99 . @emersonknapp

PR #302, opened 2020-02-27
Add filter for reading selective topics

mabelzhang:
This allows the user to specify a list of topic names to read.

Signed-off-by: Mabel Zhang <mabel@openrobotics.org>

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
@@ -28,6 +28,7 @@
 #include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
 #include "rosbag2_storage/storage_factory_interface.hpp"
+#include "rosbag2_storage/storage_filter.hpp"
 #include "rosbag2_storage/storage_interfaces/read_only_interface.hpp"
 
 // This is necessary because of using stl types here. It is completely safe, because
@@ -67,6 +68,8 @@ public:
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() override;
 
   std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() override;
+
+  void set_filter(const rosbag2_storage::StorageFilter & storage_filter);
 
   /**
    * Ask whether there is another database file to read from the list of relative
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -23,6 +23,7 @@
 #include "rcutils/types.h"
 #include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_storage/storage_filter.hpp"
 #include "rosbag2_storage/topic_metadata.hpp"
 #include "rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.hpp"
 #include "rosbag2_storage_default_plugins/visibility_control.hpp"
@@ -72,6 +73,8 @@ public:
 
   uint64_t get_minimum_split_file_size() const override;
 
+  void set_filter(const rosbag2_storage::StorageFilter & storage_filter) override;
+
 private:
   void initialize();
   void prepare_for_writing();
@@ -90,6 +93,7 @@ private:
   std::unordered_map<std::string, int> topics_;
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
+  rosbag2_storage::StorageFilter storage_filter_ {};
 };
 
 }  // namespace rosbag2_storage_plugins
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -103,6 +103,8 @@ void SqliteStorage::open(
 
   ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO_STREAM(
     "Opened database '" << relative_path_ << "' for " << to_string(io_flag) << ".");
+
+  storage_filter_.topics.clear();
 }
 
 void SqliteStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message)
@@ -211,10 +213,27 @@ void SqliteStorage::prepare_for_writing()
 
 void SqliteStorage::prepare_for_reading()
 {
-  read_statement_ = database_->prepare_statement(
-    "SELECT data, timestamp, topics.name "
-    "FROM messages JOIN topics ON messages.topic_id = topics.id "
-    "ORDER BY messages.timestamp;");
+  if (!storage_filter_.topics.empty()) {
+    // Construct string for selected topics
+    std::string topic_list{""};
+    for (auto & topic : storage_filter_.topics) {
+      topic_list += "'" + topic + "'";
+      if (&topic != &storage_filter_.topics.back()) {
+        topic_list += ",";
+      }
+    }
+
+    read_statement_ = database_->prepare_statement(
+      "SELECT data, timestamp, topics.name "
+      "FROM messages JOIN topics ON messages.topic_id = topics.id "
+      "WHERE topics.name IN (" + topic_list + ")"
+      "ORDER BY messages.timestamp;");
+  } else {
+    read_statement_ = database_->prepare_statement(
+      "SELECT data, timestamp, topics.name "
+      "FROM messages JOIN topics ON messages.topic_id = topics.id "
+      "ORDER BY messages.timestamp;");
+  }
   message_result_ = read_statement_->execute_query<
     std::shared_ptr<rcutils_uint8_array_t>, rcutils_time_point_value_t, std::string>();
   current_message_row_ = message_result_.begin();
@@ -290,6 +309,12 @@ rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
   metadata.bag_size = get_bagfile_size();
 
   return metadata;
+}
+
+void SqliteStorage::set_filter(
+  const rosbag2_storage::StorageFilter & storage_filter)
+{
+  storage_filter_ = storage_filter;
 }
 
 }  // namespace rosbag2_storage_plugins
--- a/rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
@@ -14,6 +14,8 @@
 
 #include <gmock/gmock.h>
 
+#include <rosbag2_storage/storage_filter.hpp>
+
 #include <memory>
 #include <string>
 #include <tuple>
@@ -109,6 +111,33 @@ TEST_F(StorageTestFixture, get_next_returns_messages_in_timestamp_order) {
   auto second_message = readable_storage->read_next();
   EXPECT_THAT(second_message->time_stamp, Eq(6));
   EXPECT_FALSE(readable_storage->has_next());
+}
+
+TEST_F(StorageTestFixture, read_next_returns_filtered_messages) {
+  std::vector<std::tuple<std::string, int64_t, std::string, std::string, std::string>>
+  string_messages =
+  {std::make_tuple("topic1 message", 1, "topic1", "", ""),
+    std::make_tuple("topic2 message", 2, "topic2", "", ""),
+    std::make_tuple("topic3 message", 3, "topic3", "", "")};
+
+  write_messages_to_sqlite(string_messages);
+  std::unique_ptr<rosbag2_storage::storage_interfaces::ReadOnlyInterface> readable_storage =
+    std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
+
+  auto db_filename = (rcpputils::fs::path(temporary_dir_path_) / "rosbag.db3").string();
+  readable_storage->open(db_filename);
+
+  rosbag2_storage::StorageFilter storage_filter;
+  storage_filter.topics.push_back("topic2");
+  storage_filter.topics.push_back("topic3");
+  readable_storage->set_filter(storage_filter);
+
+  EXPECT_TRUE(readable_storage->has_next());
+  auto first_message = readable_storage->read_next();
+  EXPECT_THAT(first_message->topic_name, Eq("topic2"));
+  EXPECT_TRUE(readable_storage->has_next());
+  auto second_message = readable_storage->read_next();
+  EXPECT_THAT(second_message->topic_name, Eq("topic3"));
 }
 
 TEST_F(StorageTestFixture, get_all_topics_and_types_returns_the_correct_vector) {
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -130,6 +130,12 @@ std::vector<rosbag2_storage::TopicMetadata> SequentialReader::get_all_topics_and
   throw std::runtime_error("Bag is not open. Call open() before reading.");
 }
 
+void SequentialReader::set_filter(
+  const rosbag2_storage::StorageFilter & storage_filter)
+{
+  storage_->set_filter(storage_filter);
+}
+
 bool SequentialReader::has_next_file() const
 {
   return current_file_iterator_ + 1 != file_paths_.end();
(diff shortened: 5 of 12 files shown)


rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
```diff
@@ -68,6 +69,8 @@ class ROSBAG2_CPP_PUBLIC SequentialReader
 
   std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() override;
 
+  void set_filter(const rosbag2_storage::StorageFilter & storage_filter);
```
Karsten1987: Do you think it makes sense to introduce a `reset_filter` function? Or generally the use case of I first want to read all topics `a, b, c` then want to apply a different filter for `d, e, f`.
I guess you could achieve this by simply overriding the filter, but I wonder if there are use cases where you'd want to reset the applied filter.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -90,6 +93,7 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   std::unordered_map<std::string, int> topics_;
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
+  rosbag2_storage::StorageFilter storage_filter_ {};
```
Karsten1987: nit: Remove space before braces.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -211,10 +213,27 @@ void SqliteStorage::prepare_for_writing()
...
+      topic_list += "'" + topic + "'";
+      if (&topic != &storage_filter_.topics.back()) {
+        topic_list += ",";
+      }
+    }
+
+    read_statement_ = database_->prepare_statement(
+      "SELECT data, timestamp, topics.name "
+      "FROM messages JOIN topics ON messages.topic_id = topics.id "
+      "WHERE topics.name IN (" + topic_list + ")"
```
Karsten1987: did you try to apply the same logic if `topic_list` is empty? I would assume that every topic name matches `""`. I am bringing this up because that could potentially safe us the `else` branch below.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include <gmock/gmock.h>
 
+#include <rosbag2_storage/storage_filter.hpp>
+
```
Karsten1987: style: please adhere to the applied style and include this with `""` below the system headers.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -111,6 +113,33 @@ TEST_F(StorageTestFixture, get_next_returns_messages_in_timestamp_order) {
   EXPECT_FALSE(readable_storage->has_next());
 }
 
+TEST_F(StorageTestFixture, read_next_returns_filtered_messages) {
```
Karsten1987: I would appreciate having this test actually happening in `rosbag2_cpp` so it can be applied to more than just `SQLite3`.

Karsten1987:
@mabelzhang Any update on this?

mabelzhang:
I created a standalone C++ file to test this, and for some reason whenever it calls `set_filter()`, it just terminates, without even printing the error if the file doesn't exist. It doesn't do that if I remove the `set_filter()` call. I need to look into what's wrong in the code I added.

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
mabelzhang: Added in 7d16787

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
mabelzhang: I copied the format a few lines above (lines 90-92). Should I remove the space for all of them?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
mabelzhang: So I tried that, and with empty string, no messages were returned.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
mabelzhang: I had trouble creating the test. Can you suggest a test file in `rosbag2_cpp` that I can look at as an example of how to put messages into a reader? I tried adding it in `rosbag2_cpp/test/rosbag2_cpp/test_sequential_reader.cpp`, but I'm not able to get a valid object returned from `read_next()` - I get null pointers. Maybe because the object is a `MockStorage`?

The other problem is since the actual filtering is only implemented in `sqlite3` right now, we wouldn't be able to get the right filtering behavior from a `SequentialReader` using `MockStorage`, right?

What I have done locally is I added a new class to `test_sequential_reader.cpp`, changed the `reader_` object from a `Reader` to a `SequentialReader`, since the `set_filter()` is only in `SequentialReader` right now. If we want it to be in `Reader`, it also needs to be in `BaseReadInterface` and `rosbag2_compression::SequentialCompressionReader`. Do we want to define `set_filter()` beyond `SequentialReader`?

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
Karsten1987: I didn't see that it's done as such in the lines above. Nevermind then.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: too bad. I would have hoped we could simply just initialized `topic_list` here to some magic thing to ease the code flow.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
Karsten1987: I guess you have two options here. Either you modify the mock functions in a way to take the filter logic into account or you shift the test over to `rosbag2_tests` where we have a fully set up architecture. This also runs the sqlite3 plugin and still gives you the opportunity to test the c++ API.

Do you think it makes sense to apply the filter to each individual reader instance? Then we should add the two functions (`set_filter()/reset_filter()`). I would assume that these filters are generic enough (filter by topic etc) to apply for each reader.

Karsten1987:
@mabelzhang have you had a chance to re-iterate over the tests?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
mabelzhang: I spent some time putting the tests in `rosbag2_cpp` and `rosbag2_tests`. I did both halfway, and I think I finally figured out what the tests in these 3 packages do.

In the `rosbag2_cpp` tests, `test_sequential_reader.cpp` is intentionally using a mock object in `mock_storage.cpp` so that it tests the functions in the SequentialReader only, just to verify that `reader_` calls functions inside `storage_`. The tests here are independent of the Storage object. 

So with `re/set_filter()`, in the current setup, all that should be tested in `rosbag2_cpp` is that `reader_.set_filter()` will call `storage_.set_filter()`, and no checks on whether the returned objects are the correct topic names. That can be added - I have this locally.

The tests in `rosbag2_storage` test the actual function of the storage. So `set_filter()` here should be tested with the actual returned messages' topic names. This is the present state of the PR. I think we should keep these tests, because there's nowhere else that does unit test on the actual functionality of the Storage objects.

The tests in `rosbag2_tests` execute the command line, which would require implementing the topic filter in `rosbag2_transport`. That has not been implemented yet. It would require enough changes that maybe it should be in a separate PR? With `rosbag2_transport` tests, `test_play.cpp` uses a `mock_sequential_reader.cpp` mock object, which has its functions implemented. This can be updated with `set_filter()` to mock the functionality of the storage - I have this locally.

So my question is what we want to go into this PR - I have a mess of both routes on my local computer, with new tests in all 3 packages, and I haven't pushed anything. I think both the `rosbag2_storage` and `rosbag2_cpp` tests should be kept. In that case, I would just add the few lines to `rosbag2_cpp` tests as mentioned above to this PR, and leave this test in `rosbag2_storage` as well.

If we also want to add end-to-end tests and implement a new flag on `rosbag2_transport`, I can put that in a new branch and open a new PR. Otherwise I would just get rid of those changes. That option did not exist in ROS 1 bag playback and I don't know what flag we want to make it - maybe like ROS 1 bag record where `-O` is for the file name, and a subsequent list is the selective topics.

Karsten1987: I agree with basically everything you said here.

I would still appreciate some basic tests in `rosbag2_cpp` for things like what happens if you call the `set_filter` function twice in a row or test for the exception to be thrown if the storage is not initialized. That shouldn't take too much time, I think.

The functionality can be tested in the sqlite3 implementation of it (I think that's what you meant with `rosbag2_storage`, but I think it's indeed the default storage plugin). 
But I would also assume we could get around `rosbag2_transport` or the `ros2cli` in the system tests (`rosbag2_tests`) by directly calling the `rosbag2_cpp` API. But I am okay with putting this in a separate PR to address once the option is exposed through the command line interface. I would ask you though in this case to open a new ticket for it to keep track of the outstanding work. It's up to you, really, which way you want to go. I am okay with both.

mabelzhang: Okay. I added basic tests in `rosbag2_cpp` in 5e09b58.

I tried making a new test in `rosbag2_tests` and calling the `SequentialReader::open()` in `rosbag2_cpp` to open the bag in `rosbag2_tests/resources/cdr_test`. For some reason it doesn't open:
```
[22.820s] 4: [ERROR] [1585629564.565606326] [rosbag2_storage]: Could not open 'cdr_test.db3' with 'sqlite3'. Error: Failed to read from bag: File 'cdr_test.db3' does not exist!
[22.820s] 4: [ERROR] [1585629564.565750830] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
```

In fact, I cannot even open `cdr_test` and `wrong_rmw_test` with `ros2 bag play` on the command line. I get the same error:
```
$ ros2 bag play cdr_test
[ERROR] [rosbag2_storage]: Could not open 'cdr_test.db3' with 'sqlite3'. Error: Failed to read from bag: File 'cdr_test.db3' does not exist!
[ERROR] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
[ERROR] [rosbag2_transport]: Failed to play: No storage could be initialized. Abort
```

But `ros2 bag info cdr_test` works.
Is this expected? If so, then I won't be able to call `Reader` directly from `rosbag2_tests`. I will open a new ticket to expose the topic selection to `rosbag2_transport`, and then test using the existing `test_rosbag2_play_end_to_end.cpp`.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
mabelzhang: 59d4cb1

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
```diff
@@ -142,6 +142,24 @@ std::vector<rosbag2_storage::TopicMetadata> SequentialCompressionReader::get_all
...
 
+void SequentialCompressionReader::set_filter(
+  const rosbag2_storage::StorageFilter & storage_filter)
+{
+  if (storage_) {
+    storage_->set_filter(storage_filter);
+    return;
+  }
+  throw std::runtime_error("Bag is not open. Call open() before setting "
+          "filter.");
```
Karsten1987, line 153: total bike shedding, but I somehow feel it reads easier if you invert the logic:

```
if (nullptr == storage_) {
  throw ...
}

storage_->set_filter(storage_filter);
```

same below

rosbag2_storage/include/rosbag2_storage/storage_filter.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <string>
+#include <vector>
+
+namespace rosbag2_storage
+{
+
+struct StorageFilter
+{
+  std::vector<std::string> topics;
```
Karsten1987, line 26: might deserve a few words as a comment. Mentioning e.g that it's a whitelist implementation.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -111,6 +113,33 @@ TEST_F(StorageTestFixture, get_next_returns_messages_in_timestamp_order) {
...
+  storage_filter.topics.push_back("topic2");
+  storage_filter.topics.push_back("topic3");
+  readable_storage->set_filter(storage_filter);
+
+  EXPECT_TRUE(readable_storage->has_next());
+  auto first_message = readable_storage->read_next();
+  EXPECT_THAT(first_message->topic_name, Eq("topic2"));
+  EXPECT_TRUE(readable_storage->has_next());
+  auto second_message = readable_storage->read_next();
+  EXPECT_THAT(second_message->topic_name, Eq("topic3"));
```
Karsten1987, line 140: I would expect another test after this line to see that `readable_storage->has_next()` returns `false`. This basically makes sure that really only two messages are being available and `topic1` is never visible to the caller.

rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
mabelzhang: So here I just copied the pattern in the three existing functions above these lines...

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
mabelzhang: Added in b62736a along with reset_filter() tests.

rosbag2_storage/include/rosbag2_storage/storage_filter.hpp
mabelzhang: b62736a

mabelzhang:
The checks have been failing almost every time with things like, bash failed, sudo failed, failed to fetch due to connection timeout.

CI (I kept the Ubuntu distro to default focal, hope that's right):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5499/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8040/)
* Windows [Build Status]

mabelzhang:
CI is passing:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9886/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5504/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8044/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9743/)

Note that with the tests, I'm only doing the packages I touched `--packages-select rosbag2_cpp rosbag2_storage rosbag2_storage_default_plugins rosbag2_transport`. I did not include `rosbag2_tests` in the above.

If I do everything with `--packages-select-regex rosbag2*`, then I saw Windows [Build Status](http://ci.ros2.org/job/ci_windows/9742/) with this in `rosbag2_tests`:
```
C:\ci\ws\src\mabelzhang\rosbag2\rosbag2_tests\test\rosbag2_tests\test_rosbag2_record_end_to_end.cpp(470): error: Expected: (metadata.relative_file_paths.size()) >= (1), actual: 0 vs 1
Bagfile never split!
```
from [this line](https://github.com/ros2/rosbag2/blob/master/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L470):
```
    ASSERT_GE(metadata.relative_file_paths.size(), 1) << "Bagfile never split!";
```
It looks like something with metadata. It doesn't look like it could be from this PR?

Karsten1987:
that end-to-end test is known to be failing on windows. We'll address this hopefully shortly.
CI looks good. Feel free to merge.

Issue #232, opened 2019-12-12
Improve Python API


jacobperron:
Currently, there is limited Python support for rosbag2; only a few functions are exposed via a [C-extension](https://github.com/ros2/rosbag2/blob/e2c96761c1886d14dbc470146dfd05a2528a8c35/rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp) for recording, playing, and printing info for bags. Similar to the Python API in ROS 1, it would be useful to have facilities for looping through bags, requesting a subset of topics, exporting to CSV, etc.

## Prerequisites 

In order to deal with ROS messages from bags in Python, we'll need support for serializing/deserializing messages. This can be done by exposing the `rmw_serialize` and `rmw_deserialize` messages in Python. It makes sense to me to do this in rclpy.

## Implementation considerations

We can extend the existing C-extension to expose more of the rosbag API and move it to it's own package, `rosbag2_py`. Then a new rosbag2 Python library can be built around it, using rclpy as needed.

For a first iteration, I think having the ability to read/write messages to/from bags from Python would be great; similar to [how this was done in ROS 1](http://wiki.ros.org/rosbag/Code%20API#py_api).

jacobperron:
I've implemented a proof-of-concept using pybind11 to read messages from a rosbag in a Python for loop.

Here's an [example](https://gist.githubusercontent.com/jacobperron/d2f34e74e43985c7d02ea925f23cafed/raw/da1a6fda909b66d1acd5049cf96ad38b4660d944/read_bag.py) of parsing a bag, which requires the following two branches of `rclpy` and `rosbag2`:

https://github.com/ros2/rclpy/tree/jacob/rclpy_serialize

https://github.com/ros2/rosbag2/tree/jacob/rosbag2_py

As well as `pybind11_vendor`:

https://github.com/ros2/pybind11_vendor

---

I think pybind11 looks like a promising tool for us to wrap the cpp implementation. Next, we should identify the data structures to expose/wrap in `rosbag2_py` and add a proper Python interface on top.

I'll open a PR to `rclpy` soon with the message serialization functions.

@Karsten1987 FYI

jacobperron:
`rclpy` PR adding serialization functions: https://github.com/ros2/rclpy/pull/495

Karsten1987:
@jacobperron Are you planning on working on a PR for the python API (aka `rosbag2_py`)?

mabelzhang:
[...]

PR #306, opened 2020-03-01
Add convert verb

### LINKED ISSUE #163: ros2bag convert verb
I've recently written a small converter which just uses existing functionality of the reader and writer to convert between different serialization formats. This utility would be quite helpful for example in converting old ROS 1 bags to ROS 2 format. I was wondering if it would be a good idea to add such a `convert` verb as part of the ros2bag API. Let the user choose input and output storage_id and serialization_id.

If so, I can work on a small PR since I already have a PoC running.

Kettenhoax:
Adds a convert verb to convert bags between storages and serialization formats. Tested with a conversion of rosbag2_bag_v2 to sqlite3, but I was unsure whether to add a bag to the repository for an integration test, because it would require rosbag2_bag_v2 as test dependency.

Should close #163


ros2bag/ros2bag/verb/convert.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
```
Karsten1987: year 2020

ros2bag/ros2bag/verb/convert.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def add_arguments(self, parser, cli_name):  # noqa: D102
+        parser.add_argument(
+            'bag_file', help='bag file to convert')
+        parser.add_argument(
+            '-o', '--output',
+            help='destination of the bagfile to create, \
+            defaults to a timestamped folder in the current directory')
+        parser.add_argument(
+            '-s', '--storage', default='sqlite3',
```
Karsten1987: for a `convert` verb I would consider renaming this to `--input-storage` to highlight the conversion from `input` to `output`.

Karsten1987: I am not completely sure a default for `input` makes sense here. The storage identifier might be guessed from the `metadata.yaml` if present, but otherwise this has to be specified. 
It makes more sense for me to have it defaulted to `sqlite3` on the output side.

ros2bag/ros2bag/verb/convert.py
Karsten1987: same below

togaen:
@Kettenhoax @Karsten1987 Any updates on this? It would be really great to have. I happy to help out if I can.

Kettenhoax:
I'm still interested in merging this and have been waiting on #232, which seems to add the required Python API to implement the requested changes.

jacobperron:
I'm also interested in this.

@Kettenhoax, #232 has been resolved. There's now a basic Python API. I'm not sure if it contains all the functionality needed for this feature, but I'm happy to help add/review PRs adding what's missing.

jacobperron:
@Kettenhoax Friendly ping. Let me know if you have plans to work on this, otherwise, I may have some spare cycles soon to help push it forward.

Kettenhoax:
Yes, I can work on this next few days.

@Karsten1987 or @jacobperron Looking at the current master, could you clarify if the preferred way to implement this is either

1. using the existing Reader and Writer api in rosbag_py to iterate over messages in the verb extension point, or
2. add a Converter api to rosbag_py and invoke it in ros2bag

In both cases as dependency on rosbag_py will need to be added to ros2bag.

ros2bag/ros2bag/verb/convert.py
Kettenhoax: In current master, the `play` and `info` verbs also set the input storage argument default to `sqlite3`, so I thought it best to keep the convention. I agree that it should be guessed, but the necessary api doesn't seem to be available in python, and IMO shouldn't be added in this PR.

I would rename it to `--input-storage`, remove the default and set required=True. I would also remove the short `-s` and not add `-i` to avoid confusion with `-o`, which is a filepath rather than storage specifier.

jacobperron:
@Kettenhoax For me, adding a converter API to rosbag2_py is the preferred option. This allows other Python applications to take advantage of the functionality without having to go through the CLI verb.

jacobperron:
@Kettenhoax When adding the new API to rosbag2_py, it would be best to do it in a separate PR from the changes to the CLI. This will help with the review process and maintenance :slightly_smiling_face:

jacobperron:
@Kettenhoax Friendly ping :slightly_smiling_face: Any update?

Kettenhoax:
Yes, I've started to extend rosbag2_py API by a converter. However, I've run into roadblocks when implementing compression. Currently a caller of rosbag2_py decides whether to instantiate a compression- or standard reader/writer. The way I think this API is meant to be used, a converter should be passed an already-instantiated reader and writer.

As PoC I tried to do this in the verb only, without extending rosbag2_py. Unfortunately `rosbag2_py.SequentialCompressionWriter` seems to lack a parameter for the compression options.

We also need a python-exposed API to determine whether a bag is compressed and instantiate the correct reader. Currently this is only done in `rosbag2_transport_python.cpp`.

@jacobperron It would be great if you could do the necessary rosbag2_py extensions. You can check da6c1841d4ec064574fdbe9f0764475df30959f7 for how I'm trying to use rosbag2_py

emersonknapp:
A note of interest - we have plans to consolidate the compression/standard readers/writers - having realized that the split mostly leads to code duplication that is hard to maintain. Unfortunately that work hasn't started yet - I know it would make this easier

ZhenshengLee:
Any update?

emersonknapp:
Closing this as stale in favor of the work going on by @gbiggs for https://github.com/ros2/rosbag2/issues/831 - that work will implement most if not all of this proposal.

PR #308, opened 2020-03-03
rosbag2_py reader and writer

mabelzhang:
Initial reader functionalities for rosbag2_py.

Dependencies:
```
sudo apt-get install python-pybind11
git clone https://github.com/ros2/pybind11_vendor.git
```

Example usage is ~~[here](https://gist.github.com/mabelzhang/31468244b50e6ebc5a7e374af0e01fec) (minus the lines about StorageFilter which need #302 and Python binding)~~ in the test.


rosbag2_py/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic -Werror)
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_ros REQUIRED)
+find_package(rosbag2_cpp REQUIRED)
+
+find_package(pybind11 REQUIRED)
```
piraka9011: nit: alphasort

rosbag2_py/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rosbag2_py</name>
+  <version>0.2.4</version>
+  <description>Python API for rosbag2</description>
+  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
```
piraka9011: Can you add `<maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>`

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+private:
+  rosbag2_cpp::readers::SequentialReader reader_;
+};
+
+}  // namespace rosbag2_py
+
+PYBIND11_MODULE(_rosbag2_py, m) {
+  m.doc() = "Python wrapper of rosbag2 implementation";
```
piraka9011: "... of the rosbag2_cpp API`

piraka9011:
Thanks! Not too familiar with pybind but is it possible to have this binding for the general `reader` as opposed to a specific one? That way, we can specify whatever kind of reader we want in the Python API.

rosbag2_py/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+install(
+  TARGETS _rosbag2_py
+  DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}"
+)
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
+  # TODO(jacobperron): Add pytests
```
emersonknapp: I like this TODO - I think we should probably add some basic verification of this new functionality before merging this PR

Karsten1987: @mabelzhang Do you think you could provide some basic tests for this package?

mabelzhang: Yeah. I'm going to look for some examples of setting up pytests.
I also plan on looking into wrapping around the general `Reader` mentioned in piraka9011's comment.

mabelzhang: pytests added in 54799a9
Commented lines in there are waiting for #302

mabelzhang:
@piraka9011 Thanks for the review! f8c4391 wraps around Reader. The type e.g. `"SequentialReader"` is passed in as string. Usage example is added in pytest file.

rosbag2_py/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+  find_package(ament_cmake_pytest REQUIRED)
+
+  set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE}")
+  if(WIN32 AND "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+    set(_PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
+  endif()
+  set(pythonpath "${CMAKE_CURRENT_BINARY_DIR}/rosbag2_py;${CMAKE_CURRENT_SOURCE_DIR}")
+  if(NOT WIN32)
+    string(REPLACE ";" ":" pythonpath "${pythonpath}")
+  endif()
```
piraka9011, line 60: I didn't test on windows but is this necessary?
Just wanted to know since I haven't seen other packages require this.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,119 @@
...
+  {
+    reader_->open(storage_options, converter_options);
+  }
+
+  bool has_next()
+  {
+    return reader_->has_next();
+  }
+
+  /// Return a pair containing the topic name and the serialized ROS message.
```
piraka9011, line 65: nit: returns a tuple, add the timestamp

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,65 @@
...
+from rclpy.serialization import deserialize_message
+
+from std_msgs.msg import String
+from rcl_interfaces.msg import Log
+
+import rosbag2_py._rosbag2_py as rosbag2_py
+
+def test_sequential_reader():
+
+  #print(_SRC_RESOURCES_DIR_PATH)
```
piraka9011, line 30:
```suggestion

```

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,120 @@
...
+  .def_readwrite("input_serialization_format",
+    &rosbag2_cpp::ConverterOptions::input_serialization_format)
+  .def_readwrite("output_serialization_format",
+    &rosbag2_cpp::ConverterOptions::output_serialization_format);
+
+  pybind11::class_<rosbag2_storage::TopicMetadata>(m, "TopicMetadata")
+  .def(pybind11::init())
+  .def_readwrite("name", &rosbag2_storage::TopicMetadata::name)
+  .def_readwrite("type", &rosbag2_storage::TopicMetadata::type)
+  .def_readwrite("serialization_format",
```
klintan, line 117: I might be missing something, but might be nice to add `name`, `type` and `serialization_format` attributes as arguments to the init function. I saw the `TopicMetadata` struct does not have a constructor, so this might lead to somewhat different API? but still in Python there are no way of initializing `TopicMetadata` without the somewhat convoluted:

```
topic = TopicMetadata()
topic.name = "chatter"
topic.serialization_format ="cdr"
topic.type = "std_msgs/String"
```
Would be nice to be able to just do:
```
topic = TopicMetadata(name="chatter", type="std_msgs/String", serialization_format="cdr")
```
Of course still keeping them as readwrite attributes.

mabelzhang: Yeah I was trying to keep the Python API similar to the C++ API, and yes, all the ROS message types are initialized the long way in Python.

mabelzhang:
One thing I'm not sure about is, currently the test scripts use 
```
os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), 'resources')
```
to access the `resources` directory, which is not very ROS-y. I don't know if that would work on the CI machines.

I saw somewhere that you could define something like this, and the variable would be accessible in C++, but I haven't found a Python equivalent.
`add_definitions(-D_SRC_RESOURCES_DIR_PATH="${CMAKE_CURRENT_SOURCE_DIR}/resources")`

I set this in `CMakeLists.txt` inside `ament_add_pytest_test()`:
    WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/resources"
but that still doesn't let me access that directory in Python at test time.

@jacobperron do you know of a more proper way to access a subdirectory inside a package in Python at test time? It's only used for tests, so we don't want to install that directory.

klintan:
> One thing I'm not sure about is, currently the test scripts use
> 
> ```
> os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), 'resources')
> ```
> 
> to access the `resources` subdirectory, which is not very ROS-y. I don't know if that would work on the CI machines.
> 
> I saw somewhere that you could define something like this, and the variable would be accessible in C++, but I haven't found a Python equivalent.
> 
> ```
> add_definitions(-D_SRC_RESOURCES_DIR_PATH="${CMAKE_CURRENT_SOURCE_DIR}/resources")
> ```
> 
> I set this in `CMakeLists.txt` inside `ament_add_pytest_test()`:
> 
> ```
>     WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/resources"
> ```
> 
> but that still doesn't let me access that directory in Python at test time.
> 
> @jacobperron do you know of a more proper way to access a subdirectory inside a package in Python at test time? It's only used for tests, so we don't want to install that directory.

Ah yes you are right, we could probably use:

```
from ament_index_python.packages import get_package_share_directory

get_package_share_directory("rosbag2_py"), "resources", "<filename>")
```
To get to the test data.

mabelzhang:
That gives the shared directory in the install/ directory, as opposed to the source directory, where the bag files are. Test files shouldn't be installed, so I am hesitant to install that directory since the bag files are only used for tests.

jacobperron:
> `os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), 'resources')`

@mabelzhang this is the way I would do it. I'm not aware of a ROS (or ament) specific utility for this.
I believe CI should be okay; we'll find out when we run it.

mabelzhang:
Okay I'll leave it as it is then.

rosbag2_py/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>
+  <license>Apache License 2.0</license>
+
```
jacobperron, line 10: @mabelzhang You can add yourself as an author.

rosbag2_py/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
+
+  <depend>pybind11_vendor</depend>
```
jacobperron, line 13: We should confirm, but I think we can rely on the [existing rosdep key](https://github.com/ros/rosdistro/blob/ce7f25888bae5a321ed8ebd53848ad6e85503b5e/rosdep/base.yaml#L5274) for Linux.

```suggestion
  <depend>pybind11-dev</depend>
```

For macOS and Windows, we should update our CI script and install instructions to install via pip: https://pypi.org/project/pybind11/

e.g. `python -m pip install pybind11`

Can you try updating the dependency key here? I can update https://github.com/ros2/ci to ensure that pybind11 is installed for macOS and Windows.

rosbag2_py/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <depend>rosbag2_cpp</depend>
+  <depend>rosbag2_compression</depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
```
jacobperron, line 19: I think we should add some C++ linters. Maybe just replace all of these with the common linters?
We're also missing a test depend on `ament_lint_auto`.

```suggestion
  <test_depend>ament_lint_auto</test_depend>
  <test_depend>ament_lint_common</test_depend>
```

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,78 @@
...
+from std_msgs.msg import String
+from rcl_interfaces.msg import Log
+
+from common import get_rosbag_options
+
+import rosbag2_py._rosbag2_py as rosbag2_py
+
+def test_sequential_reader():
+
+  bag_path = os.path.join(
```
jacobperron, line 32: nitpick: we use four space indentation for Python (I think almost everywhere) in ROS 2.

I'm a big surprised that the linters don't care.

rosbag2_py/test/test_sequential_writer.py
```diff
@@ -0,0 +1,103 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import glob
+import os
+import time
+import shutil
+
+from rclpy.serialization import deserialize_message, serialize_message
+import rosbag2_py._rosbag2_py as rosbag2_py
```
jacobperron, line 21: From a user perspective, I think it would be best if they didn't have to import the submodule `_rosbag2_py`.
I.e. simply this to

```python
import rosbag2_py
```
 
I vaguely recall having an issue achieving this though, but I don't recall.

rosbag2_py/test/test_sequential_reader.py
emersonknapp: The linters do care :) - but the flake8 check is temporarily disabled in the Github Actions

emersonknapp:
> os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), 'resources') 

Just as a cosmetic thing, the pathlib version might be a bit more readable.

```
from pathlib import Path
Path(__file__).parent.parent / 'resources'
```

You could reconvert it to a string when you need it to be one (`str(path)`), Path objects are a lot easier to manipulate.

rosbag2_py/CMakeLists.txt
mabelzhang: The mentioned commented lines had been implemented as of 562d2dd

rosbag2_py/package.xml
mabelzhang: That worked for me on Linux!

rosbag2_py/package.xml
mabelzhang: Niiiiice I got like a hundred errors after that. I was wondering why nothing complained...
d261c63

rosbag2_py/test/test_sequential_reader.py
mabelzhang: Adding the common linters in package.xml caught them.
d261c63

rosbag2_py/test/test_sequential_writer.py
mabelzhang: You added a TODO in CMakeLists.txt [line 29](https://github.com/mabelzhang/rosbag2/blob/rosbag2_py/rosbag2_py/CMakeLists.txt#L29) to `Consider modifying ament_python_install_module to accept targets`. Does something need to be added to ament in order to do that?

rosbag2_py/package.xml
mabelzhang: Hmm I don't know if I want to officially sign up for that...

mabelzhang:
> the pathlib version might be a bit more readable.
Cool, changed in d261c63

rosbag2_py/package.xml
jacobperron: It doesn't imply maintainership, rather giving you credit (use the `<author>` tag). Up to you though.

rosbag2_py/test/test_sequential_writer.py
jacobperron: I don't think it's related? I guess I was thinking that TODO would be better than having to install the module manually here: https://github.com/mabelzhang/rosbag2/blob/d261c63ef4a1ec1e52cf768c58667a5b8d598284/rosbag2_py/CMakeLists.txt#L40-L44

rosbag2_py/package.xml
jacobperron: I've pushed changes to ros2/ci for installing pybind11 with pip on macOS and Windows: https://github.com/ros2/ci/commit/7782acc904eb72b2874cafa9b5b8851b0d64e21a

I've triggered CI to see how it goes.

rosbag2_py/test/test_sequential_writer.py
mabelzhang: ab426d9 simplifies the import. The `__all__` trick in `__init__.py` worked. Simply importing in `__init__.py` didn't (or I couldn't figure out how).

rosbag2_py/package.xml
mabelzhang: Turns out the author tag isn't allowed by the linter. So I'll leave it off. Thanks for mentioning though!

mabelzhang:
`pybind11-dev` can be installed via `sudo apt-get install pybind11-dev`.

I think we are still deciding on which route to go though, specifically what works better for Windows and Mac.

rosbag2_py/package.xml
jacobperron: > Turns out the author tag isn't allowed by the linter.

That's surprising. We use `<author>` tags in several other core ROS 2 packages.

rosbag2_py/package.xml
jacobperron: Turns out the pip package does not ship with the necessary cmake config. Relevant upstream issue: https://github.com/pybind/pybind11/issues/1628

I think the best course of action is to create a `pybind11_cmake_module` package providing one or add it to `rosbag2_py`.

rosbag2_py/package.xml
dirk-thomas: > Turns out the author tag isn't allowed by the linter.

You need to follow the order of the tags defined by the schema to successfully validate the XML file.

mabelzhang:
@ros2/tooling_working_group Can we get another review on this? After this is in, we are planning on adding a seek feature and a Python wrapper for that (which would depend on this PR), towards having some kind of introspection in bags like in ROS 1.

rosbag2_py/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+ament_python_install_package(${PROJECT_NAME})
+
+pybind11_add_module(_rosbag2_py SHARED
+  src/rosbag2_py/_rosbag2_py.cpp
+)
+# TODO(jacobperron): Consider modifying ament_python_install_module to accept targets
+# ament_python_install_module(_rosbag2_py
+#   DESTINATION_SUFFIX "rosbag2_py"
+# )
+# target_link_libraries(_rosbag2_py
```
Karsten1987, line 33: Is that still valid? If so, I'd propose to create a follow up ticket on that.

rosbag2_py/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+
+ament_target_dependencies(_rosbag2_py PUBLIC
+  "rosbag2_cpp"
+  "rosbag2_compression"
```
Karsten1987, line 37: alpha

rosbag2_py/CMakeLists.txt
Karsten1987: what's the status on this?

rosbag2_py/package.xml
```diff
@@ -0,0 +1,24 @@
...
+  <description>Python API for rosbag2</description>
+  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
+  <maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
+
+  <depend>python11-dev</depend>
+  <depend>rosbag2_cpp</depend>
+  <depend>rosbag2_compression</depend>
```
Karsten1987, line 15: alpha

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include "rosbag2_storage/storage_filter.hpp"
+#include "rosbag2_storage/topic_metadata.hpp"
+
+namespace rosbag2_py
+{
+
+class Reader
+{
+public:
+  explicit Reader(const std::string & reader_class)
```
Karsten1987, line 43: That sort of logic will hardly scale. I understand that this might result in extra work, which we can address in another PR. But what's needed to support any type of reader implementation?

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /// Return a tuple containing the topic name, the serialized ROS message, and
+  /// the timestamp.
+  pybind11::tuple read_next()
+  {
+    const auto next = reader_->read_next();
+    rcutils_uint8_array_t rcutils_data = *next->serialized_data.get();
+    std::string serialized_data(rcutils_data.buffer,
+      rcutils_data.buffer + rcutils_data.buffer_length);
```
Karsten1987, line 75: why can't we return a uint8_t array? If we have to pack this into a STL container, I'd prefer `std::vector<uint8_t>`

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,72 @@
...
+from common import get_rosbag_options
+from rcl_interfaces.msg import Log
+from rclpy.serialization import deserialize_message
+import rosbag2_py
+from rosidl_runtime_py.utilities import get_message
+from std_msgs.msg import String
+
+
+def test_sequential_reader():
+
```
Karsten1987, line 26:
```suggestion
```

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
jacobperron: We might be able to add a template argument for the reader class. Looks like pybind11 only supports explicit template instantiations, so we could try something like this: https://github.com/pybind/pybind11/issues/199#issuecomment-323995589

rosbag2_py/CMakeLists.txt
mabelzhang: I talked to Jacob about this. He's going to think about it and ticket if necessary in the appropriate repo. This would be a todo somewhere else, not in rosbag2.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: I asked for Jacob's help looking into this. When `read_next()` is called in Python, such as in the `test_sequential_reader.py`, the message needs to be deserialized. That was enabled in a new PR in `rclpy`, which takes a byte array as input: https://github.com/ros2/rclpy/pull/495/files#diff-80dda03110b5be606b823d6b0558b5c2R4745 . Currently `read_next()` returns a `pybind11::bytes`, which translates to a Python `bytes` type and can be passed directly to `rclpy.serialization.deserialize_message()`

I think it is possible to return a `std::vector<uint8_t>`, but that would translate to a python `list` type of ints. So it would require an extra step in the Python code on the user's end to convert to a `bytes` type in order to deserialize it.

rosbag2_py/CMakeLists.txt
mabelzhang: I asked around a bit. It turns out for the first if-statement, I needed to add a dependency on `python_cmake_module` in order for `PYTHON_EXECUTABLE` variable to be defined. That is added in 133e1a4. I will run Windows CI for debug build type to check what happens.

The second if-statement is for `pythonpath` to be defined, which is necessary.

mabelzhang: Oh Windows CI is actually failing on cannot find pybind first. Duh. The cmake module for finding pybind needs to be done first before I can test this variable in Windows.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: In 08db02f, I tried making the subclasses more explicit as opposed to strings. Pybind can export inheritance
https://pybind11.readthedocs.io/en/stable/classes.html#inheritance-and-automatic-downcasting But as in this example and Jacob’s example, the subclasses still need to be explicitly defined as classes for pybind. So this latest version isn’t much better in terms of scaling, and it creates even more work for adding a new reader now, but maybe it is better in terms of making what readers are available explicit. In all three ways of doing it, new readers have to be manually added.

rosbag2_py/CMakeLists.txt
Karsten1987: @mabelzhang are you looking into this cmake module? I would expect the vendor package to provide this.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
Karsten1987: I think that's fine. I was more wondering about why using a `std::string` here for the serialized buffer.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,256 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
Karsten1987, line 1: Can we easily split this into two files? I could see that these classes (`Reader` and `Writer`) grow over time and I think it makes sense to have them separated into its own file respectively.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  void remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
+  {
+    writer_->remove_topic(topic_with_type);
+  }
+
+  /// Write a serialized message to a bag file
+  void write(
+    const std::string & topic_name, const std::string & message,
```
Karsten1987, line 144: why is the message encapsulated as a `std::string` here?

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
jacobperron: See https://github.com/mabelzhang/rosbag2/pull/2/, which uses template specialization instead of defining new classes. It's a tiny bit more compact, and I think makes it a little easier to add new specializations.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: I think it's just that the conversion between string and bytes is the easiest and native to what Python [`bytes` / `bytearray`](https://docs.python.org/3/library/functions.html#func-bytearray) do.

rosbag2_py/CMakeLists.txt
mabelzhang: It was using the `pybind11_vendor` package before, but I think we want to switch to `pybind11-dev`, see this thread https://github.com/ros2/rosbag2/pull/308#discussion_r441850368

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: The Python object being passed in is `bytes` type, and `std::string` or `char *` will be passed in without conversion
https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#passing-bytes-to-c Here since it needs to be passed to `rosbag2_storage::make_serialized_message`, which takes the size, `std::string` is easier to get the size. `pybind11:;bytes` is another option, but `rosbag2_storage::make_serialized_message` also takes a `const void *`, and using the `char *` from a string is fewer casting / conversions.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: 54f6722 splits into `_reader`, `_writer`, and `_storage` files. Let me know if these are the final naming we want for the files and pybind targets. Storage includes more than just `rosbag2_storage`, so maybe it could be called utilities.

rosbag2_py/src/rosbag2_py/_reader.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <chrono>
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
```
jacobperron, line 23: I don't think we need many of these includes after the refactoring.
```suggestion
#include <memory>
#include <string>
#include <vector>
```

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <chrono>
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
```
jacobperron, line 23: double-check, but I don't think we need these anymore
```suggestion
```

rosbag2_py/src/rosbag2_py/_writer.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <chrono>
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
```
jacobperron, line 23:
```suggestion
#include <memory>
#include <string>
```

rosbag2_py/src/rosbag2_py/_storage.cpp
mabelzhang: `pybind11/stl.h` is still needed, cleaned up the rest in b845790

rosbag2_py/package.xml
mabelzhang: I looked into this for a while and found out a few things. First, I think my test above was not thorough. I did not place a `COLCON_IGNORE` in my pybind11_vendor package, so even though I did not depend on it, colcon was likely still using it to find pybind. That was why it worked. Once I put a `COLCON_IGNORE` in pybind11_vendor, colcon fails to find pybind11Config.cmake.

Pybind only supplies the .cmake file when it is installed with Conda https://github.com/pybind/pybind11/issues/1379 There's been a lot of complaints but they aren't interested in fixing it. I got `pybind11Config.cmake`, `pybind11Tools.cmake`, `pybind11Targets.cmake`, `FindPythonLibsNew.cmake` from a Conda install of pybind, and the `pybind11Config.cmake.in` file which generates the `pybind11Config.cmake` from https://github.com/pybind/pybind11/blob/master/tools/pybind11Config.cmake.in , and I was able to hook all these files up. But the remaining problem is that I can't get `@PACKAGE_INIT@` to expand to the pybind11 installation directory. I don't see how it can ever know where that path is on the user's local machine. It expands to `../..`, which is correct if run within pybind11, but at the point it's run within rosbag2_py, find(pybind11) doesn't work yet, so we don't have the installation path of pybind11 to expand that to. That path is further used to set the `pybind11_INCLUDE_DIRS`, which is what my colcon is complaining right now - can't find `pybind/pybind.h`.

So, I've gotten a lot farther, but I don't think there's a way forward unless there's a way to figure out where pybind11 is installed locally - before pybind11Config.cmake is generated. All this is to say, I propose we go back to using `pybind11_vendor`.

jacobperron: Switching back to the vendor package sounds okay to me. Thanks for looking into it :bow: 

Please add the vendor package to the [ros2.repos](https://github.com/ros2/ros2/blob/master/ros2.repos) file and trigger CI with it and this PR.

mabelzhang:
be62274 switches back to `pybind11_vendor`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11627/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6759/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9513/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11603/)

mabelzhang:
I'm looking into the rosbag2_storage errors in the test failures here.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11651/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6779/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9530/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11627/)

rosbag2_py/test/test_sequential_writer.py
```diff
@@ -0,0 +1,101 @@
...
+    writer.create_topic(topic)
+
+
+def test_sequential_writer():
+    """
+    Test for sequential writer.
+
+    :return:
+    """
+    bag_path = str(Path(__file__).parent.parent / 'resources' / 'tmp_write_test')
```
jacobperron, line 51: I think it would be better if we didn't leave test artifacts lingering in the source tree. Also, for some reason it doesn't look like the `shutil.rmtree` call is working on my machine.

I think it's better to use a `tempfile.TemporaryDirectory` context for this whole test, e.g.

```python
    with tempfile.TemporaryDirectory() as tmpdir:
        bag_path = os.path.join(tmpdir, 'tmp_write_test')

        # rest of the test here
```

Then we don't need to worry about checking for the existence of `bag_path` or cleaning up the temporary directory afterwards.

jacobperron:
@mabelzhang It looks like the tests are failing because the `rosbag2_storage_plugins` package is not being built. I'm not sure *why* it is not being built though, since it is a transitive dependency of `rosbag2_cpp` :thinking:

rosbag2_py/test/test_sequential_writer.py
mabelzhang: Nice. `shutil.rmtree` not working explains why I was getting this other error on CI when I reran without changing anything but a printout:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11653/)
```
>       writer.open(storage_options, converter_options)
E       RuntimeError: Database direcotory already exists (/home/jenkins-agent/workspace/ci_linux/ws/src/mabelzhang/rosbag2/rosbag2_py/resources/tmp_write_test), can't overwrite existing database
```

I will try the `tempfile` and rerun CI.

mabelzhang:
Oh, interesting. I should probably add that `<exec_depend>rosbag2_storage_default_plugins</exec_depend>` line to `rosbag2_py` as well. I ran CI with `--packages-up-to rosbag2_py`

jacobperron:
> Oh, interesting. I should probably add that <exec_depend>rosbag2_storage_default_plugins</exec_depend> line to rosbag2_py as well. I ran CI with --packages-up-to rosbag2_py

I guess that would help. But I'm still puzzled, because locally the exact same `colcon build` command from the CI job will build `rosbag2_storage_default_plugins` for me...

jacobperron:
@mabelzhang Ah, I see. The CI run was using an outdated version (if I checkout the same commit locally I can reproduce the failure). It looks like the merge with master resolved the issue for me (825250c). I don't think we need to add an explicit dependency on `rosbag2_storage_default_plugins` afterall.

mabelzhang:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11666/)

rosbag2_py/test/test_sequential_writer.py
```diff
@@ -48,13 +47,9 @@ def test_sequential_writer():
...
-    bag_path = str(Path(__file__).parent.parent / 'resources' / 'tmp_write_test')
-
-    if os.path.exists(bag_path):
-        print('%s exists. It will be overwritten by new rosbag2_py test_sequential_writer run.' % (
-            bag_path))
-        shutil.rmtree(bag_path)
-    os.mkdir(bag_path)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        bag_path = os.path.join(tmpdir, 'tmp_write_test')
+    os.makedirs(bag_path)
```
jacobperron, line 52: nitpick: the TemporaryDirectory context will create the path for us. We can indent the remainder of the test to ensure the directory exists for the entire test.

I didn't think of it before, but alternatively we could use the [pytest fixture `tmp_path`](https://docs.pytest.org/en/stable/tmpdir.html) and avoid indentation altogether.

Karsten1987:
I think it further makes sense to add this package to `rosbag2` (metapackage). So we can continue using the pattern of `build: --packages-up-to rosbag2, test: --packages-select-regex rosbag2`

mabelzhang:
Added `rosbag2_py` to meta package in 76e3850.

mabelzhang:
I commented out `makedirs` in the test_sequential_writer.py, and now there are no more CI failures. But locally I get a failure.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11667/)

Locally, if I don’t call mkdir, then I get:
[2.361s] 2: >       writer.open(storage_options, converter_options)
[2.361s] 2: E       RuntimeError: No storage could be initialized. Abort
It passes if I call mkdir.

On CI, if I call mkdir, then I get
>       writer.open(storage_options, converter_options)
E       RuntimeError: Database direcotory already exists (/tmp/tmpulmafpc0/tmp_write_test), can't overwrite existing database
It passes if I comment out mkdir.

That seems weird... I merged with master locally too, and the CI is running on my fork.

jacobperron:
> That seems weird... I merged with master locally too, and the CI is running on my fork.

@mabelzhang Maybe try building in a fresh workspace (removing `build` and `install` directories for rosbag2 packages)?
It sounds like locally you may be partially overlaying on an older version of rosbag2.

mabelzhang:
That worked for me. I didn't recompile all the new rosbag2* packages I pulled!

Rerunning Mac and Windows, with `pybind11_vendor` `CMakeLists.txt` updated on my fork to get version 2.5.0.
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9544/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11645/)

mabelzhang:
Updating `pybind11_vendor` to 2.5.0 helped with the compiler warnings above, so I made a PR in `pybind11_vendor` to update the version. @jacobperron could you take a look?
https://github.com/ros2/pybind11_vendor/pull/2

Windows had trouble importing `_reader` in `__init__.py`. Added a similar wrapper to [this file](https://github.com/ros2/geometry2/blob/e813f131e5663133b5db956fae1f413b6290a303/tf2_py/tf2_py/__init__.py#L38-L41) to import modules explicitly on Windows. Retrying:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11647/)

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,72 @@
...
+    msg_counter = 0
+
+    while reader.has_next():
+        (topic, data, t) = reader.read_next()
+        msg_type = get_message(type_map[topic])
+        msg = deserialize_message(data, msg_type)
+
+        assert isinstance(msg, String)
+        assert msg.data == 'Hello, world! %d' % msg_counter
```
jacobperron, line 51: nit: In general, prefer [f-strings](https://realpython.com/python-f-strings/), which are less error-prone and more versatile. 
```suggestion
        assert msg.data == f'Hello, world! {msg_counter}'
```

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,72 @@
...
+
+    while reader.has_next():
+        (topic, data, t) = reader.read_next()
+        msg_type = get_message(type_map[topic])
+        msg = deserialize_message(data, msg_type)
+
+        assert isinstance(msg, Log) or isinstance(msg, String)
+
+        if isinstance(msg, String):
+            assert msg.data == 'Hello, world! %d' % msg_counter
```
jacobperron, line 71:
```suggestion
            assert msg.data == f'Hello, world! {msg_counter}'
```

rosbag2_py/test/test_sequential_writer.py
```diff
@@ -0,0 +1,87 @@
...
+    type_map = {topic_types[i].name: topic_types[i].type for i in range(len(topic_types))}
+
+    msg_counter = 0
+    while reader.has_next():
+        topic, data, t = reader.read_next()
+        msg_type = get_message(type_map[topic])
+        msg_deserialized = deserialize_message(data, msg_type)
+
+        assert isinstance(msg_deserialized, String)
+        assert msg_deserialized.data == 'Hello, world! %d' % msg_counter
```
jacobperron, line 84:
```suggestion
        assert msg_deserialized.data == f'Hello, world! {msg_counter}'
```

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,72 @@
...
+    reader.reset_filter()
+
+    reader = rosbag2_py.SequentialReader()
+    reader.open(storage_options, converter_options)
+
+    msg_counter = 0
+
+    while reader.has_next():
+        (topic, data, t) = reader.read_next()
+        msg_type = get_message(type_map[topic])
```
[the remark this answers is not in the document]
mabelzhang: Issue added #476

rosbag2_py/CMakeLists.txt
emersonknapp: can this be changed/removed now that we have moved back to the vendor package?

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -0,0 +1,55 @@
...
+  .def(pybind11::init())
+  .def_readwrite(
+    "input_serialization_format",
+    &rosbag2_cpp::ConverterOptions::input_serialization_format)
+  .def_readwrite(
+    "output_serialization_format",
+    &rosbag2_cpp::ConverterOptions::output_serialization_format);
+
+  pybind11::class_<rosbag2_cpp::StorageOptions>(m, "StorageOptions")
+  .def(pybind11::init())
```
emersonknapp, line 36: Are we able to bind this constructor to a "brace initializer" (https://pybind11.readthedocs.io/en/stable/advanced/classes.html#brace-initialization)? If not, can we define factory functions here in the python wrapper? I'd like for the py-side objects to be single-call constructable: e.g. instead of 

```
   storage_options = rosbag2_py.StorageOptions()
   storage_options.uri = path
   storage_options.storage_id = 'sqlite3'
```

the API would be better if we can

```
storage_options = rosbag2_py.StorageOptions(uri=path, storage_id='sqlite3')
```

This comment/question applies to all the POD structs exposed in this PR (ConverterOptions, StorageOptions, StorageFilter, TopicMetadata i think is all of them)

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,72 @@
...
+    # Set filter for topic of string type
+    storage_filter = rosbag2_py.StorageFilter()
+    storage_filter.topics = ['/topic']
+    reader.set_filter(storage_filter)
+
+    msg_counter = 0
+
+    while reader.has_next():
+        (topic, data, t) = reader.read_next()
+        msg_type = get_message(type_map[topic])
```
chaoflow, line 46: Do we need to create a new msg_type / call get_message each time or could the type_map already map to msg_type?

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pathlib import Path
+
+from common import get_rosbag_options
```
chaoflow, line 17:
```suggestion
from .common import get_rosbag_options
```

rosbag2_py/test/common.py
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rosbag2_py
+
+
+def get_rosbag_options(path, serialization_format='cdr'):
```
chaoflow, line 18: ``metadata.yaml`` and sqlite table for topics lists a serialization_format per topic. Does this mean we need one reader per serialization_format or is the some way to achieve this with one reader?

Is this a real-world scenario to have different serialization formats within one rosbag2?

rosbag2_py/test/test_sequential_reader.py
jacobperron: I had a similar comment about improving this API: https://github.com/ros2/rosbag2/pull/308#discussion_r462673026
I suggested to do this as a follow-up https://github.com/ros2/rosbag2/issues/476

rosbag2_py/test/common.py
jacobperron: IIUC, the reader will convert messages with different serialization format to the format provided by the options: https://github.com/ros2/rosbag2/blob/160af0fc73261bc7d7105f4f85cfd034bed15f44/rosbag2_cpp/include/rosbag2_cpp/reader.hpp#L81-L84

rosbag2_py/test/test_sequential_reader.py
mabelzhang: That was originally the way, but it did not work for the pytests when run using `colcon test`, so I changed it to `common`.
```
[0.808s] 1: _______________ ERROR collecting test/test_sequential_reader.py ________________
[0.808s] 1: ImportError while importing test module '/home/developer/viper/colcon_ws/src/ros2/rosbag2/rosbag2_py/test/test_sequential_reader.py'.
[0.808s] 1: Hint: make sure your test modules/packages have valid Python names.
[0.809s] 1: Traceback:
[0.809s] 1: ../test/test_sequential_reader.py:17: in <module>
[0.809s] 1:     from .common import get_rosbag_options
[0.809s] 1: E   ImportError: attempted relative import with no known parent package
```

rosbag2_py/CMakeLists.txt
mabelzhang: These are both required as in https://github.com/ros2/rosbag2/pull/308#discussion_r455455913
Windows CI looks fine with the change in 133e1a4, so I think this is good the way it is.

rosbag2_py/test/common.py
chaoflow: @jacobperron I was/am thrown of by the converter_options.input_serialization_format set below.

rosbag2_py/test/test_sequential_reader.py
chaoflow: @mabelzhang This might be fixed by adding an empty ``__init__.py`` to the same directory to turn it into a package.

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -0,0 +1,71 @@
...
+    type_map = {topic_types[i].name: topic_types[i].type for i in range(len(topic_types))}
+
+    # Set filter for topic of string type
+    storage_filter = rosbag2_py.StorageFilter()
+    storage_filter.topics = ['/topic']
+    reader.set_filter(storage_filter)
+
+    msg_counter = 0
+
+    while reader.has_next():
```
chaoflow, line 44: When doing exactly this I'm getting the following error:

``
RuntimeError: Error when preparing SQL statement 'SELECT data, timestamp, topics.name FROM messages JOIN topics ON messages.topic_id = topics.id ORDER BY messages.timestamp;'. SQLite error: (1): SQL logic error
``

Do you have any idea what could be causing this?

As this code is not yet mainline, I'm not sure where to put the question, happy to create a follow-up issue.

rosbag2_py/test/common.py
jacobperron: I think `input_serialization_format` is used by the writer; the writer will convert messages to this format when storing them.

This made me realize that in C++ there's an alternative of **not** providing storage and converter options for both readers and writers. If no `input_serialization_format` is provided, then it will use the format of the incoming messages. Similarly, if no `output_serialization_format` is provided, then messages are read in whatever format they were stored as.

For the both the readers and writers, if we want to support *not* specifying the serialization format in Python, then we should expose these overloaded `open` functions:

https://github.com/ros2/rosbag2/blob/160af0fc73261bc7d7105f4f85cfd034bed15f44/rosbag2_cpp/include/rosbag2_cpp/reader.hpp#L74

https://github.com/ros2/rosbag2/blob/160af0fc73261bc7d7105f4f85cfd034bed15f44/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L72

If this sounds like a feature we want in the Python API, it might be better to do it in a follow-up PR to avoid adding more on to this one.

rosbag2_py/test/test_sequential_reader.py
jacobperron: The error sounds unrelated to this PR, since it is just adding a wrapper around the C++ API. It could be an issue accessing the DB. Are you able to share code and steps that reproduces the error?

mabelzhang: The CI tests are passing on all OSes on this PR, so the error shouldn't happen. Do you have the latest version of `rosbag2`? Updating and recompiling everything `--packages-select-regex rosbag2* or just `--packages-select rosbag2` might fix it.

rosbag2_py/test/test_sequential_reader.py
mabelzhang: I'm not sure if that would work. That ended up confusing the package search, maybe because there are two `__init__.py` files in the package:
```
[0.614s] 1: ImportError while importing test module '/home/developer/viper/colcon_ws/src/ros2/rosbag2/rosbag2_py/test/test_sequential_reader.py'.
[0.614s] 1: Hint: make sure your test modules/packages have valid Python names.
[0.614s] 1: Traceback:
[0.614s] 1: ../test/test_sequential_reader.py:17: in <module>
[0.614s] 1:     from .common import get_rosbag_options
[0.614s] 1: ../test/common.py:15: in <module>
[0.614s] 1:     import rosbag2_py
[0.614s] 1: ../rosbag2_py/__init__.py:21: in <module>
[0.614s] 1:     from rosbag2_py._reader import \
[0.614s] 1: E   ModuleNotFoundError: No module named 'rosbag2_py._reader'
```

mabelzhang:
I forgot to paste the latest CI from yesterday
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11721/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6806/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9561/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11660/)

Mac has a test failure I'm looking into. Everything else is green.

rosbag2_py/src/rosbag2_py/_storage.cpp
mabelzhang: I tried out the brace initialization and found that it doesn't let us have keyword arguments. Some of the structs have default values like 0, and currently with the pybind brace initialization alone, I have to initialize it like this, which is kind of dumb:
```
storage_options = rosbag2_py.StorageOptions(path, 'sqlite3', 0, 0, 0)
```
The three 0s are max_bagfile_size, max_bagfile_duration, max_cache_size, which are by default 0s in C++. With the argument types specified in `<>` in the brace initialization, they aren't optional.

I think the keyword arguments might do what we want: https://pybind11.readthedocs.io/en/stable/advanced/functions.html#accepting-args-and-kwargs
I can look into this next week.

rosbag2_py/test/test_sequential_reader.py
chaoflow: The issue is caused by a version 1 rosbag2. Here in the tests a version 4 is used and it is sufficient to provide the path to the directory as ``storage_options.uri``; whereas for a version 1, the path to the db3 file needs to be provided.

By setting the version of the talker bag to 1, 2, 3 I found that:
- version 1, 2, and 3 need the path to the db3 file instead of the directory
- version 4 is fine with the directory
- version 1 produces a misleading error message ``SQL logic error``
- version 2 and 3 produce a bit more helpful messages ``RuntimeError: No storage could be initialized. Abort``

I don't think this needs to be tackled as part of this PR, but think that reader should transparently handle these and not require application logic. Currently, we use:

```python
        # TODO: workaround for older rosbag2 format versions
        if meta['version'] < 4:
            assert len(meta['relative_file_paths']) == 1
            uri = str(path.parent / meta['relative_file_paths'][0])
        else:
            uri = str(path.parent)
```

Code will soon be public. What do you think? Shall I open a separate issue for this?

rosbag2_py/test/test_sequential_reader.py
chaoflow: @mabelzhang This seems to be some peculiarity of colcon test, let's leave it the way it works then.

rosbag2_py/test/test_sequential_reader.py
jacobperron: I agree that the reader should handle this case without the application having to worry about it. Creating a separate issue sounds good to me.

mabelzhang:
The macOS test failure might be the same as #485 with fix in #487
```

>       reader.open(storage_options, converter_options)
E       RuntimeError: No storage could be initialized. Abort

../test/test_sequential_reader.py:30: RuntimeError
----------------------------- Captured stderr call -----------------------------
[ERROR] [1596250404.877517236] [rosbag2_storage]: Could not open '/Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/mabelzhang/rosbag2/rosbag2_py/resources/talker/talker.db3' with 'sqlite3'. Error: Failed to setup storage. Error: Error when processing SQL statement. SQLite error (14): unable to open database file
[ERROR] [1596250404.877601977] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
- generated xml file: /Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/rosbag2_py/test_results/rosbag2_py/test_sequential_reader_py.xunit.xml -
```

rosbag2_py/test/test_sequential_reader.py
mabelzhang: The version number is actually dealt with in `rosbag2_cpp::SequentialReader::open()`, which is an existing infrastructure the Python API just calls. Do you get this behavior only with Python, or with C++ as well? How about the command line `ros2 bag play`?

This line I believe takes care of the problem you are describing https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L95

The version is read into a field in the struct `rosbag2_storage::BagMetadata`, which can be retrieve at the reader level with `rosbag2_cpp::BaseReaderInterface::get_metadata()`. https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L36

There are actually some tests in `rosbag2_cpp` and `rosbag2_tests` that specifically test for different versions as well.

It would be worth checking whether you are getting the error separate from the Python API, to isolate whether the problem is in `rosbag2_py` or elsewhere in the existing infrastructure.

rosbag2_py/src/rosbag2_py/_storage.cpp
emersonknapp: Maybe we can do "named arguments" with defaults like the following? This may need us to define constructors or factory functions, which is maybe not ideal.

```
someclass
    .def(pybind11::init<long, double, double>(),
            pybind11::arg("arg1")=0, pybind11::arg("arg2")=0., pybind11::arg("arg3")=1.
    )
```

If it gets too difficult, maybe we do end up keeping it as a "default then set" struct - I don't love it as an API but it may be more maintainable

chaoflow: btw: It is sufficient to pass converter options without anything set:

```python
        storage_options = rosbag2.StorageOptions()
        storage_options.uri = uri
        storage_options.storage_id = 'sqlite3'
        reader.open(storage_options, rosbag2.ConverterOptions())
```

rosbag2_py/test/test_sequential_reader.py
chaoflow: @mabelzhang I currently don't have a C++ setup to test with. It is however trivial to test by simply adjusting the version set in ``talker/matadata.yaml``. My understanding is, that the additional keys of newer versions are ignored by older versions.

mabelzhang: Hmm I don't know all the nuances of the different versions to reproduce this. I believe the metadata files for different versions look a little different, for example, the qos profiles aren't there for the older versions. I can open a ticket now with limited information, but it would not be too helpful if I cannot describe the exact way to reproduce it, and whether it's a Python or C++ problem. I can look into this a little later and create a more informative issue.

rosbag2_py/src/rosbag2_py/_storage.cpp
mabelzhang: @emersonknapp Nice find! That's a lot simpler than the `kwargs` and works. 0603d27

mabelzhang:
I applied the current fix in #487 to the latest code here, but in another branch, to check CI.
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9602/)

All the tests are passing, but the build is failing with weird messages about the test XML file:
```
00:56:20 INFO: [JUnit] - 10 test report file(s) were found with the pattern 'ws/build/*/test_results/**/*.xunit.xml' relative to '/Users/osrf/jenkins-agent/workspace/ci_osx' for the testing framework 'JUnit'.
00:56:20 WARNING: The file '/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/rosbag2_py/test_results/rosbag2_py/test_sequential_reader_py.xunit.xml' is an invalid file.
00:56:20 WARNING: At line 1 of file:/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/rosbag2_py/test_results/rosbag2_py/test_sequential_reader_py.xunit.xml:cvc-complex-type.3.2.2: Attribute 'file' is not allowed to appear in element 'testcase'.
00:56:20 WARNING: At line 1 of file:/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/rosbag2_py/test_results/rosbag2_py/test_sequential_reader_py.xunit.xml:cvc-complex-type.3.2.2: Attribute 'line' is not allowed to appear in element 'testcase'.
00:56:20 ERROR: Step ‘Publish xUnit test result report’ failed: The result file '/Users/osrf/jenkins-agent/workspace/ci_osx/ws/build/rosbag2_py/test_results/rosbag2_py/test_sequential_reader_py.xunit.xml' for the metric 'JUnit' is not valid. The result file has been skipped.
00:56:20 Finished: FAILURE
```

I merged with `master` and still get this.

clalancette:
> All the tests are passing, but the build is failing with weird messages about the test XML file:

Yeah, that's because I merged https://github.com/ros2/ci/pull/497 last night.  Any package that uses pytest now has to have a pytest.ini in the package as well.  Looking here, it looks like rosbag2_py fits this description.  It should be enough to add a new file to the top-level of the package (a peer of package.xml) called `pytest.ini`, with the following contents:

```
[pytest]
junit_family=xunit2
```

mabelzhang:
Adding `pytest.ini` made Mac CI go green with #487 in another branch.
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9621/)

Once #487 is fixed, we can merge with that fix and run CI on all platforms again on this PR.

mabelzhang:
Adding .db3-shm and .db3-wal files for the bag file made macOS pass:
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9686/)

mabelzhang:
Rerun CI on all platforms:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11893/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6934/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9688/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11826/)

mabelzhang:
I think this is ready if I can get another approval.

mabelzhang:
@emersonknapp Could you take a last pass on this PR? We need one more approval before merging. Thanks!

nuclearsandwich:
I think this change has introduced a build failure in the nightly clang	job: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/513

```
00:47:40.956 In file included from /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_py/src/rosbag2_py/_writer.cpp:15:
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/install/pybind11_vendor/include/pybind11/pybind11.h:1093:9: error: expression result unused [-Werror,-Wunused-value]
00:47:40.956         PYBIND11_EXPAND_SIDE_EFFECTS(add_base<options>(record));
00:47:40.956         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/install/pybind11_vendor/include/pybind11/detail/common.h:662:47: note: expanded from macro 'PYBIND11_EXPAND_SIDE_EFFECTS'
00:47:40.956 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
00:47:40.956                                               ^                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_py/src/rosbag2_py/_writer.cpp:84:3: note: in instantiation of function template specialization 'pybind11::class_<rosbag2_py::Writer<rosbag2_cpp::writers::SequentialWriter>>::class_<>' requested here
00:47:40.956   pybind11::class_<rosbag2_py::Writer<rosbag2_cpp::writers::SequentialWriter>>(
00:47:40.956   ^
00:47:40.956 In file included from /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_py/src/rosbag2_py/_writer.cpp:15:
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/install/pybind11_vendor/include/pybind11/pybind11.h:1093:9: error: expression result unused [-Werror,-Wunused-value]
00:47:40.956         PYBIND11_EXPAND_SIDE_EFFECTS(add_base<options>(record));
00:47:40.956         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/install/pybind11_vendor/include/pybind11/detail/common.h:662:47: note: expanded from macro 'PYBIND11_EXPAND_SIDE_EFFECTS'
00:47:40.956 #define PYBIND11_EXPAND_SIDE_EFFECTS(PATTERN) pybind11::detail::expand_side_effects{ ((PATTERN), void(), false)..., false }
00:47:40.956                                               ^                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
00:47:40.956 /home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_py/src/rosbag2_py/_writer.cpp:92:3: note: in instantiation of function template specialization 'pybind11::class_<rosbag2_py::Writer<rosbag2_compression::SequentialCompressionWriter>>::class_<>' requested here
00:47:40.956   pybind11::class_<rosbag2_py::Writer<rosbag2_compression::SequentialCompressionWriter>>(
00:47:40.956   ^
00:47:40.956 2 errors generated.
00:47:40.956 make[2]: *** [CMakeFiles/_writer.dir/build.make:63: CMakeFiles/_writer.dir/src/rosbag2_py/_writer.cpp.o] Error 1
00:47:40.956 make[1]: *** [CMakeFiles/Makefile2:82: CMakeFiles/_writer.dir/all] Error 2
00:47:40.956 make: *** [Makefile:141: all] Error 2
```

nuclearsandwich:
In addition to the compilation failure mentioned above there also appear to be issues on Windows Debug which I've opened https://github.com/ros2/rosbag2/issues/504 to track.

evclark:
Noob question, but how do I actually use rosbag2_py? I have installed it as per the instructions on the rosbag2 README, 


```
$ export CHOOSE_ROS_DISTRO=crystal # rosbag2 is available starting from crystal
$ sudo apt-get install ros-$CHOOSE_ROS_DISTRO-ros2bag ros-$CHOOSE_ROS_DISTRO-rosbag2-transport
```

but I do not see it in ros2 pkg list and when I try to import it in python I get an error `ModuleNotFoundError: No module named 'rosbag2_py'`

Running on Ubuntu 20.04 on Foxy.

Thanks in advance!

mabelzhang:
Hi @evclark, for future reference, please post questions as a new Issue or on ROS Answers, as opposed to commenting on a merged PR.

First, you are using Foxy, but you installed the packages for crystal? That does not look right. Your distro should be `foxy`.
Second, you should install `ros-$CHOOSE_ROS_DISTRO-rosbag2`, which is the meta package. The two packages that you did install are only two of the packages, not the full stack.

Since this PR was merged very recently, and we had just fixed some issues and re-enabled rosbag2_py 21 days ago, I don't know if it's included in `ros-foxy-rosbag2` already. Your best bet is probably to build this repository from source.

evclark:
> Hi @evclark, for future reference, please post questions as a new Issue or on ROS Answers, as opposed to commenting on a merged PR.

Noted, will do, thanks

> First, you are using Foxy, but you installed the packages for crystal? That does not look right. Your distro should be foxy.

Ah yes, sorry, copy/paste error from the README, I did install the foxy, not crystal, packages.

> Since this PR was merged very recently, and we had just fixed some issues and re-enabled rosbag2_py 21 days ago, I don't know if it's included in ros-foxy-rosbag2 already. Your best bet is probably to build this repository from source.

It appears not to be. I'll try building from source, thanks!

jacobperron:
FWIW, I don't expect this feature to land in Foxy. While we wait for Galactic to be released, you can build from source. Once a new release is made into Rolling, you should be able to install the binary.

agutenkunst:
> FWIW, I don't expect this feature to land in Foxy. While we wait for Galactic to be released, you can build from source. Once a new release is made into Rolling, you should be able to install the binary.

Building from source including rosbag2_py on foxy and not against Rolling worked for me with
```
git clone git@github.com:ros2/rosbag2.git -b 0.3.5 && cd rosbag2 && git merge 038bae248e2f97e03010da7765bfbe9da0d8e0b9 -X theirs --no-edit
```
Only small unresolvable conflict is inside a workflow file.

Is there a better solution? I think lot of people would be happy having this feature in foxy!

Disclaimer: I did not try to read a bagfile yet.

jacobperron:
In retrospect, it may be easy to backport this new package to Foxy (assuming it's not relying on anything new in Galactic).

Pinging maintainers to see if there is any opposition, @mjeronimo @ros2/tooling-approvers

emersonknapp:
It may be easy, but the APIs that it binds to may have changed. I'm all for someone trying it out, no opposition from me - I'm just not 100% confident it will be easy

PR #168, opened 2019-10-01
zero copy api

Karsten1987:
Connects to: ros2/ros2#785

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

On a tangential related note: The GenericSubscription should actually be migrated over to rclcpp

PR #310, opened 2020-03-04
added rosidl_runtime c and cpp depencencies

ahcorde:
This PR is related to the changes introduced in this PR https://github.com/ros2/rosidl/pull/442. The full process can be followed here https://github.com/ros2/rosidl/issues/443

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rosbag2_cpp/package.xml
+++ b/rosbag2_cpp/package.xml
@@ -9,16 +9,13 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <build_depend>rosidl_runtime_cpp</build_depend>
-
-  <build_export_depend>rosidl_runtime_cpp</build_export_depend>
-
   <depend>ament_index_cpp</depend>
   <depend>pluginlib</depend>
   <depend>poco_vendor</depend>
   <depend>rcutils</depend>
   <depend>rosbag2_storage</depend>
   <depend>rosidl_runtime_c</depend>
+  <depend>rosidl_runtime_cpp</depend>
   <depend>rosidl_typesupport_cpp</depend>
   <depend>rosidl_typesupport_introspection_cpp</depend>
   <depend>shared_queues_vendor</depend>
--- a/rosbag2_cpp/package.xml
+++ b/rosbag2_cpp/package.xml
@@ -28,7 +28,6 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>test_msgs</test_depend>
   <test_depend>rosbag2_test_common</test_depend>
-  <test_depend>rosidl_runtime_cpp</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rosbag2_cpp/package.xml
+++ b/rosbag2_cpp/package.xml
@@ -9,13 +9,16 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>rosidl_runtime_cpp</build_depend>
+
+  <build_export_depend>rosidl_runtime_cpp</build_export_depend>
+
   <depend>ament_index_cpp</depend>
   <depend>pluginlib</depend>
   <depend>poco_vendor</depend>
   <depend>rcutils</depend>
   <depend>rosbag2_storage</depend>
   <depend>rosidl_runtime_c</depend>
-  <depend>rosidl_runtime_cpp</depend>
   <depend>rosidl_typesupport_cpp</depend>
   <depend>rosidl_typesupport_introspection_cpp</depend>
   <depend>shared_queues_vendor</depend>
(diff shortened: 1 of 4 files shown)


rosbag2_cpp/package.xml
```diff
@@ -19,6 +22,9 @@
   <depend>rosidl_typesupport_introspection_cpp</depend>
   <depend>shared_queues_vendor</depend>
 
+  <exec_depend>rosidl_runtime_c</exec_depend>
+  <exec_depend>rosidl_runtime_cpp</exec_depend>
```
dirk-thomas: Since the new dependencies are added as both they can use a `depend` tag.

ahcorde: `rosidl_runtime_cpp` is not generating any `.so/dll/dylib` should not be a `<depend>`, right? only `<build_depend>` and `<build_export_depend>` ?

ahcorde:
By the way, I'm not really sure why we should include this dependency here. rosidl_typesupport_cpp ( which is a dependency [here]( https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/src/rosbag2_cpp/types/introspection_message.cpp#L25) ) export in the [CMakeLists.txt](https://github.com/ros2/rosidl_typesupport/blob/ahcorde/rosidl_runtime/rosidl_typesupport_cpp/CMakeLists.txt#L23) rosidl_runtime_c/cpp but this package is not using explicitly the dependency. In this case can we remove it?

wjwwood:
> By the way, I'm not really sure why we should include this dependency here.

Just quickly searching the repo I don't see `rosidl_generator_cpp` either, could be that it should be removed here and just brought in as a transitive dependency implicitly.

Karsten1987:
I could imagine it's a leftover from this change here: https://github.com/ros2/rosbag2/pull/168/commits/b4c459a2b1227e9ba21e3fa2d17895e8854eef60

If the dependency is not being used in this repo, we should remove it.

wjwwood:
:+1: to removing it or leaving it btw, I'll leave that up to you @ahcorde and @Karsten1987.

rosbag2_cpp/package.xml
ahcorde: new changes applied: 

`rosidl_runtime_c` moved to `<depend>`.

Headers are used here https://github.com/ros2/rosbag2/blob/ec4fb3913233df00850b53a4422d9b804fbab9d7/rosbag2_cpp/include/rosbag2_cpp/converter_interfaces/serialization_format_serializer.hpp#L24

-----

`rosidl_runtime_cpp` moved to `<build_depend>` and `<build_export_depend>`

https://github.com/ros2/rosbag2/blob/ec4fb3913233df00850b53a4422d9b804fbab9d7/rosbag2_cpp/include/rosbag2_cpp/typesupport_helpers.hpp#L24

ahcorde:
# rosbag2_converter_default_plugins

this package use `rosidl_generator_cpp` headers located now in `rosidl_runtime_cpp`

https://github.com/ros2/rosbag2/blob/ec4fb3913233df00850b53a4422d9b804fbab9d7/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.hpp#L23

No need to ament_export_dependency the package or `<exec_depend>` (`rosidl_runtime_cpp` is not generating any .so/dll/dylib)

ahcorde:
I have created this other PR https://github.com/ros2/rosbag2/pull/321 (because the change is not related with this PR) Removing `rosidl_generator_cpp` in `rosbag2_transport` because it's not been used.

rosbag2_cpp/package.xml
```diff
@@ -24,6 +28,7 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>test_msgs</test_depend>
   <test_depend>rosbag2_test_common</test_depend>
+  <test_depend>rosidl_runtime_cpp</test_depend>
```
dirk-thomas: This doesn't seem necessary since this is already a build and run dependency.

ahcorde:
- Updated all repos involved in this change. 
 - Skipping `ros1_bridge`, `qt_*`, `rqt_*`, and `rviz_*` packages. 
 - Testing against `test_rclcpp` and `rosbag2_tests`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10031/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5626/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9938/)

 - One warning made the builds unstable but there is a ([PR pending](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/163))

```
CMake Deprecation Warning at /home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/ament_cmake_export_interfaces/share/ament_cmake_export_interfaces/cmake/ament_export_interfaces.cmake:37 (message):
ament_export_interfaces() is deprecated, use ament_export_targets() instead
Call Stack (most recent call first):
CMakeLists.txt:85 (ament_export_interfaces)
``` 

In MacOS
 - projectroot.cppcheck

Windows
 - some failures in test_rclcpp

PR #322, opened 2020-03-13
Replace poco dependency by rcutils

ahcorde:
As part of the effort to remove Poco dependency described in this issue https://github.com/ros2/ros2/issues/867 this PR makes use of the recent changes in rcutils https://github.com/ros2/rcutils/pull/215 to load, unload and get symbols from shared libraries instead of using Poco.

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/rosbag2_converter_default_plugins/CMakeLists.txt
+++ b/rosbag2_converter_default_plugins/CMakeLists.txt
@@ -22,10 +22,8 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_index_cpp REQUIRED)
-find_package(poco_vendor REQUIRED)
-find_package(Poco COMPONENTS Foundation)
 find_package(pluginlib REQUIRED)
-find_package(rcutils REQUIRED)
+find_package(rcpputils REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_fastrtps_cpp QUIET)
 find_package(rosbag2_cpp REQUIRED)
@@ -38,8 +36,7 @@ if(rmw_fastrtps_cpp_FOUND)
   ament_target_dependencies(${PROJECT_NAME}
     ament_index_cpp
     pluginlib
-    Poco
-    rcutils
+    rcpputils
     rmw
     rosbag2_cpp
     rosidl_generator_cpp)
--- a/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
+++ b/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
@@ -67,6 +67,8 @@ CdrConverter::CdrConverter()
   } catch (std::runtime_error &) {
     throw std::runtime_error(
             std::string("poco exception: library could not be found:") + library_path);
+  } catch (std::bad_alloc &) {
+    throw std::bad_alloc();
   }
 
   std::string serialize_symbol = "rmw_serialize";
--- a/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
+++ b/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
@@ -67,8 +67,6 @@ CdrConverter::CdrConverter()
   } catch (std::runtime_error &) {
     throw std::runtime_error(
             std::string("poco exception: library could not be found:") + library_path);
-  } catch (std::bad_alloc &) {
-    throw std::bad_alloc();
   }
 
   std::string serialize_symbol = "rmw_serialize";
--- a/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
+++ b/rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
@@ -20,8 +20,7 @@
 #include "ament_index_cpp/get_resources.hpp"
 #include "ament_index_cpp/get_package_prefix.hpp"
 
-#include "Poco/SharedLibrary.h"
-
+#include "rcpputils/shared_library.hpp"
 #include "rcutils/strdup.h"
 
 #include "rosbag2_cpp/types.hpp"
@@ -63,10 +62,9 @@ std::string get_package_library_path(const std::string & package_name)
 CdrConverter::CdrConverter()
 {
   auto library_path = get_package_library_path("rmw_fastrtps_cpp");
-  std::shared_ptr<Poco::SharedLibrary> library;
   try {
-    library = std::make_shared<Poco::SharedLibrary>(library_path);
-  } catch (Poco::LibraryLoadException &) {
+    library = std::make_shared<rcpputils::SharedLibrary>(library_path);
+  } catch (std::runtime_error &) {
     throw std::runtime_error(
             std::string("poco exception: library could not be found:") + library_path);
   }
@@ -74,26 +72,26 @@ CdrConverter::CdrConverter()
   std::string serialize_symbol = "rmw_serialize";
   std::string deserialize_symbol = "rmw_deserialize";
 
-  if (!library->hasSymbol(serialize_symbol)) {
-    throw std::runtime_error(
-            std::string("poco exception: symbol not found: ") + serialize_symbol);
+  if (!library->has_symbol(serialize_symbol.c_str())) {
+    throw std::runtime_error{std::string("rcutils exception: symbol not found: ") +
+            serialize_symbol};
   }
 
-  if (!library->hasSymbol(deserialize_symbol)) {
-    throw std::runtime_error(
-            std::string("poco exception: symbol not found: ") + deserialize_symbol);
+  if (!library->has_symbol(deserialize_symbol.c_str())) {
+    throw std::runtime_error{
+            std::string("rcutils exception: symbol not found: ") + deserialize_symbol};
   }
 
-  serialize_fcn_ = (decltype(serialize_fcn_))library->getSymbol(serialize_symbol);
+  serialize_fcn_ = (decltype(serialize_fcn_))library->get_symbol(serialize_symbol.c_str());
   if (!serialize_fcn_) {
-    throw std::runtime_error(
-            std::string("poco exception: symbol of wrong type: ") + serialize_symbol);
+    throw std::runtime_error{
+            std::string("rcutils exception: symbol of wrong type: ") + serialize_symbol};
   }
 
-  deserialize_fcn_ = (decltype(deserialize_fcn_))library->getSymbol(deserialize_symbol);
+  deserialize_fcn_ = (decltype(deserialize_fcn_))library->get_symbol(deserialize_symbol.c_str());
   if (!deserialize_fcn_) {
-    throw std::runtime_error(
-            std::string("poco exception: symbol of wrong type: ") + deserialize_symbol);
+    throw std::runtime_error{
+            std::string("rcutils exception: symbol of wrong type: ") + deserialize_symbol};
   }
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
@@ -139,6 +139,8 @@ get_typesupport(
     return type_support;
   } catch (std::runtime_error &) {
     throw std::runtime_error(rcutils_dynamic_loading_error.str() + " Library could not be found.");
+  } catch (std::bad_alloc &) {
+    throw std::bad_alloc();
   }
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
@@ -139,8 +139,6 @@ get_typesupport(
     return type_support;
   } catch (std::runtime_error &) {
     throw std::runtime_error(rcutils_dynamic_loading_error.str() + " Library could not be found.");
-  } catch (std::bad_alloc &) {
-    throw std::bad_alloc();
   }
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
@@ -15,6 +15,7 @@
 #include "rosbag2_cpp/typesupport_helpers.hpp"
 
 #include <memory>
+#include <sstream>
 #include <stdexcept>
 #include <string>
 #include <tuple>
@@ -23,7 +24,7 @@
 #include "ament_index_cpp/get_resources.hpp"
 #include "ament_index_cpp/get_package_prefix.hpp"
 
-#include "Poco/SharedLibrary.h"
+#include "rcpputils/shared_library.hpp"
 
 #include "rosidl_generator_cpp/message_type_support_decl.hpp"
 
@@ -98,39 +99,46 @@ extract_type_identifier(const std::string & full_type)
 }
 
 const rosidl_message_type_support_t *
-get_typesupport(const std::string & type, const std::string & typesupport_identifier)
+get_typesupport(
+  const std::string & type, const std::string & typesupport_identifier,
+  std::shared_ptr<rcpputils::SharedLibrary> & library)
 {
   std::string package_name;
   std::string middle_module;
   std::string type_name;
   std::tie(package_name, middle_module, type_name) = extract_type_identifier(type);
 
-  std::string poco_dynamic_loading_error = "Something went wrong loading the typesupport library "
-    "for message type " + package_name + "/" + type_name + ".";
+  std::stringstream rcutils_dynamic_loading_error;
+  rcutils_dynamic_loading_error <<
+    "Something went wrong loading the typesupport library for message type " << package_name <<
+    "/" << type_name << ".";
 
   auto library_path = get_typesupport_library_path(package_name, typesupport_identifier);
 
   try {
-    auto typesupport_library = std::make_shared<Poco::SharedLibrary>(library_path);
+    library = std::make_shared<rcpputils::SharedLibrary>(library_path);
 
     auto symbol_name = typesupport_identifier + "__get_message_type_support_handle__" +
       package_name + "__" + (middle_module.empty() ? "msg" : middle_module) + "__" + type_name;
 
-    if (!typesupport_library->hasSymbol(symbol_name)) {
-      throw std::runtime_error(poco_dynamic_loading_error + " Symbol not found.");
+    if (!library->get_symbol(symbol_name)) {
+      throw std::runtime_error{
+              rcutils_dynamic_loading_error.str() +
+              std::string(" Symbol not found.")};
     }
 
     const rosidl_message_type_support_t * (* get_ts)() = nullptr;
-    get_ts = (decltype(get_ts))typesupport_library->getSymbol(symbol_name);
-    auto type_support = get_ts();
+    get_ts = (decltype(get_ts))library->get_symbol(symbol_name);
 
-    if (!type_support) {
-      throw std::runtime_error(poco_dynamic_loading_error + " Symbol of wrong type.");
+    if (!get_ts) {
+      throw std::runtime_error{
+              rcutils_dynamic_loading_error.str() +
+              std::string(" Symbol of wrong type.")};
     }
-
+    auto type_support = get_ts();
     return type_support;
-  } catch (Poco::LibraryLoadException &) {
-    throw std::runtime_error(poco_dynamic_loading_error + " Library could not be found.");
+  } catch (std::runtime_error &) {
+    throw std::runtime_error(rcutils_dynamic_loading_error.str() + " Library could not be found.");
   }
 }
 
--- a/rosbag2_converter_default_plugins/package.xml
+++ b/rosbag2_converter_default_plugins/package.xml
@@ -11,9 +11,8 @@
 
   <depend>ament_index_cpp</depend>
   <depend>pluginlib</depend>
-  <depend>poco_vendor</depend>
   <depend>rosidl_generator_cpp</depend>
-  <depend>rcutils</depend>
+  <depend>rcpputils</depend>
   <depend>rmw</depend>
   <depend>rmw_fastrtps_cpp</depend>
   <depend>rosbag2_cpp</depend>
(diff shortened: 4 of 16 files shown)


piraka9011:
Have you properly linked `rcutils` in `rosbag2_cpp`'s `CMakeLists.txt`?

```
/home/runner/work/rosbag2/rosbag2/ros_ws/src/rosbag2/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp:26:10: fatal error: rcutils/shared_library.h: No such file or directory
   #include "rcutils/shared_library.h"
```

ahcorde:
@piraka9011 the PR is not merged yet https://github.com/ros2/rcutils/pull/215

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +105,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
...
-    auto typesupport_library = std::make_shared<Poco::SharedLibrary>(library_path);
-
-    auto symbol_name = typesupport_identifier + "__get_message_type_support_handle__" +
-      package_name + "__" + (middle_module.empty() ? "msg" : middle_module) + "__" + type_name;
-
-    if (!typesupport_library->hasSymbol(symbol_name)) {
-      throw std::runtime_error(poco_dynamic_loading_error + " Symbol not found.");
+  rcutils_shared_library_t * typesupport_library = nullptr;
+
+  typesupport_library = new rcutils_shared_library_t;
```
emersonknapp: We shouldn't be using `new` and `delete` in modern C++. I think you can use a `unique_ptr` just fine here, and then don't have to add the `delete` calls in the error cases.

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -63,37 +63,56 @@ std::string get_package_library_path(const std::string & package_name)
...
   auto library_path = get_package_library_path("rmw_fastrtps_cpp");
-  std::shared_ptr<Poco::SharedLibrary> library;
-  try {
-    library = std::make_shared<Poco::SharedLibrary>(library_path);
-  } catch (Poco::LibraryLoadException &) {
-    throw std::runtime_error(
-            std::string("poco exception: library could not be found:") + library_path);
+  rcutils_shared_library_t * library = nullptr;
+
+  library = new rcutils_shared_library_t;
```
emersonknapp: same comment here for smart ptrs

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -63,37 +63,56 @@ std::string get_package_library_path(const std::string & package_name)
...
+  *library = rcutils_get_zero_initialized_shared_library();
+
+  rcutils_ret_t ret = rcutils_load_shared_library(library, library_path.c_str());
+  if (ret != RCUTILS_RET_OK) {
+    delete library;
+    if (ret == RCUTILS_RET_BAD_ALLOC) {
+      throw std::runtime_error(
+              std::string("rcutils shared_library exception: failed to allocate memory"));
```
piraka9011:
```suggestion
      throw std::runtime_error{"rcutils shared_library exception: failed to allocate memory"};
```
ditto below.

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +105,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
   std::string type_name;
   std::tie(package_name, middle_module, type_name) = extract_type_identifier(type);
 
-  std::string poco_dynamic_loading_error = "Something went wrong loading the typesupport library "
+  std::string rcutils_dynamic_loading_error =
+    "Something went wrong loading the typesupport library "
     "for message type " + package_name + "/" + type_name + ".";
```
piraka9011:
```suggestion
  std::stringstream rcutils_dynamic_loading_error;
    rcutils_dynamic_loading_error << "Something went wrong loading the typesupport library for message type " <<  package_name <<  "/" << type_name << ".";
```

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
piraka9011: [Check this out](https://blog.quasardb.net/2017/03/05/cpp-braced-initialization)

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +105,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
...
+
+  typesupport_library = new rcutils_shared_library_t;
+  *typesupport_library = rcutils_get_zero_initialized_shared_library();
+
+  rcutils_ret_t ret = rcutils_load_shared_library(typesupport_library, library_path.c_str());
+  if (ret != RCUTILS_RET_OK) {
+    delete typesupport_library;
+    if (ret == RCUTILS_RET_BAD_ALLOC) {
+      throw std::runtime_error(
+              std::string(rcutils_dynamic_loading_error + " failed to allocate memory"));
```
piraka9011: Same suggestion as above with brace initialization.

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +105,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
...
+              std::string(
+                rcutils_dynamic_loading_error + " Library could not be found " +
+                library_path));
     }
+  }
 
-    const rosidl_message_type_support_t * (* get_ts)() = nullptr;
-    get_ts = (decltype(get_ts))typesupport_library->getSymbol(symbol_name);
-    auto type_support = get_ts();
+  auto symbol_name = typesupport_identifier + "__get_message_type_support_handle__" +
```
piraka9011: This looks complicated, can it be simplified?

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +105,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
...
-    if (!type_support) {
-      throw std::runtime_error(poco_dynamic_loading_error + " Symbol of wrong type.");
-    }
+  if (!rcutils_get_symbol(typesupport_library, symbol_name.c_str())) {
+    delete typesupport_library;
+    throw std::runtime_error(rcutils_dynamic_loading_error + " Symbol not found.");
+  }
+
+  const rosidl_message_type_support_t * (* get_ts)() = nullptr;
+  get_ts = (decltype(get_ts))rcutils_get_symbol(typesupport_library, symbol_name.c_str());
```
piraka9011: What is this trying to do?

emersonknapp: to be fair, this is identical to the preexisting code. It's declaring a function pointer and calling an rcl API to set it so that it can be called.

piraka9011: Agreed, though I was wondering if there's a better way to go about it since we're improving this anyway.

ahcorde: I will put some time on this and in a this memory leak:

```
6: =================================================================
6: ==20733==ERROR: LeakSanitizer: detected memory leaks
6: Direct leak of 100 byte(s) in 1 object(s) allocated from:
6:     #0 0x7f8b08620acb in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/liblsan.so.0+0xeacb)
6:     #1 0x7f8b07e623d0 in __default_allocate (/home/ahcorde/ros2_foxy/install/rcutils/lib/librcutils.so+0x43d0)
6:     #2 0x7f8b07e69cd0 in rcutils_load_shared_library (/home/ahcorde/ros2_foxy/install/rcutils/lib/librcutils.so+0xbcd0)
6:     #3 0x563ece336c48 in rosbag2_cpp::get_typesupport(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (/home/ahcorde/ros2_foxy/build/rosbag2_cpp/test_ros2_message+0x66c48)
6:     #4 0x563ece2f2133 in Ros2MessageTest::get_allocated_message(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (/home/ahcorde/ros2_foxy/build/rosbag2_cpp/test_ros2_message+0x22133)
6:     #5 0x563ece2ed467 in Ros2MessageTest_allocate_ros2_message_allocates_strings_message_with_empty_string_Test::TestBody() (/home/ahcorde/ros2_foxy/build/rosbag2_cpp/test_ros2_message+0x1d467)
6:     #6 0x563ece369143 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/ahcorde/ros2_foxy/build/rosbag2_cpp/test_ros2_message+0x99143)
6:     #7 0x563ece362eee in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/ahcorde/ros2_foxy/build/rosbag2_cpp/test_ros2_message+0x92eee)
 ```

Comment (ahcorde) [reply rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp]: The unique_ptr is instantiated inside the function `get_typesupport` 

https://github.com/ros2/rosbag2/blob/d2cdcc3a0fef91fbaefac6398fc1150a5e4e0cab/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp#L116-L117

but we can't free this memory because we need to use the symbol outside the function and the library needs to be open. That's why the leak is happening.

any thoughts about how can we workaround this ?

File: rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp [hunk 8]
```diff
@@ -63,37 +63,50 @@ std::string get_package_library_path(const std::string & package_name)
...
-  std::shared_ptr<Poco::SharedLibrary> library;
-  try {
-    library = std::make_shared<Poco::SharedLibrary>(library_path);
-  } catch (Poco::LibraryLoadException &) {
-    throw std::runtime_error(
-            std::string("poco exception: library could not be found:") + library_path);
+
+  std::unique_ptr<rcutils_shared_library_t> library =
+    std::make_unique<rcutils_shared_library_t>();
+  *library = rcutils_get_zero_initialized_shared_library();
```
piraka9011: Is it possible to move this to the constructor instead?

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
piraka9011: Interesting, maybe make this a `shared_ptr` then?

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
ahcorde: Maybe it makes sense to abstract this on rcpputils in a class to be able to create an destroy the object. What do you think @wjwwood ?

wjwwood: Move it to the constructor of which class?

You could add a companion api to rcpputils, but I don’t think that’s necessary unless we think lots of  places in the code will be using this in conjunction with cpp.

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
emersonknapp: The existing code was using a shared ptr that would have been freed at the end of that same function - did this leak already exist, or is this a difference in behavior between poco and the rcutils loader API?

rosbag2_converter_default_plugins/CMakeLists.txt
```diff
@@ -38,7 +37,7 @@ if(rmw_fastrtps_cpp_FOUND)
   ament_target_dependencies(${PROJECT_NAME}
     ament_index_cpp
     pluginlib
-    Poco
+    rcutils
```
Karsten1987: isn't `rcutils` listed twice here? Or is this just GitHub rendering it wrongly?

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
Karsten1987: Could that call be joined though?
```c++
auto library = std::make_unique<rcutils_shared_library_t>(rcutils_get_zero_initialized_shared_library());
```

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -63,37 +63,50 @@ std::string get_package_library_path(const std::string & package_name)
...
+  rcutils_ret_t ret = rcutils_load_shared_library(library.get(), library_path.c_str());
+  if (ret != RCUTILS_RET_OK) {
+    if (ret == RCUTILS_RET_BAD_ALLOC) {
+      throw std::runtime_error{"rcutils shared_library exception: failed to allocate memory"};
+    } else if (ret == RCUTILS_RET_INVALID_ARGUMENT) {
+      throw std::runtime_error{"rcutils shared_library exception: invalid arguments"};
+    } else {
+      throw std::runtime_error{"rcutils shared_library exception: library could not be found:" +
+              library_path};
+    }
```
Karsten1987: There is `rclcpp::throw_from_rcl_error`, maybe we can leverage this? Even though strictly speaking, `rcutils_ret_t` and `rcl_ret_t` might become different datatypes at some point.

```c++
RCLCPP_PUBLIC
void
throw_from_rcl_error [[noreturn]] (
  rcl_ret_t ret,
  const std::string & prefix = "",
  const rcl_error_state_t * error_state = nullptr,
  void (* reset_error)() = rcl_reset_error);
```

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +106,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
   std::string type_name;
   std::tie(package_name, middle_module, type_name) = extract_type_identifier(type);
 
-  std::string poco_dynamic_loading_error = "Something went wrong loading the typesupport library "
-    "for message type " + package_name + "/" + type_name + ".";
+  std::stringstream rcutils_dynamic_loading_error;
+  rcutils_dynamic_loading_error <<
+    "Something went wrong loading the typesupport library for message type " << package_name <<
+    "/" << type_name << ".";
```
Karsten1987: why is this change necessary?

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -105,33 +106,51 @@ get_typesupport(const std::string & type, const std::string & typesupport_identi
...
+                " failed to allocate memory")};
+    } else if (ret == RCUTILS_RET_INVALID_ARGUMENT) {
+      throw std::runtime_error{rcutils_dynamic_loading_error.str() + std::string(
+                " invalid arguments")};
+    } else {
+      throw std::runtime_error{
+              rcutils_dynamic_loading_error.str() + std::string(" Library could not be found ") +
+              library_path};
     }
+  }
```
Karsten1987: generally not a big fan of code duplication.

Can we factor this somewhat into a function within `typesupport_helpers.cpp` which then can also be used from the code snippet above?

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
ahcorde: It was a suggestion from @piraka9011 

https://github.com/ros2/rosbag2/pull/322#discussion_r393258319

piraka9011: @Karsten1987 It's a convention we've been using for constructing error strings in other packages.
No strong opinions but prefer a single convention.

ahcorde:
I have created an abstraction in rcpputils for rcutils shared library in this [branch](https://github.com/ahcorde/rcpputils/tree/ahcorde/feature/shared_library) and I made the changes in rosbag2 [here](https://github.com/ros2/rosbag2/tree/ahcorde/removed_poco_rcpputils). There is no leak with this solution.

@emersonknapp @piraka9011 @Karsten1987 @wjwwood  Can you have a look? Maybe we can decline this PR and I can open these two others which is more modern C++ oriented.

emersonknapp:
Took a quick look through it - over all i like that approach, I think it's a useful abstraction to have for C++. My personal vote is yes, we should split this out into these two PRs

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -20,7 +20,7 @@
 #include "ament_index_cpp/get_resources.hpp"
 #include "ament_index_cpp/get_package_prefix.hpp"
 
-#include "Poco/SharedLibrary.h"
+#include "rcutils/shared_library.h"
```
Karsten1987: nit: remove empty lines to cluster includes per package

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
Karsten1987: given that we instantiate a c-structure in the unique pointer, I would assume we'd need a custom deleter for the unique pointer which calls the appropriate `_fini` function for it.

I could imagine that the previous Poco instance had a properly defined destructor for itself in which case the default deleter of the unique pointer was sufficient.

wjwwood:
@ahcorde can you make https://github.com/ahcorde/rcpputils/commit/efc99631129947595bcdeee4923b38f5f5f9903b into a pull request? I have some comments but it's better if it is a pull request first.

ahcorde:
Sure @wjwwood https://github.com/ros2/rcpputils/pull/49

ahcorde:
rcpputils PR is merged, can you review this ? @emersonknapp @piraka9011 @Karsten1987

rosbag2_converter_default_plugins/src/rosbag2_converter_default_plugins/cdr/cdr_converter.cpp
```diff
@@ -63,37 +62,38 @@ std::string get_package_library_path(const std::string & package_name)
...
   auto library_path = get_package_library_path("rmw_fastrtps_cpp");
-  std::shared_ptr<Poco::SharedLibrary> library;
   try {
-    library = std::make_shared<Poco::SharedLibrary>(library_path);
-  } catch (Poco::LibraryLoadException &) {
+    library = std::make_shared<rcpputils::SharedLibrary>(library_path);
+  } catch (std::runtime_error &) {
     throw std::runtime_error(
             std::string("poco exception: library could not be found:") + library_path);
+  } catch (std::bad_alloc &) {
```
piraka9011, line 70: Why catch and just let it throw?

piraka9011:
This looks much cleaner than before, thanks @ahcorde 👍

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9912/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5528/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8064/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9763/)

Warnings in Windows are in `rmw_dds_common`

```
17:03:59 C:\ci\ws\src\ros2\rmw_dds_common\rmw_dds_common\test\test_graph_cache.cpp(195,8): warning C4996: 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\ci\ws\build\rmw_dds_common\test_graph_cache.vcxproj]

17:03:59 C:\ci\ws\src\ros2\rmw_dds_common\rmw_dds_common\test\test_graph_cache.cpp(388,8): warning C4996: 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details. [C:\ci\ws\build\rmw_dds_common\test_graph_cache.vcxproj]
```

ahcorde:
Feel free to merge it

dirk-thomas:
> Warnings in Windows are in `rmw_dds_common`

@ahcorde please make sure to identify the author of the causing change and notify them on that PR.

Karsten1987:
I just lost a day of my life to find out that this PR broke the `rosbag2_bag_v2_plugins`. Essentially, the converter functionality.

I haven't paid enough attention while reviewing this PR, but I think it's really bad practice to pass in a shared_pointer by reference (`std::shared_pointer<rcpputils::SharedLibrary> & library`) and  somewhat blindly overwrite it by calling `std::make_shared` on it. Also, this assumes that the pointer is empty, which isn't checked here either.

The reason why it fails in the end is that the same pointer is being passed in to load multiple typesupports, which are stored in `add_topics`. Later on, when using these typesupport symbols to create a new introspection message, these pointers have dangled.

We definitely have to come up with better tests for the converter.

wjwwood:
> but I think it's really bad practice to pass in a shared_pointer by reference (std::shared_pointer<rcpputils::SharedLibrary> & library) and somewhat blindly overwrite it by calling std::make_shared on it. 

So, I know there's some kind of taboo in our team about shared pointers and references, but what you describe is _exactly_ what the CppCoreGuidelines suggest you do if you might "reseat" (their term meaning point it to a different object) the shared pointer:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam

Karsten1987:
fair enough. I take that argument when you transfer ownership.

I still think though that in this case it's not necessarily applied correctly. The current `get_typesupport` function does two things: It transfers ownership of the passed in shared pointer and returns a raw pointer from a newly created object. I would agree if that function would be split in two, where the first one transfers ownership and instantiates a new shared_ptr to a `rcpputils::SharedLibrary`. The `getter` function should then take a correctly initialized shared pointer - `const` if possible - and return the typesupport handle. IMO, this would make things more explicit.

The moment you call the current implementation, you have to be aware that not only will the shared pointer you pass in be reset, but also that all pointers you've received from previous calls to this function are invalid. And that's the problem here.

PR #325, opened 2020-03-17
Document usage in launch

jacobperron:
Add an XML and Python example to README.md.

There's been several questions on ROS Answers about how to start rosbag2 from a launch file, so it seems like a good thing to document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ def generate_launch_description():
 Use the `ros2 launch` command line tool to launch either of the above launch scripts.
 For example, if we named the above XML launch script, `record_all.launch.xml`:
 
-```
+```sh
 $ ros2 launch record_all.launch.xml
 ```
 
--- a/README.md
+++ b/README.md
@@ -125,6 +125,39 @@ Topic information: Topic: /chatter | Type: std_msgs/String | Count: 9 | Serializ
                    Topic: /my_chatter | Type: std_msgs/String | Count: 18 | Serialization Format: cdr
 ```
 
+### Using in launch
+
+We can invoke the command line tool from a ROS launch script as an *executable* (not a *node* action).
+For example, to launch the command to record all topics you can use the following launch script:
+
+```xml
+<launch>
+  <executable cmd="ros2 bag record -a" output="screen" />
+</launch>
+```
+
+Here's the equivalent Python launch script:
+
+```python
+import launch
+
+
+def generate_launch_description():
+    return launch.LaunchDescription([
+        launch.actions.ExecuteProcess(
+            cmd=['ros2', 'bag', 'record', '-a'],
+            output='screen'
+        )
+    ])
+```
+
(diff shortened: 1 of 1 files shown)


jacobperron:
I'm not sure providing users launch files to include would save them any overhead. For example, I imagine usage would look like:

```xml
<launch>
  <include file="$(find ros2bag)/launch/record.launch.xml">
    <arg name="topics" value="/topic0" />
    <arg name="output_name" value="/path/to/output/bag" />
  </include>
</launch>
```

Compared with if they just started the tool directly

```xml
<launch>
  <arg name="topics" default="/topic0" />
  <arg name="output_name" default="/path/to/output/bag"/>
  <executable cmd="ros2 bag record $(var topics) -o $(var output_name)" output="screen" />
</launch>
```

Maybe including the file is nicer for reusability, but I think in any case users will have top-level launch configuration arguments for setting things like topics and output name.

Karsten1987:
I could see that this might not save necessarily typing overhead. What I had in mind though is that if at any point we have other solutions to running rosbag within a launchfile, other than by triggering an executable, we could make that somewhat opaque to the user. But this was only food for thoughts, really.

jacobperron:
The opaqueness argument sounds reasonable to me. I would leave adding actual launch files as a follow-up since I just discovered that trying to pass more than one topic as a launch argument in my example above does not work (I think it's a bug in launch).

README.md
```diff
@@ -125,6 +125,39 @@ Topic information: Topic: /chatter | Type: std_msgs/String | Count: 9 | Serializ
...
+            cmd=['ros2', 'bag', 'record', '-a'],
+            output='screen'
+        )
+    ])
+```
+
+Use the `ros2 launch` command line tool to launch either of the above launch scripts.
+For example, if we named the above XML launch script, `record_all.launch.xml`:
+
+```
```
piraka9011: nit: add `sh`

piraka9011:
> The opaqueness argument sounds reasonable to me. I would leave adding actual launch files as a follow-up since I just discovered that trying to pass more than one topic as a launch argument in my example above does not work (I think it's a bug in launch).

Is this in the XML or python version? I wouldn't think it to be an issue with python.

jacobperron:
> does rosbag2 have index.ros.org docs? 

It has an index page, like all other ROS packages, where it's README is visible: https://index.ros.org/r/rosbag2/github-ros2-rosbag2/#eloquent

Beyond that, I don't think there are any tutorials if that's what you are asking.

> Is this in the XML or python version? I wouldn't think it to be an issue with python.

I noticed it in XML. I'm guessing it's an issue related to the parser (possibly related to https://github.com/ros2/launch/issues/377)

README.md
jacobperron: Added 3341126

FWIW, none of the other code blocks in this file have the code hints.

PR #327, opened 2020-03-20
Create simple ros bag reader

jpmccaffery:
This request introduces a simple bag reader that builds on the SequentialReader and offers easy deserialization of messages. I'm not sure that this is the best interface for handling this. I'm looking for some feedback on whether this or something like it would be good to have in rosbag2.

@mabelzhang


piraka9011:
@jpmccaffery What does this offer that the `SequentialReader` doesn't already do? Is there a use case for this?

mabelzhang:
I think the main contribution is the deserialization function here. SequentialReader returns a serialized message, which the user then needs to figure out how to deserialize. We found that step not straight forward and requires much hunting for existing examples that are hard to find. It would be good to have a message that is already deserialized returned so that the message can be used right away, especially for non-expert users that don't necessarily need to know about serialization/deserialization.

jpmccaffery:
Yes, that's correct. The primary value of this class is in the deserialize method. I found the sequential reader was good for iterating over bag messages if all you needed was their generic contents like topic name. But I needed a straightforward way to decode the messages for what I was trying to do.

piraka9011:
Sgtm, thanks for the clarification.
Could you add tests to your PR please?

Karsten1987:
Why can't this functionality not be added to the `SequentialReader`? Or alternatively, why not letting the `SimpleBagReader` inherit from `SequentialReader` and then add the serialization functions on top?

Implementing the `BaseReaderInterface` here would also mean we lose features like compression or split bags, right?

jpmccaffery:
Regarding tests, I would be happy to add them. I had written this class for personal use and posted it here to see if this is something the ros2 community would be interested in having. If so I'd be happy to write tests for it.

jpmccaffery:
I had initially subclassed SequentialReader (I'll try and see if I can remember why I went the other way). In terms of expanding the sequential reader, I wouldn't have any objection to that.

jpmccaffery:
Now that I'm looking at it, I don't see any reason not to subclass the SequentialReader. If that's the preference here, I would be happy to make that change.

emersonknapp:
I think that's for the best, it should remove a good deal of the code in this PR :)

jpmccaffery:
I remember why I didn't subclass SequentialReader. The reason is that to deserialize the "current message" I needed to hang onto the last message read with read_next. 

The options as I see them are,
1. Leave it as is. This is kind of cluttered.
2. Have the deserialize method accept a SerializedBagMessage and then we could either subclass SequentialReader or just make this decoder a completely separate thing.
3. Modify SequentialReader to add deserialization functionality.

I'm not really wild about any of those options.

emersonknapp:
I like option 2 best of proposed.

Could you not just override `read_next` behavior from the SequentialReader in the DeserializingReader, in order to store the message that you need?

jpmccaffery:
The read_next method isn't vritual and being new to ros, I wasn't inclined to mess with any public API.

emersonknapp:
Ah, but it is virtual https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp#L45 - as you can tell from the `override` here https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp#L67

In C++ if the base declaration is virtual then so are all its overrides, even if they are not explicitly marked as `virtual` or `override`.

jpmccaffery:
Awesome. I learned something new today. Do you prefer overriding read_next to having the deserialize method accept a serialized message?

piraka9011:
>  accept a serialized message

IIUC, yes you should prefer overriding `read_next()`

jpmccaffery:
After speaking with karsten I think we're going to rework this into a serializer / deserializer instead of a new bag reader. So I'll close this pull request.

Issue #125, opened 2019-06-04
Use QoS for subscription, recording, and playback


Karsten1987:
## Feature Request

* Detect QoS for topics in order to subscribe to all topics correctly
* Record QoS of topics in the bag
* Playback bags with recorded QoS
* Option to override the recorded QoS with a different profile on playback (e.g. recorded a BEST_EFFORT, but want to play back as RELIABLE)

## Original issue report:

as the title says. When publishing topics with unreliable qos, the bag does not record any data on this topic as it expects a reliable connection - meaning all subscribers are spawned as reliable.

## Work tracking
Tracking in-progress to-do list here

QoS storage in metadata
* [x] QoS equality operators https://github.com/ros2/rclcpp/pull/1032
* [x] QoS YAML serialization https://github.com/ros2/rosbag2/pull/335
* [x] QoS field add to metadata https://github.com/ros2/rosbag2/pull/335
* [x] Query and store qos for a given topic at subscription time https://github.com/ros2/rosbag2/pull/333

Testing robustness
* [x] Run tests on all rmw implementations 
    * [x] Fix small ament bug https://github.com/ament/ament_cmake/pull/236 
    * [x] Run the rosbag2_transport tests on all rmw impls https://github.com/ros2/rosbag2/pull/349

Record with correct QoS
* [x] Allow the GenericPublisher and GenericSubscription to receive and use a QoS profile https://github.com/ros2/rosbag2/pull/340
* [x] Utilize queried qos to create subscription (if all offered profiles the same) 
    * [x] Initial implementation https://github.com/ros2/rosbag2/pull/343 
    * [x] disable tests because not working on cyclone https://github.com/ros2/rosbag2/pull/348
    * [x] More holistic implementation https://github.com/ros2/rosbag2/pull/355

Playback with correct QoS
* [x] Read saved QoS profiles from bag metadata into TopicMetadata https://github.com/ros2/rosbag2/pull/359
* [x] Expose YAML metadata to the Player https://github.com/ros2/rosbag2/pull/377
* [x] Provide adaptive QoS logic for publishing https://github.com/ros2/rosbag2/pull/379
* [ ] Use adaptive QoS to create publisher for playback https://github.com/ros2/rosbag2/pull/364

Explicit profile overrides
* [x] Let user provide override QoS file to CLI for recording #341
* [x] Let user provide override QoS file to CLI for playback #356 
* [x] Enable explicit per-topic qos profile override for recording #347
[...]

PR #330, opened 2020-03-26
Add QoS profiles field to metadata struct and provide serialization utilities

emersonknapp:
Part of #125 

But, don't _use_ this feature anywhere yet. Next step will be to query and store the QoS profiles for recorded topics.

rosbag2_transport constructs the TopicMetadata, so that's where the serialization/deserialization will need to occur (in `recorder.cpp` / `player.cpp`)


rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -33,7 +33,7 @@ struct TopicInformation
 
 struct BagMetadata
 {
-  int version = 3;  // upgrade this number when changing the content of the struct
+  int version = 4;  // upgrade this number when changing the content of the struct
```
thomas-moulard: is it the right time to update the bag metadata? Could we delay that until the feature is built completely? What's the impact if we were to release rosbag2 just after this PR gets merged?

rosbag2_storage/include/rosbag2_storage/topic_metadata.hpp
```diff
@@ -25,6 +25,8 @@ struct TopicMetadata
   std::string name;
   std::string type;
   std::string serialization_format;
+  // std::vector<rclcpp::QoS> serialized to YAML string
```
thomas-moulard: (very minor) I thought it was commented code! Maybe rephrase to add an empty line before the comment to help the reader?

rosbag2_transport/src/rosbag2_transport/qos.cpp
```diff
@@ -0,0 +1,70 @@
...
+  qos
+  .keep_last(node["depth"].as<int>())
+  .history(history)
+  .reliability(reliability)
+  .durability(durability)
+  .deadline(node["deadline"].as<rmw_time_t>())
+  .lifespan(node["lifespan"].as<rmw_time_t>())
+  .liveliness(liveliness)
+  .liveliness_lease_duration(node["liveliness_lease_duration"].as<rmw_time_t>())
+  .avoid_ros_namespace_conventions(node["avoid_ros_namespace_conventions"].as<bool>());
```
thomas-moulard: I know it's a hard problem to solve, but any idea about how we could try to keep this in sync as qos gets new fields?

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+
+namespace rosbag2_transport
+{
+/// Simple wrapper around rclcpp::QoS to provide a default constructor for YAML deserialization.
+class Rosbag2QoS : public rclcpp::QoS
+{
+public:
+  Rosbag2QoS()
+  : rclcpp::QoS(10) {}
```
thomas-moulard: Why 10? We should probably document the rationale of this constant, to make sure this knowledge is not lost. Could you also use something like QoS(10 /* arg_name */) to help the reader? I have no idea what this means here.

Do we always overwrite this 10 when the YAML obj is deserialized? If yes, maybe that's also something to explain.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include "generic_subscription.hpp"
 #include "rosbag2_node.hpp"
+#include "qos.hpp"
```
thomas-moulard: alphasort

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
emersonknapp: Yeah I was considering this, but I think that now is the time because now is when the data model actually changes. Moving forward that model will just start to hold more useful values. If we were to release this now, nothing bad would happen, it's fully backward compatible with previous metadata versions.

rosbag2_storage/include/rosbag2_storage/topic_metadata.hpp
emersonknapp: I'll rephrase the comment, I can see how that would be confusing at a glance

emersonknapp: I'll rephrase the comment, I can see how that would be confusing at a glance

rosbag2_transport/src/rosbag2_transport/qos.cpp
emersonknapp: I don't have a good answer to this. My thought as of right now is just that rosbag2 needs to have a pass to support new fields if they are added, and until that time it will only support the default values for any given policy.

I'm not aware of a way to introspect fields of a struct in C++, but that doesn't mean there isn't a way

rosbag2_transport/src/rosbag2_transport/qos.hpp
emersonknapp: Yes, we'll always overwrite it when we deserialize, it's an arbitrary number (that is used pretty widely as a de facto default), but it's part of the rclcpp api that users _must_ specify a depth, hence this wrapper for the yaml library that tries to use an empty constructor. I'll add comments to that effect

rosbag2_transport/src/rosbag2_transport/qos.cpp
emersonknapp: Ah - I thought of something we can do. It is a "change detector" like we try to avoid, but in this case of serialization method, it seems like the right thing to do. I'm adding a test that explicitly builds a rmw_qos_profile_t using {} initializer, so if new fields are added, the build will fail.

emersonknapp:
@Karsten1987 how do you feel about the structure of this? These utilities are gating the rest of the functionality, which is started but I can't finish until this is finalized.

rosbag2_transport/src/rosbag2_transport/qos.cpp
thomas-moulard: Thinking about it, `deserialize(serialize(x) == x` should do the trick?

thomas-moulard: If you're sure the qos struct will always be a POD, memcmp is an option otherwise (tread carefully as it can blow up easily)

emersonknapp: I don't think it will. If new fields are added - when I construct the object it will use the default values for those new fields. Now when I serialize, it will write out all the fields we specified here. Then, on deserializing, it will read the fields that it knows, with the unknown fields having default values. The equality will pass and new fields will have slipped through undetected

rosbag2_transport/test/rosbag2_transport/test_qos.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "qos.hpp"
+
+TEST(TestQoS, serialization)
+{
+  rosbag2_transport::Rosbag2QoS expected_qos;
+  YAML::Node offered_qos_profiles;
+  offered_qos_profiles.push_back(expected_qos);
+
+  std::string serialized = YAML::Dump(offered_qos_profiles);
+  printf("%s\n", serialized.c_str());
```
prajakta-gokhale: Remove?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -48,6 +48,7 @@ struct convert<rosbag2_storage::TopicMetadata>
     node["name"] = topic.name;
     node["type"] = topic.type;
     node["serialization_format"] = topic.serialization_format;
+    node["offered_qos_profiles"] = topic.offered_qos_profiles;
```
zmichaels11: Is the reverse required within `decode`?

rosbag2_transport/test/rosbag2_transport/test_qos.cpp
zmichaels11: I think `printf` should be removed from tests. If possible, use an EXPECT and pipe in the error message.

emersonknapp: good catch!

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: This PR only supports storing the QoS profiles, it doesn't support loading them from a bag yet - that will take a bit of extra consideration that I want to put in a separate PR

emersonknapp: The extra consideration being "how do we pass metadata version to the sub-structs?" We should expect the value if it's metadata v4+ but not try to load it for previous versions of the metadata, but right now that version value is not available in this decoding context

rosbag2_transport/test/rosbag2_transport/test_qos.cpp
emersonknapp: It wasn't intended to be there, removed.

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+
+namespace rosbag2_transport
+{
+/// Simple wrapper around rclcpp::QoS to provide a default constructor for YAML deserialization.
+class Rosbag2QoS : public rclcpp::QoS
+{
+public:
+  Rosbag2QoS()
+  : rclcpp::QoS(0) {}  // 0 history depth is always overwritten on deserializing
```
Karsten1987: I guess we could also default to `rclcpp::SystemsDefaultQos()` or something. But I think this doesn't really matter as far as I can understand your comment.

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -0,0 +1,66 @@
...
+};
+
+template<>
+struct convert<rosbag2_transport::Rosbag2QoS>
+{
+  static Node encode(const rosbag2_transport::Rosbag2QoS & qos);
+  static bool decode(const Node & node, rosbag2_transport::Rosbag2QoS & qos);
+};
+}  // namespace YAML
+
```
Karsten1987:
```suggestion
```

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -0,0 +1,66 @@
...
+# pragma warning(push)
+# pragma warning(disable:4251)
+# pragma warning(disable:4275)
+#endif
+#include "yaml-cpp/yaml.h"
+#ifdef _WIN32
+# pragma warning(pop)
+#endif
+
+
```
Karsten1987:
```suggestion
```

rosbag2_transport/src/rosbag2_transport/qos.cpp
Karsten1987: we could maybe apply something like a [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) which only allows to build the object if all parts are present. But I don't see a real way around updating the code base if the QoS struct gets more fields.

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -0,0 +1,66 @@
...
+class Rosbag2QoS : public rclcpp::QoS
+{
+public:
+  Rosbag2QoS()
+  : rclcpp::QoS(0) {}  // 0 history depth is always overwritten on deserializing
+  explicit Rosbag2QoS(const rclcpp::QoS & value)
+  : rclcpp::QoS(value) {}
+};
+}  // namespace rosbag2_transport
+
```
Karsten1987:
```suggestion
```

emersonknapp:
Thanks Karsten for the feedback - sorry we merged before you got to it, I am feeling time pressure for the API freeze and would like to start getting to the features. I'll put up a quick PR addressing these comments

And I will make sure to put a more descriptive explanation on the upcoming PRs of what part of #125 they are achieving.

Karsten1987:
no worries. I guess I was just too slow with my review ;-)

No need for an extra PR for the new lines. You might as well just put them in your next one.

rosbag2_transport/src/rosbag2_transport/qos.hpp
emersonknapp: Note that `rmw_qos_profile_system_default` is different than `rmw_qos_profile_default` - system_default is dependent on the rmw implementation, and because of that I don't think it's something anybody should ever use.

But, I will instantiate this with the default profile (not system_default)

dirk-thomas:
@emersonknapp Can you please point to the CI builds which have been run for this change before it was merged? This change seems to break the build on Windows: https://ci.ros2.org/job/ci_windows/9656/console#console-section-29

emersonknapp:
Argh, sorry, false sense of security from the automatic actions, I've reverted the change here https://github.com/ros2/rosbag2/pull/334 and will reopen this PR with a CI run on it and the windows fix

rosbag2_transport/src/rosbag2_transport/qos.cpp
```diff
@@ -0,0 +1,70 @@
...
+  Node node;
+  node["sec"] = time.sec;
+  node["nsec"] = time.nsec;
+  return node;
+}
+
+bool convert<rmw_time_t>::decode(const Node & node, rmw_time_t & time)
+{
+  time.sec = node["sec"].as<uint>();
+  time.nsec = node["nsec"].as<uint>();
```
zmichaels11:
```suggestion
  time.sec = node["sec"].as<uint64_t>();
  time.nsec = node["nsec"].as<uint64_t>();
```

zmichaels11: I think this would fix it,

PR #333, opened 2020-03-27
Query offered QoS profiles for a topic and store in metadata

emersonknapp:
Part of #125 
Moves https://github.com/ros-tooling/aws-roadmap/issues/215 to Done
This change queries the currently offered QoS profiles for a given topic at the time of subscription, and stores them in the TopicMetadata. This does not use the queried QoS profile in any way, just stores it. 

With this in, you will see the string-serialized profiles in both the metadata.yaml file and the sqlite topics table.

This does _not_ handle new publishers that join after subscribing to a topic. We do not currently have a mechanism to change TopicMetadata after initial creation, and do not plan to add this ability for Foxy.

Once we have used the queried QoS profile to subscribe (upcoming change), we will need to make sure to notice late-joining publishers that offer a different QoS profile than we used to subscribe, to warn the user that this has happened.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -143,7 +143,8 @@ void Recorder::subscribe_topics(
         topic_with_type.first,
         topic_with_type.second,
         serialization_format_,
-        serialized_offered_qos_profiles_for_topic(node_, topic_with_type.first)});
+        serialized_offered_qos_profiles_for_topic(node_, topic_with_type.first)
+      });
   }
 }
 
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -119,7 +119,8 @@ Recorder::get_missing_topics(const std::unordered_map<std::string, std::string> 
   return missing_topics;
 }
 
-namespace {
+namespace
+{
 std::string serialized_offered_qos_profiles_for_topic(
   std::shared_ptr<rosbag2_transport::Rosbag2Node> node,
   const std::string & topic_name)
@@ -137,11 +138,12 @@ void Recorder::subscribe_topics(
   const std::unordered_map<std::string, std::string> & topics_and_types)
 {
   for (const auto & topic_with_type : topics_and_types) {
-    subscribe_topic({
-      topic_with_type.first,
-      topic_with_type.second,
-      serialization_format_,
-      serialized_offered_qos_profiles_for_topic(node_, topic_with_type.first)});
+    subscribe_topic(
+      {
+        topic_with_type.first,
+        topic_with_type.second,
+        serialization_format_,
+        serialized_offered_qos_profiles_for_topic(node_, topic_with_type.first)});
   }
 }
 
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -28,7 +28,21 @@
 #include "rosbag2_transport/logging.hpp"
 
 #include "generic_subscription.hpp"
+#include "qos.hpp"
 #include "rosbag2_node.hpp"
+
+#ifdef _WIN32
+// This is necessary because of a bug in yaml-cpp's cmake
+#define YAML_CPP_DLL
+// This is necessary because yaml-cpp does not always use dllimport/dllexport consistently
+# pragma warning(push)
+# pragma warning(disable:4251)
+# pragma warning(disable:4275)
+#endif
+#include "yaml-cpp/yaml.h"
+#ifdef _WIN32
+# pragma warning(pop)
+#endif
 
 namespace rosbag2_transport
 {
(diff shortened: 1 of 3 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -105,11 +119,31 @@ Recorder::get_missing_topics(const std::unordered_map<std::string, std::string>
...
+  const std::string & topic_name)
+{
+  YAML::Node offered_qos_profiles;
+  auto publishers_info = node->get_publishers_info_by_topic(topic_name);
+  for (auto info : publishers_info) {
+    offered_qos_profiles.push_back(rosbag2_transport::Rosbag2QoS(info.qos_profile()));
+  }
+  return YAML::Dump(offered_qos_profiles);
+}
+}  // unnamed namespace
```
zmichaels11: Does the linter accept this? I thought for anonymous namespaces it wanted `// namespace`

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -105,11 +119,31 @@ Recorder::get_missing_topics(const std::unordered_map<std::string, std::string>
...
   const std::unordered_map<std::string, std::string> & topics_and_types)
 {
   for (const auto & topic_with_type : topics_and_types) {
-    subscribe_topic({topic_with_type.first, topic_with_type.second, serialization_format_, ""});
+    subscribe_topic(
+      {
+        topic_with_type.first,
+        topic_with_type.second,
+        serialization_format_,
+        serialized_offered_qos_profiles_for_topic(node_, topic_with_type.first)});
```
piraka9011: nit: should the braces be on a newline?

zmichaels11: I saw that, iirc they don't have to be.
My personal preference is that they should, but its only a preference.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -64,3 +64,20 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
...
+  pub_man_.add_publisher<test_msgs::msg::Strings>(topic, string_message, 2);
+  run_publishers();
+  stop_recording();
+
+  MockSequentialWriter & writer =
+    static_cast<MockSequentialWriter &>(writer_->get_implementation_handle());
+  auto recorded_topics = writer.get_topics();
+  std::string serialized_profiles = recorded_topics.at(topic).offered_qos_profiles;
+  // Basic smoke test that the profile was serialized into the metadata as a string.
+  EXPECT_THAT(serialized_profiles, ContainsRegex("reliability"));
```
piraka9011: Why not make this a bit more strict like the [`rosbag2 info` tests](https://github.com/ros2/rosbag2/blob/master/rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.cpp#L43)?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
emersonknapp: Note: I learned recently that the technical term is "unnamed namespace" and that "anonymous namespace" isn't a concept in the C++ spec. https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces. I've been calling them anonymous for years lol.

But, yes, the linter accepts it

rosbag2_transport/src/rosbag2_transport/recorder.cpp
emersonknapp: Do you mean the closing brace?

I wanted the opening brace to be

```
subscribe_topic({
```

but uncrustify doesn't like that

rosbag2_transport/test/rosbag2_transport/test_record.cpp
emersonknapp: Yeah, good idea, I'll make it a little more strict

rosbag2_transport/src/rosbag2_transport/recorder.cpp
zmichaels11: TIL: I've been calling it the wrong name for a whole decade

emersonknapp: Maybe an old version of the spec used that terminology, idk. Maybe it's the mandela effect

zmichaels11: https://docs.microsoft.com/en-us/cpp/cpp/namespaces-cpp?view=vs-2019#anonymous-or-unnamed-namespaces

emersonknapp:
@Karsten1987 ping for monday review :) I put more detail on the description for this one.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/8a25fa1084cd436020bf28fc35b0077d/raw/f1ace117ae842c9262eca05929f27756f8bf4e6b/ros2.repos
TEST args: --packages-select rosbag2_transport rosbag2_storage rosbag2_storage_default_plugins rosbag2_tests
Job: ci_launcher

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5446/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7988/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9682/)

Issue #329, opened 2020-03-24
rosbag2_transport crashes due to unsafe threading


zmichaels11:
## Description
Original issue: https://github.com/ros2/rmw_cyclonedds/issues/118
It looks like the E2E tests for `rosbag2_transport` crash due to unsafe usage of `std::async`.

### Expected Behavior
`colcon test --packages-select rosbag2_transport` should pass when using `rmw_cyclonedds`

### Actual Behavior
*The test case "RecordIntegrationTestFixture.published_messages_from_multiple_topics_are_recorded" sometimes fails.

sometimes the output looks like this:
```
colcon test-result      
/opt/ros/master/build/rosbag2_transport/Testing/20200324-2148/Test.xml: 15 tests, 0 errors, 1 failure, 0 skipped
- test_record
  <<< failure message
    -- run_test.py: invoking following command in '/opt/ros/master/build/rosbag2_transport':
     - /opt/ros/master/build/rosbag2_transport/test_record --gtest_output=xml:/opt/ros/master/clion/test_results/rosbag2_transport/test_record.gtest.xml
    Running main() from gmock_main.cc
    [==========] Running 1 test from 1 test case.
    [----------] Global test environment set-up.
    [----------] 1 test from RecordIntegrationTestFixture
    [ RUN      ] RecordIntegrationTestFixture.published_messages_from_multiple_topics_are_recorded
    
    >>> [rcutils|error_handling.c:108] rcutils_set_error_state()
    This error state is being overwritten:
    
      'array_list is already initialized, at /opt/ros/master/src/ros2/rcutils/src/array_list.c:61, at /opt/ros/master/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:188'
    
    with this new error message:
    
      'Failed to add publisher to map., at /opt/ros/master/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:190'
    
    rcutils_reset_error() should be called after error handling to avoid this.
    <<<
    [INFO] [1585086552.363023089] [rosbag2_transport]: Listening for topics...
    unknown file: Failure
    C++ exception with description "failed to initialize rcl node: Failed to add publisher to map., at /opt/ros/master/src/ros2/rcl/rcl/src/rcl/logging_rosout.c:190" thrown in the test body.
  >>>

Summary: 15 tests, 0 errors, 1 failure, 0 skipped
```

Sometimes the output looks like this:
```
colcon test-result
[...]

PR #335, opened 2020-03-27
Add QoS to metadata (re-do #330)

emersonknapp:
Part of #125 
This is an update of https://github.com/ros2/rosbag2/pull/330, which was reverted
* This PR adds a field `offered_qos_profiles` to TopicMetadata. The field is not yet filled by anything, and is is not deserialized on loading. 
* bumps the metadata version to 4 to reflect the change in data model

Next step will be to query and store the QoS profiles for recorded topics. The reason we are not deserializing is because the metadata version will need to be passed to deserialization, or the deserialization will need to just be tolerant to the missing field. We'll address that decision in its own PR. 

rosbag2_transport constructs the TopicMetadata, so that's where the serialization/deserialization will need to occur (in `recorder.cpp` / `player.cpp`)


emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9803/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5424/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7968/)
* Windows [Build Status]

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9804/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5425/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7971/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9666/)

It was a symbol visibility issue for the test executable to link the qos utility functions. Added the explicit yaml dependency while I was poking around

emersonknapp:
Sanity re-test for windows - the failures seem really unrelated and are potentially just flaky [Build Status](https://ci.ros2.org/job/ci_windows/9670/)

emersonknapp:
@Karsten1987 the buildfarm is showing the exact same test failures on windows on master as with this PR. do you think we're good to merge?

emersonknapp:
@Karsten1987 this particular test is failing on the nightly build as well https://ci.ros2.org/view/nightly/job/nightly_win_deb/1573/testReport/junit/rosbag2_tests/RecordFixture/record_end_to_end_test_with_zstd_file_compression_compresses_files/ - in order to not block this feature due to unrelated issues, I'm going to merge now

Karsten1987:
while debugging #329 I've noticed the following test error with cyclonedds:
```
[1.348s] 7: /home/karsten/ros2_ws/src/ros2/rosbag2/rosbag2_transport/test/rosbag2_transport/test_record.cpp:99: Failure
[1.349s] 7: Value of: serialized_profiles
[1.349s] 7: Expected: contains regular expression "- history: 3\n  depth: 0\n  reliability: 1\n  durability: 2\n  deadline:\n    sec: 2147483647\n    nsec: 4294967295\n  lifespan:\n    sec: 2147483647\n    nsec: 4294967295\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 2147483647\n    nsec: 4294967295\n  avoid_ros_namespace_conventions: false"
[1.349s] 7:   Actual: "- history: 1\n  depth: 10\n  reliability: 1\n  durability: 2\n  deadline:\n    sec: 9223372036\n    nsec: 854775807\n  lifespan:\n    sec: 9223372036\n    nsec: 854775807\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 9223372036\n    nsec: 854775807\n  avoid_ros_namespace_conventions: true"
[1.352s] 7: [  FAILED  ] RecordIntegrationTestFixture.qos_is_stored_in_metadata (172 ms)
```

emersonknapp:
Thanks for pointing that out! I think we should call this a bug in rmw_cyclone.

See how the different values are the first two, around history.
* History/depth policy have no bearing on compatibility, so it is not useful to know what the endpoint uses
* The other RMW implementations return RMW_QOS_HISTORY_POLICY_UNKNOWN and depth=0, which makes all endpoint info consistent, which makes it much easier to respect the above point

Karsten1987:
do you mind reporting this on https://github.com/ros2/rmw_cyclonedds? I feel you have more insights on QoS than me to give a meaningful issue report ;-)

However, just for my understanding: If `history/depth` have no influence, why are we testing for it then? We rely on every rmw implementation to adhere to some sort of default then. Are these defaults documented?

emersonknapp:
I think it's reasonable to remove the check for it in the test you saw failing. I'm worried about cross-rmw compatibility, seeing these values, though

PR #337, opened 2020-03-30
Allow GenericPublisher / GenericSubscription to take a QoS profile

emersonknapp:
Part of #125 
Part of https://github.com/ros-tooling/aws-roadmap/issues/216
This PR adds QoS profiles to the GenericSubscription / GenericPublisher internal API, so that they do not always use the default. It continues to set the default explicitly, so behavior is not yet changed. 

Note on testing: I validated that both added tests fail before the rest of the change is introduced. The reason I added the timeout is that the failing case waits the full 60 second rostest timeout otherwise, which is very overkill in all cases, for passing 2 messages.

The followups after this change will be:
* use the queried QoS profiles (from #333) to determine what QoS profile to pass to `create_generic_subscription` for recording behavior. 
* use the deserialized QoS profiles that are read in from TopicMetadata to pass to `create_generic_publisher` for playback behavior

### DIFF
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -52,6 +52,7 @@ ament_target_dependencies(${PROJECT_NAME}
   rosbag2_compression
   rosbag2_cpp
   shared_queues_vendor
+  yaml_cpp_vendor
 )
 
 include(cmake/configure_python.cmake)
@@ -126,22 +127,25 @@ if(BUILD_TESTING)
   endif()
 
   ament_add_gmock(test_rosbag2_node
-    test/rosbag2_transport/test_rosbag2_node.cpp
     src/rosbag2_transport/generic_publisher.cpp
     src/rosbag2_transport/generic_subscription.cpp
+    src/rosbag2_transport/qos.cpp
     src/rosbag2_transport/rosbag2_node.cpp
+    test/rosbag2_transport/test_rosbag2_node.cpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
   if(TARGET test_rosbag2_node)
     target_include_directories(test_rosbag2_node
       PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/rosbag2_transport>
       $<INSTALL_INTERFACE:include>)
     ament_target_dependencies(test_rosbag2_node
       ament_index_cpp
-      rosbag2_cpp
       rclcpp
+      rosbag2_cpp
+      rosbag2_test_common
       test_msgs
-      rosbag2_test_common)
+      yaml_cpp_vendor)
   endif()
 
   ament_add_gmock(test_formatter
@@ -179,15 +183,18 @@ if(BUILD_TESTING)
   endif()
 
   ament_add_gmock(test_qos
+    src/rosbag2_transport/qos.cpp
     test/rosbag2_transport/test_qos.cpp
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     ${SKIP_TEST})
   if(TARGET test_qos)
-    target_link_libraries(test_qos rosbag2_transport)
     target_include_directories(test_qos
       PUBLIC
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/rosbag2_transport>)
-    ament_target_dependencies(test_qos rosbag2_test_common)
+    ament_target_dependencies(test_qos
+      rosbag2_test_common
+      yaml_cpp_vendor)
   endif()
 
 endif()
--- a/rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
@@ -17,14 +17,25 @@
 #include <memory>
 #include <string>
 
+namespace
+{
+rcl_publisher_options_t rosbag2_get_publisher_options(const rclcpp::QoS & qos)
+{
+  auto options = rcl_publisher_get_default_options();
+  options.qos = qos.get_rmw_qos_profile();
+  return options;
+}
+}  // unnamed namespace
+
 namespace rosbag2_transport
 {
 
 GenericPublisher::GenericPublisher(
   rclcpp::node_interfaces::NodeBaseInterface * node_base,
-  const std::string & topic,
-  const rosidl_message_type_support_t & type_support)
-: rclcpp::PublisherBase(node_base, topic, type_support, rcl_publisher_get_default_options())
+  const rosidl_message_type_support_t & type_support,
+  const std::string & topic_name,
+  const rclcpp::QoS & qos)
+: rclcpp::PublisherBase(node_base, topic_name, type_support, rosbag2_get_publisher_options(qos))
 {}
 
 void GenericPublisher::publish(std::shared_ptr<rmw_serialized_message_t> message)
--- a/rosbag2_transport/src/rosbag2_transport/generic_publisher.hpp
+++ b/rosbag2_transport/src/rosbag2_transport/generic_publisher.hpp
@@ -28,8 +28,9 @@ class GenericPublisher : public rclcpp::PublisherBase
 public:
   GenericPublisher(
     rclcpp::node_interfaces::NodeBaseInterface * node_base,
-    const std::string & topic,
-    const rosidl_message_type_support_t & type_support);
+    const rosidl_message_type_support_t & type_support,
+    const std::string & topic_name,
+    const rclcpp::QoS & qos);
 
   virtual ~GenericPublisher() = default;
 
--- a/rosbag2_transport/src/rosbag2_transport/generic_subscription.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/generic_subscription.cpp
@@ -22,6 +22,16 @@
 
 #include "rosbag2_transport/logging.hpp"
 
+namespace
+{
+rcl_subscription_options_t rosbag2_get_subscription_options(const rclcpp::QoS & qos)
+{
+  auto options = rcl_subscription_get_default_options();
+  options.qos = qos.get_rmw_qos_profile();
+  return options;
+}
+}  // unnamed namespace
+
 namespace rosbag2_transport
 {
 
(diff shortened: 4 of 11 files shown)


rosbag2_transport/src/rosbag2_transport/generic_publisher.hpp
```diff
@@ -28,8 +28,9 @@ class GenericPublisher : public rclcpp::PublisherBase
 public:
   GenericPublisher(
     rclcpp::node_interfaces::NodeBaseInterface * node_base,
-    const std::string & topic,
-    const rosidl_message_type_support_t & type_support);
+    const rosidl_message_type_support_t & type_support,
+    const std::string & topic_name,
+    const rclcpp::QoS & qos);
```
piraka9011: Is this change to the API acceptable?

emersonknapp: This is an internal API (notice `src/` directory) - I changed the argument order to be consistent with (GenericSubscription)[https://github.com/ros2/rosbag2/pull/337/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.hpp#diff-edcbd943e37cb6ab8b38a9e64d0d46feR48] because it was just confusing that they were not in the same order

rosbag2_transport/CMakeLists.txt
```diff
@@ -130,18 +131,21 @@ if(BUILD_TESTING)
...
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/rosbag2_transport>
       $<INSTALL_INTERFACE:include>)
     ament_target_dependencies(test_rosbag2_node
       ament_index_cpp
       rosbag2_cpp
       rclcpp
       test_msgs
-      rosbag2_test_common)
+      rosbag2_test_common
```
thomas-moulard: alphasort

rosbag2_transport/CMakeLists.txt
```diff
@@ -180,14 +184,15 @@ if(BUILD_TESTING)
 
   ament_add_gmock(test_qos
     test/rosbag2_transport/test_qos.cpp
+    src/rosbag2_transport/qos.cpp
```
thomas-moulard: alphasort (if possible)

rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
```diff
@@ -17,14 +17,25 @@
 #include <memory>
 #include <string>
 
+namespace
+{
+rcl_publisher_options_t rosbag2_get_publisher_options(const rclcpp::QoS & qos)
+{
+  rcl_publisher_options_t options = rcl_publisher_get_default_options();
```
thomas-moulard:
```suggestion
  auto options = rcl_publisher_get_default_options();
```

nit

rosbag2_transport/src/rosbag2_transport/generic_subscription.cpp
```diff
@@ -22,19 +22,30 @@
 
 #include "rosbag2_transport/logging.hpp"
 
+namespace
+{
+rcl_subscription_options_t rosbag2_get_subscription_options(const rclcpp::QoS & qos)
+{
+  rcl_subscription_options_t options = rcl_subscription_get_default_options();
```
thomas-moulard:
```suggestion
  auto options = rcl_subscription_get_default_options();
```

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -26,7 +26,8 @@
 #include "test_msgs/message_fixtures.hpp"
 #include "test_msgs/msg/basic_types.hpp"
 
-#include "../../src/rosbag2_transport/rosbag2_node.hpp"
+#include "rosbag2_node.hpp"
```
thomas-moulard: alphasort

rosbag2_transport/src/rosbag2_transport/generic_subscription.cpp
zmichaels11:
```suggestion
  auto options = rcl_subscription_get_default_options();
```

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -125,6 +135,68 @@ TEST_F(RosBag2NodeFixture, publisher_and_subscriber_work)
...
+  std::string type = "test_msgs/Strings";
+  rclcpp::QoS qos = rclcpp::SensorDataQoS();
+  auto publisher = node_->create_publisher<test_msgs::msg::Strings>(topic_name, qos);
+
+  auto subscriber_future_ = std::async(
+    std::launch::async, [this, topic_name, type, qos] {
+      return subscribe_raw_messages(1, topic_name, type, qos);
+    });
+  // Give time to the subscriber to start.
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
thomas-moulard: This test is time dependent, and will be flaky.

Better: switch an atomic bool in the lambda and do some busy wait + a sleep here
Best: condition variable, but I understand it's harder to write

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -125,6 +135,68 @@ TEST_F(RosBag2NodeFixture, publisher_and_subscriber_work)
...
+
+  for (const auto & message : test_messages) {
+    test_msgs::msg::Strings string_message;
+    string_message.string_value = message;
+    publisher->publish(string_message);
+    // This is necessary because sometimes, the subscriber is initialized very late
```
thomas-moulard: Do you smell flaky? Because I smell flaky :)

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -158,7 +159,8 @@ void Player::prepare_publishers()
   for (const auto & topic : topics) {
     publishers_.insert(
       std::make_pair(
-        topic.name, rosbag2_transport_->create_generic_publisher(topic.name, topic.type)));
+        topic.name, rosbag2_transport_->create_generic_publisher(
+          topic.name, topic.type, Rosbag2QoS())));
   }
```
zmichaels11:
```suggestion
    publishers_.insert({
      topic.name, 
      rosbag2_transport_->create_generic_publisher(topic.name, topic.type, Rosbag2Qos())
    });
```
I think this works. 
Also, can you replace `Rosbag2Qos()` with `Rosbag2Qos{}`?

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -125,6 +135,68 @@ TEST_F(RosBag2NodeFixture, publisher_and_subscriber_work)
...
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+
+  auto subscribed_messages = subscriber_future_.get();
+  ASSERT_THAT(subscribed_messages, SizeIs(Not(0)));
+  EXPECT_THAT(subscribed_messages[0], StrEq("Hello World"));
+}
+
+TEST_F(RosBag2NodeFixture, generic_publisher_uses_qos)
```
thomas-moulard: ditto here

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
emersonknapp: Copying existing test structure from this file which has been working reliably - I'm personally willing to take on the tech debt of "possibly flaky test" right now in order to hit the Foxy release date, over restructuring the tests now. What do you say?

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -125,6 +135,68 @@ TEST_F(RosBag2NodeFixture, publisher_and_subscriber_work)
   EXPECT_THAT(subscribed_messages[0], StrEq("Hello World"));
 }
 
+TEST_F(RosBag2NodeFixture, generic_subscription_uses_qos)
+{
+  // If the GenericSubscription does not use the provided QoS profile,
+  // its request will be incompatible with the Publisher's offer and no messages will be passed.
+  using namespace std::chrono_literals;
+  std::vector<std::string> test_messages = {"Hello World", "Hello World"};
```
mm318: Shouldn't you test with different messages?

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -59,20 +60,28 @@ class RosBag2NodeFixture : public Test
...
-    size_t expected_messages_number, const std::string & topic_name, const std::string & type)
+    size_t expected_messages_number,
+    const std::string & topic_name,
+    const std::string & type,
+    const rclcpp::QoS & qos = rosbag2_transport::Rosbag2QoS())
   {
+    using namespace std::chrono_literals;
+    auto start = std::chrono::system_clock::now();
     std::vector<std::string> messages;
     size_t counter = 0;
```
zmichaels11: Should `counter` be atomic here since the callback is incrementing it while you busy wait on the current thread?

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
zmichaels11: Is there a way to not depend on `sleep_for`?

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
emersonknapp: I'll remove the contents check - all I care about is whether any message got through

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: No problem on the `{}` (can you point me at some literature why that's better?). I'm going to leave `make_pair` because it's outside the scope of my change

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
Karsten1987: It's definitely flaky and part of the reason why some end to end tests are disabled. I am happy to track this in a follow-up issue, but I think it's a rabbit hole we shouldn't go down here in this PR. But that's just my 2cts.

rosbag2_transport/src/rosbag2_transport/player.cpp
zmichaels11: Using `{}` to invoke ctors prevents it from invoking anything but a ctor.
Its in the [isocpp core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax)

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
emersonknapp: I was able to greatly simplify these tests to avoid the sleeps

PR #343, opened 2020-04-01
Subscribe to topics using the common offered QoS

emersonknapp:
Part of #125 
Moves https://github.com/ros-tooling/aws-roadmap/issues/216 to Done

Subscribe to topics using the common offered QoS if there is one, with sufficient warning in the cases where we can't.

As we discussed in the original design, we will not try to resolve compatibility at all. So, if all publishers at the time of subscription offer the same QoS profile (likely vast majority of cases), then that profile will be used to subscribe.

However, if there are different profiles being offered for a topic, fallback to the preexisting default. This means rosbag2 acts the same for these cases as it does before this PR. But, now we print a warning about it.

If new publishers join after our subscription was created, and they offer a different QoS profile than we used to subscribe, also print a warning about that case because it may result in loss of data.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -160,7 +160,7 @@ void Recorder::subscribe_topic(const rosbag2_storage::TopicMetadata & topic)
   // that callback called before we reached out the line: writer_->create_topic(topic)
   writer_->create_topic(topic);
 
-  Rosbag2QoS subscription_qos(common_qos_or_fallback(topic.name));
+  Rosbag2QoS subscription_qos{common_qos_or_fallback(topic.name)};
   auto subscription = create_subscription(topic.name, topic.type, subscription_qos);
   if (subscription) {
     subscriptions_.insert({topic.name, subscription});
@@ -219,7 +219,7 @@ rclcpp::QoS Recorder::common_qos_or_fallback(const std::string & topic_name)
   }
   ROSBAG2_TRANSPORT_LOG_WARN_STREAM(
     "Topic " << topic_name << " has publishers offering different QoS settings. "
-      "Can't guess what QoS to request, falling back to default QoS profile."
+      "Cannot determine what QoS to request, falling back to default QoS profile."
   );
   topics_warned_about_incompatibility_.insert(topic_name);
   return Rosbag2QoS{};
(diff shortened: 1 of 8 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.hpp
```diff
@@ -46,6 +48,14 @@ class Recorder
   explicit Recorder(std::shared_ptr<rosbag2_cpp::Writer> writer, std::shared_ptr<Rosbag2Node> node);
 
   void record(const RecordOptions & record_options);
+  const std::unordered_set<std::string> & topics_using_fallback_qos() const
+  {
+    return topics_warned_about_incompatibility_;
+  }
+  const std::unordered_map<std::string, std::shared_ptr<GenericSubscription>> & subscriptions()
```
thomas-moulard, line 55:
```suggestion
  const std::unordered_map<std::string, std::shared_ptr<GenericSubscription>> & subscriptions() const
```

rosbag2_transport/test/rosbag2_transport/test_rosbag2_node.cpp
```diff
@@ -251,3 +259,50 @@ TEST_F(RosBag2NodeFixture, get_all_topics_with_types_returns_all_topics)
...
+  auto success_condition = [recorder, topic]() -> bool {
+    return recorder->topics_using_fallback_qos().count(topic) != 0;
+  };
+  while (!success_condition()) {
+    if ((clock::now() - start) > timeout) {
+      timed_out = true;
+      break;
+    }
+    // rate limit the busy-wait and let other threads work
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
thomas-moulard, line 302: 10ms as you already included chrono_literals?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -154,26 +159,26 @@ void Recorder::subscribe_topic(const rosbag2_storage::TopicMetadata & topic)
   // callback for subscription we are calling writer_->write(bag_message); and it could happened
   // that callback called before we reached out the line: writer_->create_topic(topic)
   writer_->create_topic(topic);
-  auto subscription = create_subscription(topic.name, topic.type);
 
+  Rosbag2QoS subscription_qos(common_qos_or_fallback(topic.name));
```
zmichaels11, line 163:
```suggestion
  Rosbag2QoS subscription_qos{common_qos_or_fallback(topic.name)};
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -196,4 +201,54 @@ void Recorder::record_messages() const
...
+  auto existing_subscription = subscriptions_.find(topic_name);
+  if (existing_subscription == subscriptions_.end()) {
+    // Not subscribed yet
+    return;
+  }
+  if (topics_warned_about_incompatibility_.count(topic_name) > 0) {
+    // Already warned about this topic
+    return;
+  }
+  const rclcpp::QoS & used_qos = existing_subscription->second->qos_profile();
```
zmichaels11, line 239:
```suggestion
  const auto & used_qos = existing_subscription->second->qos_profile();
```

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9896/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5512/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9750/)

emersonknapp:
The windows build is again the known test failure

PR #346, opened 2020-04-02
Override Subscriber QoS - Record

piraka9011:
* Add a `qos_profiles` string to the recording options
* Expose Publisher's QoS profile option in `PublisherManager`
* Override a topic's QoS profile if an override is specified.

Part of #125

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
@@ -78,7 +78,7 @@ public:
     const std::string & topic_name,
     std::shared_ptr<T> message,
     size_t expected_messages = 0,
-    const rclcpp::QoS & qos = rclcpp::QoS(10))
+    const rclcpp::QoS & qos = rclcpp::QoS{10})
   {
     auto node_name = std::string("publisher") + std::to_string(counter_++);
     auto publisher_node = std::make_shared<rclcpp::Node>(
--- a/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
@@ -75,13 +75,16 @@ public:
 
   template<class T>
   void add_publisher(
-    const std::string & topic_name, std::shared_ptr<T> message, size_t expected_messages = 0)
+    const std::string & topic_name,
+    std::shared_ptr<T> message,
+    size_t expected_messages = 0,
+    const rclcpp::QoS & qos = rclcpp::QoS(10))
   {
     auto node_name = std::string("publisher") + std::to_string(counter_++);
     auto publisher_node = std::make_shared<rclcpp::Node>(
       node_name,
       rclcpp::NodeOptions().start_parameter_event_publisher(false).enable_rosout(false));
-    auto publisher = publisher_node->create_publisher<T>(topic_name, 10);
+    auto publisher = publisher_node->create_publisher<T>(topic_name, qos);
 
     publisher_nodes_.push_back(publisher_node);
     publishers_.push_back(
--- a/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
@@ -78,7 +78,7 @@ public:
     const std::string & topic_name,
     std::shared_ptr<T> message,
     size_t expected_messages = 0,
-    const rclcpp::QoS & qos = rclcpp::QoS{10})
+    const rclcpp::QoS & qos = rclcpp::QoS{rclcpp::KeepAll()})
   {
     auto node_name = std::string("publisher") + std::to_string(counter_++);
     auto publisher_node = std::make_shared<rclcpp::Node>(
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -15,6 +15,9 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic -Werror)
 endif()
 
+# TODO(piraka9011) Remove once testing is finished
+add_definitions(-DROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION)
+
 # Windows supplies macros for min and max by default. We should only use min and max from stl
 if(WIN32)
   add_definitions(-DNOMINMAX)
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -153,10 +153,10 @@ function(create_tests_for_rmw_implementation)
 
   rosbag2_transport_add_gmock(test_record
     test/rosbag2_transport/test_record.cpp
-    ${SKIP_TEST}
-    LINK_LIBS rosbag2_transport
     AMENT_DEPS test_msgs rosbag2_test_common
-    INCLUDE_DIRS src/rosbag2_transport)
+    INCLUDE_DIRS src
+    LINK_LIBS rosbag2_transport
+    ${SKIP_TEST})
 
   rosbag2_transport_add_gmock(test_play
     test/rosbag2_transport/test_play.cpp
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -155,7 +155,8 @@ function(create_tests_for_rmw_implementation)
     test/rosbag2_transport/test_record.cpp
     ${SKIP_TEST}
     LINK_LIBS rosbag2_transport
-    AMENT_DEPS test_msgs rosbag2_test_common)
+    AMENT_DEPS test_msgs rosbag2_test_common
+    INCLUDE_DIRS src/rosbag2_transport)
 
   rosbag2_transport_add_gmock(test_play
     test/rosbag2_transport/test_play.cpp
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -151,7 +151,7 @@ function(create_tests_for_rmw_implementation)
   rosbag2_transport_add_gmock(test_record
     test/rosbag2_transport/test_record.cpp
     AMENT_DEPS test_msgs rosbag2_test_common
-    INCLUDE_DIRS src
+    INCLUDE_DIRS $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/rosbag2_transport>
     LINK_LIBS rosbag2_transport
     ${SKIP_TEST})
 
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -15,9 +15,6 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic -Werror)
 endif()
 
-# TODO(piraka9011) Remove once testing is finished
-add_definitions(-DROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION)
-
 # Windows supplies macros for min and max by default. We should only use min and max from stl
 if(WIN32)
   add_definitions(-DNOMINMAX)
--- a/rosbag2_transport/include/rosbag2_transport/record_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/record_options.hpp
@@ -17,7 +17,10 @@
 
 #include <chrono>
 #include <string>
+#include <unordered_map>
 #include <vector>
+
+#include "rclcpp/rclcpp.hpp"
 
 namespace rosbag2_transport
 {
@@ -32,7 +35,7 @@ public:
   std::string node_prefix = "";
   std::string compression_mode = "";
   std::string compression_format = "";
-  std::string qos_profiles = "";
+  std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides{};
   bool include_hidden_topics = false;
 };
 
--- a/rosbag2_transport/include/rosbag2_transport/record_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/record_options.hpp
@@ -32,6 +32,7 @@ public:
   std::string node_prefix = "";
   std::string compression_mode = "";
   std::string compression_format = "";
+  std::string qos_profiles = "";
   bool include_hidden_topics = false;
 };
 
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -150,12 +150,7 @@ void Recorder::subscribe_topic(const rosbag2_storage::TopicMetadata & topic)
   // that callback called before we reached out the line: writer_->create_topic(topic)
   writer_->create_topic(topic);
 
-  // TODO(emersonknapp) re-enable subscription_qos_for_topic once the cyclone situation is resolved
-  #ifdef ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
   Rosbag2QoS subscription_qos{subscription_qos_for_topic(topic.name)};
-  #else
-  Rosbag2QoS subscription_qos{};
-  #endif  // ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
   auto subscription = create_subscription(topic.name, topic.type, subscription_qos);
   if (subscription) {
     subscriptions_.insert({topic.name, subscription});
@@ -213,6 +208,8 @@ rclcpp::QoS Recorder::subscription_qos_for_topic(const std::string & topic_name)
     subscription_qos = topic_qos_profile_overrides_.at(topic_name);
     ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding subscription profile for " << topic_name);
   }
+  // TODO(emersonknapp) re-enable subscription_qos_for_topic once the cyclone situation is resolved
+#ifdef ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
   auto endpoints = node_->get_publishers_info_by_topic(topic_name);
   if (!endpoints.empty() && all_qos_same(endpoints)) {
     return Rosbag2QoS(endpoints[0].qos_profile()).default_history();
@@ -222,6 +219,7 @@ rclcpp::QoS Recorder::subscription_qos_for_topic(const std::string & topic_name)
       "Cannot determine what QoS to request, falling back to default QoS profile."
   );
   topics_warned_about_incompatibility_.insert(topic_name);
+#endif  // ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
   return subscription_qos;
 }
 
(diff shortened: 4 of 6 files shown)


rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
```diff
@@ -75,13 +75,16 @@ class PublisherManager
 
   template<class T>
   void add_publisher(
-    const std::string & topic_name, std::shared_ptr<T> message, size_t expected_messages = 0)
+    const std::string & topic_name,
+    std::shared_ptr<T> message,
+    size_t expected_messages = 0,
+    const rclcpp::QoS & qos = rclcpp::QoS(10))
```
zmichaels11, line 81:
```suggestion
    const rclcpp::QoS & qos = rclcpp::QoS{10})
```

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,49 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+    "    nsec: 0\n"
+    "  liveliness: 0\n"
+    "  liveliness_lease_duration:\n"
+    "    sec: 0\n"
+    "    nsec: 0\n"
+    "  avoid_ros_namespace_conventions: false\n";
+
+  // Create two publishers on the same topic with different QoS profiles.
+  // If no override is specified, then the recorder cannot see any published messages.
+  auto profile1 = rosbag2_transport::Rosbag2QoS().best_effort().durability_volatile();
```
zmichaels11, line 175:
```suggestion
  auto profile1 = rosbag2_transport::Rosbag2QoS{}.best_effort().durability_volatile();
```

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,49 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+  auto profile2 = rosbag2_transport::Rosbag2QoS().best_effort().transient_local();
+  pub_man_.add_publisher<test_msgs::msg::Strings>(strict_topic, strict_msg, 3, profile1);
+  pub_man_.add_publisher<test_msgs::msg::Strings>(strict_topic, strict_msg, 3, profile2);
+
+  start_recording(record_options);
+  run_publishers();
+  stop_recording();
+
+  MockSequentialWriter & writer =
```
zmichaels11, line 184:
```suggestion
auto & writer =
```

rosbag2_transport/include/rosbag2_transport/record_options.hpp
```diff
@@ -32,6 +32,7 @@ struct RecordOptions
   std::string node_prefix = "";
   std::string compression_mode = "";
   std::string compression_format = "";
+  std::string qos_profiles = "";
```
emersonknapp, line 35: Can we try to have the API provide something more structured? Maybe `std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides;` Since this is a public API and we may have external code writing calls against it, this is probably too error prone.

Worst case, let's name this something a little more descriptive, e.g. `yaml_serialized_per_topic_qos_profile_overrides_map`. But as a last resort.

rosbag2_transport/src/rosbag2_transport/recorder.hpp
```diff
@@ -101,6 +114,7 @@ class Recorder
   std::unordered_map<std::string, std::shared_ptr<GenericSubscription>> subscriptions_;
   std::unordered_set<std::string> topics_warned_about_incompatibility_;
   std::string serialization_format_;
+  YAML::Node qos_profile_overrides_;
```
emersonknapp, line 117: This should be a map at this point - we should be storing structured data, the C++ YAML node is an intermediate format

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -175,10 +163,15 @@ std::shared_ptr<GenericSubscription>
 Recorder::create_subscription(
   const std::string & topic_name, const std::string & topic_type, const rclcpp::QoS & qos)
 {
+  Rosbag2QoS subscription_qos(qos);
+  if (qos_profile_overrides_[topic_name]) {
```
emersonknapp, line 167: rather than putting this logic here, I think we should update `common_qos_or_fallback` to be the single source of truth for what QoS to use for a topic, that way the logic isn't distributed.

Maybe rename it as `subscription_qos_for_topic` and update the docstring to clarify that it does a priority ordering of "override, common qos, fallback"

piraka9011:
Remember that the `ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION` flag must be enabled.

```
add_definitions(-DROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION)
# Or
colcon build ... --cmake-args -DROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
```

rosbag2_transport/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic -Werror)
 endif()
 
+# TODO(piraka9011) Remove once testing is finished
```
zmichaels11, line 18: When is testing going to be finished? Is there an issue to track this temp change?

rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
```diff
@@ -75,13 +75,16 @@ class PublisherManager
 
   template<class T>
   void add_publisher(
-    const std::string & topic_name, std::shared_ptr<T> message, size_t expected_messages = 0)
+    const std::string & topic_name,
+    std::shared_ptr<T> message,
+    size_t expected_messages = 0,
+    const rclcpp::QoS & qos = rclcpp::QoS{10})
```
dabonnie, line 81: Use `SystemDefaultsQoS`?

emersonknapp: SystemDefault != Default and should never be used, it has undefined behavior because it does not define a default but rather shops the value out the specific rmw implementation.

emersonknapp: But you're right that we can't just have the 10 constant sitting in here. Best to initialize with `rmw_qos_profile_default`

piraka9011: Apparently `SystemDefaultsQoS` doesn't guarantee consistency across tests since each RMW implementation has it's own definition of System Default.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -218,8 +201,15 @@ std::string Recorder::serialized_offered_qos_profiles_for_topic(const std::strin
   return YAML::Dump(offered_qos_profiles);
 }
 
-rclcpp::QoS Recorder::common_qos_or_fallback(const std::string & topic_name)
+rclcpp::QoS Recorder::subscription_qos_for_topic(const std::string & topic_name)
 {
+  rclcpp::QoS subscription_qos{10};
```
dabonnie, line 206: Same comment regarding the QoS default. `10` is used in multiple places, use a common default instead (no magic numbers)?

rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
piraka9011: It looks like the [QoS constructor](https://github.com/ros2/rclcpp/blob/01a6befdde7f78b6f7788e6180cc5ed38390e955/rclcpp/include/rclcpp/qos.hpp#L60) uses `rmw_qos_profile_default` by default. It still needs a `QoSInitialization` so maybe something along the lines of:

```C++
const size_t DEFAULT_HISTORY = 10;
...
const rclcpp::QoS & qos = rclcpp::QoS{KeepLast(DEFAULT_HISTORY)};
```

Since the default in [`qos_profiles.h`](https://github.com/ros2/rmw/blob/2f954d77a102042fa5c13540e2db300d549d7ad3/rmw/include/rmw/qos_profiles.h#L54) is 10.

rosbag2_transport/CMakeLists.txt
thomas-moulard: Consider adding a link to an issue instead so that we can track it.

rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp
piraka9011: Update: Using instead

```c++
const rclcpp::QoS & qos = rclcpp::QoS{KeepAll()};
```

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,37 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
 #endif  // ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
+
+TEST_F(RecordIntegrationTestFixture, topic_qos_overrides)
+{
+  auto strict_msg = std::make_shared<test_msgs::msg::Strings>();
+  strict_msg->string_value = "strict";
+  std::string strict_topic = "/strict_topic";
+
+  rosbag2_transport::RecordOptions record_options =
+  {false, false, {strict_topic}, "rmw_format", 100ms};
```
thomas-moulard, line 152: Consider using aggregated initialization instead: https://en.cppreference.com/w/cpp/language/aggregate_initialization
I.e. `{ .field1 = ..., .field2 = ...}`.

For those `struct` containing options which are mainly primitive types, you're really at risk of failing to the usual blunder: someone will add a new field somewhere, and all assignments are shifted by one, and the last attribute is uninitialized. Stating the fields explicitly at least guarantee you that each value won't end up being shifted.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,37 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+  // 0 means system default for all options
+  const auto profile_override = rclcpp::QoS{10}
```
thomas-moulard, line 154: why 10?

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,37 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+
+TEST_F(RecordIntegrationTestFixture, topic_qos_overrides)
+{
+  auto strict_msg = std::make_shared<test_msgs::msg::Strings>();
+  strict_msg->string_value = "strict";
+  std::string strict_topic = "/strict_topic";
+
+  rosbag2_transport::RecordOptions record_options =
+  {false, false, {strict_topic}, "rmw_format", 100ms};
+  // 0 means system default for all options
```
thomas-moulard, line 153: Not sure I understand this comment?

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,37 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+  std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {
+    {"/strict_topic", profile_override}
+  };
+  record_options.topic_qos_profile_overrides = topic_qos_profile_overrides;
+
+  // Create two publishers on the same topic with different QoS profiles.
+  // If no override is specified, then the recorder cannot see any published messages.
+  auto profile1 = rosbag2_transport::Rosbag2QoS{}.best_effort().durability_volatile();
+  auto profile2 = rosbag2_transport::Rosbag2QoS{}.best_effort().transient_local();
+  pub_man_.add_publisher<test_msgs::msg::Strings>(strict_topic, strict_msg, 3, profile1);
```
thomas-moulard, line 165: what is 3?

rosbag2_transport/CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
+add_definitions(-DROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION)
```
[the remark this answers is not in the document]
piraka9011: We need to keep this after #348 was introduced.
I don't know of another way to test this with the GitHub Actions CI, however, we can add the flag when running the ROS2 CI.

rosbag2_transport/CMakeLists.txt
```diff
@@ -152,7 +155,8 @@ function(create_tests_for_rmw_implementation)
     test/rosbag2_transport/test_record.cpp
     ${SKIP_TEST}
     LINK_LIBS rosbag2_transport
-    AMENT_DEPS test_msgs rosbag2_test_common)
+    AMENT_DEPS test_msgs rosbag2_test_common
+    INCLUDE_DIRS src/rosbag2_transport)
```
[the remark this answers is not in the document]
piraka9011: This test requires `qos.hpp` which is in `src/rosbag2_transport` and not exposed publicly.
The previous suggestion was to change the `include_directories` in the CMakeLists and not use relative paths in the test  (`../../src/rosbag2_transport/qos.hpp`).

Not sure if we should add `src/rosbag2_transport` to all tests by default.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -218,8 +201,15 @@ std::string Recorder::serialized_offered_qos_profiles_for_topic(const std::strin
...
+  if (topic_qos_profile_overrides_.count(topic_name)) {
+    subscription_qos = topic_qos_profile_overrides_.at(topic_name);
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding subscription profile for " << topic_name);
+  }
+  // TODO(emersonknapp) re-enable subscription_qos_for_topic once the cyclone situation is resolved
+#ifdef ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
```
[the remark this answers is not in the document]
piraka9011: Nope, this is from #348.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
piraka9011: So it doesn't look like we can use this in C++14?

```
 error: ISO C++ does not allow C99 designated initializers [-Werror=pedantic]
```

rosbag2_transport/CMakeLists.txt
emersonknapp: No, I mean that the resolution is to remove it now, before we approve the PR

rosbag2_transport/CMakeLists.txt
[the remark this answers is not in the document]
emersonknapp: This feature should be completely independent of the adaptive logic. Like I suggested moving the override block into the "qos finder" function and putting the ifdef check only around the adaptive part. You should not need this definition to test your code.

rosbag2_transport/CMakeLists.txt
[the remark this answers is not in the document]
emersonknapp: No, not all tests need (or should have access to) internal interfaces. Some are intended to only touch the public API. This is fine here

rosbag2_transport/test/rosbag2_transport/test_record.cpp
emersonknapp: Correct, this is a C99 syntax and is not supported in C++ until C++20

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -31,19 +31,6 @@
 #include "qos.hpp"
 #include "rosbag2_node.hpp"
 
-#ifdef _WIN32
-// This is necessary because of a bug in yaml-cpp's cmake
-#define YAML_CPP_DLL
```
emersonknapp, line 36: this block removal is unrelated to this PR

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -218,8 +201,15 @@ std::string Recorder::serialized_offered_qos_profiles_for_topic(const std::strin
...
-rclcpp::QoS Recorder::common_qos_or_fallback(const std::string & topic_name)
+rclcpp::QoS Recorder::subscription_qos_for_topic(const std::string & topic_name)
 {
+  rclcpp::QoS subscription_qos{rclcpp::KeepAll()};
+  if (topic_qos_profile_overrides_.count(topic_name)) {
+    subscription_qos = topic_qos_profile_overrides_.at(topic_name);
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding subscription profile for " << topic_name);
+  }
+  // TODO(emersonknapp) re-enable subscription_qos_for_topic once the cyclone situation is resolved
+#ifdef ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
```
emersonknapp, line 212: minor nit: can you keep these indented so the function contents are at the same level?

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -141,3 +141,39 @@ TEST_F(RecordIntegrationTestFixture, records_sensor_data)
...
+  rosbag2_transport::RecordOptions record_options =
+  {false, false, {strict_topic}, "rmw_format", 100ms};
+  const auto profile_override = rclcpp::QoS{rclcpp::KeepAll()}
+  .best_effort().durability_volatile().avoid_ros_namespace_conventions(false);
+  std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {
+    {strict_topic, profile_override}
+  };
+  record_options.topic_qos_profile_overrides = topic_qos_profile_overrides;
+
+  // Create two publishers on the same topic with different QoS profiles.
```
emersonknapp, line 161:
```suggestion
  // Create two BEST_EFFORT publishers on the same topic with different Durability policies.
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -46,6 +46,8 @@
 
 namespace
 {
+// TODO(emersonknapp) re-enable subscription_qos_for_topic once the cyclone situation is resolved
+#ifdef ROSBAG2_ENABLE_ADAPTIVE_QOS_SUBSCRIPTION
```
piraka9011, line 50: I had to introduce this `ifdef` since I was getting an unused function compile time error.

piraka9011:
@ros2/aws-oncall @prajakta-gokhale - please run this CI job
Gist: https://gist.githubusercontent.com/piraka9011/42402c78b9ffea330dd6b88a3841b6d2/raw/aed855a901d65a26df9ff5eadc8e452ef6c20a2a/ros2.repos
Job: ci_launcher

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9956/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5563/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8100/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9826/)

emersonknapp:
The Windows cmake warning seems ephemeral, there is no actual output except mayybe something around `spdlog`. All tests passed on Windows, which I am not used to seeing. Considering this a clean CI run and merging

PR #347, opened 2020-04-03
Add QoS Profile override to CLI

piraka9011:
* Add a `--qos-profile-overrides` CLI option
* Parse a yaml file with overrides into a `QoSProfile`
* Test different scenarios (complete, partial, and non-existent yaml overrides).

This PR as it stands is non-functional.
However, it ties in to the PR's related to #125 by overriding the `qos_profiles` introduced in the `RecordOptions`.

Part of #125


rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -69,7 +71,7 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
   bool include_hidden_topics = false;
   if (
     !PyArg_ParseTupleAndKeywords(
-      args, kwargs, "ssssss|bbKKKOb", const_cast<char **>(kwlist),
+      args, kwargs, "ssssss|bbKKKObs", const_cast<char **>(kwlist),
```
emersonknapp, line 74: We will need to be passing these QoS profiles around as actual objects. Are you familiar with the "capsule" concept used in Python<->C bindings to hold opaque C pointers?

If not yet, let's discuss. I don't know off the top of my head what the implementation looks like, but we should be able to pass either:
* a map of topic -> rmw_qos_profile_t
* or a list/array of pairs of (topic, rmw_qos_profile_t)

That way we've already deserialized here and the rosbag2_transport side can just receive structured data.

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +85,30 @@ def create_bag_directory(self, uri):
...
 
+    def validate_qos_profiles(self, qos_profile_path):
+        """Validate the QoS profile yaml file path and its structure."""
+        qos_params = ['history', 'depth', 'reliability', 'durability',
+                      'deadline', 'lifespan', 'liveliness', 'liveliness_lease_duration',
+                      'avoid_ros_namespace_conventions']
+        if os.path.isfile(qos_profile_path):
+            with open(qos_profile_path, 'r') as file:
+                # Load as a dict first to verify contents
+                qos_profile = yaml.safe_load(file)
```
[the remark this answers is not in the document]
piraka9011: I use it again later on to read the file as a string. I have an optimization in mind (later) to read in the dictionary directly in the transport layer.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
piraka9011: Familiar with Python's C-API [Object protocol](https://docs.python.org/3/c-api/object.html).

I was thinking something along the lines of:

```
Rosbag2Qos qos{};
qos.history = *PyString_AsString(PyMapping_GetItemString(pDict, "history"))
...
```

piraka9011: Also though on using [`pybind`](https://pybind11.readthedocs.io/en/master/advanced/cast/stl.html) now that it's part of the repo?

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +91,34 @@ def create_bag_directory(self, uri):
...
+        topic_names = list(qos_profile_dict)
+        for name in topic_names:
+            try:
+                profile = qos_profile_dict.get(name)
+                # Convert dict to Duration. Required for construction
+                profile['deadline'] = self._dict_to_duration(profile.get('deadline'))
+                profile['lifespan'] = self._dict_to_duration(profile.get('lifespan'))
+                profile['liveliness_lease_duration'] = self._dict_to_duration(
+                    profile.get('liveliness_lease_duration'))
+                QoSProfile(**qos_profile_dict.get(name))
```
emersonknapp, line 115: you already have the variable `profile`, no need to call `get` again

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -105,6 +160,17 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
     rosbag2_compression::compression_mode_from_string(record_options.compression_mode)
   };
 
+  if (qos_profile_overrides) {
+    PyObject * key, * value;
+    Py_ssize_t pos = 0;
+    std::unordered_map<std::string, rclcpp::QoS> topic_qos_overrides{};
+    while (PyDict_Next(qos_profile_overrides, &pos, &key, &value)) {
```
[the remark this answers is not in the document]
piraka9011: We technically do that check in the Python layer but I can add a check here as well.
FYI the default value of the override is an empty dict `{}`.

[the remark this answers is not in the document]
emersonknapp: That's currently the only thing that calls this, but this C++ layer is technically a standalone API

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +91,34 @@ def create_bag_directory(self, uri):
...
+    def validate_qos_profile_overrides(self, qos_profile_path: str) -> Dict[str, Dict]:
+        """Validate the QoS profile yaml file path and its structure."""
+        if not os.path.isfile(qos_profile_path):
+            raise ValueError('[ERROR] [ros2bag]: {} does not exist.'.format(qos_profile_path))
+        with open(qos_profile_path, 'r') as file:
+            qos_profile_dict = yaml.safe_load(file)
+        topic_names = list(qos_profile_dict)
+        for name in topic_names:
+            try:
+                profile = qos_profile_dict.get(name)
```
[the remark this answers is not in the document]
piraka9011: I tried this with a simple yaml file and it seems that the overriding occurs already in the yaml load so there's not much we can do there.
Example:

```yaml
topic_a:
  history: 1
  durability: 2
  reliability: 0
topic_b:
  history: 1
  durability: 2
topic_a:
  history: 2
  durability: 1
```

will be loaded as:

```python
map = {
  'topic_a': {'history': 2, 'durability': 1},
  'topic_b': {'history': 1, 'durability': 2}
}
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +91,34 @@ def create_bag_directory(self, uri):
...
+            return Duration()
+
+    def validate_qos_profile_overrides(self, qos_profile_path: str) -> Dict[str, Dict]:
+        """Validate the QoS profile yaml file path and its structure."""
+        if not os.path.isfile(qos_profile_path):
+            raise ValueError('[ERROR] [ros2bag]: {} does not exist.'.format(qos_profile_path))
+        with open(qos_profile_path, 'r') as file:
+            qos_profile_dict = yaml.safe_load(file)
+        topic_names = list(qos_profile_dict)
+        for name in topic_names:
```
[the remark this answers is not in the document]
piraka9011: Automagically resolved by taking a break 🙃

ros2bag/ros2bag/verb/record.py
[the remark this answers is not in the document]
piraka9011: Offline discussion: Not much we can do here so we'll keep as is.

ros2bag/package.xml
```diff
@@ -17,6 +17,8 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <test_depend>launch_testing</test_depend>
+  <test_depend>launch_testing_ros</test_depend>
```
thomas-moulard, line 21: alphasort

ros2bag/ros2bag/verb/record.py
```diff
@@ -13,11 +13,29 @@
...
 from ros2cli.node import NODE_NAME_PREFIX
+import yaml
+
+
+POLICY_MAP = {
+    'history': rclpy.qos.QoSHistoryPolicy.get_from_short_key,
+    'reliability': rclpy.qos.QoSReliabilityPolicy.get_from_short_key,
+    'durability': rclpy.qos.QoSDurabilityPolicy.get_from_short_key,
+    'liveliness': rclpy.qos.QoSLivelinessPolicy.get_from_short_key
+}
```
thomas-moulard, line 35: How do we make sure those two will stay in-sync? Can't we just get the user to pass the short key?

ros2bag/ros2bag/verb/record.py
```diff
@@ -71,7 +89,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         )
         parser.add_argument(
             '--include-hidden-topics', action='store_true',
-            help='record also hidden topics.')
+            help='record also hidden topics.'
+        )
+        parser.add_argument(
+            '--qos-profile-overrides', type=str, default='',
```
thomas-moulard, line 95: `--qos-profile-overrifes-path` - please make explicit to the user that this is a "path", otherwise this is confusing.

ros2bag/ros2bag/verb/record.py
```diff
@@ -71,7 +89,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            help='Path to a yaml file defining overrides of the QoS profile for specific topics.'
+                 'See docs/qos_profiles_overrides.md for more info.'
```
thomas-moulard, line 97: Once the software is installed, this path won't exist, so this is not helpful.

ros2bag/ros2bag/verb/record.py
thomas-moulard, line 95: The correct type here is `argparse.FileType('r')`.
See https://docs.python.org/3/library/argparse.html#argparse.FileType

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +104,32 @@ def create_bag_directory(self, uri):
...
 
+    def _dict_to_duration(self, time_dict: Optional[Dict[str, int]]) -> Duration:
+        if not all(key in time_dict for key in ['sec', 'nsec']):
+            raise ValueError(
+                'Time overrides must include both seconds (sec) and nanoseconds (nsec).')
+        if time_dict:
+            return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
+        else:
+            return Duration()
+
```
thomas-moulard, line 115: Move to top-level as this does not use `self`

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +104,32 @@ def create_bag_directory(self, uri):
         except OSError:
             return "[ERROR] [ros2bag]: Could not create bag folder '{}'.".format(uri)
 
+    def _dict_to_duration(self, time_dict: Optional[Dict[str, int]]) -> Duration:
+        if not all(key in time_dict for key in ['sec', 'nsec']):
```
thomas-moulard, line 108: The logic in this function is really confusing. Let's remove all those negations:

```python

  if time_dict:
    if all(...):
       return ...
    else:
      raise ...
  else:
    return Duration()
```

thomas-moulard: Additionally, consider moving your `all(...)` into another function to make the logic clearer.

`def _is_dict_valid_duration(dict) -> bool`

ros2bag/ros2bag/verb/record.py
```diff
@@ -80,6 +104,32 @@ def create_bag_directory(self, uri):
...
+        for name in qos_profile_dict.keys():
+            profile = qos_profile_dict[name]
+            # Convert dict to Duration. Required for construction
+            conversion_keys = ['deadline', 'lifespan', 'liveliness_lease_duration']
+            for k in conversion_keys:
+                profile[k] = self._dict_to_duration(profile.get(k))
+            for policy in POLICY_MAP.keys():
+                profile[policy] = POLICY_MAP[policy](profile.get(policy, 'system_default'))
+            qos_profile_dict[name] = QoSProfile(**profile)
+        return qos_profile_dict
```
thomas-moulard, line 131: Move to top-level, you don't use `self` here.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -32,6 +35,29 @@
 #include "rosbag2_transport/storage_options.hpp"
 #include "rmw/rmw.h"
 
+namespace
+{
+/// Convert a Python3 unicode string to a native C++ std::string
+std::string PyObject_AsStdString(PyObject * object)
+{
+  PyObject * python_string;
```
thomas-moulard, line 43: Initialize to nullptr!

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -52,6 +78,7 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
     "max_cache_size",
     "topics",
     "include_hidden_topics",
+    "qos_profile_overrides",
```
thomas-moulard, line 81: alphasort?

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -105,6 +134,21 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
     rosbag2_compression::compression_mode_from_string(record_options.compression_mode)
   };
 
+  if (PyDict_Check(qos_profile_overrides)) {
+    PyObject * key, * value;
```
thomas-moulard, line 138: nullptr!

ros2bag/ros2bag/verb/record.py
piraka9011: It was a previous suggestion to fail early than later, which is why we have that negation statement upfront.

thomas-moulard: How is this making a difference?

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
thomas-moulard: Sure, sgtm. Please link the PR back here once it's done.

ros2bag/ros2bag/verb/record.py
piraka9011: We can have the user specify in the overrid yaml file the enum integer for each policy (ex. `reliability: 0` which is the same as `reliability: system_default`) but that's not good UX.
I don't think there's a 'better' way to get these values beside using these helper methods provided by each policy.
Maybe using [`importlib`](https://docs.python.org/3/library/importlib.html) to import all modules with the name `*Policy` in `rclpy.qos` but that's hacky IMO.

thomas-moulard: I see, we should at least add a comment explaining this needs to be updated when rclpy.qos is updated.
Consider also using a `frozenset()` here as you don't really need a map to store `x => x.foo` for multiple values of `x`, unless you plan to have elements that would break this pattern, but it's probably YAGNI.

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/b0eaa784659cc6e98bb09ea175d97367/raw/f99f2a68258a11bac3d4e9c09ff6e723d6d05ee1/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_transport
TEST args: --packages-select ros2bag rosbag2_transport
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9957/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5564/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8101/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9827/)

piraka9011:
@emersonknapp Looks like the same issue as before [here](https://github.com/ros2/rosbag2/pull/346#issuecomment-610582670). Let me know when it's ready to merge (or feel free to do so).

PR #349, opened 2020-04-03
Test all RMW implementations for rosbag2_transport

emersonknapp:
Depends on https://github.com/ament/ament_cmake/pull/236 - there is a bug in `ament_add_gmock` that shows up when I start using `ENV`

Iterate over all RMW implementations for the rosbag2_transport tests, because it's been the case recently that some tests will pass on fastrtps but fail on cyclonedds.

### DIFF
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -28,6 +28,7 @@ find_package(rcutils REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rosbag2_compression REQUIRED)
 find_package(rosbag2_cpp REQUIRED)
+find_package(rmw_implementation_cmake REQUIRED)
 find_package(shared_queues_vendor REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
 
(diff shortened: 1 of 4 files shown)


rosbag2_transport/CMakeLists.txt
```diff
@@ -84,63 +85,72 @@ ament_export_interfaces(export_${PROJECT_NAME})
...
-  find_package(ament_cmake_gmock REQUIRED)
-  find_package(ament_index_cpp REQUIRED)
-  find_package(ament_lint_auto REQUIRED)
-  find_package(test_msgs REQUIRED)
-  find_package(rosbag2_test_common REQUIRED)
-  # explicitly disable cppcheck
-  set(ament_cmake_cppcheck_FOUND TRUE)
-  ament_lint_auto_find_test_dependencies()
+function(create_tests_for_rmw_implementation)
+  message(STATUS "Creating tests for '${rmw_implementation}'")
```
thomas-moulard, line 89: drop?

rosbag2_transport/CMakeLists.txt
```diff
@@ -84,63 +85,72 @@ ament_export_interfaces(export_${PROJECT_NAME})
...
-  if(TARGET test_info)
-    target_link_libraries(test_info rosbag2_transport)
-    ament_target_dependencies(test_info rosbag2_test_common)
+    ENV "${rmw_implementation_env_var}"
+    WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
+  if(TARGET ${test_info})
+    target_link_libraries(${test_info} rosbag2_transport)
+    ament_target_dependencies(${test_info}
+      ${rmw_implementation} rosbag2_test_common)
   endif()
```
thomas-moulard, line 101: Let's extract this logic into a single macro like `declare_rosbag2_test(test_info, test/rosbag2_transport/test_info.cpp)` or something? The current diff is copying and pasting the same logic everywhere.

rosbag2_transport/CMakeLists.txt
```diff
@@ -149,45 +159,68 @@ if(BUILD_TESTING)
...
-    target_link_libraries(test_play rosbag2_transport)
-    ament_target_dependencies(test_play test_msgs rosbag2_test_common)
+  if(TARGET ${test_play})
+    target_link_libraries(${test_play} rosbag2_transport)
+    ament_target_dependencies(${test_play}
+      ${rmw_implementation} test_msgs rosbag2_test_common)
+  endif()
+endfunction()
+
+macro(test_rmw_implementation_wrapper)
```
thomas-moulard, line 194: macro -> function? Do you need to export things out of the scope of the macro? If not, a function is preferable as it defines a scope and avoids naming clash.

rosbag2_transport/CMakeLists.txt
emersonknapp: fair enough. done

rosbag2_transport/CMakeLists.txt
emersonknapp: was able to  skip this entirely. i was copying `rcl`'s tests but they actually import the rmw implementation, we don't have to

emersonknapp:
Ok - I moved it to a separate file and added documentation

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9935/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5545/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9784/)

PR #350, opened 2020-04-03
Remove assumptions and checks around QoS policies that don't affect compatibility

emersonknapp:
Part of #125 
Depends on https://github.com/ros2/rclcpp/pull/1049

Some of the policies in QoS profiles don't affect compatibility, only local behavior, so it's not worth comparing those values. The different RMW implementations report these values inconsistently currently, so though it is useful to record everything, it is not meaningful to test for specific values.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/qos.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/qos.cpp
@@ -82,21 +82,21 @@ bool operator==(const rmw_time_t & left, const rmw_time_t & right)
 
 namespace rosbag2_transport
 {
-  // TODO(emersonknapp) it is the long term goal to have similar logic exist within rmw/rcl/rclcpp,
-  //  but it's not yet determined where that logic will go. We should remove this function before
-  //  G-Turtle
-  bool Rosbag2QoS::compatibility_policies_exactly_equal(const rclcpp::QoS & other) const
-  {
-    // Note that these policies do not affect QoS compatibility
-    // * History + Depth (relevant only for local behavior)
-    // * Lifespan (only relevant locally on Publishers)
-    // * avoid ros namespace conventions (this affects choice of topic names, not quality of service)
-    const auto & pl = get_rmw_qos_profile();
-    const auto & pr = other.get_rmw_qos_profile();
-    return pl.reliability == pr.reliability &&
-           pl.durability == pr.durability &&
-           pl.deadline == pr.deadline &&
-           pl.liveliness == pr.liveliness &&
-           pl.liveliness_lease_duration == pr.liveliness_lease_duration;
-  }
+// TODO(emersonknapp) it is the long term goal to have similar logic exist within rmw/rcl/rclcpp,
+//  but it's not yet determined where that logic will go. We should remove this function before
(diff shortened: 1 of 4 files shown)


thomas-moulard:
I think keeping a trace of how things have been recorded is valuable for debugging later. If we have this feature already, I find it a bit counter-productive to throw this data away.

Example use case: "rosbag2 has troubles recording everything, and is dropping messages? oh yeah the history depth was set to 1 erroneously, let's change this behavior". I'm not sure if we already have a rosbag2 info like in ros1, but that's something we could pretty print there.

I understand that some RMW implementations may not be "there yet" in terms of reporting data correctly, but we should continue pushing them to get better, instead of trying to workaround their quirks, unless there is a clear loss for our customers.

I think what's missing maybe from your PR message, is why you're doing this? What problem are you fixing here for users?

emersonknapp:
@thomas-moulard yeah I think you're right. My main problem had to do with checking for equality. The real solution here is not to check for exact values, but continue recording them as yes, they are potentially-useful information.

rosbag2_transport/src/rosbag2_transport/qos.hpp
```diff
@@ -46,6 +46,13 @@ class Rosbag2QoS : public rclcpp::QoS
     keep_last(rmw_qos_profile_default.depth);
     return *this;
   }
+
+  /// Determine if all policies that affect QoS compatibility are equal.
+  /*
+    Note that this method does not check if the QoS policies are compatible.
+    This is a simple check for exact equality in all policies that affect compatibility.
+  */
```
zmichaels11, line 54: I don't think this is the proper comment formatting

rosbag2_transport/src/rosbag2_transport/qos.cpp
```diff
@@ -68,3 +68,35 @@ bool convert<rosbag2_transport::Rosbag2QoS>::decode(
...
+}
+}  // unnamed namespace
+
+
+namespace rosbag2_transport
+{
+// TODO(emersonknapp) it is the long term goal to have similar logic exist within rmw/rcl/rclcpp,
+//  but it's not yet determined where that logic will go. We should remove this function before
+//  G-Turtle
+bool Rosbag2QoS::compatibility_policies_exactly_equal(const rclcpp::QoS & other) const
```
piraka9011, line 88: Does it make sense to have this as an `==` operator instead?

emersonknapp: I think it's too specific to use ==, I don't want anyone accidentally thinking it's full equality

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5555/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8093/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9811/)

emersonknapp:
I am closing this PR in favor of a more complete approach that will cause better behavior on average (with a more accurate branch name)

PR #351, opened 2020-04-04
replace info with launch testing

### LINKED ISSUE #271: Resolve workarounds in rosbag2_tests for Windows
`test_rosbag2_record_end_to_end` defines a workaround for Windows in multiple places due to Windows not supporting a mechanism to programmatically send a SIGINT.

These tests should either be rewritten to not require the workaround or a helper function should be written so less platform-specific code is required.

Karsten1987:
fixes #271 by replacing the end to end tests in `rosbag2_tests` with launch_testing.


rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
```diff
@@ -0,0 +1,93 @@
...
+            assert command.exit_code == launch_testing.asserts.EXIT_OK
+        # Using the SequentialStdout context manager asserts that the following stdout
+        # happened in the same order that it's checked
+        with launch_testing.asserts.assertSequentialStdout(proc_output, rosbag_info_process) as cm:
+            cm.assertInStdout('Files:             cdr_test.db3')
+            # cm.assertInStdout('Bag size:          .*B')
+            cm.assertInStdout('Storage id:        sqlite3')
+            cm.assertInStdout('Duration:          0.155s')
+            # cm.assertInStdout('Start:             Sep 18 2018 .*:.*:44.241 (1537282604.241)')
+            # cm.assertInStdout('End                Sep 18 2018 .*:.*:44.397 (1537282604.397)')
```
Karsten1987, line 73: These values need to be regex'ed because they are displayed slightly differently on Windows.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
```diff
@@ -0,0 +1,93 @@
...
+            cm.assertInStdout('Messages:          7')
+            cm.assertInStdout(
+                ('Topic information: Topic: /test_topic | Type: test_msgs/BasicTypes '
+                 '| Count: 3 | Serialization Format: cdr'))
+            cm.assertInStdout(
+                ('Topic: /array_topic | Type: test_msgs/Arrays '
+                 '| Count: 4 | Serialization Format: cdr'))
+        # TODO(Karsten1987): Regex doesn't seem to work.
+        # launch_testing.asserts.assertInStdout(
+        #     proc_output, 'Files:\s+cdr_test.db3', rosbag_info_process)
```
Karsten1987, line 83: in connection to the above comments, https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/asserts/assert_output.py#L46 mentions support for regex, but I wasn't able to find out a syntax for it.

Karsten1987:
to get an idea about windows:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9947/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5554/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8092/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9810/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9949/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5556/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9812/)

Karsten1987:
so this doesn't sound good:
```
18:38:53 3: [WARNING] [ros2-1]: 'SIGINT' sent to process[ros2-1] not supported on Windows, escalating to 'SIGTERM'
18:38:53 3: [INFO] [ros2-1]: sending signal 'SIGTERM' to process[ros2-1]
```

@hidmic the `test_play_end_to_end.py` is timing out. It might be related to the above warning/error. Do you have any idea on how to work around this?

Any help is appreciated. I'd like to have this sorted out before digging deeper into porting all tests to launch testing.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
```diff
@@ -0,0 +1,91 @@
...
+             r'Storage id:\s+sqlite3\n'
+             r'Duration:\s+0.155s\n'
+             r'Start:\s+Sep 18 2018 \d{2}:\d{2}:44.241 \(1537282604.241\)\n'
+             r'End:\s+Sep 18 2018 \d{2}:\d{2}:44.397 \(1537282604.397\)\n'
+             r'Messages:\s+7\n'
+             r'Topic information:\s+Topic: /test_topic \| Type: test_msgs/BasicTypes '
+             r'\| Count: 3 \| Serialization Format: cdr\n'
+             r'\s+Topic: /array_topic \| Type: test_msgs/Arrays '
+             r'\| Count: 4 \| Serialization Format: cdr'))
+        launch_testing.asserts.assertInStdout(proc_output, p, rosbag_info_process)
```
hidmic, line 77: @Karsten1987 FYI you can also assert on `command.output`, and there's [`launch_testing.tools.output.expect_output`](https://github.com/ros2/launch/blob/0ce19ac6b9e02e93d58ba4ddfd31e3a2d3fac36f/launch_testing/launch_testing/tools/output.py#L87) to aid that if you want.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
```diff
@@ -0,0 +1,91 @@
...
+
+class TestInfo(unittest.TestCase):
+
+    def test_ros2_bag_info(self, launch_service, proc_info, proc_output, rosbag_info_process):
+        """Test terminating_proc without command line arguments."""
+        print('Running ros2 bag info')
+        with launch_testing.tools.launch_process(
+              launch_service, rosbag_info_process, proc_info, proc_output) as command:
+            assert command.wait_for_shutdown(timeout=3)
+            assert command.exit_code == launch_testing.asserts.EXIT_OK
```
hidmic, line 64: @Karsten1987 nit: I'd move this statement outside the context manager scope

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_ros2_bag_record(self, launch_service, proc_info, proc_output, proc_env):
+        """Test terminating_proc without command line arguments."""
+        print('Running ros2 bag record')
+
+        dir_path = 'rosbag2_tmp_bag'
+        try:
+            shutil.rmtree(dir_path)
+        except OSError as e:
+            print("Error: %s : %s" % (dir_path, e.strerror))
```
hidmic, line 64: @Karsten1987 you may want to use [`setUp()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) and [`tearDown()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown) methods to create and destroy temporary resources.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.py
```diff
@@ -0,0 +1,87 @@
...
+        rosbag_record_process = launch.actions.ExecuteProcess(
+            cmd=['ros2', 'bag', 'record', '/rosbag2_record_topic1', '/rosbag2_record_topic2', '-o', dir_path],
+            output='screen', env=proc_env,
+        )
+        with launch_testing.tools.launch_process(
+              launch_service, rosbag_record_process, proc_info, proc_output) as command:
+            time.sleep(10)
+        proc_info.assertWaitForShutdown(rosbag_record_process, timeout=5)
```
hidmic, line 72: @Karsten1987 FYI you can use `command.wait_for_shutdown` too.

hidmic:
> @hidmic the `test_play_end_to_end.py` is timing out. It might be related to the above warning/error. Do you have any idea on how to work around this?

Hmm, I looked at it for a while but I don't see how it got into that state. Exit codes are unrelated. Can you reproduce in a Windows box? If you run `launch_test` with `-v` that may help us understand what is happening.

wjwwood:
> @wjwwood do you recall why CTRL_C_EVENT was never an option in Windows? I recall it wouldn't work right away, but that's as far as my memory goes.

No I don’t remember the details, only that it is asymmetric between windows and Linux.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
Karsten1987: how do I access the exit code if the `command` is out of scope?

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
Karsten1987: but `expect_output` does not accept a regex pattern, does it?

hidmic: Yeap, it does. All these `assertIn*` functions use `expect_output` under the hood.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.py
hidmic: `command` doesn't leave scope after the `with` statement is over :)

Karsten1987:
Are you guys aware of any workarounds for the `ctrl-c` dilemma? I just played around locally with stuff like shown here: https://gist.github.com/rdp/f51fb274d69c5c31b6be

Which works pretty decently and rosbag behaves accurately, meaning it receives the `ctrl-c` and writes the bagfiles correctly. I found the equivalent in python using `ctypes`:
https://stackoverflow.com/a/58802146/4583130

Would you have an idea if that route is somewhat feasible?

Karsten1987:
After talking with @hidmic I've tried a couple of more things (`CTRL_C_EVENT` instead of `SIGTERM`, playing around with `creationflags` (https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags)) but that all seemed to lead into a deep rabbit hole which is most likely not worth it.

In order to stabilize the end to end tests, I have a couple of ideas I'd like to run by you (i.e. @emersonknapp, @piraka9011, @zmichaels11). 
* introduce a new CLI option which dictates the duration of recording (could possibly also be used for replaying). That would allow the record e2e tests to automatically stop after `n` seconds and we wouldn't need to rely on a custom `ctrl-c` event.
* convert the internal `Rosbag2Node` into a lifecycle node. I think this could help us overcome parts of the flaky discovery problems we encounter especially in the replay tests. In the configure step of the lifecycle node, we could prepare and advertise the publishers on the system. Likewise, the subscriptions in the test can be started. After that, the activate step really only triggers the call to publish. 
* similar to the approach above, I've started work a while ago on a CLI option called `--paused` where the rosbag would start to prepare everything, but don't publish. Either by a service call or `spacebar` the replay would start. As a gut feeling, I am in favor of the lifecycle node though.

Any opinions?

piraka9011:
I too played around with this ctrl-c business on windows (See [this branch](https://github.com/aws-ros-dev/rosbag2/tree/demo-ctrl-c), specifically, this [file](https://github.com/aws-ros-dev/rosbag2/blob/224d5079d6baa39c274e7ca16763baf69cb5c3a2/rosbag2_tests/test/rosbag2_tests/test_compression_with_splitting.py)). However, my approach kills `colcon test` when it attempts to raise `ctrl-c` (I opened an issue [here](https://github.com/colcon/colcon-core/issues/303)).

I came to the conclusion that the right way to go about this is to introduce keyboard commands to control the rosbag2 node, specifically things like `s`/`stop`, `q`/`quit` and `p`/`pause`.

The lifecycle node approach is also good, it allows a user to programmatically control the rosbag2 node. I think keyboard/cli/lifecycle node approaches are all valid and we should decide what is the most useful approach (probably lifecycle).

piraka9011:
FYI, Also played around with the windows creation flags, trying out different combinations one by one...
Unfortunately it did not resolve the issue 🙃

zmichaels11:
Instead of invoking the ros2 cli, can the tests instead just invoke the commands that the cli runs?

Karsten1987:
>I think keyboard/cli/lifecycle node approaches are all valid and we should decide what is the most useful approach (probably lifecycle).

I don't see them as mutual exclusive. I could see how the lifecycle generally offers a way to steer the communication and the keyboard inputs are really only a trigger to the state transitions of that lifecycle. That would make sense to me.

> However, my approach kills colcon test when it attempts to raise ctrl-c

I ran into the same thing. The closest I came to a solution was trying to use `CREATE_NEW_CONSOLE`, but this also didn't really yield any results.

> Instead of invoking the ros2 cli, can the tests instead just invoke the commands that the cli runs?

I guess we could tackle the e2e tests starting from `rosbag2_transport`. Cause, really, the CLI is only forwarding the command line options to the constructor of the transport class.

PR #353, opened 2020-04-06
QoS Profile Overrides - Player

piraka9011:
* Add playback option to override QoS profiles for topics.
* Enable `SusbcriptionManager` to accept a QoS profile.

Part of #125

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
@@ -43,7 +43,7 @@ public:
   void add_subscription(
     const std::string & topic_name,
     size_t expected_number_of_messages,
-    const rclcpp::QoS & qos = rclcpp::QoS{rclcpp::KeepAll()})
+    const rclcpp::QoS & qos = rclcpp::QoS{rmw_qos_profile_default.depth})
   {
     expected_topics_with_size_[topic_name] = expected_number_of_messages;
 
--- a/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
@@ -40,13 +40,11 @@ public:
   }
 
   template<typename MessageT>
-  void add_subscription(const std::string & topic_name, size_t expected_number_of_messages)
+  void add_subscription(
+    const std::string & topic_name,
+    size_t expected_number_of_messages,
+    const rclcpp::QoS & qos = rclcpp::QoS{rclcpp::KeepAll()})
   {
-    auto qos = rclcpp::QoS(rclcpp::KeepAll());
-    qos.get_rmw_qos_profile().depth = 4;
-    qos.reliability(RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT);
-    qos.durability(RMW_QOS_POLICY_DURABILITY_SYSTEM_DEFAULT);
-    qos.avoid_ros_namespace_conventions(false);
     expected_topics_with_size_[topic_name] = expected_number_of_messages;
 
     auto options = rclcpp::SubscriptionOptions();
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -154,15 +154,22 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
   }
 }
 
+rclcpp::QoS Player::publisher_qos_for_topic(const std::string & topic_name)
+{
+  auto qos_it = topic_qos_profile_overrides_.find(topic_name);
+  if (qos_it != topic_qos_profile_overrides_.end()) {
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic_name);
+    return Rosbag2QoS{topic_qos_profile_overrides_.at(topic_name)};
+  } else {
+    return Rosbag2QoS{};
+  }
+}
+
 void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
-    Rosbag2QoS topic_qos{};
-    if (topic_qos_profile_overrides_.count(topic.name)) {
-      topic_qos = Rosbag2QoS{topic_qos_profile_overrides_.at(topic.name)};
-      ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic.name);
-    }
+    rclcpp::QoS topic_qos = publisher_qos_for_topic(topic.name);
     publishers_.insert(
       std::make_pair(
         topic.name, rosbag2_transport_->create_generic_publisher(
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -59,6 +59,7 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play(const PlayOptions & options)
 {
+  topic_qos_profile_overrides_ = options.topic_qos_profile_overrides;
   prepare_publishers();
 
   storage_loading_future_ = std::async(
@@ -157,10 +158,14 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
+    rclcpp::QoS topic_qos{10};
+    if (topic_qos_profile_overrides_.count(topic.name)) {
+      topic_qos = topic_qos_profile_overrides_.at(topic.name);
+    }
     publishers_.insert(
       std::make_pair(
         topic.name, rosbag2_transport_->create_generic_publisher(
-          topic.name, topic.type, Rosbag2QoS{})));
+          topic.name, topic.type, topic_qos)));
   }
 }
 
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -158,9 +158,9 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
-    rclcpp::QoS topic_qos{rclcpp::KeepAll()};
+    Rosbag2QoS topic_qos{};
     if (topic_qos_profile_overrides_.count(topic.name)) {
-      topic_qos = topic_qos_profile_overrides_.at(topic.name);
+      topic_qos = Rosbag2QoS{topic_qos_profile_overrides_.at(topic.name)};
       ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic.name);
     }
     publishers_.insert(
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -158,9 +158,10 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
-    rclcpp::QoS topic_qos{10};
+    rclcpp::QoS topic_qos{rclcpp::KeepAll()};
     if (topic_qos_profile_overrides_.count(topic.name)) {
       topic_qos = topic_qos_profile_overrides_.at(topic.name);
+      ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic.name);
     }
     publishers_.insert(
       std::make_pair(
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -18,7 +18,7 @@
 #include <memory>
 #include <queue>
 #include <string>
-#include <vector>
+#include <unordered_map>
 #include <utility>
 
 #include "rcl/graph.h"
@@ -35,6 +35,31 @@
 #include "qos.hpp"
 #include "rosbag2_node.hpp"
 #include "replayable_message.hpp"
+
+namespace
+{
+/**
+ * Determine which QoS to offer for a topic.
+ * The priority of the profile selected is:
+ *   1. The override specified in play_options (if one exists for the topic).
+ *   2. The default RMW QoS profile.
+ *
+ * \param topic_name The full name of the topic, with namespace (ex. /arm/joint_status).
+ * \param topic_qos_profile_overrides A map of topic to QoS profile overrides.
+ * @return The QoS profile to be used for subscribing.
+ */
+rclcpp::QoS publisher_qos_for_topic(
+  const std::string & topic_name,
+  std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides)
+{
+  auto qos_it = topic_qos_profile_overrides.find(topic_name);
+  if (qos_it != topic_qos_profile_overrides.end()) {
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic_name);
+    return rosbag2_transport::Rosbag2QoS{qos_it->second};
+  }
+  return rosbag2_transport::Rosbag2QoS{};
+}
+}  // namespace
 
 namespace rosbag2_transport
 {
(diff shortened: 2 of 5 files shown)


rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,58 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+{
+  const auto topic1 = "topic1";
+  const auto topic2 = "topic2";
+  const auto msg_type = "test_msgs/BasicTypes";
+  auto basic_msg = get_messages_basic_types()[0];
+  basic_msg->int32_value = 42;
+
+  auto topic_types = std::vector<rosbag2_storage::TopicMetadata>{
+    {topic1, msg_type, "", ""},
+    {topic2, msg_type, "", ""},
```
thomas-moulard, line 134: Please document what each field represents, as it's not really readable right now.

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,58 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+  };
+
+  std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> messages =
+  {serialize_test_message(topic1, 500, basic_msg),
```
thomas-moulard, line 138: Why 500, 700? meaning?

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,58 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+
+  // Due to a problem related to the subscriber, we play many (3) messages but make the subscriber
+  // node spin only until 2 have arrived. Hence the 2 as `launch_subscriber()` argument.
+  sub_->add_subscription<test_msgs::msg::BasicTypes>("/topic1", 2);
+  sub_->add_subscription<test_msgs::msg::BasicTypes>("/topic2", 2);
+
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  const auto profile1 = rclcpp::QoS{0}.keep_all().durability_volatile();
+  const auto profile2 = rclcpp::QoS{0}.keep_last(10).reliable();
```
thomas-moulard, line 157: why those options?

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,58 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+    std::pair<std::string, rclcpp::QoS>{"/topic2", profile2}
+  };
+  play_options_.topic_qos_profile_overrides = topic_qos_profile_overrides;
+
+  Rosbag2Transport rosbag2_transport(reader_, writer_, info_);
+  rosbag2_transport.play(storage_options_, play_options_);
+
+  await_received_messages.get();
+
+  auto received_messages_1 = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
```
thomas-moulard, line 169:
```suggestion
  const auto received_messages_1 = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,58 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+    "/topic1");
+
+  auto received_messages_2 = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
```
thomas-moulard, line 172:
```suggestion
  const auto received_messages_2 = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -157,10 +158,14 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
+    rclcpp::QoS topic_qos{10};
```
thomas-moulard, line 161: 10?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -157,10 +158,15 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
+    rclcpp::QoS topic_qos{rclcpp::KeepAll()};
```
thomas-moulard, line 161: KeepAll of default profile?

piraka9011: It's for the QoS History setting. Constructing a new `rclcpp::QoS` profile requires either `KeepAll()` or `KeepLast(n)`. Everything is set to the ros2 default.

thomas-moulard: Don't we have a default value for `KeepLast`? Or do we need to respecify that every time?

thomas-moulard: rosbag2_transport::Rosbag2QoS should be used AFAICT

rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
```diff
@@ -40,13 +40,11 @@ class SubscriptionManager
   }
 
   template<typename MessageT>
-  void add_subscription(const std::string & topic_name, size_t expected_number_of_messages)
+  void add_subscription(
+    const std::string & topic_name,
+    size_t expected_number_of_messages,
+    const rclcpp::QoS & qos = rclcpp::QoS{rclcpp::KeepAll()})
```
Karsten1987, line 46: Does systems default not make sense here? or does `KeepAll()` represent all values you removed in this particular change?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -157,10 +158,15 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
+    Rosbag2QoS topic_qos{};
+    if (topic_qos_profile_overrides_.count(topic.name)) {
```
Karsten1987, line 162: consider using `find`.
```
auto qos_it = topic_qos_profile_overrides_.find(topic.name);
if (qos_it != topic_qos_profile_overrides_.end()) {
  publishers_.insert(..., *qos_it);
} else {
  publishers_.insert(..., Rosbag2QoS{})
}
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -119,3 +120,49 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+  reader_ = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
+
+  const auto qos_request = rclcpp::QoS{rclcpp::KeepAll()}.durability_volatile();
+  sub_->add_subscription<test_msgs::msg::BasicTypes>(topic_name, num_msgs, qos_request);
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  // The previous subscriber requested durability VOLATILE which is the default in rosbag2.
+  // We override the requested durability to TRANSIENT_LOCAL so that we can receive messages.
+  // If the previous subscription requested TRANSIENT_LOCAL and we overrode with VOLATILE, then we
+  // would not receive any messages.
```
Karsten1987, line 153: is it feasible to test this? Saying we set up a conflicting setup and see that after a specific amount of time, no messages have been received.

rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
emersonknapp: This has come up a few times - "SystemDefault" has undefined behavior and shouldn't be used. It's not the same as "Default" - which we do define as `rosbag2_transport::Rosbag2QoS{}` which is what should be used here

Karsten1987: Sorry, didn't read through the rest of the reviews here on this ticket. If the default shouldn't be used, can we escalade this to `rclcpp` where it's defined? Or does it really only have UB in the context of rosbags?

emersonknapp: "SYSTEM_DEFAULT" in the rmw API means "let the rmw implementation use its own default values", theoretically the DDS spec defines defaults but the RMW API does not define or enforce that they are consistent across implementations. `rmw_qos_profile_system_default` is a different value than `rmw_qos_profile_default`. We are wrapping `rmw_qos_profile_default` in `Rosbag2QoS` with a default constructor and a few other rosbag2-specific utilities.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -157,10 +158,15 @@ void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
+    Rosbag2QoS topic_qos{};
```
emersonknapp, line 161: can you move this logic out into a `qos_for_publisher(topic_name)` function like in the recorder? I'll have similar adaptive logic to add to it

rosbag2_transport/test/rosbag2_transport/test_play.cpp
emersonknapp, line 153: We can add "wait for condition" checks with timeouts to check that "something doesn't happen". My only complaint with these tests is that they take a minimum of.. say 3 seconds as a reasonable timeout, so the test suite starts to take a long time when we have a lot of them. I can't think of a better way, though, given the mechanics of DDS discovery...

piraka9011: It should be possible.
I've also been trying to identify some flakiness in fastrtps from this test.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -154,15 +154,22 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
   }
 }
 
+rclcpp::QoS Player::publisher_qos_for_topic(const std::string & topic_name)
+{
+  auto qos_it = topic_qos_profile_overrides_.find(topic_name);
+  if (qos_it != topic_qos_profile_overrides_.end()) {
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic_name);
+    return Rosbag2QoS{topic_qos_profile_overrides_.at(topic_name)};
```
zmichaels11:
```suggestion
    return Rosbag2QoS{qos_it->second};
```
you've already searched and found it

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -154,15 +154,22 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
 void Player::prepare_publishers()
 {
   auto topics = reader_->get_all_topics_and_types();
   for (const auto & topic : topics) {
-    Rosbag2QoS topic_qos{};
-    if (topic_qos_profile_overrides_.count(topic.name)) {
-      topic_qos = Rosbag2QoS{topic_qos_profile_overrides_.at(topic.name)};
-      ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic.name);
-    }
+    rclcpp::QoS topic_qos = publisher_qos_for_topic(topic.name);
```
zmichaels11, line 172:
```suggestion
   auto topic_qos = publisher_qos_for_topic(topic.name);
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -153,16 +160,44 @@ TEST_F(RosBag2PlayTestFixture, topic_qos_profiles_overriden)
...
+  // The previous subscriber requested durability TRANSIENT_LOCAL.
+  // We override the requested durability to VOLATILE.
+  // Since they are incompatible policies, we will not receive any messages.
+  const auto qos_override = rclcpp::QoS{rclcpp::KeepAll()}.durability_volatile();
+  const auto topic_qos_profile_overrides = std::unordered_map<std::string, rclcpp::QoS>{
+    std::pair<std::string, rclcpp::QoS>{topic_name_, qos_override},
+  };
+  play_options_.topic_qos_profile_overrides = topic_qos_profile_overrides;
+
+  Rosbag2Transport rosbag2_transport(reader_, writer_, info_);
```
zmichaels11, line 192:
```suggestion
  Rosbag2Transport rosbag2_transport{reader_, writer_, info_};
```

rosbag2_transport/src/rosbag2_transport/player.cpp
piraka9011: Yup, I realized that when going through the PR again :sweat_smile:

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -59,6 +61,15 @@ class Player
   void play_messages_until_queue_empty(const PlayOptions & options);
   void prepare_publishers();
 
+  /**
+   * Determine which QoS to offer for a topic.
+   * Overrides the QoS for a topic if one is specified in the play_options.
```
Karsten1987, line 66: Does it though? As far as I can see this function only returns a QOS, but doesn't do any further actions such as overriding anything.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -153,14 +154,26 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
+rclcpp::QoS Player::publisher_qos_for_topic(const std::string & topic_name)
+{
+  auto qos_it = topic_qos_profile_overrides_.find(topic_name);
+  if (qos_it != topic_qos_profile_overrides_.end()) {
+    ROSBAG2_TRANSPORT_LOG_INFO_STREAM("Overriding QoS profile for topic " << topic_name);
+    return Rosbag2QoS{qos_it->second};
+  } else {
+    return Rosbag2QoS{};
+  }
+}
```
emersonknapp, line 166: nit: instead of putting in an else block it's better to make this final fall-through case stand alone
```suggestion
  } 
  return Rosbag2QoS{};
}
```

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -59,6 +61,17 @@ class Player
...
+  /**
+   * Determine which QoS to offer for a topic.
+   * The priority of the profile selected is:
+   *   1. The override specified in play_options (if one exists for the topic).
+   *   2. The default RMW QoS profile.
+   *
+   * \param topic_name The full name of the topic, with namespace (ex. /arm/joint_status).
+   * @return The QoS profile to be used for subscribing.
+   */
+  rclcpp::QoS publisher_qos_for_topic(const std::string & topic_name);
```
Karsten1987, line 73: I've just seen that this function is private. You might want to consider replacing this with a free function, taking two arguments (the overrides and topic name).

piraka9011: The overrides are specified in the `play_options` which I think is a suitable entrypoint for users. I don't think we'd want to expose the internals here otherwise (Unless a user want's to change QoS policies dynamically...?)

emersonknapp: I'd agree with Anas that this function is not useful outside the context of the player. It only contains logic that deals with the internal mechanics of the Player. I'll be having it call out to an adaptive function in the case an override is not provided, but that adaptive logic will live in a separate utility.

Karsten1987: I guess I wasn't clear. I would still keep the (free) function within context of player. But only have it defined in the `player.cpp` and not exposed in the header. Basically for the same reasons you all said, that it doesn't make much sense for users to call this function (which they couldn't anyway since it's private). 

The only class member variable this function is using is `topic_qos_profile_overrides_` which could easily be passed into the free function.

emersonknapp: Oh, yeah that would be fine

piraka9011:
@emersonknapp I was able to resolve the fastrtps issue by using reliability policies instead of durability. Might need to open an issue against fastrtps for this.

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/61645b2f212b620d7b4ad528ee410b87/raw/4222ca24ef476e6593aaa38d45e555c8feadc6e8/ros2.repos
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10053/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5652/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8182/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9963/)

Karsten1987:
There are quite a few tests failing on Windows. Is that known?

emersonknapp:
Aware now, not sure where it came from. Investigating - insights welcome

piraka9011:
@Karsten1987 @emersonknapp they all appear to be the same 

```
Failed to find symbol 'rosidl_typesupport_connext_c__get_message_type_support_handle__rcl_interfaces__msg__Log' in library
[ERROR] [1586489902.463660500] [rosbag2_transport]: Failed to play: failed to initialize rcl node: type support handle implementation 'rosidl_typesupport_c' 
...
```

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5687/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8222/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10013/)

PR #355, opened 2020-04-07
Intelligently subscribe to topics according to their QoS profiles

emersonknapp:
Part of #125 
Should probably go after https://github.com/ros2/rosbag2/pull/346

This PR handles subscribing to topics intelligently (across all rmw implementations), even if there are slightly different QoS profiles on offer. We still record offered profiles in full for completeness of information, but we do not test against their values for most policies because the different rmw implementations do not yet consistently report all values the same.

Only try to adapt to Reliability and Durability, because they are the only policies that actually affect delivery of messages. Since they each have two values, a "stricter" and "looser" value, we subscribe using the strict value if all publishers offer it, otherwise we fall back to the looser value.

History and Lifespan do not affect compatibility at all (history is purely local to both publishers and subscriptions, lifespan is local to publishers only and doesn't do anything on subscriptions), so we ignore them and use sensible defaults for rosbag2 use case.

Deadline and Liveliness don't affect delivery of messages, their purpose is to generate callbacks when the contract is broken. Since we do not record these callbacks, we can always use the loosest request and be compatible with all publishers on these policies.

This PR allows rosbag2 to now receive latched messages, and subscribe to Best Effort topics.


rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -80,31 +93,31 @@ TEST_F(RecordIntegrationTestFixture, qos_is_stored_in_metadata)
   auto recorded_topics = writer.get_topics();
   std::string serialized_profiles = recorded_topics.at(topic).offered_qos_profiles;
   // Basic smoke test that the profile was serialized into the metadata as a string.
+  EXPECT_THAT(serialized_profiles, ContainsRegex("reliability: 1\n"));
+  EXPECT_THAT(serialized_profiles, ContainsRegex("durability: 2\n"));
+  EXPECT_THAT(
+    serialized_profiles, ContainsRegex(
+      "deadline:\n"
+      "    sec: .+\n"
+      "    nsec: .+\n"
```
thomas-moulard, line 102: string contains "deadline:"?

emersonknapp: i'm slightly interested in the rmw_time_t being present as well, but not in the specific values

emersonknapp:
I agree it feels a little ad hoc, but I think it solves the problem robustly for the Foxy feature set, in not very much code.

It should be possible to define an internal ordering for each individual QoS policy that ROS 2 exposes, which could provide a more programmatic way to construct an adapted QoS profile. Ideally it would be defined in or near the `rmw` layer, where the APIs around this behavior are defined. For now, given time constraints, I think we are better off defining this here, as is, rather than building the more general solution. The `rmw` API specification needs some refining to make sure that the assumptions we would be making in those utilities are obligated by the API contract - right now there are inconsistencies in reporting from the different rmw implementations that we are not in a good position to track down before the distro release.

Long way to say, yeah, I feel similarly, but I think this is the best move right now.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5559/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8097/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9821/)

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -218,18 +195,75 @@ std::string Recorder::serialized_offered_qos_profiles_for_topic(const std::strin
...
+  request_qos.default_history();
+
+  // Policy: reliability
+  if (reliability_reliable_endpoints_count == num_endpoints) {
+    request_qos.reliable();
+  } else {
+    if (reliability_reliable_endpoints_count > 0) {
+      ROSBAG2_TRANSPORT_LOG_WARN_STREAM(
+        "Some, but not all, publishers on topic \"" << topic_name << "\" "
+          "are offering Reliable reliability. "
```
piraka9011, line 227:
```suggestion
          "are offering a RELIABLE reliability QoS policy. "
```
Wondering how we can make this helpful in the context of other ros2 nodes running. I'd prefer to be more verbose.

piraka9011: Ditto with logs below

emersonknapp: I don't think we can be _much_ more verbose, it's pretty wordy as it is. I updated the setting names to use exact rmw enum values, for extra specificity.

We do have the node names publishing, but for now i'd rather not print quite so much. The user has the ability to it down via `ros2 topic info`.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9965/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5570/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8108/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9836/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5575/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8114/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9843/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9974/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8115/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9843/)

emersonknapp:
@Karsten1987 the failing test on windows is `record_end_to_end_test` with a 60s timeout. My assessment is that the test is flaky on windows, the part of the test that waits is around metadata file creation, which is not touched here. There has been a backup on the windows CI executors all day, what are your thoughts on merging this?

Karsten1987:
It'll take me a bit to review this, but feel free to go ahead and merge.
I don't think the test is flaky, it's legit broken given that it constantly fails.

PR #358, opened 2020-04-07
Refactor utility functions in ros2bag

piraka9011:
* Move utility function to `ros2bag`'s API module.
* Rename tests to be more descriptive

### DIFF
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -32,21 +32,19 @@ _QOS_POLICY_FROM_SHORT_NAME = {
     'liveliness': QoSLivelinessPolicy.get_from_short_key
 }
 _DURATION_KEYS = ['deadline', 'lifespan', 'liveliness_lease_duration']
-_ENUM_KEYS = ['history', 'reliability', 'durability', 'liveliness']
 _VALUE_KEYS = ['depth', 'avoid_ros_namespace_conventions']
 
 
-def is_dict_valid_duration(duration_dict: Dict[str, int]) -> bool:
-    """Check if dictionary contains the keys `sec` and `nsec`."""
-    return all(key in duration_dict for key in ['sec', 'nsec'])
+def print_error(string: str) -> str:
+    return "[ERROR] [ros2bag]: {}".format(string)
 
 
 def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
     """Convert a QoS duration profile from YAML into an rclpy Duration."""
     if time_dict:
         try:
-            return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
-        except TypeError:
+            return Duration(seconds=time_dict['sec'], nanoseconds=time_dict['nsec'])
+        except KeyError:
             raise ValueError(
                 'Time overrides must include both seconds (sec) and nanoseconds (nsec).')
     else:
@@ -59,12 +57,12 @@ def interpret_dict_as_qos_profile(qos_profile_dict: Dict) -> QoSProfile:
     for policy_key, policy_value in qos_profile_dict.items():
         if policy_key in _DURATION_KEYS:
             new_profile_dict[policy_key] = dict_to_duration(policy_value)
-        elif policy_key in _ENUM_KEYS:
+        elif policy_key in _QOS_POLICY_FROM_SHORT_NAME:
             new_profile_dict[policy_key] = _QOS_POLICY_FROM_SHORT_NAME[policy_key](policy_value)
         elif policy_key in _VALUE_KEYS:
             new_profile_dict[policy_key] = policy_value
         else:
-            raise ValueError('Unexpected key for QoS profile.')
+            raise ValueError('Unexpected key `{}` for QoS profile.'.format(policy_key))
     return QoSProfile(**new_profile_dict)
 
 
@@ -81,7 +79,7 @@ def create_bag_directory(uri: str) -> str:
     try:
         os.makedirs(uri)
     except OSError:
-        return "[ERROR] [ros2bag]: Could not create bag folder '{}'.".format(uri)
+        return print_error("Could not create bag folder '{}'.".format(uri))
 
 
 def check_positive_float(value: float) -> float:
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -74,7 +74,7 @@ def convert_yaml_to_qos_profile(qos_profile_dict: Dict) -> Dict[str, QoSProfile]
     return topic_profile_dict
 
 
-def create_bag_directory(uri: str) -> str:
+def create_bag_directory(uri: str) -> Optional[str]:
     """Create a directory."""
     try:
         os.makedirs(uri)
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from argparse import ArgumentTypeError
-import datetime
 import os
 from typing import Any
 from typing import Dict
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 from argparse import ArgumentTypeError
+import datetime
 import os
+from typing import Any
 from typing import Dict
 from typing import Optional
 
@@ -82,12 +84,22 @@ def create_bag_directory(uri: str) -> Optional[str]:
         return print_error("Could not create bag folder '{}'.".format(uri))
 
 
-def check_positive_float(value: float) -> float:
+def check_positive_float(value: Any) -> float:
     """Argparse validator to verify that a value is a float and positive."""
     try:
         fvalue = float(value)
         if fvalue <= 0.0:
-            raise ArgumentTypeError('%s is not in the valid range (> 0.0)' % value)
+            raise ArgumentTypeError('{} is not in the valid range (> 0.0)'.format(value))
         return fvalue
     except ValueError:
-        raise ArgumentTypeError('%s is not of the valid type (float)' % value)
+        raise ArgumentTypeError('{} is not the valid type (float)'.format(value))
+
+
+def check_path_exists(value: Any) -> str:
+    """Argparse validator to verify a path exists."""
+    try:
+        if os.path.exists(value):
+            return value
+        raise ArgumentTypeError("Bag file '{}' does not exist!".format(value))
+    except ValueError:
+        raise ArgumentTypeError('{} is not the valid type (string)'.format(value))
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -45,16 +45,15 @@ def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
     """Convert a QoS duration profile from YAML into an rclpy Duration."""
     if time_dict:
         try:
-            if is_dict_valid_duration(time_dict):
-                return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
-        except ValueError:
+            return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
+        except TypeError:
             raise ValueError(
                 'Time overrides must include both seconds (sec) and nanoseconds (nsec).')
     else:
         return Duration()
 
 
-def interpret_dict_as_qos_profile(qos_profile_dict: Dict) -> Dict[str, Dict]:
+def interpret_dict_as_qos_profile(qos_profile_dict: Dict) -> QoSProfile:
     """Sanitize a user provided dict of a QoS profile and verify all keys are valid."""
     new_profile_dict = {}
     for policy_key, policy_value in qos_profile_dict.items():
@@ -66,15 +65,14 @@ def interpret_dict_as_qos_profile(qos_profile_dict: Dict) -> Dict[str, Dict]:
             new_profile_dict[policy_key] = policy_value
         else:
             raise ValueError('Unexpected key for QoS profile.')
-    return qos_profile_dict
+    return QoSProfile(**new_profile_dict)
 
 
 def convert_yaml_to_qos_profile(qos_profile_dict: Dict) -> Dict[str, QoSProfile]:
     """Convert a YAML file to use rclpy's QoSProfile."""
     topic_profile_dict = {}
     for topic, profile in qos_profile_dict.items():
-        qos_profile = interpret_dict_as_qos_profile(profile)
-        topic_profile_dict[topic] = QoSProfile(**qos_profile)
+        topic_profile_dict[topic] = interpret_dict_as_qos_profile(profile)
     return topic_profile_dict
 
 
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -36,7 +36,7 @@ _VALUE_KEYS = ['depth', 'avoid_ros_namespace_conventions']
 
 
 def print_error(string: str) -> str:
-    return "[ERROR] [ros2bag]: {}".format(string)
+    return '[ERROR] [ros2bag]: {}'.format(string)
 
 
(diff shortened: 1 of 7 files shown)


ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+def is_dict_valid_duration(duration_dict: Dict[str, int]) -> bool:
+    """Check if dictionary contains the keys `sec` and `nsec`."""
+    return all(key in duration_dict for key in ['sec', 'nsec'])
+
+
+def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
+    """Convert a QoS duration profile from YAML into an rclpy Duration."""
+    if time_dict:
+        if is_dict_valid_duration(time_dict):
```
emersonknapp, line 44: EAFP would be more pythonic, e.g.

```
return Duration(seconds=time_dict['sec'], nanoseconds=time_dict['nsec']) if time_dict else Duration()
```

If you want to rewrite the message, catch the KeyError and re-raise

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+from rclpy.duration import Duration
+from rclpy.qos import QoSDurabilityPolicy
+from rclpy.qos import QoSHistoryPolicy
+from rclpy.qos import QoSLivelinessPolicy
+from rclpy.qos import QoSProfile
+from rclpy.qos import QoSReliabilityPolicy
+
+# This map needs to be updated when new policies are introduced
+POLICY_MAP = {
```
emersonknapp, line 28: Make this internal?
```suggestion
_POLICY_MAP = {
```

emersonknapp: Also this name is uninformative. `PolicyNameToGetterFunction` perhaps?

piraka9011:
Gist: https://gist.githubusercontent.com/piraka9011/b7b7b3b22274d37cd1342ad0523974a7/raw/3d6ecc423d7c495ec16fa2c48c92275268b2606f/ros2.repos
BUILD args: --packages-up-to ros2bag
TEST args: --packages-select ros2bag
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5572/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8110/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9838/)

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        if is_dict_valid_duration(time_dict):
+            return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
+        else:
+            raise ValueError(
+                'Time overrides must include both seconds (sec) and nanoseconds (nsec).')
+    else:
+        return Duration()
+
+
+def validate_qos_profile_overrides(qos_profile_dict: Dict) -> Dict[str, Dict]:
```
emersonknapp, line 53: This name is misleading. I would expect "validate" to return a bool, and the dict logic is not really necessary. Let's have this do the minimal useful functionality for a single item.

```
def interpret_yaml_as_qos_profile(qos_profile_dict: dict) -> rclpy.qos.QoSProfile:
```

The users of this function can do name: profile mapping themselves

```
qos_profile_overrides = {
  key: interpret_yaml_as_qos_profile(value) for key, value in qos_profile_dict
}
```

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    """Validate the QoS profile yaml file path and its structure."""
+    for name in qos_profile_dict.keys():
+        profile = qos_profile_dict[name]
+        # Convert dict to Duration. Required for construction
+        conversion_keys = ['deadline', 'lifespan', 'liveliness_lease_duration']
+        for k in conversion_keys:
```
emersonknapp, line 59: I think this logic should also validate that no unexpected keys have been found, by checking every value in the profile instead of checking expected values

```
for policy_key, policy_value in profile.items():
  if policy_key in duration_keys:
    duration_conversion
  elif policy_key in enum_keys:
    enum_conversion
  else:
    raise ValueError('unexpected key')
```

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+def check_positive_float(value: float) -> float:
+    """Verify value a float and positive."""
+    try:
+        fvalue = float(value)
+        if fvalue <= 0.0:
+            raise ArgumentTypeError('%s is not in the valid range (> 0.0)' % value)
+        return fvalue
+    except ValueError:
+        raise ArgumentTypeError('%s is not of the valid type (float)' % value)
```
emersonknapp, line 83: Can you note that this is an argparse validator/converter? raising ArgumentTypeError for a function with this name is not an immediately obvious thing to do

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+
+# This map needs to be updated when new policies are introduced
+_QOS_POLICY_FROM_SHORT_NAME = {
+    'history': QoSHistoryPolicy.get_from_short_key,
+    'reliability': QoSReliabilityPolicy.get_from_short_key,
+    'durability': QoSDurabilityPolicy.get_from_short_key,
+    'liveliness': QoSLivelinessPolicy.get_from_short_key
+}
+_DURATION_KEYS = ['deadline', 'lifespan', 'liveliness_lease_duration']
+_ENUM_KEYS = ['history', 'reliability', 'durability', 'liveliness']
```
emersonknapp, line 35:
```suggestion

```
This data does not need to be duplicated

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+        return Duration()
+
+
+def interpret_dict_as_qos_profile(qos_profile_dict: Dict) -> Dict[str, Dict]:
+    """Sanitize a user provided dict of a QoS profile and verify all keys are valid."""
+    new_profile_dict = {}
+    for policy_key, policy_value in qos_profile_dict.items():
+        if policy_key in _DURATION_KEYS:
+            new_profile_dict[policy_key] = dict_to_duration(policy_value)
+        elif policy_key in _ENUM_KEYS:
```
emersonknapp, line 62:
```suggestion
        elif policy_key in _QOS_POLICY_FROM_SHORT_NAME:
```

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
+    """Convert a QoS duration profile from YAML into an rclpy Duration."""
+    if time_dict:
+        try:
+            return Duration(seconds=time_dict.get('sec'), nanoseconds=time_dict.get('nsec'))
+        except TypeError:
+            raise ValueError(
+                'Time overrides must include both seconds (sec) and nanoseconds (nsec).')
+    else:
+        return Duration()
```
emersonknapp, line 53: As-is this won't throw because you use `get`, and you're catching the wrong type, it would be a `KeyError`.

```suggestion
def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
    try:
        return Duration(seconds=time_dict['sec'], nanoseconds=time_dict['nsec']) if time_dict else Duration()
    except KeyError:
        raise ValueError('Durations require both keys "sec" and "nsec")
```

Let's add some unit tests for these utilities

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    'reliability': QoSReliabilityPolicy.get_from_short_key,
+    'durability': QoSDurabilityPolicy.get_from_short_key,
+    'liveliness': QoSLivelinessPolicy.get_from_short_key
+}
+_DURATION_KEYS = ['deadline', 'lifespan', 'liveliness_lease_duration']
+_ENUM_KEYS = ['history', 'reliability', 'durability', 'liveliness']
+_VALUE_KEYS = ['depth', 'avoid_ros_namespace_conventions']
+
+
+def is_dict_valid_duration(duration_dict: Dict[str, int]) -> bool:
```
emersonknapp, line 39: remove?

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    new_profile_dict = {}
+    for policy_key, policy_value in qos_profile_dict.items():
+        if policy_key in _DURATION_KEYS:
+            new_profile_dict[policy_key] = dict_to_duration(policy_value)
+        elif policy_key in _ENUM_KEYS:
+            new_profile_dict[policy_key] = _QOS_POLICY_FROM_SHORT_NAME[policy_key](policy_value)
+        elif policy_key in _VALUE_KEYS:
+            new_profile_dict[policy_key] = policy_value
+        else:
+            raise ValueError('Unexpected key for QoS profile.')
```
emersonknapp, line 67: Include the name of the offending key

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+    topic_profile_dict = {}
+    for topic, profile in qos_profile_dict.items():
+        topic_profile_dict[topic] = interpret_dict_as_qos_profile(profile)
+    return topic_profile_dict
+
+
+def create_bag_directory(uri: str) -> str:
+    """Create a directory."""
+    try:
+        os.makedirs(uri)
```
emersonknapp, line 82: I don't see the exception-case return value used anywhere.

In the successful case this function hits end of scope without returning a value. 

Is the intended return type `None` (and supposed to log) or is it `Optional[str]`?

ros2bag/ros2bag/api/__init__.py
piraka9011: Duration throws a `TypeError` if the nano/seconds is none. Your approach is also valid tho.

ros2bag/ros2bag/api/__init__.py
piraka9011: I moved this from the `play` verb and I believe it's intention is to simply log the result.

ros2bag/test/resources/incomplete_qos_duration.yaml
```diff
@@ -0,0 +1,7 @@
+/test_topic:
+  history: keep_all
+  depth: 0
+  reliability: reliable
+  durability: volatile
+  deadline:
+    sec: 2147483647
```
zmichaels11, line 7: what is the significance of this value?
Can you insert a yaml comment describing why it was chosen?

piraka9011: `ULONG_MAX`. No significance tho.

ros2bag/ros2bag/api/__init__.py
emersonknapp: LGTM'd the PR, but the returned value here doesn't actually ever get printed. You can move forward though, if this is maintaining previous behavior

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10000/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5600/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8135/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9883/)

piraka9011:
Another round with new config and gist

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10003/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5603/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8138/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9887/)

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    'reliability': QoSReliabilityPolicy.get_from_short_key,
+    'durability': QoSDurabilityPolicy.get_from_short_key,
+    'liveliness': QoSLivelinessPolicy.get_from_short_key
+}
+_DURATION_KEYS = ['deadline', 'lifespan', 'liveliness_lease_duration']
+_VALUE_KEYS = ['depth', 'avoid_ros_namespace_conventions']
+
+
+def print_error(string: str) -> str:
+    return '[ERROR] [ros2bag]: {}'.format(string)
```
thomas-moulard, line 39: Why reimplementing `logging.error`?

ros2bag/ros2bag/api/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+    try:
+        os.makedirs(uri)
+    except OSError:
+        return print_error("Could not create bag folder '{}'.".format(uri))
+
+
+def check_positive_float(value: float) -> float:
+    """Argparse validator to verify that a value is a float and positive."""
+    try:
+        fvalue = float(value)
```
thomas-moulard, line 88: seems to contradict your typing? (value is a floating-point value)

ros2bag/ros2bag/verb/play.py
```diff
@@ -34,22 +35,14 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
-            fvalue = float(value)
-            if fvalue <= 0.0:
-                raise ArgumentTypeError('%s is not in the valid range (> 0.0)' % value)
-            return fvalue
-        except ValueError:
-            raise ArgumentTypeError('%s is not of the valid type (float)' % value)
-
     def main(self, *, args):  # noqa: D102
         bag_file = args.bag_file
         if not os.path.exists(bag_file):
```
thomas-moulard, line 43: Why not reusing argparse.FileType instead of reimplementing the logic?

ros2bag/ros2bag/verb/record.py
```diff
@@ -122,41 +79,36 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
     def main(self, *, args):  # noqa: D102
         if args.all and args.topics:
-            return 'Invalid choice: Can not specify topics and -a at the same time.'
+            return print_error('Invalid choice: Can not specify topics and -a at the same time.')
 
         uri = args.output or datetime.datetime.now().strftime('rosbag2_%Y_%m_%d-%H_%M_%S')
 
         if os.path.isdir(uri):
-            return "[ERROR] [ros2bag]: Output folder '{}' already exists.".format(uri)
+            return print_error("Output folder '{}' already exists.".format(uri))
```
thomas-moulard, line 94: Consider moving the argument validation in the `arpgarse` logic if you keep modifying this code. This will take care for you of reporting errors to the users the "proper way". You could just create a function checking if the directory exists, and call `strftime` to pass a default value as it's built here.

ros2bag/ros2bag/api/__init__.py
piraka9011: I was debating between using `logging` vs print statements and it appears that the ros2 verb extension needs to `return` when something goes wrong/finishes.
If we use logging, we'd need to log followed by a `return` with the same statement which is duplicate work.
If there's a proper approach for `return`ing from a ros2 verb lmk (I'm following what other packages did).

ros2bag/ros2bag/api/__init__.py
piraka9011: I wasn't sure what to put the type as since this is supposed to be used as an argparse validator. (Is `Any` valid?)

ros2bag/ros2bag/verb/play.py
piraka9011: Good point

piraka9011:
One more round ...

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10041/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5637/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8171/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9952/)

piraka9011:
There's an unrelated warning:

```sh
ament_export_interfaces() is deprecated, use ament_export_targets() instead
```

I will address this in a follow up PR.

PR #359, opened 2020-04-07
Read serialized qos profiles out of the metadata

emersonknapp:
Part of #125 

This PR introduces the ability to read the new field `offered_qos_profiles` out of a bag metadata. This required making a way to pass version information to sub-structs of BagMetadata so that backwards compatibility can be maintained.

### DIFF
--- a/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
+++ b/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
@@ -39,6 +39,28 @@
 
 namespace YAML
 {
+/// Pass metadata version to the sub-structs of BagMetadata for deserializing.
+/**
+  * Encoding should always use the current metadata version, so it does not need this value.
+  * We cannot extend the YAML::Node class to include this, so we must call it
+  * as a function with the node as an argument.
+  */
+template<typename T>
+T decode_for_version(const Node & node, int version)
+{
+  static_assert(
+    std::is_default_constructible<T>::value,
+    "Type passed to decode_for_version that has is not default constructible.");
+  if (!node.IsDefined()) {
+    throw TypedBadConversion<T>(node.Mark());
+  }
+  T value{};
+  if (convert<T>::decode(node, value, version)) {
+    return value;
+  }
+  throw TypedBadConversion<T>(node.Mark());
+}
+
 template<>
 struct convert<rosbag2_storage::TopicMetadata>
 {
@@ -52,12 +74,16 @@ struct convert<rosbag2_storage::TopicMetadata>
     return node;
   }
 
-  static bool decode(const Node & node, rosbag2_storage::TopicMetadata & topic)
+  static bool decode(const Node & node, rosbag2_storage::TopicMetadata & topic, int version)
   {
     topic.name = node["name"].as<std::string>();
     topic.type = node["type"].as<std::string>();
     topic.serialization_format = node["serialization_format"].as<std::string>();
-
+    if (version >= 4) {
+      topic.offered_qos_profiles = node["offered_qos_profiles"].as<std::string>();
+    } else {
+      topic.offered_qos_profiles = "";
+    }
     return true;
   }
 };
@@ -73,10 +99,38 @@ struct convert<rosbag2_storage::TopicInformation>
     return node;
   }
 
-  static bool decode(const Node & node, rosbag2_storage::TopicInformation & metadata)
+  static bool decode(const Node & node, rosbag2_storage::TopicInformation & metadata, int version)
   {
-    metadata.topic_metadata = node["topic_metadata"].as<rosbag2_storage::TopicMetadata>();
+    metadata.topic_metadata = decode_for_version<rosbag2_storage::TopicMetadata>(
+      node["topic_metadata"], version);
     metadata.message_count = node["message_count"].as<uint64_t>();
+    return true;
+  }
+};
+
+template<>
+struct convert<std::vector<rosbag2_storage::TopicInformation>>
+{
+  static Node encode(const std::vector<rosbag2_storage::TopicInformation> & rhs)
+  {
+    Node node(NodeType::Sequence);
+    for (const auto & value : rhs) {
+      node.push_back(value);
+    }
+    return node;
+  }
+
+  static bool decode(
+    const Node & node, std::vector<rosbag2_storage::TopicInformation> & rhs, int version)
+  {
+    if (!node.IsSequence()) {
+      return false;
+    }
+
+    rhs.clear();
+    for (const auto & value : node) {
+      rhs.push_back(decode_for_version<rosbag2_storage::TopicInformation>(value, version));
+    }
     return true;
   }
 };
@@ -148,7 +202,8 @@ struct convert<rosbag2_storage::BagMetadata>
       .as<std::chrono::time_point<std::chrono::high_resolution_clock>>();
     metadata.message_count = node["message_count"].as<uint64_t>();
     metadata.topics_with_message_count =
-      node["topics_with_message_count"].as<std::vector<rosbag2_storage::TopicInformation>>();
+      decode_for_version<std::vector<rosbag2_storage::TopicInformation>>(
+      node["topics_with_message_count"], metadata.version);
 
     if (metadata.version >= 3) {  // fields introduced by rosbag2_compression
       metadata.compression_format = node["compression_format"].as<std::string>();
--- a/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
+++ b/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
@@ -113,7 +113,7 @@ struct convert<std::vector<rosbag2_storage::TopicInformation>>
 {
   static Node encode(const std::vector<rosbag2_storage::TopicInformation> & rhs)
   {
-    Node node(NodeType::Sequence);
+    Node node{NodeType::Sequence};
     for (const auto & value : rhs) {
       node.push_back(value);
     }
--- a/rosbag2_storage/test/rosbag2_storage/test_metadata_serialization.cpp
+++ b/rosbag2_storage/test/rosbag2_storage/test_metadata_serialization.cpp
@@ -93,3 +93,42 @@ TEST_F(MetadataFixture, test_writing_and_reading_yaml)
     Eq(expected_second_topic.topic_metadata.serialization_format));
   EXPECT_THAT(actual_second_topic.message_count, Eq(expected_second_topic.message_count));
 }
+
+TEST_F(MetadataFixture, metadata_reads_v3_check_offered_qos_profiles_empty)
+{
+  // Make sure that when reading a v3 metadata, the deserialization
+  // does not attempt to fill the offered_qos_profiles field
+  const std::string offered_qos_profiles = "qos_profile_string_data";
+  const size_t message_count = 100;  // arbitrary value
+
+  BagMetadata metadata{};
+  metadata.version = 3;
+  metadata.topics_with_message_count.push_back(
+    {{"topic", "type", "rmw", offered_qos_profiles}, message_count});
+  metadata_io_->write_metadata(temporary_dir_path_, metadata);
+  auto read_metadata = metadata_io_->read_metadata(temporary_dir_path_);
+  ASSERT_THAT(
+    read_metadata.topics_with_message_count,
+    SizeIs(metadata.topics_with_message_count.size()));
+  auto actual_first_topic = read_metadata.topics_with_message_count[0];
+  EXPECT_TRUE(actual_first_topic.topic_metadata.offered_qos_profiles.empty());
+}
+
+TEST_F(MetadataFixture, metadata_reads_v4_fills_offered_qos_profiles)
+{
+  // Make sure when reading a v4 metadata that the deserialization fills "offered_qos_profiles"
+  const std::string offered_qos_profiles = "qos_profile_string_data";
+  const size_t message_count = 100;  // arbitrary value
+
+  BagMetadata metadata{};
+  metadata.version = 4;
+  metadata.topics_with_message_count.push_back(
+    {{"topic", "type", "rmw", offered_qos_profiles}, message_count});
+  metadata_io_->write_metadata(temporary_dir_path_, metadata);
+  auto read_metadata = metadata_io_->read_metadata(temporary_dir_path_);
+  ASSERT_THAT(
+    read_metadata.topics_with_message_count,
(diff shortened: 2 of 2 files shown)


rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -39,6 +39,24 @@
 
 namespace YAML
 {
+/// Allow pass a metadata version to the sub-structs of BagMetadata for deserializing.
```
thomas-moulard, line 42: Pass metada version...

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -39,6 +39,24 @@
...
+/**
+  * Encoding should always use the current metadata version, so it does not need this value.
+  * We cannot extend the YAML::Node class to include this, so we must call it
+  * as a function with the node as an argument.
+  */
+template<typename T>
+T versioned_decode(const Node & node, int version)
```
thomas-moulard, line 49: `decode_for_version` ? Not clear what this means.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -39,6 +39,24 @@
...
+{
+  if (!node.IsDefined()) {
+    throw TypedBadConversion<T>(node.Mark());
+  }
+  T t;
```
thomas-moulard, line 54: Is allocating `T` here OK performance wise? `t` is not a good variable name, what can this represent? Also allocating it here instead of passing it in forces `T` to be default constructible. Not sure if this could be an issue.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -52,12 +70,14 @@ struct convert<rosbag2_storage::TopicMetadata>
...
-  static bool decode(const Node & node, rosbag2_storage::TopicMetadata & topic)
+  static bool decode(const Node & node, rosbag2_storage::TopicMetadata & topic, int version)
   {
     topic.name = node["name"].as<std::string>();
     topic.type = node["type"].as<std::string>();
     topic.serialization_format = node["serialization_format"].as<std::string>();
-
+    if (version >= 4) {
+      topic.offered_qos_profiles = node["offered_qos_profiles"].as<std::string>();
+    }
```
thomas-moulard, line 80: do you want to reset `offered_qos_profiles` in an `else` branch to be on the safe side? Could `topic` contain a bogus value for this field when we start?

piraka9011:
This is pretty interesting, how does it work for future changes to the metadata? Do we just add a layer on top of decode?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -39,6 +39,24 @@
...
 {
+/// Allow pass a metadata version to the sub-structs of BagMetadata for deserializing.
+/**
+  * Encoding should always use the current metadata version, so it does not need this value.
+  * We cannot extend the YAML::Node class to include this, so we must call it
+  * as a function with the node as an argument.
+  */
+template<typename T>
+T versioned_decode(const Node & node, int version)
+{
```
thomas-moulard, line 50: Consider adding a static_assert to check that T is "correct".

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -73,14 +93,42 @@ struct convert<rosbag2_storage::TopicInformation>
...
+      node.push_back(value);
+    }
+    return node;
+  }
+
+  static bool decode(
+    const Node & node, std::vector<rosbag2_storage::TopicInformation> & rhs, int version)
+  {
+    if (!node.IsSequence()) {
+      return false;
```
thomas-moulard, line 121: if we return false, do we abort? do we have any testing outside the happy path for this? I'd like to simplify this problem by just throwing and not catching if we can live with that for now. It's better than having libyaml catch this and hand us half-correct structures.

emersonknapp:
You just add if blocks for data added in new versions. That's the way that I have historically seen this done for similar data format parsing

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -73,14 +93,42 @@ struct convert<rosbag2_storage::TopicInformation>
...
+    }
+
+    rhs.clear();
+    for (const_iterator it = node.begin(); it != node.end(); ++it) {
```
thomas-moulard, line 125: auto?

rosbag2_storage/test/rosbag2_storage/test_metadata_serialization.cpp
```diff
@@ -93,3 +93,39 @@ TEST_F(MetadataFixture, test_writing_and_reading_yaml)
     Eq(expected_second_topic.topic_metadata.serialization_format));
   EXPECT_THAT(actual_second_topic.message_count, Eq(expected_second_topic.message_count));
 }
+
+TEST_F(MetadataFixture, metadata_reads_v3)
```
thomas-moulard, line 97: It would help the reader to have more specific test cases description: "metadata_reads_v3_check_offered_qos_profiles_empty", ditto for next one.

A line or two of comments explaining that this is a new feature in v4, and we make sure that the new attribute is properly initialized to zero would also be super helpful.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Ah, good catch. When I first wrote this I specialized a `yaml-cpp` class that was not called directly and throws when convert returns false. I'll change it to throw that exception

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: it can't right now because it's a std::string default initialized immediately called before this function.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: It's the logic that yaml-cpp uses, I'd prefer to not get too far from it. Can rename the typevar. But yes, allocating T is fine. This is a fixed cost at startup, it doesn't affect runtime performance, and the total amount of data involved is small even for complex bags, a few KB at most

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Oh wait, never mind. This is not the line i thought it was. These return values are checked to throw a `TypedBadConversion` in the yaml-cpp implementation. We don't catch the exception, it will halt the program (correctly)

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: I don't understand, where would this assert go?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Done

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Sure, done

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Done

rosbag2_storage/test/rosbag2_storage/test_metadata_serialization.cpp
emersonknapp: Done

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
Karsten1987: this assert could be the first line of the function. Alternatively it could go in the template declaration with `std::enable_if`, but I don't think that's needed.
Can we limit the amount of `T`s for this function?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -39,6 +39,25 @@
...
+  * We cannot extend the YAML::Node class to include this, so we must call it
+  * as a function with the node as an argument.
+  */
+template<typename T>
+T decode_for_version(const Node & node, int version)
+{
+  if (!node.IsDefined()) {
+    throw TypedBadConversion<T>(node.Mark());
+  }
+  T value;
```
Karsten1987, line 54: can this be default initialized? `T value{}`?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -73,14 +96,42 @@ struct convert<rosbag2_storage::TopicInformation>
...
   }
 };
 
+template<>
+struct convert<std::vector<rosbag2_storage::TopicInformation>>
+{
+  static Node encode(const std::vector<rosbag2_storage::TopicInformation> & rhs)
+  {
+    Node node(NodeType::Sequence);
+    for (auto value : rhs) {
```
Karsten1987, line 114:
```suggestion
    for (auto value & : rhs) {
```

Given that this function copies one vector into, what seems like, another vector, can this be constructed in place in the `Node` constructor? Alternatively, I would recommend `reserve`ing the node sequence.
Obviously, please disregard all this if `Node` doesn't allow us to do this.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -73,14 +96,42 @@ struct convert<rosbag2_storage::TopicInformation>
...
+    const Node & node, std::vector<rosbag2_storage::TopicInformation> & rhs, int version)
+  {
+    if (!node.IsSequence()) {
+      return false;
+    }
+
+    rhs.clear();
+    for (const auto & value : node) {
+      rhs.push_back(decode_for_version<rosbag2_storage::TopicInformation>(value, version));
+    }
```
Karsten1987, line 130: same comment as above. Can we initialize `rhs` in place instead of iterating over node?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: Sorry, could one of you use the github suggestion feature or a code block, I don't understand what you're looking to accomplish

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: yeah

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
Karsten1987: So further below, we call this function with `T = rosbag2_storage::TopicInformation`. For this type, the function is well defined.
```
decode_for_version<std::vector<rosbag2_storage::TopicInformation>>
```

With  a `static_assert` we could verify that this function is not called with a type T which doesn't behave correctly. One which comes to my mind here is `std::is_trivially_default_constructible<T>` as we at one point call `T value`. The first line could then be something like:
```
static_assert(std::is_trivially_default_constructible<T>::value);
```
That would prevent a compilation with a wrong type T.

That is all to say, that it's most likely not being needed given that all of this is in a `.cpp` file in the first place and can't be called by anybody else than this file.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: There's a vector but it's behind a few levels of `private`, we don't have access to reserve or iterator-insert, we only have `push_back`.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: unfortunately it does not seem possible with `yaml-cpp`

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
emersonknapp: I have added an assertion for `std::is_default_constructible`, the `trivially` version doesn't work here.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9978/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5579/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9852/)

jacobperron:
@emersonknapp FYI, we're have a windows build issue that just started happening. I cleared a build from the queue and edited your comment to point to the right build number, but I think you might hit the same unrelated build failure in `resource_retriever`.

emersonknapp:
Thanks for the heads up!

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -73,14 +99,42 @@ struct convert<rosbag2_storage::TopicInformation>
...
     return true;
   }
 };
 
+template<>
+struct convert<std::vector<rosbag2_storage::TopicInformation>>
+{
+  static Node encode(const std::vector<rosbag2_storage::TopicInformation> & rhs)
+  {
+    Node node(NodeType::Sequence);
```
zmichaels11, line 116:
```suggestion
    Node node{NodeType::Sequence};
```

emersonknapp:
Running CI again before merging since a few minor changes have gone into this and a lot has gone into the upstream 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10015/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5614/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9904/)

emersonknapp:
The test failures are clearly unrelated, having to do with typesupport for connext and cyclone, on windows,  in the last two runs. We need to fix that, but I'm going to merge this change to unblock the followup work.

PR #360, opened 2020-04-08
use ament_export_targets()

dirk-thomas:
Related to ament/ament_cmake#237.

### DIFF
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -81,8 +81,8 @@ install(
 )
 
 ament_export_include_directories(include)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Note: the `build_and_test` action doesn't use the same branch from other repos which is required to pass.

emersonknapp:
We can wait until the new feature exists (I assume it is deprecating, not replacing, the existing function?) and then this PR will be able to stand alone from other repositories' branches, correct?

Karsten1987:
I would assume that there'll be a single CI run for all PRs listed in https://github.com/ament/ament_cmake/issues/237.

With that CI passing, I think it's safe to be merged.

Issue #292, opened 2020-02-10
Loop playback


emersonknapp:
## Description

When playing back a bagfile, I would like to be able to specify a commandline option so that it starts back at the beginning on reaching the end, and plays indefinitely.

## Related Issues

N/A

## Completion Criteria

* [ ] A user can specify a commandline option to loop playback indefinitely

## Implementation Notes / Suggestions

none

## Testing Notes / Suggestions

none

ketatam:
Hi, I am starting to work on this issue.

ketatam:
Hi, is it okay that we reopen the database every time we reach the end and replay the bagfile or do we have to only open it once at the beginning? Thank you!

PR #361, opened 2020-04-09
Add loop option to rosbag play

ketatam:
With the option -l or --loop it's now possible to play a bagfile in loop.
The current implementation is as simple as possible, but that means that the database has to be reopened every time the bagfile starts from the beginning. But also with this implementation we don't rewrite timestamps, so the loop will jump back in time on every new opening.
If the database should not be reopened every time, I can think of two ways to fix it:
we can add a function start_reading_from_beginning() to the SequentialReader.
Or we can add a flag to the storage on 'open', so that 'read_next' returns the first message after the last.
In these Cases, should we start the time stamp at the beginning? Or increase the timestamp continuously?
Any propositions or ideas are very welcome.
Closes https://github.com/ros2/rosbag2/issues/292.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
+++ b/rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
@@ -31,6 +31,7 @@ public:
   {
     (void) storage_options;
     (void) converter_options;
+    num_read_ = 0;
   }
 
   void reset() override {}
--- a/rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
@@ -65,7 +65,7 @@ TEST_F(RosBag2PlayTestFixture, messages_played_in_loop) {
   bool loop_playback = true;
   std::thread loop_thread(&rosbag2_transport::Rosbag2Transport::play, rosbag2_transport_ptr,
     storage_options_,
-    rosbag2_transport::PlayOptions{1000, "", 1.0, {}, loop_playback});
+    rosbag2_transport::PlayOptions{1000, "", 1.0, {}, {}, loop_playback});
   std::this_thread::sleep_for(std::chrono_literals::operator""s(1));
   rclcpp::shutdown();
   loop_thread.join();
--- a/rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
@@ -1,0 +1,83 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
+// Copyright 2020, TNG Technology Consulting GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include <chrono>
+#include <future>
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
+#include <utility>
+
+#include "rclcpp/rclcpp.hpp"
+
+#include "rosbag2_test_common/subscription_manager.hpp"
+
+#include "rosbag2_transport/rosbag2_transport.hpp"
+
+#include "test_msgs/msg/arrays.hpp"
+#include "test_msgs/msg/basic_types.hpp"
+#include "test_msgs/message_fixtures.hpp"
+
+#include "rosbag2_play_test_fixture.hpp"
+
+using namespace ::testing;  // NOLINT
+using namespace rosbag2_transport;  // NOLINT
+using namespace std::chrono_literals;  // NOLINT
+using namespace rosbag2_test_common;  // NOLINT
+
+TEST_F(RosBag2PlayTestFixture, messages_played_in_loop) {
+  auto primitive_message1 = get_messages_basic_types()[0];
+  primitive_message1->int32_value = 42;
+
+  auto topic_types = std::vector<rosbag2_storage::TopicMetadata>{
+    {"loop_test_topic", "test_msgs/BasicTypes", "", ""}
+  };
+
+  std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> messages =
+  {serialize_test_message("loop_test_topic", 700, primitive_message1)};
+
+  auto prepared_mock_reader = std::make_unique<MockSequentialReader>();
+  prepared_mock_reader->prepare(messages, topic_types);
+  reader_ = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
+
+  //  any number greater than 2 would be logic to test the loop_playback functionality,
+  //  for example 4.
+  size_t expected_number_of_messages = 4;
+  sub_->add_subscription<test_msgs::msg::BasicTypes>(
+    "/loop_test_topic",
+    expected_number_of_messages);
+
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  auto rosbag2_transport_ptr = std::make_shared<Rosbag2Transport>(reader_, writer_, info_);
+  bool loop_playback = true;
+  std::thread loop_thread(&Rosbag2Transport::play, rosbag2_transport_ptr, storage_options_,
+    rosbag2_transport::PlayOptions{1000, "", 1.0, loop_playback});
+  std::this_thread::sleep_for(1s);
+  rclcpp::shutdown();
+  loop_thread.join();
+
+  await_received_messages.get();
+  auto replayed_test_primitives = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
+    "/loop_test_topic");
+
+  EXPECT_THAT(replayed_test_primitives.size(), Ge(4u));
+  EXPECT_THAT(
+    replayed_test_primitives,
+    Each(Pointee(Field(&test_msgs::msg::BasicTypes::int32_value, 42))));
+}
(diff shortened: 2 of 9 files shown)


rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <vector>
+#include <utility>
+
+#include "rclcpp/rclcpp.hpp"
+
+#include "rosbag2_test_common/subscription_manager.hpp"
+
+#include "rosbag2_transport/rosbag2_transport.hpp"
+
+#include "test_msgs/msg/arrays.hpp"
```
emersonknapp, line 32: i don't think you use this include

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "test_msgs/msg/basic_types.hpp"
+#include "test_msgs/message_fixtures.hpp"
+
+#include "rosbag2_play_test_fixture.hpp"
+
+using namespace ::testing;  // NOLINT
```
emersonknapp, line 38: I'd say remove all these `using`, there's a reason the linter warns against it. You could put one inside a test function, if it saves a lot of space, but I'd prefer avoiding most of them entirely.

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,83 @@
...
+  loop_thread.join();
+
+  await_received_messages.get();
+  auto replayed_test_primitives = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
+    "/loop_test_topic");
+
+  EXPECT_THAT(replayed_test_primitives.size(), Ge(4u));
+  EXPECT_THAT(
+    replayed_test_primitives,
+    Each(Pointee(Field(&test_msgs::msg::BasicTypes::int32_value, 42))));
```
emersonknapp, line 82: 42 is a magic value here - declare a `const int test_value = 42;` for this test

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,83 @@
...
+    rosbag2_transport::PlayOptions{1000, "", 1.0, loop_playback});
+  std::this_thread::sleep_for(1s);
+  rclcpp::shutdown();
+  loop_thread.join();
+
+  await_received_messages.get();
+  auto replayed_test_primitives = sub_->get_received_messages<test_msgs::msg::BasicTypes>(
+    "/loop_test_topic");
+
+  EXPECT_THAT(replayed_test_primitives.size(), Ge(4u));
```
emersonknapp, line 79:
```suggestion
  EXPECT_THAT(replayed_test_primitives.size(), Ge(expected_number_of_messages));
```

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> messages =
+  {serialize_test_message("loop_test_topic", 700, primitive_message1)};
+
+  auto prepared_mock_reader = std::make_unique<MockSequentialReader>();
+  prepared_mock_reader->prepare(messages, topic_types);
+  reader_ = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
+
+  //  any number greater than 2 would be logic to test the loop_playback functionality,
```
emersonknapp, line 58: I only see one message in `messages` - perhaps you should define `num_messages`, use it to fill `messages`, and say `expected_number_of_messages = num_messages * 2`

rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
```diff
@@ -31,6 +31,7 @@ class MockSequentialReader : public rosbag2_cpp::reader_interfaces::BaseReaderIn
   {
     (void) storage_options;
     (void) converter_options;
+    num_read_ = 0;
```
emersonknapp, line 34: Remove unrelated change

ketatam: Actually the MockReader needs this in order to be able to read the messages from the beginning when it has read all messages. Otherwise the has_next() method would return false after the first iteration in the loop and the MockReader wouldn't be able to reread the first message.

Karsten1987:
@ketatam I think your branch has to be rebased on top of master. There's currently merge conflicts.

ketatam:
> @ketatam I think your branch has to be rebased on top of master. There's currently merge conflicts.

I have rebased my branch on top of master.
But the record_all test failed because now when running the test there is this warning `[WARN] [1586949432.325929033] [publisher1]: This rmw implementation does not support ON_REQUESTED_INCOMPATIBLE_QOS events, you will not be notified when Subscriptions request an incompatible QoS profile from Publishers on the same topic.` and it's published on the topic /rosout and so the test records one message more than expected.. It would be helpful if you could fix this

ketatam:
> @ketatam I think your branch has to be rebased on top of master. There's currently merge conflicts.

I have another time my branch rebased on top of master. But the error in the CI doesn't exist on my machine so I don't know how to fix it..

emersonknapp:
The failing test is the one I've reported here https://github.com/ros2/rosbag2/issues/381

emersonknapp:
@emersonknapp 
Gist: https://gist.githubusercontent.com/emersonknapp/25387ca2689095a075a0129803431ea2/raw/b0501de81ad866c30345ec743fa526013b310992/ros2.repos
TEST args: --packages-select-regex rosbag2
Job: ci_launcher

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10256/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5794/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8338/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10141/)

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,81 @@
...
+  const int test_value = 42;
+  primitive_message1->int32_value = test_value;
+
+  auto topic_types = std::vector<rosbag2_storage::TopicMetadata>{
+    {"loop_test_topic", "test_msgs/BasicTypes", "", ""}
+  };
+
+  size_t num_messages = 3;
+  std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> messages(num_messages,
+    serialize_test_message("loop_test_topic", 700, primitive_message1));
```
emersonknapp, line 48: what is 700 - looks like another magic value.

can you declare and define all your test constants at the top of the test so that their meaning and choice of values is obvious?

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  auto prepared_mock_reader = std::make_unique<MockSequentialReader>();
+  prepared_mock_reader->prepare(messages, topic_types);
+  reader_ = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
+
+  size_t expected_number_of_messages = num_messages * 2;
```
emersonknapp, line 54: move to top of test

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,81 @@
...
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  auto rosbag2_transport_ptr = std::make_shared<rosbag2_transport::Rosbag2Transport>(
+    reader_,
+    writer_,
+    info_);
+  bool loop_playback = true;
+  std::thread loop_thread(&rosbag2_transport::Rosbag2Transport::play, rosbag2_transport_ptr,
+    storage_options_,
+    rosbag2_transport::PlayOptions{1000, "", 1.0, {}, {}, loop_playback});
```
emersonknapp, line 68: What are magic values `1000` and `1.0`?

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -0,0 +1,81 @@
...
+  std::this_thread::sleep_for(std::chrono_literals::operator""s(1));
```
emersonknapp, line 69: Why are we sleeping here? Is this robust to slow machines?

Additionally, instead of trying to do a literal, you can supply `std::chrono::seconds(1)` - which should fix the OSX build

emersonknapp: I believe that should also fix the windows warnings

ketatam: Yes, I actually added the sleep_for to make sure that enough messages are played for the test to pass, but if you think it wouldn't be necessary I can remove it

emersonknapp: I think that there are better ways. Sleep-based tests are inherently flaky and prone to randomly breaking CI. You'd be much better off waiting for a specific condition to be met (up to a reasonable timeout)

emersonknapp: See the [`spin_and_wait_for` utility](https://github.com/ros2/rosbag2/blob/master/rosbag2_test_common/include/rosbag2_test_common/wait_for.hpp#L25) and [an example of its usage](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/test/rosbag2_transport/test_record.cpp#L126)

This way, rather than waiting a fixed amount of time, your test can finish as soon as the condition is met, and time eventually if the condition is never met.

ketatam: I changed this a little bit, I noticed that the `SubscriptionManager` returns a future and when we call `await_received_messages` it will automatically wait until the messages are received, and this should be equivalent with what you suggested.

Martin-Idel:
I took the liberty to restart CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10316/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5845/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10221/)

PR #366, opened 2020-04-10
rename rosidl_generator_cpp namespace to rosidl_runtime_cpp

dirk-thomas:
Related to ros2/rosidl#456.


dirk-thomas:
The `build_and_test` failure is expected since this change require the change in the referenced PR.

PR #368, opened 2020-04-11
rename rosidl_generator_c namespace to rosidl_runtime_c

dirk-thomas:
Related to ros2/rosidl#458.


dirk-thomas:
The `build_and_test` action is expected to fail since they don't use the changes in the referenced PRs.

PR #373, opened 2020-04-14
Update GenericSubscription's handle_message signature

emersonknapp:
Fixes the build since https://github.com/ros2/rclcpp/pull/1047

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

PR #375, opened 2020-04-14
updates since changes to message_info in rclcpp

wjwwood:
Follow up of ros2/rclcpp#1047, which broke nightly.


wjwwood:
I had these changes staged (and forgot to push them to a pr...), but they're the same as https://github.com/ros2/rosbag2/pull/373, so I'll close this on in favor of that one, I'm just opening it for visibility of the branch.

PR #376, opened 2020-04-14
Fix Action CI tests to pass reliably

emersonknapp:
Note: this description is updated over the original PR

In our regular CI runs, since the upgrade to Fast-RTPS 1.10.x (), we have been seeing consistent failures in the following tests on our Action CI:

```
2020-04-14T20:21:18.5513377Z The following tests FAILED:
2020-04-14T20:21:18.5513644Z 	 11 - test_record_all__rmw_fastrtps_cpp (Failed)
2020-04-14T20:21:18.5513914Z 	 18 - test_play__rmw_fastrtps_cpp (Failed)
2020-04-14T20:21:18.5514188Z 	 20 - test_record_all__rmw_fastrtps_dynamic_cpp (Failed)
2020-04-14T20:21:18.5514448Z 	 27 - test_play__rmw_fastrtps_dynamic_cpp (Failed)
```

The `test_play` failure is receiving one fewer message than it testing that a specific number of messages arrive, when we only really care that any messages arrive at all for that test. The [upgrade to Fast-RTPS 1.10.x](https://github.com/ros2/ros2/pull/888) introduced a some nondeterministic behavior on shutdown and made it uncertain whether the final message in a playback would be delivered. This particular test doesn't care about that behavior so it has been modified to expect fewer messages. 

The `test_record_all` failure is receiving _more_ messages than it expects. The extra message is a warning passed on `/rosout` (introduced here https://app.chime.aws/meetings/7596499512) - the warning is irrelevant to the test, I change it to expect _at least_ the number of messages we are looking for, the specific by-type checks are sufficient.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -27,7 +27,6 @@
 #include "rosbag2_test_common/subscription_manager.hpp"
 
 #include "rosbag2_transport/rosbag2_transport.hpp"
-#include "rosbag2_transport/logging.hpp"
 
 #include "test_msgs/msg/arrays.hpp"
 #include "test_msgs/msg/basic_types.hpp"
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -27,6 +27,7 @@
 #include "rosbag2_test_common/subscription_manager.hpp"
 
 #include "rosbag2_transport/rosbag2_transport.hpp"
+#include "rosbag2_transport/logging.hpp"
 
 #include "test_msgs/msg/arrays.hpp"
 #include "test_msgs/msg/basic_types.hpp"
@@ -127,6 +128,15 @@ public:
   RosBag2PlayQosOverrideTestFixture()
   : RosBag2PlayTestFixture()
   {
+    // Because this test only cares about compatibility (receiving any messages at all)
+    // We publish one more message than we expect to receive, to avoid caring about
+    // shutdown edge behaviors that are not explicitly being tested here.
+    const size_t num_msgs_to_publish = num_msgs_to_wait_for_ + 1;
+    topic_timestamps_ms_.reserve(num_msgs_to_publish);
+    for (size_t i = 0; i < num_msgs_to_publish; i++) {
+      topic_timestamps_ms_.push_back(start_time_ms_ + message_spacing_ms_ * i);
+    }
+
     messages_.reserve(topic_timestamps_ms_.size());
     for (const auto topic_timestamp : topic_timestamps_ms_) {
       messages_.push_back(serialize_test_message(topic_name_, topic_timestamp, basic_msg_));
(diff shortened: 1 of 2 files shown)


emersonknapp:
OK - so my goal here was to make these tests pass reliably, the incorrect logic was just something I saw.

What we saw on this Action CI run was the following

```
2020-04-14T20:21:18.5513377Z The following tests FAILED:
2020-04-14T20:21:18.5513644Z 	 11 - test_record_all__rmw_fastrtps_cpp (Failed)
2020-04-14T20:21:18.5513914Z 	 18 - test_play__rmw_fastrtps_cpp (Failed)
2020-04-14T20:21:18.5514188Z 	 20 - test_record_all__rmw_fastrtps_dynamic_cpp (Failed)
2020-04-14T20:21:18.5514448Z 	 27 - test_play__rmw_fastrtps_dynamic_cpp (Failed)
```

So we are effectively seeing two failure cases, both on FastRTPS.  CycloneDDS performs these tests very reliably (hundreds of repetitions locally without problem), whereas I am able to reproduce the FastRTPS failures on repeat, usually on just the second iteration.

Specifically, the problem tests are
1. `RosBag2PlayQosOverrideTestFixture.topic_qos_profiles_overridden`
1. `RecordIntegrationTestFixture.published_messages_from_multiple_topics_are_recorded`



### Case: `topic_qos_profiles_overridden`

When this test fails, it always has received 2 out of the 3 expected messages. This is with reliable transport, so delivery retries are expected. This occurs even when adding a large initial padding time to aid discovery.

When checking the order of messages - the *first* 2 published messages are received, and the third message is never received, then the test times out.

I am now investigating further why this would happen.

### Case: `published_messages_from_multiple_topics_are_recorded`

This test is failing right now after receiving 5 messages, when only 4 are expected!

I am also reproducing this locally, investigating further why this would happen.

emersonknapp:
Update on `play` test: If I publish one more message (4) than I want to receive (3), I always receive that third message, but never the last one because the test gets shut down.

What this is starting to look like is that the publisher is being torn down before it is actually able to deliver the last message in the bag (but `publish` is definitely called). If I add a `std::this_thread::sleep_for(std::chrono::milliseconds(1));` before the player is torn down [here](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/rosbag2_transport.cpp#L101) - the test starts working reliably for me. 

This makes it seem that there is a very slight timing condition in some thread (in Fast-RTPS?) that is able to complete its work to deliver the final message if we give it a little bit of time before destroying the Player.

emersonknapp:
Something in the upgrade to Fast-RTPS 1.10.x has caused this change in behavior (https://github.com/ros2/ros2/pull/888)

If I check out Fast-RTPS 1.9.x and Fast-CDR 1.0.11 then the test works reliably again.

emersonknapp:
Latest update to this fixes the `test_play` issue. Still investigating `test_record_all`

piraka9011:
Maybe this has to do with the default QoS for the publisher manager [here](https://github.com/ros2/rosbag2/blob/23fa2e6428e7e02d9ceec66b25de8fe46deb7d94/rosbag2_test_common/include/rosbag2_test_common/publisher_manager.hpp#L81)

emersonknapp:
None of the failures ended up being QoS-related. See my latest version of the PR description.

@Karsten1987 @thomas-moulard this is read for re-review, I believe the Action CI will pass now

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10220/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5768/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8308/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10108/)

emersonknapp:
Action CI now green again (finally) - and above ci.ros2.org instability is the ros-tracing CMake warning across the board, plus the known compression windows tests failing on Windows.

Merging

PR #377, opened 2020-04-15
Expose BaseReaderInterface's BagMetadata 

emersonknapp:
Add `get_metadata` interface to BaseReaderInterface (so that Player can use it) - and provide the TopicMetadata list from it (which comes from YAML file), instead of calling the storage plugin (e.g. database) which provides an outdated version of the data.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
@@ -25,6 +25,7 @@
 #include "rosbag2_cpp/visibility_control.hpp"
 
 #include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_storage/bag_metadata.hpp"
 #include "rosbag2_storage/topic_metadata.hpp"
 
 // This is necessary because of using stl types here. It is completely safe, because
@@ -88,12 +89,20 @@ public:
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next();
 
   /**
+    * Ask bagfile for its full metadata.
+    *
+    * \return a const reference to a BagMetadata owned by the Reader
+    * \throws runtime_error if the Reader is not open.
+    */
+  const rosbag2_storage::BagMetadata & get_metadata() const;
+
+  /**
    * Ask bagfile for all topics (including their type identifier) that were recorded.
    *
    * \return vector of topics with topic name and type as std::string
    * \throws runtime_error if the Reader is not open.
    */
-  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types();
+  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() const;
 
   reader_interfaces::BaseReaderInterface & get_implementation_handle() const
   {
--- a/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
@@ -24,8 +24,8 @@
 #include "rosbag2_cpp/storage_options.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
-#include "rosbag2_storage/serialized_bag_message.hpp"
 #include "rosbag2_storage/bag_metadata.hpp"
+#include "rosbag2_storage/serialized_bag_message.hpp"
 #include "rosbag2_storage/topic_metadata.hpp"
 
 // This is necessary because of using stl types here. It is completely safe, because
--- a/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
@@ -47,6 +47,7 @@ public:
   virtual std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() = 0;
 
   virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
+
   virtual std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() = 0;
 
   virtual void set_filter(const rosbag2_storage::StorageFilter & storage_filter) = 0;
--- a/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
@@ -22,6 +22,7 @@
 #include "rosbag2_cpp/storage_options.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
+#include "rosbag2_storage/bag_metadata.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
 #include "rosbag2_storage/storage_filter.hpp"
 #include "rosbag2_storage/topic_metadata.hpp"
@@ -45,6 +46,7 @@ public:
 
   virtual std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() = 0;
 
+  virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
   virtual std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() = 0;
 
   virtual void set_filter(const rosbag2_storage::StorageFilter & storage_filter) = 0;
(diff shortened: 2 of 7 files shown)


rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
```diff
@@ -45,6 +46,7 @@ class ROSBAG2_CPP_PUBLIC BaseReaderInterface
 
   virtual std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() = 0;
 
+  virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
```
Karsten1987, line 49:
```suggestion
  virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
  
```

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
```diff
@@ -67,6 +67,7 @@ class ROSBAG2_CPP_PUBLIC SequentialReader
 
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() override;
 
+  const rosbag2_storage::BagMetadata & get_metadata() const override;
```
Karsten1987, line 70:
```suggestion
  const rosbag2_storage::BagMetadata & get_metadata() const override;
  
```

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -152,12 +152,19 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::read_ne
...
 {
-  if (storage_) {
-    return storage_->get_all_topics_and_types();
+  topics_metadata_.clear();
+  topics_metadata_.reserve(metadata_.topics_with_message_count.size());
+  for (const auto & topic_information : metadata_.topics_with_message_count) {
+    topics_metadata_.push_back(topic_information.topic_metadata);
   }
-  throw std::runtime_error("Bag is not open. Call open() before reading.");
+  return topics_metadata_;
```
Karsten1987, line 167: I think we could do this operation once - when the metadata is read initially. Don't think we need to do that every time we ask for all topics given that this data shouldn't change.

emersonknapp: Yeah, you're right - changed to fill only once.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5774/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8315/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10114/)

rosbag2_cpp/include/rosbag2_cpp/reader.hpp
```diff
@@ -25,6 +25,7 @@
 #include "rosbag2_cpp/visibility_control.hpp"
 
 #include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_storage/bag_metadata.hpp"
```
Karsten1987, line 28: alpha

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
```diff
@@ -130,10 +132,13 @@ class ROSBAG2_CPP_PUBLIC SequentialReader
   std::unique_ptr<Converter> converter_{};
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_{};
   rosbag2_storage::BagMetadata metadata_{};
+  std::vector<rosbag2_storage::TopicMetadata> topics_metadata_{};
   std::vector<std::string> file_paths_{};  // List of database files.
   std::vector<std::string>::iterator current_file_iterator_{};  // Index of file to read from
 
 private:
+  void fill_topics_and_types();
```
Karsten1987, line 140: does this have to be part of a public API?

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -152,10 +153,27 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::read_ne
...
+}
+
+
+void SequentialReader::fill_topics_and_types()
+{
+  topics_metadata_.reserve(metadata_.topics_with_message_count.size());
+  for (const auto & topic_information : metadata_.topics_with_message_count) {
+    topics_metadata_.push_back(topic_information.topic_metadata);
+  }
+}
```
Karsten1987, line 171: I think this could be a free function (inline) which takes the metadata as input and returns the topics_metadata

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -152,10 +153,27 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::read_ne
...
 }
 
-std::vector<rosbag2_storage::TopicMetadata> SequentialReader::get_all_topics_and_types()
+const rosbag2_storage::BagMetadata & SequentialReader::get_metadata() const
 {
   if (storage_) {
-    return storage_->get_all_topics_and_types();
+    return metadata_;
+  }
+  throw std::runtime_error("Bag is not open. Call open() before reading.");
```
zmichaels11, line 161:
```c++
rcpputils::check_true(storage_ != nullptr, "Bag is not open. Call open() before reading.");
return metadata_;
```
You can simplify it with this.

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
emersonknapp: Under this philosophy, are private methods ever appropriate?

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
emersonknapp: thx

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
Karsten1987: haha, maybe I read too much Scott Meyers' effective c++. 
I am not totally against private functions. I just believe in free functions if easy possible. In this case, this function is getting called once per application and doesn't really depend on an internal runtime state of the class, except the bag metadata which could easily be passed in as an argument.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -24,6 +24,19 @@
 #include "rosbag2_cpp/logging.hpp"
 #include "rosbag2_cpp/readers/sequential_reader.hpp"
 
+namespace
+{
+void fill_topics_and_types(
+  const rosbag2_storage::BagMetadata & metadata,
+  std::vector<rosbag2_storage::TopicMetadata> & topics_and_types)
+{
+  topics_and_types.reserve(metadata.topics_with_message_count.size());
```
Karsten1987, line 33: should we warn here if the incoming vector is not empty?

emersonknapp: I think we should clear it. Updated the PR to do that, lmk if you disagree with that approach.

Karsten1987: looks good.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10239/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5786/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8325/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10125/)

emersonknapp:
CMake warning from ros_tracing and the preexisting Windows build failure being tracked here https://github.com/ros2/rosbag2/issues/380

Merging

PR #382, opened 2020-04-17
fix CMake warning about using uninitialized variables

dirk-thomas:
When invoking CMake with `--warn-uninitialized`.


dirk-thomas:
@ros2/tooling_working_group Maybe someone can clarify if the CI failure is expected?

dirk-thomas:
Same failure already on master. Merging...

PR #386, opened 2020-04-22
use serialized message

Karsten1987:
Connects to https://github.com/ros2/rclcpp/pull/1081


Karsten1987:
This certainly should be refined, but I am using this commit to make CI pass and iterate over this in a follow up

PR #387, opened 2020-04-22
avoid new deprecations

wjwwood:
Connects to ros2/rclcpp#1083


wjwwood:
This was an intermediate change that is no longer needed.

Issue #326, opened 2020-03-18
Topic Remapping in play


yossioo:
## Description

An option to remap a topic in a way similar to `ros2 bag play bag_dir /topic1:=/topic2`.

## Implementation Notes / Suggestions

Maybe this can be done similarily to `run` functionality:

```
ros2 run examples_rclpy_minimal_publisher publisher_member_function /topic:=/new_topic
```

Issue #381, opened 2020-04-16
Unstable tests on Action CI


emersonknapp:
## Description

The regularly scheduled CI builds are seeing periodic failure of the test 

See the following overnight builds for example:
* https://github.com/ros2/rosbag2/runs/592733016?check_suite_focus=true
* https://github.com/ros2/rosbag2/runs/592032435?check_suite_focus=true

It does not fail every time, but has failed in approximately half of today's hourly runs.

The test times out when it fails, this suggests it is waiting for messages that are never delivered, or were delivered before the Subscriptions came online.

## System (please complete the following information)
 - OS: Ubuntu Bionic
 - ROS 2 Distro: Foxy
 - Version: master, reporting at cb44cdd1879508f5151

emersonknapp:
From the last day or so these are the test results, it's not always the same tests failing. The list of unstable tests seems to include:

* test_play_topic_remap__rmw_fastrtps_cpp (Timeout)
* test_play_topic_remap__rmw_fastrtps_dynamic_cpp (Timeout)

* test_record__rmw_fastrtps_cpp (Failed)
* test_record__rmw_fastrtps_dynamic_cpp (Failed)
* test_record__rmw_cyclonedds_cpp (Failed)
* test_play_topic_remap__rmw_fastrtps_dynamic_cpp (Failed)

* test_rosbag2_play_end_to_end (Timeout)

The timeouts probably have something to do with discovery, or shutdown behavior not receiving the final message published by a publisher.

The failures, I'm guessing are segmentation faults of some kind. It seems common with fastrtps, and occurred once in cyclone in this list.

Here's my output from the script I used to scrape, for context:

```
TEST RUN 93323352 - 2020-05-01T19:03:53Z
b'2020-05-01T19:38:31.9162579Z The following tests FAILED:'
b'2020-05-01T19:38:31.9162907Z \t 22 - test_play_topic_remap__rmw_fastrtps_cpp (Timeout)'
b'2020-05-01T19:38:31.9163376Z \t 30 - test_record__rmw_fastrtps_dynamic_cpp (Failed)'
b'2020-05-01T19:38:31.9164216Z \t 33 - test_play_topic_remap__rmw_fastrtps_dynamic_cpp (Timeout)'
b'2020-05-01T19:38:31.9164349Z Errors while running CTest'

b'2020-05-01T19:40:03.0866840Z The following tests FAILED:'
[...]

PR #388, opened 2020-04-22
add topic remapping option to rosbag2 play

ketatam:
With the option --remap or -m it's now possible to do topic remapping in rosbag2 play.
This is done by passing the remapping options to the play node.

Closes https://github.com/ros2/rosbag2/issues/326.

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -52,6 +52,10 @@ class PlayVerb(VerbExtension):
             '-l', '--loop', action='store_true',
             help='enables loop playback when playing a bagfile: it starts back at the beginning '
                  'on reaching the end and plays indefinitely.')
+        parser.add_argument(
+            '--remap', '-m', default='', nargs='+',
+            help='list of topics to be remapped: in the form '
+                 '"old_topic1:=new_topic1 old_topic2:=new_topic2 etc." ')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -77,4 +81,5 @@ class PlayVerb(VerbExtension):
             rate=args.rate,
             topics=args.topics,
             qos_profile_overrides=qos_profile_overrides,
-            loop=args.loop)
+            loop=args.loop,
+            topic_remapping=args.remap)
--- a/rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
@@ -85,7 +85,9 @@ public:
   void print_bag_info(const std::string & uri, const std::string & storage_id);
 
 private:
-  std::shared_ptr<Rosbag2Node> setup_node(std::string node_prefix = "");
+  std::shared_ptr<Rosbag2Node> setup_node(
+    std::string node_prefix = "",
+    const std::vector<std::string> & topic_remapping_options = {});
 
   std::shared_ptr<rosbag2_cpp::Reader> reader_;
   std::shared_ptr<rosbag2_cpp::Writer> writer_;
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -169,6 +169,11 @@ function(create_tests_for_rmw_implementation)
     LINK_LIBS rosbag2_transport
     AMENT_DEPS test_msgs rosbag2_test_common)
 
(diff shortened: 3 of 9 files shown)


rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
```diff
@@ -74,7 +74,9 @@ class Rosbag2Transport
    * \param play_options Options regarding the playback (e.g. queue size)
    */
   ROSBAG2_TRANSPORT_PUBLIC
-  void play(const StorageOptions & storage_options, const PlayOptions & play_options);
+  void play(
+    const StorageOptions & storage_options, const PlayOptions & play_options,
+    const std::vector<std::string> & topic_remapping_options = {});
```
emersonknapp, line 79: Do you think it would be more appropriate to add `topic_name_map` to `PlayOptions`/`RecordOptions` rather than adding new arguments to the API?

ros2bag/ros2bag/verb/play.py
```diff
@@ -52,6 +52,10 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '-l', '--loop', action='store_true',
             help='enables loop playback when playing a bagfile: it starts back at the beginning '
                  'on reaching the end and plays indefinitely.')
+        parser.add_argument(
```
emersonknapp, line 55: should we call this `--ros-args` (and call out that it enables topic name remapping in the `help` string), since you can pass any `ros-args` here and it'll work?

Martin-Idel: Yes, we were wondering that. My reasoning not to go that route was the following:
- `--ros-args` was introduced in particular to avoid name clashes with user defined flags on user ros2 nodes, in other words to separate flags defined by rclcpp or rclpy, etc. from flags defined on nodes deriving from those
- Rosbag isn't the classical node and I don't see a scenario where you'd want to provide a custom rosbag node, so separating "namespaces" of flags is not a big deal here
- It could be interesting to expose more flags from rosbag's node, but I wonder whether it is a good idea to expose all of them. Therefore, I'd say: keep things hidden, so we'd only expose `--ros-args --remap` for now. I also wondered whether at that point it isn't confusing if the notation is the same for a general node but not all options you can generally pass to `--ros-args` are available.
What do you think about this?

ketatam: I have added added a new topic remapping options to the play options. Please let me know if there are still some other problems

rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
```diff
@@ -74,7 +74,8 @@ class Rosbag2Transport
    * \param play_options Options regarding the playback (e.g. queue size)
    */
   ROSBAG2_TRANSPORT_PUBLIC
-  void play(const StorageOptions & storage_options, const PlayOptions & play_options);
+  void play(
+    const StorageOptions & storage_options, const PlayOptions & play_options);
```
emersonknapp, line 78: Remove unrelated change
```suggestion
  void play(const StorageOptions & storage_options, const PlayOptions & play_options);
```

ketatam: done!

Martin-Idel:
CI (gist: https://gist.githubusercontent.com/Martin-Idel/f92c7fb9c27def6447c87d222532d75c/raw/c42febac209f511b0b28d03628e2ccd2b6571e42/ros2_rosbag_topic_remapping.repos):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10515/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5982/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8567/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10389/)

ketatam:
@emersonknapp  the CI is unstable but I think it's not because of my code. So can you please check this and if I do have to change somthing just let me know. Thanks!

emersonknapp:
I took a look through the builds and yes, I don't see anything here that's specifically related to this PR.

The Action CI is the continuing https://github.com/ros2/rosbag2/issues/381
OSX and Windows build warnings are all in Fast-RTPS
The consistent test failure across platforms seems to be an issue with Connext type support, I don't want to block this change on any changes happening upstream there.

We'll need to keep an eye on these issues leading up to the release candidate package, but for now I think we're good to merge.

Issue #307, opened 2020-03-02
Ability to playback bags with unknown message types


clalancette:
## Description

Sometimes I am handed a bagfile that has messages I don't have the sources for.  Right now if I try to play those bag files back with `ros2 bag play`, I get an error message like:

```
[ERROR] [1583183596.808870285] [rosbag2_transport]: Failed to play: package 'custom_pkg' not found, searching: [/home/ubuntu/ros2_ws/install/rosbag2, /home/ubuntu/ros2_ws/install/ros2bag, ...]
```

It would be nice if I could still play back these bag files, and `ros2 bag play` would just play the ones it could deserialize.  Basically I am asking for this error to be downgraded to a warning, and skipping the messages that can't be played back.

I'm currently working around this by manually using sqlite3 to remove the messages I don't have the sources for.

## Related Issues

None.

## Completion Criteria

* [ ] Bag file with unknown messages can be played back, skipping the message types that are unknown. This errors out normally, but can be enabled with a flag (maybe `--ignore-unknown-types`)

## Implementation Notes / Suggestions

Make the initial error messages about unknown messages a warning, and then skip those messages that can be deserialized during playback.

## Testing Notes / Suggestions

Make sure there is a bag file with unknown message types, and that it can be played back.

emersonknapp:
Good idea, perhaps an explicit argument to `play`, something like `--ignore-unknown-types`

clalancette:
> Good idea, perhaps an explicit argument to `play`, something like `--ignore-unknown-types`

I'd lean towards making it the default, with a warning, but a command-line option would work too.

kielczykowski:
Shouldn't this issue be closed due to https://github.com/ros2/rosbag2/pull/592 and https://github.com/ros2/rosbag2/pull/394 or is the implementation still needed?

emersonknapp:
Good catch - this is closed by #592

PR #394, opened 2020-04-28
add the possibility to play back bagfiles with unknown message types

ketatam:
With the option --ignore-unknown-types it is now possible to skip messages, for which we don't have the sources.

The implementation is based on topic filtering, in the sense that at first we check which topics can be actually played and those are filtered, and we create publishers only for these topics.
One small problem is that with the current implementation of the topic_filter an empty topics_to_filter vector means that all topics should be played, but with the idea of my implementation when the topics_to_filter vector is empty that would mean that we can not play any topic. I found a solution for this but I think it would be better to change the meaning of an empty topics_to_filter vector so that it would mean no topic should be played.
Closes https://github.com/ros2/rosbag2/issues/307.

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -56,6 +56,10 @@ class PlayVerb(VerbExtension):
             '--remap', '-m', default='', nargs='+',
             help='list of topics to be remapped: in the form '
                  '"old_topic1:=new_topic1 old_topic2:=new_topic2 etc." ')
+        parser.add_argument(
+            '--ignore-unknown-types', action='store_true',
+            help='enables playing back bag files with unknown messages, skipping the message '
+                 'types that are unknown.')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -82,4 +86,5 @@ class PlayVerb(VerbExtension):
             topics=args.topics,
             qos_profile_overrides=qos_profile_overrides,
             loop=args.loop,
-            topic_remapping=args.remap)
+            topic_remapping=args.remap,
+            ignore_unknown_types=args.ignore_unknown_types)
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -1,4 +1,5 @@
 // Copyright 2018, Bosch Software Innovations GmbH.
+// Copyright 2020, TNG Technology Consulting GmbH.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,6 +22,8 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+
+#include "ament_index_cpp/get_resource.hpp"
 
 #include "rcl/graph.h"
 
@@ -95,6 +98,10 @@ void Player::play(const PlayOptions & options)
 {
   topic_qos_profile_overrides_ = options.topic_qos_profile_overrides;
   prepare_publishers(options);
+
+  if (publishers_.size() == 0) {
+    return;
+  }
 
   storage_loading_future_ = std::async(
     std::launch::async,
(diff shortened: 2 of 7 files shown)


ros2bag/ros2bag/verb/play.py
```diff
@@ -56,6 +56,10 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '--remap', '-m', default='', nargs='+',
             help='list of topics to be remapped: in the form '
                  '"old_topic1:=new_topic1 old_topic2:=new_topic2 etc." ')
+        parser.add_argument(
+            '--ignore-unknown-types', action='store_true',
+            help='enables playing back bag files with unknown messages, skipping the message '
+                 'types that are unknown.')
```
Karsten1987, line 62: I am actually not sure if that command line option is really needed. I think it makes sense to simply skip over the non-playable messages and issue a warning with it.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -188,19 +195,65 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
   }
 }
 
+bool is_package_found(rosbag2_storage::TopicMetadata topic)
```
Karsten1987, line 198: that name is a bit misleading in my opinion. What about `message_type_available`?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -188,19 +195,65 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
 {
+  auto topics_to_filter = options.topics_to_filter;
+  if (options.ignore_unknown_types) {
+    topics_to_filter = update_topics_to_filter_to_play_messages_with_unknwon_types(
+      options.topics_to_filter);
+    if (topics_to_filter.size() == 0) {
+      ROSBAG2_TRANSPORT_LOG_ERROR("No topic to be sucribed to could be found.");
+      return;
+    }
+  }
```
Karsten1987, line 238: I don't think that's necessary. We apply the filter, get all topics and types and filter then once which of the topics can actually be replayed.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -188,19 +195,65 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
-    auto topic_qos = publisher_qos_for_topic(topic, topic_qos_profile_overrides_);
-    publishers_.insert(
-      std::make_pair(
-        topic.name, rosbag2_transport_->create_generic_publisher(
-          topic.name, topic.type, topic_qos)));
+    //  We have to create a publisher only for the topics we want to play.
+    bool publisher_should_be_created = topics_to_filter.size() == 0 ||
+      std::find(
+      topics_to_filter.begin(), topics_to_filter.end(),
+      topic.name) != topics_to_filter.end();
```
Karsten1987, line 249: I would simply invert this.
```c++
if (std::find_if(topics.begin(), topics.end(), message_type_available(topic)) != topics.end()) {
  // create pub
} else {
  // issue warning about topic being skipped
}
```

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -62,6 +63,8 @@ class Player
   void prepare_publishers(const PlayOptions & options);
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
   static const std::chrono::milliseconds queue_read_wait_period_;
+  std::vector<std::string> update_topics_to_filter_to_play_messages_with_unknwon_types(
+    const std::vector<std::string> & topics_to_filter);
```
Karsten1987, line 67: as mentioned above, I don't think we need that function.

rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
```diff
@@ -42,7 +42,7 @@ class MockSequentialReader : public rosbag2_cpp::reader_interfaces::BaseReaderIn
       return num_read_ < messages_.size();
     }
 
-    while (num_read_ < messages_.size()) {
+    while (num_read_ < messages_.size() - 1) {
```
Karsten1987, line 45: why that change?

rosbag2_transport/src/rosbag2_transport/player.cpp
Karsten1987: alternatively, we filter out all topics which we can publish and pass these in via storage_filters. We can then later on be sure that all topics can be replayed and safely create publishers for it.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -188,19 +195,65 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
+      topics_to_filter_updated.emplace_back(topic.name);
+    } else if (!is_package_found(topic)) {
+      std::string package_name = std::get<0>(rosbag2_cpp::extract_type_identifier(topic.type));
+      ROSBAG2_TRANSPORT_LOG_WARN(
+        "All messages on topic '%s' will be ignored because package '%s' is not found.",
+        topic.name.c_str(), package_name.c_str());
+    }
+  }
+  return topics_to_filter_updated;
+}
```
Karsten1987, line 226: isn't that just `std::set_intersection`?

emersonknapp:
@ketatam I know this was opened a long time ago, is it something you are still interested in following through on?

emersonknapp:
Closing in favor of https://github.com/ros2/rosbag2/pull/592 which implements this feature

PR #400, opened 2020-05-02
Fix typesupport loading

Karsten1987:
The change to `rcpputils::SharedLibrary` in #322 introduced a regression to the converter logic. 

copied from #322 
> I haven't paid enough attention while reviewing this PR, but I think it's really bad practice to pass in a shared_pointer by reference (std::shared_pointer<rcpputils::SharedLibrary> & library) and somewhat blindly overwrite it by calling std::make_shared on it. Also, this assumes that the pointer is empty, which isn't checked here either.
The reason why it fails in the end is that the same pointer is being passed in to load multiple typesupports, which are stored in add_topics. Later on, when using these typesupport symbols to create a new introspection message, these pointers have dangled.

This PR is fixing this by extending the `ConverterTypeSupport` struct with the shared library. It is currently using a shared pointer to it, even though I don't understand why this is needed and a reference isn't sufficient. Maybe @ahcorde can give some more insights here. 

This is needed to unblock https://github.com/ros2/rosbag2_bag_v2/pull/28

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
@@ -98,11 +98,25 @@ extract_type_identifier(const std::string & full_type)
   return std::make_tuple(package_name, middle_module, type_name);
 }
 
-const rosidl_message_type_support_t *
-get_typesupport(
-  const std::string & type, const std::string & typesupport_identifier,
-  std::shared_ptr<rcpputils::SharedLibrary> & library)
+std::shared_ptr<rcpputils::SharedLibrary>
+get_typesupport_library(const std::string & type, const std::string & typesupport_identifier)
 {
+  auto package_name = std::get<0>(extract_type_identifier(type));
+  auto library_path = get_typesupport_library_path(package_name, typesupport_identifier);
+  return std::make_shared<rcpputils::SharedLibrary>(library_path);
+}
+
+const rosidl_message_type_support_t *
+get_typesupport_handle(
+  const std::string & type,
+  const std::string & typesupport_identifier,
+  std::shared_ptr<rcpputils::SharedLibrary> library)
+{
+  if (nullptr == library) {
+    throw std::runtime_error(
+            "rcpputils::SharedLibrary not initialized. Call get_typesupport_library first.");
+  }
+
   std::string package_name;
   std::string middle_module;
   std::string type_name;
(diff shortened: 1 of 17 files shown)


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10616/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6055/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8648/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10511/)

rosbag2_cpp/src/rosbag2_cpp/typesupport_helpers.cpp
```diff
@@ -116,8 +116,11 @@ get_typesupport(
   auto library_path = get_typesupport_library_path(package_name, typesupport_identifier);
 
   try {
-    library = std::make_shared<rcpputils::SharedLibrary>(library_path);
-
+    // TODO(karsten1987) This logic is pretty fragile.
+    // The library should not be passed in via a reference to a shared pointer.
```
emersonknapp, line 120: So should we just fix the interface now?

Karsten1987: I was thinking about splitting the function into two parts. One which initializes the library and a second which extracts the type support handle from it. 

What do you think?

emersonknapp: That sounds good to me - I feel like it makes sense to handle it immediately rather than put a todo - unless it's significantly more work

Karsten1987: I'll try to get to this by the end of today.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10626/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6060/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8653/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10529/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10639/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6069/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8662/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10547/)

Karsten1987:
That is up for review.

I am still a bit concerned about this as every introspection message is now strongly coupled with the `rcpputils::SharedLibrary` in order to be cleaned up correctly. However, this would need a bigger round of code review and in order to get the beta going and the `rosbag2_bag_v2_plugin` unblocked, this should be good to go.

Karsten1987:
@emersonknapp @zmichaels11 friendly ping for review

PR #401, opened 2020-05-02
Update GitHub workflows to use Docker, and Focal

thomas-moulard:
This pull request introduces two changes:

a/ lint.yml is using the latest version of action-ros-lint, and runs it on
   a Docker image built by ros-tooling/setup-ros-docker.
   Using a docker image instead of ros-tooling/setup-ros removes the
   need for running APT, and other commands which requires internet
   connectivity and may become a source of flakyness in the CI pipeline.
   Instead, the worker runs a Docker image with the linter
   pre-installed.

b/ test.yml now uses a Focal docker image. This also allows us to skip
   invoking ros-tooling/setup-ros, which speeds up the CI, and reduces
   risks of flakyness.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,8 @@
 # Run linters automatically on pull requests.
+#
+# ros-tooling/actino-ros-lint is relying on the latest APT binary packages.
+# As of 2020-05-01, Eloquent is the latest release for which APT binary
+# packages are available, which is why the Docker image is based on Eloquent.
 name: Lint rosbag2
 on:
   pull_request:
@@ -6,11 +10,14 @@ on:
 jobs:
   ament_copyright:
     name: ament_copyright
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container:
+      image: rostooling/setup-ros-docker:ubuntu-bionic-ros-eloquent-ros-base-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: ros-tooling/setup-ros@0.0.13
-    - uses: ros-tooling/action-ros-lint@0.0.5
+    # TODO(setup-ros-docker#7): calling chown is necessary for now
+    - run: sudo chown -R rosbuild:rosbuild "$HOME" .
+    - uses: actions/checkout@v2
+    - uses: ros-tooling/action-ros-lint@0.0.6
       with:
         linter: copyright
         package-name: |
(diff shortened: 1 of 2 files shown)


thomas-moulard:
First attempt failed for the following reason (retrying to run the CI):

```
  	  2 - test_rosbag2_play_end_to_end (Timeout)
```

emersonknapp:
That test seems to be known unstable - it seems to pass less frequently than it succeeds on the Action runners, but almost always passes on the buildfarm CI. I'm not sure why - it is recorded here https://github.com/ros2/rosbag2/issues/381 - maybe we want to move forward with this PR, see if it makes the other tests pass more reliably, and track down this last one as a followup?

thomas-moulard:
@Karsten1987 WDYT? Would you be OK going forward with this, or is it better to make the test less flaky first, and then merge this?
My gut feeling is that the change of platform is making the test more unstable but it's unrelated to this change.

emersonknapp:
Oh - this exact test appears to be failing in the nightly builds, at least last night's - https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1543/testReport/ i looked at amd64 and aarch64, it failed in both. so we can safely say it's not related to this change. I think we're better off matching the buildfarm Focal environment than staying on Bionic - I can't think of a reason to hold back the OS

Karsten1987:
I am okay with migrating the GH Actions to focal. Makes sense to stay as close as possible to the ROS2 buildfarm. 
Question though: Why are the linters using bionic and eloquent? Why not using the ROS2 master version for linting as well?

Regarding the tests, they are failing quite frequently and I would really appreciate if we could take time to either fix them or disable them.

thomas-moulard:
We could be building the linters from source and run the linters only in `action-ros-lint`. The reason it has not been a priority is because `action-ros-ci` is already doing that (linters are part of the test suite).

At heart, `action-ros-lint` is really an optimization for PRs only, you are less precise by using an APT package, but it's useful to have near instantaneous feedback that you are breaking some lint rule. If you are OK waiting, better using `action-ros-ci` as it runs `colcon-test` and this is the source of truth.

We could, I guess, either build a Docker image every night with the linters only, or compile only the linters from source, but it has not been a priority until now as the current approach is "acceptable enough" for us, and other improvements have been deemed more urgent, but I understand your point of view, and a PR adding those features would be welcome.

+1 on fixing the test - @emersonknapp @prajakta-gokhale can we get that him in the sprint as part of the QA push?

PR #403, opened 2020-05-04
export targets

### LINKED ISSUE #402: make rosbag2 related packages relocatable
Atm various `.cmake` files installed by `rosbag2_*` packages and its dependencies contain absolute paths to the install space which makes them non-relocatable. See:

* https://github.com/ros2/ros2/issues/915#issuecomment-623552072
* https://github.com/ros2/ros2/issues/904

Karsten1987:
Closes #402 
Requires https://github.com/ros/pluginlib/pull/190


Karsten1987:
This is currently work in progress. 

I am facing the following CMake error which I have to figure out what's happening:

```
--- stderr: rosbag2_storage
CMake Error in CMakeLists.txt:
  Target "rosbag2_storage" INTERFACE_INCLUDE_DIRECTORIES property contains
  path:

    "/Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_storage/include"

  which is prefixed in the source directory.


CMake Error: install(EXPORT "export_rosbag2_storage" ...) includes target "rosbag2_storage" which requires target "tinyxml2_vendor" that is not in any export set.
CMake Generate step failed.  Build files cannot be regenerated correctly.
---
Failed   <<< rosbag2_storage	[ Exited with code 1 ]
```

Karsten1987:
The above error could be resolved by https://github.com/ros/pluginlib/pull/190. With that PR, I am putting this PR out of draft mode.

I think the exported targets look decent so far. There's no abosolute path in it. The absolute path for the include directories are coming from other packages as far as I can tell.

```cmake
set_target_properties(rosbag2_transport::rosbag2_transport PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/Users/karsten/workspace/ros2/ros2_master/install/include;/Users/karsten/workspace/ros2/ros2_master/install/opt/yaml_cpp_vendor/lib/cmake/yaml-cpp/../../../include"
  INTERFACE_LINK_LIBRARIES "rcl::rcl;rclcpp::rclcpp;rcutils::rcutils;rmw::rmw;rosbag2_compression::rosbag2_compression;rosbag2_compression::rosbag2_compression_zstd;rosbag2_cpp::rosbag2_cpp;yaml-cpp"
)
```

Karsten1987:
@emersonknapp @zmichaels11 Maybe you can help me out. I honestly don't really get how that `zstd_vendor` package is working. Especially the line in here: https://github.com/ros2/rosbag2/blob/69c58939bc5a873c6a25b7e36af81f9410c48ce0/zstd_vendor/CMakeLists.txt#L52
What is `zstd` in this case? 

From what I can see [here](https://github.com/facebook/zstd/blob/7120d97cb3d3ad65986fb653f0fdec7fd7a346b8/build/cmake/lib/CMakeLists.txt#L182-L185) zstd does use already exported cmake targets and we should actually just leverage these. However, I am somewhat confused that there is no `find_package(zstd)` within `rosbag2_compression`.

zmichaels11:
> @emersonknapp @zmichaels11 Maybe you can help me out. I honestly don't really get how that `zstd_vendor` package is working. Especially the line in here:
> 
> https://github.com/ros2/rosbag2/blob/69c58939bc5a873c6a25b7e36af81f9410c48ce0/zstd_vendor/CMakeLists.txt#L52
> 
> What is `zstd` in this case?
> 
> From what I can see [here](https://github.com/facebook/zstd/blob/7120d97cb3d3ad65986fb653f0fdec7fd7a346b8/build/cmake/lib/CMakeLists.txt#L182-L185) zstd does use already exported cmake targets and we should actually just leverage these. However, I am somewhat confused that there is no `find_package(zstd)` within `rosbag2_compression`.

@piraka9011 do you have input on this design choice?

emersonknapp:
I'm not too familiar with the compression code either - it seems like it's just declaring the preexisting library, either system installed or built via ExternalProject - as an ament target?

Karsten1987:
So I worked around this now by providing a `Findzstd.cmake` similar to what is done in the `sqlite3_vendor` package. I've figured out that version 1.4.4 of zstd does not provide any modern CMake logic (neither a `find_package` so to speak).

piraka9011:
@Karsten1987 `zstd` does not have modern CMake exports as you described so instead I exported the library as [`zstd_vendor`](https://github.com/ros2/rosbag2/blob/69c58939bc5a873c6a25b7e36af81f9410c48ce0/rosbag2_compression/CMakeLists.txt#L28) and linked using the [`zstd_vendor_INCLUDE_DIRS`](https://github.com/ros2/rosbag2/blob/69c58939bc5a873c6a25b7e36af81f9410c48ce0/rosbag2_compression/CMakeLists.txt#L38) and [`zstd_vendor_LIBRARIES`](https://github.com/ros2/rosbag2/blob/master/rosbag2_compression/CMakeLists.txt#L39) CMake variables.

dirk-thomas:
I would suggest to split the non-export target related changes into a separate PR to make this easier to review.

sqlite3_vendor/CMakeLists.txt
```diff
@@ -33,7 +33,9 @@ if (NOT SQLite3_FOUND)
     )
 
   # The external project will install to the build folder, but we'll install that on make install.
-  install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/sqlite3_install/ DESTINATION ${CMAKE_INSTALL_PREFIX})
+  install(
+    DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/sqlite3_install/
+    DESTINATION .)
```
clalancette, line 38: All the rest of our vendor packages use `${CMAKE_INSTALL_PREFIX}` here:

* https://github.com/ros2/spdlog_vendor/blob/3cae65ab249d1ef3c9bfa203497007d8c7ef299f/CMakeLists.txt#L78
* https://github.com/ros2/yaml_cpp_vendor/blob/efe04baa5b3903fa238204f291f50f05d4fcdd18/CMakeLists.txt#L62
* https://github.com/ros2/rosbag2/blob/bfa63199421e53b8d229b9403ac7dd4ad4f155cb/sqlite3_vendor/CMakeLists.txt#L36
* https://github.com/ros2/rosbag2/blob/bfa63199421e53b8d229b9403ac7dd4ad4f155cb/shared_queues_vendor/CMakeLists.txt#L38
* https://github.com/ros2/tinyxml_vendor/blob/b7a50bf7af9f7ae8c517dd06552533d4e5f4a726/CMakeLists.txt#L37
* https://github.com/ros2/rviz/blob/7ca470831304a1cd29ec207f2c07e62ddb8552f1/rviz_assimp_vendor/CMakeLists.txt#L51
* https://github.com/ros2/rviz/blob/7ca470831304a1cd29ec207f2c07e62ddb8552f1/rviz_ogre_vendor/CMakeLists.txt#L61
* https://github.com/ros2/poco_vendor/blob/3f23e18f8d71e979e03395d860bc3053d3162ae6/CMakeLists.txt#L87
* https://github.com/ament/uncrustify_vendor/blob/0291bc988c45846400aca50cb973bf04a4de3d56/CMakeLists.txt#L63
* https://github.com/ros/resource_retriever/blob/d6a6818edeebcc3af73b77ebf88d3116d489089e/libcurl_vendor/CMakeLists.txt#L64

That doesn't make the rest of them automatically correct, but I'm not sure why this one is different.  Do we need to update all of those to use `.`, or can we just leave this one as it was?

Karsten1987: I did this based on this comment: https://github.com/ros2/console_bridge_vendor/pull/9#pullrequestreview-405300466

dirk-thomas: It was changed under the impression that older CMake versions had an issue otherwise. I would guess that was won't the problem was coming from somewhere else. So referring that change sounds reasonable.

Karsten1987:
I am okay with splitting this PR.

I think I could do the following, which could hopefully be merged individually: 
* changing to modern cmake targets 
* make rosbag2 relocatable (specifically their vendor packages)
* general touchups

shared_queues_vendor/CMakeLists.txt
```diff
@@ -5,8 +5,9 @@ find_package(ament_cmake REQUIRED)
 
 include(ExternalProject)
 # Single producer single consumer queue by moodycamel - header only, don't build, install
-ExternalProject_Add(singleproducerconsumer
-  DOWNLOAD_DIR ${CMAKE_CURRENT_BINARY_DIR}/single
+ExternalProject_Add(ext-singleproducerconsumer
```
Karsten1987, line 8: I deliberately left in these changes because the cmake target of the `ExternalProject_Add` would collide with `add_library`.

Karsten1987:
I would ask for another round of review here. I hope I could strip down the packages well enough to make the changes easier to grasp.

zstd_vendor/cmake/Modules/Findzstd.cmake
```diff
@@ -0,0 +1,41 @@
...
+
+    if(NOT TARGET zstd::zstd)
+        add_library(zstd::zstd UNKNOWN IMPORTED)
+        set_property(TARGET zstd::zstd PROPERTY IMPORTED_LOCATION ${zstd_LIBRARY})
+        set_property(TARGET zstd::zstd PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${zstd_INCLUDE_DIR})
+    endif()
+    list(APPEND zstd_TARGETS zstd::zstd)
+else()
+    message(FATAL_ERROR "Unable to find zstd")
+endif()
```
dirk-thomas, line 41: I guess this package doesn't use linters yet.

rosbag2_cpp/CMakeLists.txt
```diff
@@ -87,16 +87,19 @@ install(
...
   RUNTIME DESTINATION bin)
 
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
+ament_export_targets(export_${PROJECT_NAME})
 ament_export_dependencies(pluginlib
   rosbag2_storage
   rosidl_runtime_c
   rosidl_runtime_cpp
+  rosidl_typesupport_cpp
```
dirk-thomas, line 102: How is this related to the change?

zstd_vendor/cmake/Modules/Findzstd.cmake
```diff
@@ -0,0 +1,41 @@
...
+    set(zstd_LIBRARIES ${zstd_LIBRARY})
+
+    if(NOT TARGET zstd::zstd)
+        add_library(zstd::zstd UNKNOWN IMPORTED)
+        set_property(TARGET zstd::zstd PROPERTY IMPORTED_LOCATION ${zstd_LIBRARY})
+        set_property(TARGET zstd::zstd PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${zstd_INCLUDE_DIR})
+    endif()
+    list(APPEND zstd_TARGETS zstd::zstd)
+else()
+    message(FATAL_ERROR "Unable to find zstd")
```
dirk-thomas, line 40: Nitpick: I usually try to exit early to avoid indentation, like:

```
if(NOT zstd_FOUND)
  message(FATAL_ERROR "Unable to find zstd")
endif()

...
```

rosbag2_cpp/CMakeLists.txt
Karsten1987: From my understanding, when importing a cmake target, all dependencies have to present at cmake's generation step as opposed to using `${<pkg>_LIBRARIES}` which is applied during linking. It's therefore important to export all necessary dependencies.

zstd_vendor/cmake/Modules/Findzstd.cmake
Karsten1987: it does not and but can be added in a follow up PR.

zstd_vendor/cmake/Modules/Findzstd.cmake
Karsten1987: Yes, exiting early is preferred. 

There was actually another problem in here though, in that it would fail (`FATAL_ERROR`) regardless whether `find_package(zstd)` was called with `REQUIRED` or not. So I've changed that. With this however, I don't see a good way of exiting earlier anymore.

Issue #404, opened 2020-05-06
rosbag2_compression relies on relative header in another package


jacobperron:
## Description

rosbag2_compression includes relative headers from a different package:

https://github.com/ros2/rosbag2/blob/69c58939bc5a873c6a25b7e36af81f9410c48ce0/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_reader.cpp#L26-L29

This causes a compilation error in packaging jobs:

```
17:29:00 /tmp/binarydeb/ros-foxy-rosbag2-compression-0.2.5/test/rosbag2_compression/test_sequential_compression_reader.cpp:26:10: fatal error: ../../rosbag2_cpp/test/rosbag2_cpp/mock_converter_factory.hpp: No such file or directory
17:29:00    26 | #include "../../rosbag2_cpp/test/rosbag2_cpp/mock_converter_factory.hpp"
17:29:00       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The bug is causing the Debian packaging job to fail: http://build.ros2.org/view/Fbin_uF64/job/Fbin_uF64__rosbag2_compression__ubuntu_focal_amd64__binary/5/

To remove these relative includes, we should either duplicate the needed mock objects in rosbag2_compression or install them to be shared by multiple packages.

Karsten1987:
clsoed via #405

jacobperron:
Thanks @zmichaels11 and @Karsten1987 !

PR #405, opened 2020-05-06
Remove relative paths from includes

zmichaels11:
## Description
~~Adds `rosbag2_cpp/test` path to includes in `rosbag2_compression` to replace relative includes.~~
Duplicate mocks from `rosbag2_cpp` in `rosbag2_compression` so relative includes can be removed.

This should resolve this issue: https://github.com/ros2/rosbag2/issues/404

Follow-up work should be done to de-duplicate this code.

Signed-off-by: Zachary Michaels <zmichaels11@gmail.com>

### DIFF
--- a/rosbag2_compression/test/rosbag2_compression/mock_converter.hpp
+++ b/rosbag2_compression/test/rosbag2_compression/mock_converter.hpp
@@ -1,0 +1,45 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2_COMPRESSION__MOCK_CONVERTER_HPP_
+#define ROSBAG2_COMPRESSION__MOCK_CONVERTER_HPP_
+
+#include <gmock/gmock.h>
+
+#include <memory>
+#include <string>
+
+#include "rosbag2_cpp/converter_interfaces/serialization_format_converter.hpp"
+
+#include "rosbag2_storage/serialized_bag_message.hpp"
+
+class MockConverter : public rosbag2_cpp::converter_interfaces::SerializationFormatConverter
+{
+public:
+  MOCK_METHOD3(
+    deserialize,
+    void(
+      std::shared_ptr<const rosbag2_storage::SerializedBagMessage>,
+      const rosidl_message_type_support_t *,
+      std::shared_ptr<rosbag2_cpp::rosbag2_introspection_message_t>));
+
+  MOCK_METHOD3(
+    serialize,
+    void(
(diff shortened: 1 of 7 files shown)


rosbag2_compression/CMakeLists.txt
```diff
@@ -140,12 +140,14 @@ if(BUILD_TESTING)
   ament_add_gmock(test_sequential_compression_reader
     test/rosbag2_compression/test_sequential_compression_reader.cpp)
   target_include_directories(test_sequential_compression_reader PUBLIC include)
+  target_include_directories(test_sequential_compression_reader PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/../rosbag2_cpp/test")
```
jacobperron, line 143: I don't think this will work. I believe we'll run into the same underlying issue that `rosbag2_cpp` is not accessible from the source tree of `rosbag2_compression`. The source debs are created from [this release branch](https://github.com/ros2-gbp/rosbag2-release/tree/release/foxy/rosbag2_compression), which does not contain the source code for `rosbag2_cpp`.

thomas-moulard, line 143:
```suggestion
  target_include_directories(test_sequential_compression_reader PRIVATE "${CMAKE_SOURCE_DIR}/test")
```

rosbag2_compression/CMakeLists.txt
```diff
@@ -140,12 +140,14 @@ if(BUILD_TESTING)
...
   target_link_libraries(test_sequential_compression_reader ${PROJECT_NAME})
   ament_target_dependencies(test_sequential_compression_reader rosbag2_cpp)
 
   ament_add_gmock(test_sequential_compression_writer
     test/rosbag2_compression/test_sequential_compression_writer.cpp)
   target_include_directories(test_sequential_compression_writer PUBLIC include)
+  target_include_directories(test_sequential_compression_writer PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/../rosbag2_cpp/test")
```
thomas-moulard, line 150: ditto

rosbag2_compression/CMakeLists.txt
zmichaels11: Unfortunately that won't work since its in `rosbag2_cpp` and not `rosbag2_compression`

zmichaels11: hmm, in that case I think I'll need to just duplicate the code into `rosbag2_compression`

jacobperron: Either that or install the test headers in rosbag2_cpp. Maybe it could be a separate “test” target.

piraka9011:
Why not put these in `rosbag2_test_common` instead?

zmichaels11:
> Why not put these in `rosbag2_test_common` instead?

I think it would create a circular dependency since the involved classes are mocks of objects declared in `rosbag2_cpp` and `rosbag2_cpp` already includes `rosbag2_test_common`.

jacobperron:
Consider opening a ticket to track the follow-up task of de-duplicating code.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10656/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6085/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8671/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10562/)

Karsten1987:
CI comes back as expected. Going to merge and cut a new release to unblock the Foxy beta release.

PR #409, opened 2020-05-13
Add split by time to recording

jhdcs:
This is my first attempt at adding splitting bags by time functionality to Rosbag 2's recording. Currently having some difficulty with adding the new functionality to the command line for testing.

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -60,6 +60,12 @@ class RecordVerb(VerbExtension):
                   'Default it is zero, recording written in single bagfile and splitting '
                   'is disabled.'
         )
+        parser.add_argment(
+            '-d', '--max-bag-duration', type=int, default=0,
+            help='maximum duration in nanosecnds before the bagfile will be split. '
+                  'Default is zero, recording written in single bagfile and splitting '
+                  'is disabled.'
+        )
         parser.add_argument(
             '--max-cache-size', type=int, default=0,
             help='maximum amount of messages to hold in cache before writing to disk. '
@@ -129,6 +135,7 @@ class RecordVerb(VerbExtension):
                 no_discovery=args.no_discovery,
                 polling_interval=args.polling_interval,
                 max_bagfile_size=args.max_bag_size,
+                max_bagfile_duration=args.max_bag_duration,
                 max_cache_size=args.max_cache_size,
                 include_hidden_topics=args.include_hidden_topics,
                 qos_profile_overrides=qos_profile_overrides)
@@ -150,6 +157,7 @@ class RecordVerb(VerbExtension):
                 no_discovery=args.no_discovery,
                 polling_interval=args.polling_interval,
                 max_bagfile_size=args.max_bag_size,
+                max_bagfile_duration=args.max_bag_duration,
                 max_cache_size=args.max_cache_size,
                 topics=args.topics,
                 include_hidden_topics=args.include_hidden_topics,
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -60,7 +60,7 @@ class RecordVerb(VerbExtension):
                   'Default it is zero, recording written in single bagfile and splitting '
                   'is disabled.'
         )
-        parser.add_argment(
+        parser.add_argument(
             '-d', '--max-bag-duration', type=int, default=0,
             help='maximum duration in nanosecnds before the bagfile will be split. '
                   'Default is zero, recording written in single bagfile and splitting '
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -64,7 +64,8 @@ class RecordVerb(VerbExtension):
             '-d', '--max-bag-duration', type=int, default=0,
             help='maximum duration in seconds before the bagfile will be split. '
                   'Default is zero, recording written in single bagfile and splitting '
-                  'is disabled.'
+                  'is disabled. If both splitting by size and duration are enabled, '
+                  'the bag will split at whichever threshold is reached first.'
         )
         parser.add_argument(
             '--max-cache-size', type=int, default=0,
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -62,7 +62,7 @@ class RecordVerb(VerbExtension):
         )
         parser.add_argument(
             '-d', '--max-bag-duration', type=int, default=0,
-            help='maximum duration in nanosecnds before the bagfile will be split. '
+            help='maximum duration in seconds before the bagfile will be split. '
                   'Default is zero, recording written in single bagfile and splitting '
                   'is disabled.'
         )
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -61,7 +61,7 @@ SequentialWriter::SequentialWriter(
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
   max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
-  max_bagfile_duration(std::chrono::nanoseconds(0)),
+  max_bagfile_duration(std::chrono::seconds(0)),
   topics_names_to_info_(),
   metadata_()
 {}
@@ -86,7 +86,7 @@ void SequentialWriter::open(
 {
   base_folder_ = storage_options.uri;
   max_bagfile_size_ = storage_options.max_bagfile_size;
-  max_bagfile_duration = std::chrono::nanoseconds(storage_options.max_bagfile_duration);
+  max_bagfile_duration = std::chrono::seconds(storage_options.max_bagfile_duration);
   max_cache_size_ = storage_options.max_cache_size;
 
   cache_.reserve(max_cache_size_);
@@ -239,7 +239,9 @@ bool SequentialWriter::should_split_bagfile() const
   } else if (max_bagfile_duration != std::chrono::seconds(
       rosbag2_storage::storage_interfaces::MAX_BAGFILE_DURATION_NO_SPLIT)) {
     // Splitting by time
-    return (std::chrono::system_clock::now() - metadata_.starting_time) > max_bagfile_duration;
+    auto max_duration_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+      max_bagfile_duration);
+    return (std::chrono::system_clock::now() - metadata_.starting_time) > max_duration_ns;
   } else {
     // Not splitting by time or size
     return false;
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -246,7 +246,8 @@ bool SequentialWriter::should_split_bagfile() const
 
   // Splitting by time
   if (max_bagfile_duration != std::chrono::seconds(
-      rosbag2_storage::storage_interfaces::MAX_BAGFILE_DURATION_NO_SPLIT)) {
+      rosbag2_storage::storage_interfaces::MAX_BAGFILE_DURATION_NO_SPLIT))
+  {
     auto max_duration_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
       max_bagfile_duration);
     should_split = should_split ||
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -86,6 +86,7 @@ void SequentialWriter::open(
 {
   base_folder_ = storage_options.uri;
   max_bagfile_size_ = storage_options.max_bagfile_size;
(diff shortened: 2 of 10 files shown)


emersonknapp:
You're almost there - you're just missing an addition to the parsing command at [line 125](https://github.com/ros2/rosbag2/pull/409/files#diff-fa7b2593ea899bd8b6c1fd104229ec63R125)

See the string for parsing:
```
"ssssss|bbKKKObO"
```

That is the types of the arguments that are expected, `s` for string, `b` for bool etc, see https://docs.python.org/3/c-api/arg.html - from the looks of it you need an additional `K` in the middle for the `unsigned long long` you've added.

jhdcs:
Thanks for the hint! I've added the missing K to the parse string, but I'm still getting an error saying "ros2: error: unrecognized arguments: --max-bag-duration". I think I'm probably still missing a spot that I need to add the command in...

emersonknapp:
Taking a look - in order to help understand more easily in general, can you paste the exact command you ran when mentioning output?

And, a note that this feature will need tests before merging.

emersonknapp:
When I build this and try it, I get this result -

```
-> /ros_ws $ ros2 bag record
  File "/ros_ws/install_x86_64/ros2cli/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.9.2', 'console_scripts', 'ros2')()
  File "/ros_ws/install_x86_64/ros2cli/lib/python3.8/site-packages/ros2cli/cli.py", line 39, in main
    add_subparsers_on_demand(
  File "/ros_ws/install_x86_64/ros2cli/lib/python3.8/site-packages/ros2cli/command/__init__.py", line 249, in add_subparsers_on_demand
    extension.add_arguments(
  File "/ros_ws/install_x86_64/ros2bag/lib/python3.8/site-packages/ros2bag/command/bag.py", line 26, in add_arguments
    add_subparsers_on_demand(
  File "/ros_ws/install_x86_64/ros2cli/lib/python3.8/site-packages/ros2cli/command/__init__.py", line 249, in add_subparsers_on_demand
    extension.add_arguments(
  File "/ros_ws/install_x86_64/ros2bag/lib/python3.8/site-packages/ros2bag/verb/record.py", line 63, in add_arguments
    parser.add_argment(
AttributeError: 'ArgumentParser' object has no attribute 'add_argment'
```

Based on your result, however, it sounds like either you didn't build up to `ros2bag` to get your new argument, or forgot to `source install/setup.bash` before running the command

ros2bag/ros2bag/verb/record.py
```diff
@@ -60,6 +60,12 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
                   'Default it is zero, recording written in single bagfile and splitting '
                   'is disabled.'
         )
+        parser.add_argment(
```
emersonknapp, line 63:
```suggestion
        parser.add_argument(
```

jhdcs: Wow, I can't believe I missed that... Thanks for the catch!

ros2bag/ros2bag/verb/record.py
```diff
@@ -60,6 +60,12 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            '-d', '--max-bag-duration', type=int, default=0,
+            help='maximum duration in nanosecnds before the bagfile will be split. '
```
emersonknapp, line 65:
```suggestion
            help='maximum duration in nanoseconds before the bagfile will be split. '
```

We probably want to specify what happens when both splitting arguments are provided - I'm guessing it's something like "will be split at the duration or the size, whichever comes first"

Also nanoseconds are a bit unwieldy since likely users are thinking in minutes for bagfile durations, so they'll have to type at least 10 digits. I think that seconds would be a better atomic unit for splitting.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -61,6 +61,7 @@ SequentialWriter::SequentialWriter(
   metadata_io_(std::move(metadata_io)),
   converter_(nullptr),
   max_bagfile_size_(rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT),
+  max_bagfile_duration(std::chrono::nanoseconds(0)),
```
emersonknapp, line 64: This should use a constant like the above `max_bagfile_size`

ros2bag/ros2bag/verb/record.py
jhdcs: I started in nanoseconds since that's what the internal times were set in. But if people are thinking in minutes, then I completely agree seconds would be a better unit of time. I'll set to changing that now.

jhdcs:
Made a few changes based on your suggestions. Also implemented a test, but I'm not positive it's actually running correctly. It's not generating an error (that I can see), but still seems to be exiting early. Still investigating.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -414,6 +414,72 @@ TEST_F(RecordFixture, record_end_to_end_with_splitting_splits_bagfile) {
...
+
+      // There is no guarantee that the bagfile split expected_split times
+      // due to possible io sync delays. Instead, assert that the bagfile split
+      // at least once
+      if (rcpputils::fs::path(bag_file_path).exists()) {
+        metadata.relative_file_paths.push_back(bag_file_path);
+      }
+    }
+
+    ASSERT_GE(metadata.relative_file_paths.size(), 1) << "Bagfile never split!";
```
emersonknapp, line 469: This is all copy-paste from another test, but those tests aren't running that reliably so I'm not sure we should be following in their footsteps.

Regardless, should this asserting about the metadata contents be done outside the `ifdef` block? Is the condition something that should be true no matter what? It seems like it is currently _only_ testing "does vector.push_back work?", unless I'm misreading it.

jhdcs: Didn't know those tests were unreliable. I'm a bit new to the testing scene, so is there a different test I should be following as a guideline?

jhdcs: And judging by the TODO comments, I think it's testing if the metadata got written in Windows. Not sure why Windows would prevent metadata from being written, though, but I suppose it's enough of a problem that they needed to write a case for it.

jhdcs:
Finally found the problem with the test, and I'm actually a little embarrassed it took so long for me to find. Turns out I was trying to divide seconds by milliseconds. Last I checked, Physics doesn't agree with that. Anyways, converted the split time to milliseconds, and it seems to be running correctly now.

jhdcs:
@emersonknapp Sorry for bothering you again, but now that Foxy has been released, could I get another review on this?

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/4b7201d9af882bbd754a32bf3d31b748/raw/e9836cd526e7df9cec412fe2d254f3f2e5d68463/ros2.repos
TEST args: --packages-select-regex rosbag2 ros2bag
Job: ci_launcher

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11243/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9203/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11155/)

emersonknapp:
You'll need to rebase on master, this branch is pretty out of date

jhdcs:
Rebase complete. We'll see what the checks say.

Karsten1987:
From the Windows CI output:
```
09:44:54 C:\ci\ws\src\jhdcs\rosbag2\rosbag2_cpp\src\rosbag2_cpp\writers\sequential_writer.cpp(255,42): error C2676: binary '-': 'std::chrono::system_clock::time_point' does not define this operator or a conversion to a type acceptable to the predefined operator [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
```

jhdcs:
I think what's going on is that the compiler is treating `std::chrono::high_resolution_clock` in `std::chrono::time_point<std::chrono::high_resolution_clock>` within the BagMetadata struct as a `system_clock` in Linux, and as a `steady_clock` in Mac & Windows. Interesting quirk. I've replaced its clock with a `std::chrono::system_clock` to make _sure_ it only gets treated the way I'm expecting. But the only way to tell is to try and build it on those OSes again.

Can I get another CI on this?

Karsten1987:
It doesn't compile locally on my OSX:

```
--- stderr: rosbag2_storage
/Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp:201:28: error: no viable overloaded '='
    metadata.starting_time = node["starting_time"]
    ~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/chrono:1326:28: note: candidate function (the implicit copy assignment operator) not viable: no known conversion from 'time_point<std::__1::chrono::steady_clock, duration<[...], ratio<[...], 1000000000>>>' to 'const time_point<std::__1::chrono::system_clock, duration<[...], ratio<[...], 1000000>>>' for 1st argument
class _LIBCPP_TEMPLATE_VIS time_point
                           ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/chrono:1326:28: note: candidate function (the implicit move assignment operator) not viable: no known conversion from 'time_point<std::__1::chrono::steady_clock, duration<[...], ratio<[...], 1000000000>>>' to 'time_point<std::__1::chrono::system_clock, duration<[...], ratio<[...], 1000000>>>' for 1st argument
In file included from /Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp:35:
In file included from /Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/yaml.h:17:
/Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/node/impl.h:216:14: error: implicit instantiation of undefined template 'YAML::convert<std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > >'
  AssignData(convert<T>::encode(rhs));
             ^
/Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/node/impl.h:201:3: note: in instantiation of function template specialization 'YAML::Node::Assign<std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > >' requested here
  Assign(rhs);
  ^
/Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp:184:27: note: in instantiation of function template specialization 'YAML::Node::operator=<std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > >' requested here
    node["starting_time"] = metadata.starting_time;
                          ^
/Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/node/node.h:142:8: note: template is declared here
struct convert;
       ^
In file included from /Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp:35:
In file included from /Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/yaml.h:17:
/Users/karsten/workspace/ros2/ros2_master/install/include/yaml-cpp/node/impl.h:216:24: error: incomplete definition of type 'YAML::convert<std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000> > > >'
  AssignData(convert<T>::encode(rhs));
             ~~~~~~~~~~^~
3 errors generated.
```

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11272/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6529/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9229/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11178/)

jhdcs:
I think this might have been that flaky test you were warning me about. I've applied Karsten's fix from elsewhere in the testing suite, hopefully that makes it more reliable.

Not sure about the test_qos_simple test, though. I don't remember having touched that test, and it seems to be trying to access a database that's in use by another test?

jhdcs:
I'm still not finding what could be causing the test_qos_simple failure. Near as I can guess, it might be failing because the other tests are failing. Maybe.

@emersonknapp can I get another CI on this when you get a chance? Hopefully the updates to the windows testing code should fix it.

Karsten1987:
A new round of CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11395/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6616/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11289/)

@emersonknapp can you take a look at this once more?

PR #396, opened 2020-04-30
Fix #381 unstable play_end_to_end test

mabelzhang:
I think this should fix #381. The failing test was `play_filters_by_topic` in `rosbag2_tests`.

I took the same kind of fix from `rosbag2_transport` https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/test/rosbag2_transport/test_play.cpp#L140 , so I think it should make the tests more stable.

I ran GitHub Actions on this branch in my fork (temporarily the default branch to run the scheduled Actions) https://github.com/mabelzhang/rosbag2/actions , and the ros-tooling/action-ros-ci passed most of the time. The occasional times it fails are at `ament_cmake_core`, which I think is a glitch in Actions. The next Actions check, `actions/upload-artifact`, always fails in my fork. I think that's not relevant to the test failures. The tests are passing.

For example, these are passing runs scheduled ~5 minutes apart
348 https://github.com/mabelzhang/rosbag2/actions/runs/91812093
349 https://github.com/mabelzhang/rosbag2/actions/runs/91814897
350 https://github.com/mabelzhang/rosbag2/actions/runs/91818893
351 https://github.com/mabelzhang/rosbag2/actions/runs/91821981
352 https://github.com/mabelzhang/rosbag2/actions/runs/91825552
354 https://github.com/mabelzhang/rosbag2/actions/runs/91839933
355 https://github.com/mabelzhang/rosbag2/actions/runs/91843979
356 https://github.com/mabelzhang/rosbag2/actions/runs/91847046

As @Karsten1987 mentioned, the problem is probably due to publishers and subscribers not starting at the same time. A real fix is to revive and finish the `start_paused` branch, to add a feature that starts bag playback paused.

PR #410, opened 2020-05-13
Disable play_filters_by_topic test

mabelzhang:
Disabling the test still failing on Linux after #396.

That seems to always be the one that times out. If there are also others, will disable them when we see them.


tfoote:
+1 for disabling the test, if it's known to fail. However, it should have a ticket open to track reenabling it and a todo in the code pointing to that ticket.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp
```diff
@@ -120,7 +120,7 @@ TEST_F(PlayEndToEndTestFixture, play_fails_gracefully_if_needed_coverter_plugin_
     error_output, HasSubstr("Requested converter for format 'wrong_format' does not exist"));
 }
 
-#ifndef _WIN32
+/*
```
tfoote, line 123:
```suggestion
/*
todo(tfoote) Disabling failing test. Ticketed to follow up at: https://github.com/ros2/rosbag2/issues/408
```

Signed-off-by: Tully Foote <tfoote@osrfoundation.org>

PR #411, opened 2020-05-15
Explicitly add DLL directories for Windows before importing

jacobperron:
New in Python 3.8, we should call os.add_dll_directory for directories
containing the DLLs we intend to import as well as their recursive
dependencies.

Connects to https://github.com/ros2/ci/pull/432

### DIFF
--- a/rosbag2_transport/rosbag2_transport/__init__.py
+++ b/rosbag2_transport/rosbag2_transport/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import importlib
+import os
 
 from rpyutils import add_dll_directories_from_env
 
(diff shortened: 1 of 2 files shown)


rosbag2_transport/rosbag2_transport/__init__.py
```diff
@@ -28,5 +33,4 @@ def _import(name):
                          '#import-failing-even-with-library-present-on-the-system')
         raise
 
-
```
dirk-thomas, line 31: This seems unrelated and the linter might complain.

jacobperron: Fixed 3447ca4

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10797/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6182/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8790/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/10691/)

PR #416, opened 2020-05-20
use consistent quotes in help messages

dirk-thomas:
Follow up of this discussion: https://github.com/ros2/ros2cli/pull/516/files#r428193690


dirk-thomas:
Unfortunately the custom CI of this repo seems to be broken. Since it isn't even passing for the target branch (and already for a while) I will go ahead and merge this trivial change.

PR #419, opened 2020-05-27
Rosbag2_py sequential_writer

klintan:
Building on top of @mabelzhang excellent efforts https://github.com/ros2/rosbag2/pull/308 to get the python API going for rosbags, I tried to get the sequential writer part going.  However I'm still a beginner c++ developer so there are both some stylistic and other issues. 

Most pressing though which is why I wanted to create this draft PR (even though I know that is usually not how things are done :) ) I get a segmentation fault from the `rosbag2_storage_default_plugins` and `std::shared_ptr<SqliteStatementWrapper>
SqliteStatementWrapper::bind(std::shared_ptr<rcutils_uint8_array_t> value)` method, I'm not sure what happens, but the pointer address is not the same as it is in the `write` method in the rosbag2_py.cpp wrapper file. Not sure if it goes out of scope and gets discarded or if there is something else I'm messing up.  (I tried making the shared ptr a private class variable and assign the incoming ptr to that, but to no avail)

Would love some feedback on all of these things and perhaps someone sees the issue and what is missing (probably something fairly obvious for experienced c++ devs).

Regardless of what happens, could perhaps serve as the starting point of someone elses implementation. I need the writer to create the `kitti2rosbag2` :) 

edit: also reason why all changes from PR 308 is included is because those would probably be merged to master long before this one regardless, so hopefully an easier merge if they are similar.


rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,210 @@
...
+  /// Return a mapping from topic name to topic type.
+  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() {
+    return reader_->get_all_topics_and_types();
+  }
+
+private:
+  std::unique_ptr<rosbag2_cpp::Reader> reader_;
+};
+
+class SerializedBagMessage : public rosbag2_storage::SerializedBagMessage {
```
mabelzhang, line 84: Does the `SerializedBagMessage` struct need a wrapper class, or can it simply be wrapped with `def_readwrite()` like `TopicMetadata`? Then you would be able to use the `rosbag2_storage::SerializedBagMessage` directly. I don't know if that would solve the seg fault, but it wouldn't require the `memcpy` here which could lead to something like different pointer addresses. See the other comment for how I think it can be done.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+  /// Return a tuple containing the topic name, the serialized ROS message, and
+  /// the timestamp.
+  pybind11::tuple read_next() {
+    const auto next = reader_->read_next();
+    rcutils_uint8_array_t rcutils_data = *next->serialized_data.get();
+    std::string serialized_data(
+        rcutils_data.buffer, rcutils_data.buffer + rcutils_data.buffer_length);
+    return pybind11::make_tuple(
+        next->topic_name, pybind11::bytes(serialized_data), next->time_stamp);
```
mabelzhang, line 72: It would be nice for `write()` to take a tuple of 3 arguments, parallel to what `read_next()` returns. Then the Python script wouldn't need access to a `SerializedBagMessage` type. With the current writer test script, the call would look like `writer.write('chatter', serialize_message(msg), time_stamp)`. All the work with `SerializedBagMessage` can then be done within `_rosbag2_py.cpp`, and then hopefully you wouldn't need to wrap `SerializedBagMessage`.

klintan: Great suggestion, makes much more sense, I'll try it out :)  Thanks for helping out!

klintan:
> The reason #308 isn't merged is actually because we wanted to add a writer to make the API complete before merging the reader. So thanks for making this PR!
> 
> A few comments regarding `SerializedBagMessage` that could make the code cleaner and maybe address the seg fault.

Ah ok, let me know if you want me to to this PR on your fork so that it will be included in that PR.
Another reason for keeping this in a separate one was actually that i thought that one might have finished the review, so didn't want you do need to do another double review on code that was already finished.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
klintan: Yay it worked :) Thanks! 

Let me know what you think is best for this PR: as described below, create a PR to your fork or just finish up this one as a separate one for a proper review now that it's working?

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  /// Write a serialized message to a bag file
+  void write(pybind11::tuple & message) {
+    auto bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
+
+    std::string data = std::string(pybind11::str(*message[1]));
+
+    bag_message->topic_name = std::string(pybind11::str(*message[0]));
+    bag_message->serialized_data = rosbag2_storage::make_serialized_message(&data, data.length());
+    bag_message->time_stamp = pybind11::cast<long>(*message[2]);
```
mabelzhang, line 126: Is it possible to cast to `rcutils_time_point_value_t`, which `time_stamp` is defined to be? In case the typedef of that custom type ever changes.

klintan:
> I think it's fine to leave this PR separate as it is. You're right that the other PR is mostly reviewed. We'll see how they want to merge these two.
> 
> > * python tests, what should be asserted when the DB writes are done? Read it again and make sure we have the same message?
> 
> Yeah I think that makes sense.
> 
> > * Style
> > * write method argument style and that function is not very neat.
> 
> I did another pass. Hope I caught the obvious things.

Thanks, sounds good. I'll add "sensor_msgs/msg/Image" message type write as a test as well, because as I was trying it out on https://github.com/klintan/kitti2bag2 all message types were written without issues, except for images, got a seg fault on those. 

I'll take a look at all the feedback tomorrow or this weekend.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  void remove_topic(const rosbag2_storage::TopicMetadata &topic_with_type) {
+    writer_->remove_topic(topic_with_type);
+  }
+
+  /// Write a serialized message to a bag file
+  void write(pybind11::tuple & message) {
+    auto bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
+
+    std::string data = std::string(pybind11::str(*message[1]));
```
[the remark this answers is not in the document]
klintan: Added the throw runtime_error. Could you elaborate on `define the indices 0 1 2 as const private class variables that describe what they are.`

Perhaps it would be better to define the 3 input arguments, either as kwargs or just named so it can be used similar to the other methods `reader.open(storage_options, converter_options)` returning the tuple makes sense since python3 has the automatic tuple unpacking on returns. That way we could just call it like so: write(topic_name, serialize(data), timestamp) instead of the wrapped tuple. That way I can define the types in the input in the arguments (instead of the indices)

rosbag2_py/resources/talker/metadata.yaml
```diff
@@ -0,0 +1,31 @@
+  rosbag2_bagfile_information:
```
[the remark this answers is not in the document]
klintan: Ah yes, I think this was a copy pasta error, I think it should be fixed :)

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
klintan: fixed.

rosbag2_py/resources/talker/metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+          offered_qos_profiles: "- history: 3\n  depth: 0\n  reliability: 1\n  durability: 1\n  deadline:\n    sec: 2147483647\n    nsec: 4294967295\n  lifespan:\n    sec: 10\n    nsec: 0\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 2147483647\n    nsec: 4294967295\n  avoid_ros_namespace_conventions: false"
+        message_count: 10
+      - topic_metadata:
+          name: /parameter_events
+          type: rcl_interfaces/msg/ParameterEvent
+          serialization_format: cdr
+          offered_qos_profiles: "- history: 3\n  depth: 0\n  reliability: 1\n  durability: 2\n  deadline:\n    sec: 2147483647\n    nsec: 4294967295\n  lifespan:\n    sec: 2147483647\n    nsec: 4294967295\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 2147483647\n    nsec: 4294967295\n  avoid_ros_namespace_conventions: false"
+        message_count: 0
+    compression_format: ""
+    compression_mode: ""
```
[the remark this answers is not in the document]
klintan: Added

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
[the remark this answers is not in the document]
mabelzhang: Oh yeah! I think that makes more sense. That would be clearer too.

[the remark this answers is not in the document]
klintan: Ok changed this and added working tests for the writer where it reads the data afterwards and checks that it's ok!  let me know what you think and/or if you have any additional feedback or thoughts.

rosbag2_py/test/common.py
```diff
@@ -0,0 +1,13 @@
+import rosbag2_py._rosbag2_py as rosbag2_py
```
mabelzhang, line 1: This file will need a copyright block on top.

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
```diff
@@ -0,0 +1,177 @@
...
+  void create_topic(const rosbag2_storage::TopicMetadata &topic_with_type) {
+    writer_->create_topic(topic_with_type);
+  }
+
+  void remove_topic(const rosbag2_storage::TopicMetadata &topic_with_type) {
+    writer_->remove_topic(topic_with_type);
+  }
+
+  /// Write a serialized message to a bag file
+  void write(pybind11::str &topic_name, pybind11::bytes &message, pybind11::int_ &time_stamp) {
```
mabelzhang, line 118: Do all the args need to be `pybind::` types? I would think you could use plain C++ types for them like the other methods do, e.g. `const std::string &topic_name`. This is a question. I don't know the answer.

klintan: Ok you are right! seems you can pass the c++ types at least for certain types https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html and `bytes` immediately get `utf-8` encoded into `std::string` . 

Changed it.

rosbag2_py/test/common.py
klintan: Fixed!

rosbag2_py/src/rosbag2_py/_rosbag2_py.cpp
mabelzhang: Cool thanks! I have no more comments.

emersonknapp:
Is the goal to direct this PR to merge into the mabelzhang:rosbag2_py branch? If so, can we set this up so that it is just that diff?

klintan:
> Is the goal to direct this PR to merge into the mabelzhang:rosbag2_py branch? If so, can we set this up so that it is just that diff?

If that is the preferred way I'll create a PR to @mabelzhang fork.

emersonknapp:
Ah, yes I see now that the other branch is on the fork. Makes sense to want to keep the PR here, but that makes it slightly difficult to review the increment. In that case - I'm not sure exactly how to handle this - probably still a PR on the fork makes sense and then we can have the final pass review here all in one?

klintan:
> Ah, yes I see now that the other branch is on the fork. Makes sense to want to keep the PR here, but that makes it slightly difficult to review the increment. In that case - I'm not sure exactly how to handle this - probably still a PR on the fork makes sense and then we can have the final pass review here all in one?

Hm ok, so now actually all my changes would go into @mabelzhang fork and PR and the incremental changes would be done in that PR. Hopefully it's not to bad, since the new changes only pertains to the writer pretty much (some very small changes in the reader tests to consolidate some code)

Hopefully @mabelzhang will have time to merge that soon.

mabelzhang:
I merged this into my fork https://github.com/mabelzhang/rosbag2/pull/1
The diffs are in https://github.com/ros2/rosbag2/pull/308/commits/563fe4d0c4c0411d789a4900e517e9d8a0c3486f in #308

mabelzhang:
Closing this one as its contents are now in #308

PR #421, opened 2020-05-28
Use latest versions from ament_lint which fix flake8

emersonknapp:
We're currently using the default, which is eloquent, which has an 0.8.x release of ament_lint, whereas the newest 0.9.4 has the flake8 fix that is making all PRs fail linting.

Fixes ... #x - we didn't have an issue tracking this. Let's get these builds to green!

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
     name: ament_copyright
     runs-on: ubuntu-latest
     container:
-      image: rostooling/setup-ros-docker:ubuntu-bionic-ros-eloquent-ros-base-latest
+      image: rostooling/setup-ros-docker:ubuntu-focal-ros-foxy-ros-base-latest
     steps:
     # TODO(setup-ros-docker#7): calling chown is necessary for now
     - run: sudo chown -R rosbuild:rosbuild "$HOME" .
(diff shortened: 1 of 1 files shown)


emersonknapp:
Ah - this isn't going to work until the Foxy packages are on packages.ros.org/ros2 - for now they are there, but only available in packages.ros.org/ros2-testing. Related to https://github.com/ros-tooling/setup-ros-docker/pull/6

dirk-thomas:
I would recommend that any kind of CI uses the `testing` repo. That is also what the Jenkins-based ROS buildfarms are doing.

Karsten1987:
@emersonknapp is there any update on this? If not, I would propose to disable the GH actions until this is resolved.

emersonknapp:
@Karsten1987 you're right that we're better off disabling checks that currently provide no benefit. We're going through some process change right now that has left us in this bad state for too long (clearly unacceptable), and we'll be putting in place a system that doesn't allow for it in the future. Until that happens, I've opened https://github.com/ros2/rosbag2/pull/440

Sorry for letting this stay broken so long. With these disabled, we'll be in no worse a state than we were before they existed, and when we do re-enable them we plan to have some explicit guarantees around fixing them much more promptly.

emersonknapp:
As for the state of this PR - it does now fix the flake8 check but I don't know what's going on with ament_cppcheck and don't have the bandwidth to figure it out right now. I'd like to leave this draft PR open though to continue working towards the fixed state.

Karsten1987:
sounds good. Thanks for the follow ups. 
Let me know if I can help in any way.

Karsten1987:
@emersonknapp can this branch be deleted?

Issue #439, opened 2020-06-12
Flake8 Action CI broken


emersonknapp:
## Description
The Flake8 linter tests in our Action CI for this package is broken. This happened due to using a version from Eloquent and installing Flake8 via pip, which got us flake8 version 3.8. The release (as of the time of this writing) in Eloquent for ament_lint is not flake8 3.8 compliant.

A fix has already been backported into eloquent, but we should now be running against the latest Foxy builds anyways.

### Expected Behavior
* Flake8 passes on Pull Requests
* We are using ament_lint from ros2-testing for Foxy to run the linter test

### Actual Behavior
Always red

## To Reproduce
Open a Pull Request

## System (please complete the following information)
GitHub

Karsten1987:
@emersonknapp I am closing this with #440 being merged. Feel free to reopen.

emersonknapp:
It was disabled by #440, fixed upstream, then re-enabled by #467 - just for the context to all be here :)

PR #440, opened 2020-06-12
Disable broken tests for pull requests

emersonknapp:
Until the following two issues are resolved, disabling consistently red Action CI checks
* https://github.com/ros2/rosbag2/issues/381
* https://github.com/ros2/rosbag2/issues/439

These checks are currently actively detrimental to contributions because they require special knowledge that they're broken, and admin privileges to override for merging.

Once the above issues are fixed, we can re-enable these checks.

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -87,7 +87,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [flake8, pep257]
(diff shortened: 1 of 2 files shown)


Karsten1987:
After merging, do you mind providing a backport of this PR to the foxy branch?

PR #443, opened 2020-06-17
Playback pause feature

mabelzhang:
### Usage
```
ros2 bag play <bagfile> --paused
```
Press space bar during playback to pause/unpause.

Sample output with numeric printout for readability:
```
$ ros2 bag play rosbag2_2020_04_02-18_23_54/ --paused
[INFO] [1592375522.558032913] [rosbag2_storage]: Opened database 'rosbag2_2020_04_02-18_23_54//rosbag2_2020_04_02-18_23_54_0.db3' for READ_ONLY.
[INFO] [1592375522.664716232] [rosbag2_transport]: Paused
 [INFO] [1592375523.266395695] [rosbag2_transport]: Resumed
publishing message 1
publishing message 2
publishing message 3
publishing message 4
publishing message 5
publishing message 6
 [INFO] [1592375524.766716700] [rosbag2_transport]: Paused
 [INFO] [1592375525.879586586] [rosbag2_transport]: Resumed
publishing message 7
publishing message 8
publishing message 9
publishing message 10
publishing message 11
publishing message 12
 [INFO] [1592375527.379393673] [rosbag2_transport]: Paused
 [INFO] [1592375528.161453555] [rosbag2_transport]: Resumed
publishing message 13
publishing message 14
publishing message 15
publishing message 16
publishing message 17
publishing message 18
 [INFO] [1592375529.661435598] [rosbag2_transport]: Paused
 [INFO] [1592375530.413582545] [rosbag2_transport]: Resumed
publishing message 19
publishing message 20
```

### Implementation
Implementation is by changing `Rosbag2Node` to a `LifecycleNode`.
Tests run one thread for playing and another for activating/deactivating the lifecycle node.
Keyboard handling is in a separate thread from playback in `play.cpp`.

### Feedback needed
- Keyboard interaction and terminal modification are copied directly from ROS 1 bag [player.cpp](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag/src/player.cpp#L731). I don't have a Windows setup to test this, and the keyboard interaction has to be manually tested, right? Could someone help me test this in Windows, or is there a way to programmatically test the interaction? I looked in ROS 1 [test_rosbag](https://github.com/ros/ros_comm/tree/noetic-devel/test/test_rosbag/test) and did not find anything like that.
- Currently the sleep time to wait for key press is 30 ms. While the period of pause is taken into account, I don't know how rigorous the timestamp of the original bag file is followed, given that there is delay in activate/deactivate, pause/resume sleep time. The messages in `test_play_timing` are timed 1 second apart currently. I could try to add a more closely timed test if needed. Or I can decrease the sleep time that waits for a lifecycle state change.

I tried to use the pause feature to fix the `recorded_messages_are_played_for_filtered_topics` test, which publishes 3 messages and expects 2. We were hoping if we pause it in the beginning, we can expect 3. Unfortunately, it was not fixed. Locally, the Cyclone test passes always, but the FastRTPS test fails about 1 out of 4, even if I wait for 2 seconds for the subscriber to come up.

### Future work
- Currently only "Paused" and "Resumed" are printed. In ROS 1, the timestamp is printed at all times during playback, and the timestamp freezes when paused. Once the timestamp is implemented for rosbag2, it would be useful to mirror ROS 1 behavior.


mabelzhang:
Had to fix DCO 13 commits back X_X

Karsten1987:
I think this could be rebased on top of master to limit the amount of commits.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -102,6 +109,21 @@ void Player::play(const PlayOptions & options)
...
   wait_for_filled_queue(options);
 
+  rosbag2_transport_->configure();
+  // If playback should not be paused in the beginning, activate lifecycle node right away.
+  // If playback should be paused, move on.
+  // Playback will detect when node is activated, either from here or externally.
+  if (!options.paused) {
+    rosbag2_transport_->activate();
+  } else {
+    paused_ = true;
```
Karsten1987, line 119: do we really need that `paused_` variable for bookkeeping? Can't we rely on the state of the lifecycle node?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -102,6 +109,21 @@ void Player::play(const PlayOptions & options)
 
   wait_for_filled_queue(options);
 
+  rosbag2_transport_->configure();
+  // If playback should not be paused in the beginning, activate lifecycle node right away.
+  // If playback should be paused, move on.
+  // Playback will detect when node is activated, either from here or externally.
+  if (!options.paused) {
+    rosbag2_transport_->activate();
```
Karsten1987, line 117: should this set `paused_` to `false`?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -181,7 +207,29 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
   while (message_queue_.try_dequeue(message) && rclcpp::ok()) {
     std::this_thread::sleep_until(
       start_time_ + std::chrono::duration_cast<std::chrono::nanoseconds>(
-        1.0 / rate * message.time_since_start));
+        1.0 / rate * message.time_since_start + paused_duration_));
```
Karsten1987, line 210: why do we sleep longer here?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -204,4 +252,110 @@ void Player::prepare_publishers(const PlayOptions & options)
...
+        } else {
+          rosbag2_transport_->deactivate();
+        }
+        paused_ = !paused_;
+        break;
+    }
+  }
+}
+
+void Player::setup_terminal()
```
Karsten1987, line 275: Could that remaining part here be extracted in a separate `KeyboardHandler` or some other entity? I don't really see them necessarily being part of the `Player`.

rosbag2_transport/src/rosbag2_transport/player.hpp
```diff
@@ -22,6 +22,14 @@
...
 #include <unordered_map>
 
+// Terminal
+#if !defined(_MSC_VER)
+  #include <termios.h>
+  #include <unistd.h>
+#else
+  #include <windows.h>
+#endif
+
```
Karsten1987, line 32: as per previous comment, I think it makes sense to hide these `#if/#else` behind a new header/class for handling all specifics with the keyboard.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: Agreed! I've seen a request for Action/Service control of rosbag playback, so if we abstract the controls into a common API that should be a lot easier to do as well.

rosbag2_transport/src/rosbag2_transport/player.cpp
mabelzhang: The `paused_duration_` accounts for the wall duration of the pause. Otherwise, if you have a 10-second bag, and you `ros2 bag play --paused`, wait for 10 seconds, then resume, then the bag will immediately play all the messages at once when you resume.

rosbag2_transport/src/rosbag2_transport/player.cpp
mabelzhang: Oh yeah you're right. I messed up somewhere. 9c92972

rosbag2_transport/src/rosbag2_transport/player.cpp
mabelzhang: Removed `paused_` in 9c92972

rosbag2_transport/src/rosbag2_transport/player.hpp
mabelzhang: fd97872

rosbag2_transport/src/rosbag2_transport/player.cpp
mabelzhang: fd97872
Currently it is refactored to a separate `KeyboardHandler` class.
The Action/Service control would probably use interfaces associated with the LifecycleNode that Rosbag2Transport is now, so it will still want access to the `rosbag2_transport_` object in the player.
This may or may not be related. For some reason, I never see the Rosbag2Transport node listed in `ros2 node list` even before, nor now with `ros2 lifecycle nodes`. I don't know if that would affect the services. I was looking for the node to see if I can trigger a service in the tests, but not finding the node listed, I triggered it internally through new methods in Rosbag2Transport instead.

mabelzhang:
Re rebasing on master: I figured it'd get squashed into a single commit when it's merged anyway, so I didn't bother messing with rebasing.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -181,7 +200,29 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
   while (message_queue_.try_dequeue(message) && rclcpp::ok()) {
     std::this_thread::sleep_until(
       start_time_ + std::chrono::duration_cast<std::chrono::nanoseconds>(
-        1.0 / rate * message.time_since_start));
+        1.0 / rate * message.time_since_start + paused_duration_));
+
+    if (rosbag2_transport_->get_current_state().id() !=
+      lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE)
```
Karsten1987, line 206: what happened if in state `unconfigured` or `finalized`?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -181,7 +200,29 @@ void Player::play_messages_until_queue_empty(const PlayOptions & options)
...
+    {
+      ROSBAG2_TRANSPORT_LOG_INFO("Paused");
+      std::chrono::time_point<std::chrono::system_clock> pause_start =
+        std::chrono::system_clock::now();
+
+      // Sleep until activated externally
+      while (rosbag2_transport_->get_current_state().id() !=
+        lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE && rclcpp::ok())
+      {
+        std::this_thread::sleep_for(std::chrono::milliseconds(30));
```
Karsten1987, line 216: can we expose this as a constant?

Karsten1987:
@mabelzhang which OSX SDK version are you compiling with?

I ran into the following compiler error:
```
Starting >>> rosbag2_transport
--- stderr: rosbag2_transport
/Users/karsten/workspace/ros2/ros2_master/src/ros2/rosbag2/rosbag2_transport/src/rosbag2_transport/keyboard_handler.cpp:85:3: error: use of undeclared identifier 'bcopy'
  FD_COPY(&stdin_fdset_, &testfd);
  ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/sys/_types/_fd_copy.h:29:25: note: expanded from macro 'FD_COPY'
#define FD_COPY(f, t)   __DARWIN_FD_COPY(f, t)
                        ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/sys/_types/_fd_def.h:75:33: note: expanded from macro '__DARWIN_FD_COPY'
#define __DARWIN_FD_COPY(f, t)  bcopy(f, t, sizeof(*(f)))
                                ^
1 error generated.
make[2]: *** [CMakeFiles/rosbag2_transport.dir/src/rosbag2_transport/keyboard_handler.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/rosbag2_transport.dir/all] Error 2
make: *** [all] Error 2
---
Failed   <<< rosbag2_transport	[ Exited with code 2 ]
```

Which could be fixed by including `strings.h` as noted in here: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/bcopy.3.html

Seen this before?

rosbag2_transport/src/rosbag2_transport/keyboard_handler.cpp
```diff
@@ -0,0 +1,118 @@
...
+#else
+  const int fd = fileno(stdin);
+  tcsetattr(fd, TCSANOW, &orig_flags_);
+#endif
+  terminal_modified_ = false;
+}
+
+int KeyboardHandler::read_char_from_stdin()
+{
+#ifdef __APPLE__
```
Karsten1987, line 83: why the different style between `#ifdef` and `#if defined`?

rosbag2_transport/include/rosbag2_transport/rosbag2_transport.hpp
```diff
@@ -84,6 +84,18 @@ class Rosbag2Transport
   ROSBAG2_TRANSPORT_PUBLIC
   void print_bag_info(const std::string & uri, const std::string & storage_id);
 
+  /**
+   * Activate underlying lifecycle node.
+   */
+  ROSBAG2_TRANSPORT_PUBLIC
+  void activate_lifecycle();
```
Karsten1987, line 91: I think it makes more sense to call it `resume` and `pause` at this level.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -184,7 +186,9 @@ Recorder::create_subscription(
 
 void Recorder::record_messages() const
 {
-  spin(node_);
+  rclcpp::executors::MultiThreadedExecutor exe;
+  exe.add_node(node_->get_node_base_interface());
+  exe.spin();
```
Karsten1987, line 191: So while this allows me to use the traditional `lifecycle` command line argument, this will not work during playback time. Playback as well has to have a `spin` thread where the lifecycle services are executed.

emersonknapp:
I'm a little worried about the architecture here - this may be more appropriate as smaller PRs that are easier to reason about, since it encompasses a few different features.

First, it seems like the commandline argument to start paused is a follow-up feature to the ability to pause and resume in the first place.

Next, I'm concerned that we shouldn't be building keyboard handling, with its platform switches, into the rosbag2_transport layer. This seems like it may be appropriate for the `ros2bag` CLI layer instead, and any input handling code that requires platform-specific logic should be pulled out into a separate utility so that logic is not tied up in rosbag2 business logic. The keyboard handling feels like its own standalone PR, since it introduces a new feature set entrypoint that may become much larger as it opens the doors to other runtime controls.

Karsten1987:
I'd second @emersonknapp concerns. Having the changes for starting/stopping as well as the lifecycle node in a separate PR from the keyboard handler makes a lot of sense to me. Also, having the keyboard handler live in ros2cli sounds very reasonable.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
mabelzhang: Are you saying we should add a spin thread to `player.cpp`, maybe inside `play()`?
I actually added the spin above for `recorder.cpp` because the test wouldn't pass without it. Nothing happens.

On a related note, as I mentioned [here](https://github.com/ros2/rosbag2/pull/443#discussion_r445903866), I actually didn't get anything with `ros2 lifecycle nodes` on the command line, same for `ros2 bag record`. Is there something else needed for that to work?

mabelzhang:
Okay, so sounds like we want to break this into 2 PRs (as opposed to 3), one for lifecycle pause/resume, and another for keyboard handling.

I did some refactoring, and I realized it doesn't make too much sense to move the command line `--paused` to a 3rd PR, because the lifecycle node is activated initially using that flag in `PlayOptions`. The tests also uses that flag to create a few different cases to test pausing and resuming.

Currently, other than directly accessing the `Rosbag2Node` lifecycle node in code, there is no way to pause/resume, since the nodes don't show up in `ros2 lifecycle`. So if we decouple the `--paused` command line flag from this PR, then the feature is not really useful to anyone. With the command line flag, at least the user can pause at the beginning (which was the reason for this PR), they just cannot resume during playback. The 2nd PR with keyboard handling will add the capability of pausing and resuming during playback.

I have removed keyboard handling from this PR. If people are okay with just 2 PRs, I will create the 2nd PR with the keyboard handling.

danthony06:
This would be a really nice feature to have back, is there anything I can do to help move the progress along?

mabelzhang:
Actually there is! Thank you for asking. I'm getting pulled to a different project with a tight timeline for the next 2-3 weeks at least, and most of my rosbag time has been spent on trying to fix a Windows debug mode failure (don't ask me how that's going). We would really like to get the LifeCycle node feature in, so any help on that end is appreciated!

Specifically, I have not had time to address Karsten's PR comments above. I think the most time-consuming is to learn about which states of a lifecycle node can transfer to which other state? Then some logic checking the states can be put in to make sure this comment https://github.com/ros2/rosbag2/pull/443#discussion_r451034011 is addressed.

Other than that, my other lingering concern is that the lifecycle nodes don't show up in `ros2 lifecycle nodes` on the command line, which no one has replied yet. It would be good to look into why that is the case, and whether it is needed for lifecycle functions to be called. It could be that it's not needed? Then all is well.

Then I think all the comments above would be addressed more or less. We are planning on just getting the LifeCycle stuff in first, and move all the keyboard-related things to a separate PR later. That is a second area that we can use help on. Let me know if anything is unclear.

This PR currently comes from my fork, which is inconvenient for other contributors. As the PR already has some comments, hmm you can either open a PR to my fork (not ideal), or maybe you can base your stuff on my fork and when you're ready, we can close this PR? Anyway, we can figure that out.

danthony06:
@mabelzhang Okay, I have some available time, and I'll start working on it and work through the issues you raised.

mabelzhang:
@danthony06 Awesome thanks!! Let us know if you run into blockers.

rosbag2_transport/src/rosbag2_transport/player.cpp
danthony06: With respect to your first comment, the ROS 1 bag implementation had a time_translator class that took care of manipulating the time and keeping track of time pauses and associated math. Do you think this is a better approach?

mabelzhang: I think the question is referring to https://github.com/ros2/rosbag2/pull/443#discussion_r451010655
@Karsten1987 Is the time_translator way raised above preferred for accounting for the pause duration?

Karsten1987:
@danthony06 please apologize, your comment flew under my radar. Also, I feel Github renders the chronological order of inline comments weird, so let me just answer here:

>With respect to your first comment, the ROS 1 bag implementation had a time_translator class that took care of manipulating the time and keeping track of time pauses and associated math. Do you think this is a better approach?

I think it makes sense to abstract the timing behavior out of the `player` class, which handles things like the duration we've been paused or the ratio in which messages are being replayed. I guess we can also leverage this later on when replaying with simulated time.
I am generally just concerned whether "sleeping" is the right way of seeking for the next message to be ready, but I guess we can roll this way as for now. How do you envision the `time_translator` class API to look like? Returning a delta for each loop until when we sleep?

rosbag2_transport/src/rosbag2_transport/player.cpp
danthony06: > @danthony06 please apologize, your comment flew under my radar. Also, I feel Github renders the chronological order of inline comments weird, so let me just answer here:
> 
> > With respect to your first comment, the ROS 1 bag implementation had a time_translator class that took care of manipulating the time and keeping track of time pauses and associated math. Do you think this is a better approach?
> 
> I think it makes sense to abstract the timing behavior out of the `player` class, which handles things like the duration we've been paused or the ratio in which messages are being replayed. I guess we can also leverage this later on when replaying with simulated time.
> I am generally just concerned whether "sleeping" is the right way of seeking for the next message to be ready, but I guess we can roll this way as for now. How do you envision the `time_translator` class API to look like? Returning a delta for each loop until when we sleep?

Sorry, I'm even tardier in responding to this. I'm working through it, but I was thinking about something like this:
https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag/src/time_translator.cpp
Which would allow us to pass in the current time, and get back the delta for sleeping.

Kaju-Bubanja:
What the status of this PR? Is it working on Ubuntu?

danthony06:
I think I have a fix and I'm testing it now.

Kaju-Bubanja:
Sorry to bother you again, did you get around to test the fix you have? Can I assist with something? If yes, could you brief me shortly what tasks still need to be done?

danthony06:
I'm planning on spending the rest of the week on it.

Kaju-Bubanja:
I realized that I forgot to compile rosbag2_transport. Now I can pause the bag at the beginning, but as mentioned above there is no way to resume or pause the bag during playback. I checked out the PRs/Issues of ros2cli but didn't find anything related to this PR. As I read the comments from @mabelzhang the code to pause was once here but got factored out into ros2cli? @danthony06 are you working on the keyboard handling? If yes in what repo/PR/issue? Also is there the ability to skip 1 frame ahead with a key? This is kinda a blocker for me now and I will be working on it in the coming weeks, so any guidance would be appreciated

danthony06:
I've got a test branch set up on my fork of the repo. I'd appreciate help
testing it out.

Karsten1987:
@danthony06 Can you open a PR against @mabelzhang 's PR here? That way we have all the modifications and conversations in one PR rather than distributed across multiple contributors.

mabelzhang:
This PR is from my fork (didn't have access to this repo then). Do you want me to move it to this repo?

danthony06:
I can open a PR. I'd really appreciate help testing it as well.

emersonknapp:
> This PR is from my fork (didn't have access to this repo then). Do you want me to move it to this repo?

I think that would make sense so that others could contribute to it - you have access to push a branch here?

mabelzhang:
Looks like I do. Branch pushed to this repo: https://github.com/ros2/rosbag2/tree/start_paused_lifecycle

danthony06:
Fixed my PR to the main repo here: https://github.com/ros2/rosbag2/pull/666

emersonknapp:
Closing in favor of #666

PR #426, opened 2020-06-02
use interface_include_directories

Karsten1987:
fixes #422

PR #444, opened 2020-06-17
[Foxy Backport]  use interface_include_directories #426

Karsten1987:
backport of https://github.com/ros2/rosbag2/pull/426


emersonknapp:
Is the "Fpr__rosbag2__ubuntu_focal_amd64" job sufficient for our CI needs on these backports, or should we also be doing full CI runs against the other supported platforms?

emersonknapp:
Also, we should possible add branch protection to the distro branches to require at least 1 approving review before merge. What do you think? I don't have admin access to this repo so I can't just go do it.

Karsten1987:
I think it's safe, given that all commits on master got backported already. With this backport, master is equivalent to foxy.

Karsten1987:
@emersonknapp fyi, I enabled branch protection on foxy

Issue #449, opened 2020-06-29
How to write data to rosbag2 with API


henry0572:
## Description
It is not possible to write data directly to rosbag2 using ros2 foxy.
I get the same error as below.

[ROS2]How to write data to rosbag2 in ros2 foxy
https://answers.ros.org/question/355764/ros2how-to-write-data-to-rosbag2-in-ros2-foxy/

I was able to load the data based on the following.

How to read data from rosbag2 in ros2
https://stackoverflow.com/questions/60282296/how-to-read-data-from-rosbag2-in-ros2

### Expected Behavior
Writing succeeds.

### Actual Behavior

`writer.write(message)` is what cause all the error messages:
```
terminate called after throwing an instance of 'std::logic_error' what(): basic_string::_M_construct null not valid
```

## System
 - OS: Ubunutu 20.04
 - ROS 2 Distro: Foxy
 - Version: release

emersonknapp:
@henry0572 thanks for bringing this up. The C++ API is not straightforward. You can see in @Karsten1987 's PR ^ linked above this comment, a project to reduce the complexity of using the C++ API directly. It also acts as a good example of how to use it as it is currently.

Karsten1987:
The following test should illustrate how to use the C++ API as of now. There will be further modifications done which allows users to directly write non-serialized messages in the future. 
https://github.com/ros2/rosbag2/blob/master/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_api.cpp

I'll go ahead and close this issue as it's possible to write data to rosbag2 via the C++ API.

PR #451, opened 2020-07-02
minimal c++ API test

Karsten1987:
That's a first step into improving the general C++ API for writing and reading rosbags.

The set of API calls in the test are the minimal set I could come up with in order to write a MessageT into the bag file. IMHO, there's quite some room for making this more user friendly. 

* Making the `SequentialReader` and `SequentialWriter` the default implementation. That would allow for `rosbag2_cpp::Writer writer;` to simply work.
* Default the converter options. As seen in [here](https://github.com/ros2/rosbag2/issues/449) and [here](https://answers.ros.org/question/355764/ros2how-to-write-data-to-rosbag2-in-ros2-foxy/) the current converter API is rather confusing than helpful. A call to `open` with only specifying the `StorageOptions` should suffice and the converter functionality should be bypassed as such.
* Every topic has to be created beforehand. However, if you forget that you'll get a uncaught exception such as `unknown file: Failure C++ exception with description "unordered_map::at: key not found" thrown in the test body.` I've addressed that in this PR.
* Write bag messages which are allocated on the stack: There's currently no way of writing serialized data which was allocated on the stack, but has to be passed in via a shared pointer. That results in specifying a custom deleter which avoids to deallocate the data (`malloc: *** error for object 0x7ffeefbfdc88: pointer being freed was not allocated`).

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -206,8 +206,8 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
     ++topics_names_to_info_.at(message->topic_name).message_count;
   } catch (const std::out_of_range & oor) {
     std::stringstream errmsg;
-    errmsg << "Failed to write on topic '" << message->topic_name
-      << "'. Call create_topic() before first write.";
+    errmsg << "Failed to write on topic '" << message->topic_name <<
+      "'. Call create_topic() before first write.";
     throw std::runtime_error(errmsg.str());
   }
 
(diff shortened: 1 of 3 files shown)


Karsten1987:
running CI before merging:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11417/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6633/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9360/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11319/)

Karsten1987:
GH actions are showing the linters being fixed. Merging.

dirk-thomas:
The Windows CI build shows a compiler warning (https://ci.ros2.org/job/ci_windows/11319/msbuild/) which is also present in the last nightly builds.

ros-discourse:

PR #452, opened 2020-07-02
introduce defaults

Karsten1987:
This PR sits on top of #451 to show case the difference when decent defaults are chosen.
I'd like to hear from others whether these defaults are sensible or not.

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reader.cpp
@@ -16,6 +16,7 @@
 #include "rosbag2_cpp/reader.hpp"
 
 #include <memory>
+#include <string>
 #include <utility>
 #include <vector>
 
@@ -32,6 +33,18 @@ Reader::Reader(std::unique_ptr<reader_interfaces::BaseReaderInterface> reader_im
 Reader::~Reader()
 {
   reader_impl_->reset();
+}
+
+void Reader::open(const std::string & uri)
+{
+  rosbag2_cpp::StorageOptions storage_options;
+  storage_options.uri = uri;
+  storage_options.storage_id = "sqlite3";
+  storage_options.max_bagfile_size = 0;  // default
+  storage_options.max_cache_size = 0;  // default
+
+  rosbag2_cpp::ConverterOptions converter_options{};
+  return open(storage_options, converter_options);
 }
 
 void Reader::open(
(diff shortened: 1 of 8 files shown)


Karsten1987:
@emersonknapp sorry, I've added one other thing to this PR. I introduce an overload for `open` for reader and writer which take a single `uri` string to a rosbag and internally fills the `storage_options`. Given that we internally set the storage options to the default, which is `sqlite3`, I believe this introduces a `exec_depend` on `rosbag2_storage_default_plugins`. I do think that's a reasonable thing to do though. But I'd like to hear your opinion about it, whether that's sensible to introduce as a dependency.

ros-discourse:

rosbag2_cpp/src/rosbag2_cpp/reader.cpp
```diff
@@ -34,6 +35,18 @@ Reader::~Reader()
   reader_impl_->reset();
 }
 
+void Reader::open(const std::string & uri)
+{
+  rosbag2_cpp::StorageOptions storage_options;
+  storage_options.uri = uri;
+  storage_options.storage_id = "sqlite3";
```
emersonknapp, line 42: minor nit - should these 3 default values ("sqlite3", 0, 0) be defined as constants somewhere?

Karsten1987: the last two are directly "hardcoded" in the storage options struct - meaning I could technically leave out the following two lines of code. 
Where would you define the other constant, "sqlite3"?

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6731/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11552/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6743/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9487/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11567/)

rosbag2_cpp/src/rosbag2_cpp/reader.cpp
Karsten1987: @emersonknapp I've excluded the `sqlite3` into a constant within the same file. Let me know if that works for you.

emersonknapp: LGTM

PR #456, opened 2020-07-08
More reliable topic remapping test

### LINKED ISSUE #455: Topic remap test is failing regularly
The topic remapping test is failing regularly on CI due to timeouts, e.g.:

```
02:07:44 22: Running main() from gmock_main.cc
02:07:44 22: [==========] Running 1 test from 1 test case.
02:07:44 22: [----------] Global test environment set-up.
02:07:44 22: [----------] 1 test from RosBag2PlayTestFixture
02:07:44 22: [ RUN      ] RosBag2PlayTestFixture.recorded_message_is_played_on_remapped_topic
02:08:44 1/1 Test #22: test_play_topic_remap__rmw_fastrtps_cpp ...***Timeout  60.10 sec
02:08:44 
02:08:44 0% tests passed, 1 tests failed out of 1
02:08:44 
02:08:44 Label Time Summary:
02:08:44 gmock    =  60.10 sec*proc (1 test)
02:08:44 
02:08:44 Total Test time (real) =  60.11 sec
02:08:44 
02:08:44 The following tests FAILED:
02:08:44 	 22 - test_play_topic_remap__rmw_fastrtps_cpp (Timeout)
02:08:44 Errors while running CTest
02:08:44 
```
https://ci.ros2.org/view/nightly/job/nightly_osx_extra_rmw_release/831/consoleFull#console-section-43

The problem is as usual a indeterministic execution of ROS2, which doesn't allow to verify an exact number of messages being transported.

Karsten1987:
fixes #455 
As mentioned on the issue ticket, as long as we don't have a deterministic execution in ROS2, we can't test for an exact number of messages to be transported. 
I lowered the ration of expected messages to only 60 percent, which seems to be fairly reliable and still fulfills the purpose of the test, which is that topics being remapped upon replay.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play_topic_remap.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_topic_remap.cpp
@@ -37,10 +37,7 @@ TEST_F(RosBag2PlayTestFixture, recorded_message_is_played_on_remapped_topic) {
   const std::string remapped_topic = "/topic_after_remap";
   const int32_t test_value = 42;
   const int64_t dt_in_milliseconds = 100;
-  // With ROS2 communication being undeterministic, we can only verify that topics arrive
-  // We can't verify how many though; The 60% below seem to be reliable.
   const size_t expected_number_of_outgoing_messages = 5;
-  const size_t expected_number_of_incoming_messages = 3;
   play_options_.topic_remapping_options =
   {"--ros-args", "--remap", "topic_before_remap:=topic_after_remap"};
 
@@ -60,8 +57,10 @@ TEST_F(RosBag2PlayTestFixture, recorded_message_is_played_on_remapped_topic) {
   prepared_mock_reader->prepare(messages, topic_types);
   reader_ = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
 
+  // We can't reliably test for an exact number of messages to arrive.
+  // Therefore, we test that at least one message is being received over the remapped topic.
   sub_->add_subscription<test_msgs::msg::BasicTypes>(
-    remapped_topic, expected_number_of_incoming_messages);
+    remapped_topic, 1u);
   auto await_received_messages = sub_->spin_subscriptions();
 
   rosbag2_transport::Rosbag2Transport rosbag2_transport(reader_, writer_, info_);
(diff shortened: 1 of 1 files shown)


Karsten1987:
CI (rosbag2_transport, retest-until-fail 10):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11401/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6621/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9347/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11299/)

Karsten1987:
The topic remapping test seems to be addressed with this. So I'll put this up for review.

rosbag2_transport/test/rosbag2_transport/test_play_topic_remap.cpp
```diff
@@ -32,14 +31,16 @@
...
 TEST_F(RosBag2PlayTestFixture, recorded_message_is_played_on_remapped_topic) {
   //  test constants
   const std::string original_topic = "/topic_before_remap";
   const std::string remapped_topic = "/topic_after_remap";
   const int32_t test_value = 42;
-  const int64_t timestamp_in_milliseconds = 700;
-  const size_t expected_number_of_messages = 1;
+  const int64_t dt_in_milliseconds = 100;
+  // With ROS2 communication being undeterministic, we can only verify that topics arrive
+  // We can't verify how many though; The 60% below seem to be reliable.
```
emersonknapp, line 41: Does this count really matter for this test, or should we be happy to see just a single message arrive? I would think that a more reliable test would be to publish an arbitrarily large number of messages and interrupt execution when we receive a single message on the subscription. (IIRC we don't have a good way to stop playback yet, though, so maybe the point is just "we should expect 1 message")

Karsten1987: I was thinking about this. And while I am not settled on the 60% either, I was arguing with myself to apply the "1, 2, many" to rule out there're no leftovers in the system. That's not to say that this is reasonable in the first place :P So I am happy to modify the test in a way to expect at least 1 message.

emersonknapp: I think that's the right thing to do for now. It's unfortunate that we can't depend on the specific numbers, but that's not the specific thing being tested here so we might as well loosen the restrictions all the way to only require the exact behavior being tested to work.

Karsten1987:
testing the failing test specifically (test-until-fail 30)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11410/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6627/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9353/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11312/)

PR #457, opened 2020-07-10
use rclcpp serialized messages to write data

Karsten1987:
~Next on the list towards a user-friendly C++ API. Sits on top of "defaults" branch.~

The implementation here is pretty straightforward even though we have to iterate over the existing data structures used here at some point, namely the `rosbag2_storage::SerializedBagMessage`. We're dealing with shared pointers here in order to ease the cleanup of the underlying `uint8` buffer. However, this prevents any stack allocations and makes it pretty cumbersome to integrate with the existing `rclcpp::SerializedMessage` API.

As a second point, I personally don't see a way around the `rclcpp` dependency at this point. Even though I'd like to keep the number of dependencies here fairly small, in order to eventually come up with a `template<class MsgT> void write(MsgT & msg)` API, we'd need to have that dependency.

I'll open this PR as a draft to start a discussion.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -21,6 +21,10 @@
 #include <unordered_map>
 #include <vector>
 
+#include "rclcpp/serialization.hpp"
+#include "rclcpp/serialized_message.hpp"
+#include "rclcpp/time.hpp"
+
 #include "rosbag2_cpp/converter_options.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 #include "rosbag2_cpp/writers/sequential_writer.hpp"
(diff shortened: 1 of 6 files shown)


emersonknapp:
Maybe I'm missing some context - now that rclcpp::SerializedMessage exists (and if we pull a dependency on rclcpp) - do we still need the rosbag2 SerializedBagMessage structure? Or do they provide the same functionality

Karsten1987:
fair point. I am a big fan of making the serialized bag message a POD, essentially removing the shared pointer and only having the actual buffer pointer. 

We could get around rclcpp if the rosbag2 api only introduces API which takes a serialized data buffer (`uint8_t*`), but yes, if we introduce `rclcpp`, I think it makes sense to remove the serialized bag message altogether.

Karsten1987:
oh, also to be clear. The `rclcpp` dependency will only be introduced at `rosbag2_cpp`, not below. So we still need to keep the `SerializedBagMessage` struct around. It should really only be a container for the data buffer IMO.

emersonknapp:
Makes sense. I think it's fine to add the `rclcpp` dependency. We were trying to limit it to just `rosbag2_transport`, right? However, it does seem reasonable to say "you're using a ROS2-related C++ API, it requires the C++ client library"

ros-discourse:

emersonknapp:
@Karsten1987 did this already get implemented in a different PR? Any need to keep this draft open?

Karsten1987:
@emersonknapp please have a look at this. I've just finished the reader and writer API in conjunction with the `rclcpp` addition. The API should now look relatively close to the ROS1 cookbook snippets.

rosbag2_cpp/package.xml
```diff
@@ -12,6 +12,7 @@
 
   <depend>ament_index_cpp</depend>
   <depend>pluginlib</depend>
+  <depend>rclcpp</depend>
```
emersonknapp, line 15: We avoided adding this dependency for a long time :) but it seems unavoidable

rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -126,6 +130,45 @@ class ROSBAG2_CPP_PUBLIC Writer final
     const std::string & type_name,
     const std::string & serialization_format = "cdr");
 
+  /**
+   * Write a serialized message to a bagfile.
+   * Opposing a call to \sa write without topic metadata, the topic will be created.
```
emersonknapp, line 135: nit: I know this was already on the existing signature - but I found the wording confusing. Maybe this would be clearer:

```suggestion
   * The topic will be created if it has not been created already 
```

emersonknapp (approved):
This LGTM

Karsten1987:
I never run CI on this:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14052/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8873/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11745/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14143/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14054/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8875/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11747/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14145/)

PR #461, opened 2020-07-15
reenable cppcheck for rosbag2_transport

Karsten1987:
I wasn't aware of it and not sure why it got disabled in the first place, but with the following patch `cppcheck` passes successfully locally on my OSX.
Alternatively, one can place a  `// cppcheck-suppress unknownMacro` above the two `RCLCPP` macros. 

This PR was inspired by https://github.com/ros2/rclpy/pull/577

CI (only testing `cppcheck` within `rosbag2_transport`):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11472/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6677/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11365/)

PR #463, opened 2020-07-16
Use foxy testing apt repos to install linters for Actions

emersonknapp:
We were still using the Eloquent release of the linters, which is not accurate to the current code.

Depends on https://github.com/ros2/rosbag2/pull/461

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,8 +1,4 @@
 # Run linters automatically on pull requests.
-#
-# ros-tooling/actino-ros-lint is relying on the latest APT binary packages.
-# As of 2020-05-01, Eloquent is the latest release for which APT binary
-# packages are available, which is why the Docker image is based on Eloquent.
 name: Lint rosbag2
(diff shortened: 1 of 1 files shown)


Karsten1987:
Would it make sense to use the linters from the master branch? I mean, the moment something changes upstream we'll have the same inconsistency as right now in which the binary packages are different from ci.ros2.org. Is that right?

emersonknapp:
I think that it does make sense - but for the moment action-ros-lint operates by expecting to install the linters via `apt`, and we don't have a "master apt" build available. This at least gets us a little closer to in sync.

Then, there is https://github.com/ros-tooling/action-ros-lint/issues/142 which requests we install the linters via `pip` instead, which would provide us an avenue to get the latest version of the linters and probably be lighter weight.

Karsten1987:
Can you rebase this on top of master? I'd expect the `cppcheck` action to pass then.

PR #465, opened 2020-07-17
Re-enable Action CI for PRs

emersonknapp:
We disabled Action CI a while back because the tests were unreliable (much more so here than on ci.ros2.org) and did not want to block/confuse contributors. They seem to be working well now so this PR tests re-enabling that check.

I am still seeing ~1/6 (4/24) of the last day's runs failing https://github.com/ros2/rosbag2/actions?query=workflow%3A%22Test+rosbag2%22 - so maybe this isn't quite ready for prime-time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Test rosbag2
 on:
-  # pull_request:
+  pull_request:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


emersonknapp:
The Action CI is still failing ~1/6 times, we don't feel this is good enough yet for contributors to have confidence in their results.

PR #466, opened 2020-07-18
move wait_for_shutdown() call out of the context manager

dirk-thomas:
Hopefully fixes (no, it doesn't) #454 this time.

The subprocess started by `launch_testing.tools.launch_process()` is only signaled to terminate when the context exits. Therefore calling `wait_for_shutdown()` within the `with` block doesn't do anything but wait for the timeout. Afterwards it was a race if the process terminated quickly enough so that the file handles from the subprocess where released before the tear down function of the test tries to delete the temporary directory.

Moving the `wait_for_shutdown()` call outside the `with` block ensures to give the shutdown event enough time to be processed.

Windows build testing `ros2bag` with `--retest-until-fail 10`: [Build Status](https://ci.ros2.org/job/ci_windows/11481/)

Full CI just because:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11545/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6708/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9443/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11480/)

Issue #454, opened 2020-07-08
ros2bag QOS test uses several sleeps on windows


Karsten1987:
Test is "reliably" failing on windows:
https://ci.ros2.org/view/nightly/job/nightly_win_rel/1605/testReport/ros2bag.test/test_record_qos_profiles/test_record_qos_profiles/

```
FAIL: test_record_qos_profiles.TestRos2BagRecord.test_incomplete_qos_profile
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\python38\lib\shutil.py", line 613, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmphur3j5pi\\ros2bag_test_incomplete_0.db3'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\python38\lib\tempfile.py", line 803, in onerror
    _os.unlink(path)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmphur3j5pi\\ros2bag_test_incomplete_0.db3'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\ci\ws\src\ros2\rosbag2\ros2bag\test\test_record_qos_profiles.py", line 87, in test_incomplete_qos_profile
    assert not matches, print('ros2bag CLI did not produce the expected output')
```

@emersonknapp do you have an idea?

emersonknapp:
I don't know immediately but will begin investigating

dirk-thomas:
@emersonknapp Any update on this?

dirk-thomas:
Please see #462 which modifies the test to only create one temp dir for all tests within the test class. That allows it to be cleanly removed on Windows.

dirk-thomas:
The test is still failing (even though it passed CI on the PR) - reopening...

emersonknapp:
Noted - we are pulling this in as top priority for Monday

jaisontj:
@dirk-thomas Could you take a look at: #525 ? It gets rid of most of the `time.sleep` except for the one inside `tearDownClass`. Was not able to find a better way around it. However, given the title of this issue, we no longer use several sleeps in the test, can this issue be closed?

[...]

PR #470, opened 2020-07-23
add wait for closed file handles on Windows

dirk-thomas:
Follow up of https://github.com/ros2/rosbag2/pull/466#issuecomment-661122726.

Only testing `ros2bag` on Windows: [Build Status](https://ci.ros2.org/job/ci_windows/11570/)


dirk-thomas:
> we definitely want to get rid of all these `sleep(3)` calls asap.

Absolutely agreed. I updated the title of the existing ticket which can be used to track this: #454.

PR #472, opened 2020-07-24
catch parent exception

dirk-thomas:
Follow up of #470.

I am not planning to run CI builds for this change since the PR build ensures that the syntax is valid.

A single Windows build won't say anything if this actually catches the problem in the test since this is subject to a race. Hopefully the next nightlies will show an improvement.


dirk-thomas:
The failing test in the PR build seems unrelated to this change.

dirk-thomas:
This Windows build shows no test failure anymore: [Build Status](https://ci.ros2.org/job/ci_windows/11599/) (just an unrelated CMake warning).

PR #479, opened 2020-07-31
fix missing target dependencies

dirk-thomas:
Fix build failure reported in https://answers.ros.org/question/358314/colcon-build-include-path-from-overlay-workspace/.


dirk-thomas:
I just also added an explicit dependency on `rosbag2_storage` instead of relying it to be transitive dependency: dc1575b3967f410b7ea56ec1af99dc0531ebe051.

dirk-thomas:
If the PR builds are green I would suggest merging this since the CI builds from ci.ros2.org won't provide any more information. The actual change can only be tested with an overlay workspace.

Karsten1987:
I can confirm that this solves the issue. The packages within rosbag2 can be built correctly in an overlay workspace.
Thanks @dirk-thomas for the fix.

PR #480, opened 2020-07-31
Fix topc id map

Stapelzeiger:
The `topic_id` field in the messages table is sometimes incorrect when logging many topics (\~20) at higher rate (\~ few 100Hz). This results in `bag info` reporting that the affected topic has recorded messages, but `bag play` will never publish them since the `topic_id` is wrong.
I've tracked the source of this down to the `database_->get_last_insert_id()` reporting the wrong number. My suspicion is that this might be a race condition with some other topic writing to the messages table.

I have replaced the `get_last_insert_id()` by a query for the id of the topic entry after it is written. This fixes the bug.

See log for debug output with the patch (this usually happens once every a few recordings):
```
...
[INFO] [1596216662.035733631] [rosbag2_storage]: new topic /accel
[INFO] [1596216662.035835423] [rosbag2_storage]: id from query 10
[INFO] [1596216662.035868673] [rosbag2_storage]: last insert id 10
[INFO] [1596216662.037794547] [rosbag2_transport]: Subscribed to topic '/accel'
[INFO] [1596216662.038935838] [rosbag2_storage]: new topic /twist
[INFO] [1596216662.039000880] [rosbag2_storage]: id from query 11
[INFO] [1596216662.039022755] [rosbag2_storage]: last insert id 36
[INFO] [1596216662.040468838] [rosbag2_transport]: Subscribed to topic '/twist'
...
```

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -237,7 +237,17 @@ void SqliteStorage::create_topic(const rosbag2_storage::TopicMetadata & topic)
     insert_topic->bind(
       topic.name, topic.type, topic.serialization_format, topic.offered_qos_profiles);
     insert_topic->execute_and_reset();
-    topics_.emplace(topic.name, static_cast<int>(database_->get_last_insert_id()));
+
+    ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO("new topic %s", topic.name.c_str());
+    auto statement = database_->prepare_statement(
+      "SELECT id FROM topics WHERE topics.name IN (?);");
+    statement->bind(topic.name);
+    auto query_results = statement->execute_query<int>(); // not sure int is the right type
+    auto result = std::begin(query_results);
+    assert(result != std::end(query_results));
+    topics_.emplace(topic.name, static_cast<int>(std::get<0>(*result)));
+    ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO("id from query %d", static_cast<int>(std::get<0>(*result)));
+    ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO("last insert id %d", static_cast<int>(database_->get_last_insert_id()));
   }
 }
 


Stapelzeiger:
Note this seems to happen much more frequently with cyclonedds, but that might just be due to the higher throughput I am getting with it.

Stapelzeiger:
see #486 description for a launch file to reproduce this.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -237,7 +237,17 @@ void SqliteStorage::create_topic(const rosbag2_storage::TopicMetadata & topic)
     insert_topic->bind(
       topic.name, topic.type, topic.serialization_format, topic.offered_qos_profiles);
     insert_topic->execute_and_reset();
-    topics_.emplace(topic.name, static_cast<int>(database_->get_last_insert_id()));
+
+    ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_INFO("new topic %s", topic.name.c_str());
```
emersonknapp, line 241: Let's either remove these debug print statements, or format more informatively and bump down to `_DEBUG` severity.

Can you add a comment here as to why the logic is necessary? It's not obvious to me what the purpose of the SELECT statement is at a glance. For readability as well, it may be nice to write the result to a named variable e.g.

```
queried_topic_id = static_cast<int>(std::get<0>(*result)));
topics_.emplace(topic.name, queried_topic_id);
```

Stapelzeiger: Yes, I will remove the debug prints. They were there mostly to illustrate the issue. 
I am not sure if a db query is the right way to solve this. I will refactor it to be more readable if we decide to keep it.

Stapelzeiger:
There is another concurrency related issue (#486). I am wondering if the topic entry creation and message writing were supposed to happen concurrently.
Depending on how we fix #486, the solution here might look differently.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -237,7 +237,17 @@ void SqliteStorage::create_topic(const rosbag2_storage::TopicMetadata & topic)
...
+    auto statement = database_->prepare_statement(
+      "SELECT id FROM topics WHERE topics.name IN (?);");
+    statement->bind(topic.name);
+    auto query_results = statement->execute_query<int>(); // not sure int is the right type
```
Stapelzeiger, line 245: Should this be int64_t?

Stapelzeiger:
The storage plugins are not designed to be thread safe. Since this PR fixes an issue that occurs only from misuse of the API in the recorder, I will close this PR.
I opened a PR that fixes the problem in the recorder (#491).

PR #481, opened 2020-07-31
[foxy backport] fix missing target dependencies (#479)

Karsten1987:
backport of #479


Karsten1987:
@emersonknapp fyi: the dev job showed some double free corruption in the QoS tests.

emersonknapp:
That may be what's happening in our Action CI. About 1/6 times one fastrtps qos test exits with -11, maybe the double free can return that?

Issue #445, opened 2020-06-18
fix troubleshooting URL


dirk-thomas:
Since the page was moved in ros2/ros2_documentation#700. Same as ros2/rclpy#579

See https://github.com/ros2/rosbag2/blob/e9d006b6a88bddc42b5a167e121a8fc950f6cced/rosbag2_transport/rosbag2_transport/__init__.py#L33

Instead of fixing the URL in the code I suggest to deduplicate this copied logic with the one existing in `rclpy` and make the functionality reusable instead.

The fix needs to be backported to all active distros.

emersonknapp:
AWS team picking this up for planned completion by next WG meeting

PR #482, opened 2020-08-01
Use rpyutils import_c_library function for rosbag2_transport_py

emersonknapp:
Use rpyutils import_c_library function for rosbag2_transport_py - instead of re-defining the logic here (with outdated information)

Depends on https://github.com/ros2/rpyutils/pull/4
Closes https://github.com/ros2/rosbag2/issues/445

### DIFF
--- a/rosbag2_transport/rosbag2_transport/__init__.py
+++ b/rosbag2_transport/rosbag2_transport/__init__.py
@@ -12,27 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import importlib
-import os
+from rpyutils import import_c_library
 
-from rpyutils import add_dll_directories_from_env
-
-
-def _import(name):
-    try:
-        # Since Python 3.8, on Windows we should ensure DLL directories are explicitly added
-        # to the search path.
-        # See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
-        with add_dll_directories_from_env('PATH'):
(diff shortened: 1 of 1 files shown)


emersonknapp:
using https://gist.githubusercontent.com/emersonknapp/03048af0c22cf44348e76a3da1e1dd1b/raw/5fc3c523f2e6446a0d7e941d7cff9d90c13f4eaa/ros2.repos

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11783/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6863/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9622/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11726/)

Karsten1987:
@emersonknapp this branch can be deleted, I think.

Karsten1987:
Also, the URL should be corrected in older/existing distros. I guess this PR can't really be used to be backported, can it?

emersonknapp:
Ah - this I didn't think all the way through - we could backport this only into Foxy because rpyutils was added in that distro.

What do you think about a single backport patch commit to Foxy, Eloquent, and Dashing that just copies the contents of the new rpyutils function?

Karsten1987:
> What do you think about a single backport patch commit to Foxy, Eloquent, and Dashing that just copies the contents of the new rpyutils function?

I think it makes sense to backport this PR to Foxy. For Eloquent and Dashing I would actually really just change the URL.

PR #483, opened 2020-08-01
Validate QoS profile values are not negative.

### LINKED ISSUE #431: Negative values allowed in QoS overrides file
## Description
When defining an [overrides file](https://index.ros.org/doc/ros2/Tutorials/Ros2bag/Overriding-QoS-Policies-For-Recording-And-Playback/) negative values are allowed for the following fields:
- depth 
- deadline
 - lifespan
 - liveliness_lease_duration

when using `ros2 bag record` or `ros2 bag play` 

### Expected Behavior
The CLI throws an error / exception stating that negative values are not allowed for these fields.

### Actual Behavior
The CLI accepts the values and attempts to record / play.

## To Reproduce
1. Define an overrides file with negative values, e.g., 

```
/test:
  history: keep_last
  depth: -1
```
2. Run `ros2 bag record -a --qos-profile-overrides-path negative_depth.yaml `

 - OS: OSX 10.14, Ubuntu Focal
 - ROS 2 Distro: Foxy pre release binaries

jikawa-az:
Closes #431 

Signed-off-by: Jesse Ikawa <jikawa@amazon.com>


jikawa-az:
Need to add unit test

jikawa-az:
After confirming this test case works, split into 4 test cases to individually test: depth, deadline, lifespan, liveliness_lease_duration

ros2bag/test/test_play_qos_profiles.py
```diff
@@ -82,3 +82,15 @@ def test_qos_incomplete(self):
         expected_string_regex = re.compile(ERROR_STRING)
         matches = expected_string_regex.search(bag_command.output)
         assert not matches, print('ros2bag CLI did not produce the expected output')
+        
+    def test_qos_negative(self):
```
emersonknapp, line 86: This is a very heavy test and doesn't hit a case that i think right now is incorrect. I think this would be better structured as a unit test on the `interpret_dict_as_qos_profile` function, passing in a dict of values rather than going through launch and subprocess and the filesystem. 

You should have at least 4 separate tests with only a single invalid value in each. I believe the duration values are not being checked correctly right now, since they go through `dict_to_duration`, as they are not `_VALUE_KEYS`. Adding this extra lighter weight unit test should help you catch this error.

ros2bag/test/test_play_qos_profiles.py
```diff
@@ -82,3 +82,4 @@ def test_qos_incomplete(self):
         expected_string_regex = re.compile(ERROR_STRING)
         matches = expected_string_regex.search(bag_command.output)
         assert not matches, print('ros2bag CLI did not produce the expected output')
+
```
emersonknapp, line 85: should revert this change - may want to set your editor to always ensure newlines if you can

ros2bag/test/test_api.py
```diff
@@ -74,3 +74,17 @@ def test_convert_yaml_to_qos_profile(self):
         assert qos_profiles[topic_name_2].avoid_ros_namespace_conventions == expected_convention
         assert qos_profiles[topic_name_2].history == \
             QoSHistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_ALL
+
+    def test_interpret_dict_as_qos_profile_negative(self):
```
emersonknapp, line 78: have you run this test, does it pass?

jikawa-az: all four tests are passing now :)

jikawa-az:
Will 'sec' and 'nsec' be the only time duration values for QoS profile in the future? If not I should revise to cover any time duration value.

ros2bag/ros2bag/api/__init__.py
```diff
@@ -44,6 +44,10 @@ def dict_to_duration(time_dict: Optional[Dict[str, int]]) -> Duration:
     """Convert a QoS duration profile from YAML into an rclpy Duration."""
     if time_dict:
         try:
+            if time_dict['sec'] < 0:
+                raise ValueError('`sec` may not be a negative value.')
+            if time_dict['nsec'] < 0:
```
dabonnie, line 49: If `sec` is zero couldn't `nsec` be zero? I think you'll want to use all possible time inputs and convert to a duration.

jikawa-az: Do you mean that 0 seconds and 0 nsecs is not a valid input?

dabonnie: Sorry let me rephrase and dismiss: I was thinking that it could be possible to have a negative `nsec` but not sure why someone would enter the value this way, because you're pulling the values directly from the user input. My mistake.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11764/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6840/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9598/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11706/)

Karsten1987:
@jikawa-az @emersonknapp can this PR be backported? If so, it would be great if you could come up with a backport PR today so we can cut a new release for rosbag2.

Issue #485, opened 2020-08-02
Playback error on macOS


Stapelzeiger:
## Description
Playing back a bag that contains only the *.db3 file (without the *.db3-shm, *.db3-wal) fails on macOS with `"SQLite error (14): unable to open database file"`

### Expected Behavior
Bag plays correctly.

### Actual Behavior
fails to play with error message described above.

## To Reproduce
open a bag that doesn't contain *.db3-shm/*.db3-wal files on macOS:
`ros2 bag play <name-of-bag>`

to get such a bag either create it on aarch64 ubuntu (those don't contain the files for some reason)
or open the .db3 in https://sqlitebrowser.org and close it. This removes the *.db3-shm/*.db3-wal files.

## System
 - OS: macOS 10.14
 - ROS 2 Distro: Foxy
 - Version: binary patch 1

## Additional context

The problem error comes from [this line](https://github.com/ros2/rosbag2/blob/6ff9d61535c59ee539d0f70f1e1a017f467254d6/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp#L48).
From what I understand, the database is in WAL mode and the *.db3-shm, *.db3-wal have to be created if they don't exist. This seems to fail when opening in read-only mode for some reason.

Changing to `SQLITE_OPEN_READWRITE` [in the sqlite3_open_v2](https://github.com/ros2/rosbag2/blob/6ff9d61535c59ee539d0f70f1e1a017f467254d6/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp#L40) fixes the error, the extra files are created and the bag plays correctly.

I believe this fixes half of the issue #153 (the playback of an armhf bag on macOS), but I haven't tested it yet.

Karsten1987:
very interesting. Thanks for reporting this. Do you mind opening a PR with your proposed fix?

Stapelzeiger:
Sure, I'll do that.

MichaelOrlov:
@Stapelzeiger At some point we changed `PRAGMA journal_mode=MEMORY` by default instead of `PRAGMA journal_mode=WAL`.
Therefore database not in `WAL` mode by default any more. 
Therein this issue shouldn't be reproducible any more.  
However I don't have macOS to reproduce or test it.

I am closing this issue as stale and fixed via workaround. Feel free to reopen it if issue still reproduces.

PR #487, opened 2020-08-04
Open sqlite-db in RW for reading

Stapelzeiger:
This fixes playback on macOS when the *.db3-shm, *.db3-wal don't exist yet. See #485

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
@@ -37,7 +37,7 @@ SqliteWrapper::SqliteWrapper(
   if (io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY) {
     int rc = sqlite3_open_v2(
       uri.c_str(), &db_ptr,
-      SQLITE_OPEN_READONLY | SQLITE_OPEN_NOMUTEX, nullptr);
+      SQLITE_OPEN_READWRITE | SQLITE_OPEN_NOMUTEX, nullptr);
     if (rc != SQLITE_OK) {
       std::stringstream errmsg;
       errmsg << "Could not read-only open database. SQLite error (" <<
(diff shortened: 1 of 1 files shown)


rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -37,7 +37,7 @@ SqliteWrapper::SqliteWrapper(
   if (io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY) {
     int rc = sqlite3_open_v2(
       uri.c_str(), &db_ptr,
-      SQLITE_OPEN_READONLY | SQLITE_OPEN_NOMUTEX, nullptr);
+      SQLITE_OPEN_READWRITE | SQLITE_OPEN_NOMUTEX, nullptr);
```
emersonknapp, line 40: Does this pose any danger of misuse? Would it be better to do a quick "open-close" to create the proper files, then still open as readonly?

Stapelzeiger: Good point.  I did a quick test, opening in RW, executing the first statement (`PRAGMA schema_version;`), closing it and reopening as RO works, and the *.db3-shm, *.db3-wal are created. I can implement it like that, but we're doubling the number of lines of code.

I'm wondering why it needs RW anyway. I'm not familiar with sqlite, so I don't understand how/why these WAL files are used. Maybe we are using it wrong?

Karsten1987:
We might use it incorrectly. According to the [documentation](https://sqlite.org/wal.html#read_only_databases) older versions of sqlite3 did not support WAL in RO mode. The way I understand it, the proposed “best-practice” is to apply journal_mode DELETE before closing the connection after the last write. But we might be able to apply the same pragma when opening the db in RO.

Stapelzeiger:
So you're suggesting to open the db in RW, set the journal_mode=DELETE, close and reopen as RO?

Stapelzeiger:
@Karsten1987 something like the above? (92b51a4)

gbiggs:
@Stapelzeiger What's the status of this PR? Is this still needed? If so we should move forward with getting it fixed and merged.

MichaelOrlov:
Closing as will not done, since relevant issue has been closed. Please refer to the closing comment https://github.com/ros2/rosbag2/issues/485#issuecomment-1233269755

PR #491, opened 2020-08-06
Mutex around writer access in recorder

### LINKED ISSUE #486: Database access race condition when recording
## Description

Recording rosbags with lots of topics and high message rate causes race condidtions when writing to the database.
I believe the problem is that new topic entries are written while at the same time message data from other topics is coming in.

### Expected Behavior
`ros2 bag record -a` records all topics without errors

### Actual Behavior

The problem manifests itself in about 1 in 10 runs (with the launchfile below) in one of the following three ways:
- recording seems fine, then right after hitting ctrl-C the it logs an error and exits, the bag is almost empty (I believe something gets stuck in the topic setup)
```
...
^C[INFO] [1596518181.782926826] [rclcpp]: signal_handler(signal_value=2)
[ERROR] [1596518182.245972174] [rosbag2_transport]: Failed to record: Error when processing SQL statement. SQLite error (5): database is locked
```
- segfault a few seconds after recording starts, while still adding new topics
- error "database disk image is malformed" a few seconds after recording starts, again while adding topics (this seems to be the most common)
```
...
[INFO] [1596518141.207651288] [rosbag2_transport]: Subscribed to topic '/topic57'
[ERROR] [1596518141.560762687] [rosbag2_transport]: Failed to record: Error when processing SQL statement. SQLite error (11): database disk image is malformed
```

## To Reproduce

run many publishers. The launch file below seems to work, although not as well as my robot which is publishing at higher data rate (but fewer topics).
```
from launch import LaunchDescription
from launch_ros.actions import Node

def generate_launch_description():
    return LaunchDescription([
        Node(
            package='examples_rclcpp_minimal_publisher',
            executable='publisher_lambda',
            name='pub'+str(i),
            remappings=[
                ('/topic', '/topic'+str(i))
            ]
        ) for i in range(70)
    ])
```
then run `ros2 bag record -a`

- OS: macOS 10.14, ubuntu 20.4 (aarch64)
- ROS 2 Distro: Foxy
- Version: binary patch 1

## Additional context

[Opening the database](https://github.com/ros2/rosbag2/blob/6ff9d61535c59ee539d0f70f1e1a017f467254d6/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp#L52) in `SQLITE_OPEN_FULLMUTEX` mode instead of `SQLITE_OPEN_NOMUTEX ` fixes this. I've also tried manually adding a lock around the topic entry write and the messages write, which seems to work as well.

See also #480 for a related issue. (`database_->get_last_insert_id()` is sometimes incorrect, also caused by multiple threads using the database)

Stapelzeiger:
The storage plugins are not designed to be thread safe. The recorder must therefore use a mutex when accessing the writer from different contexts. This happens for topic creation during discovery as well as in message callbacks from subscribed topics.

This fixes #486 and #480.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -16,10 +16,10 @@
 #define ROSBAG2_CPP__WRITER_HPP_
 
 #include <memory>
+#include <mutex>
 #include <string>
 #include <unordered_map>
 #include <vector>
-#include <mutex>
 
 #include "rosbag2_cpp/converter_options.hpp"
 #include "rosbag2_cpp/storage_options.hpp"
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <mutex>
 
 #include "rosbag2_cpp/converter_options.hpp"
 #include "rosbag2_cpp/storage_options.hpp"
@@ -115,6 +116,7 @@ public:
   }
 
 private:
+  std::mutex writer_mutex_;
   std::unique_ptr<rosbag2_cpp::writer_interfaces::BaseWriterInterface> writer_impl_;
 };
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writer.cpp
@@ -60,16 +60,19 @@ void Writer::open(
 
 void Writer::create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
 {
+  std::lock_guard<std::mutex> writer_lock(writer_mutex_);
   writer_impl_->create_topic(topic_with_type);
 }
 
 void Writer::remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
 {
+  std::lock_guard<std::mutex> writer_lock(writer_mutex_);
   writer_impl_->remove_topic(topic_with_type);
 }
 
 void Writer::write(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message)
 {
+  std::lock_guard<std::mutex> writer_lock(writer_mutex_);
   writer_impl_->write(message);
 }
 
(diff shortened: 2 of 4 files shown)


dawonn-haval:
Just wanted to leave a note that this patch seems to have made an improvement on my system. Before this patch I was only able to record sensor data bag files for a 5~10 seconds before it would hang and stop recording, but now rosbag constantly runs 1~2 minutes before hanging! 

FWIW: Foxy, Ubuntu 20.04, 5 lidar + 1 camera.

Stapelzeiger:
@Karsten1987 I added locking in the `rosbag2_cpp` writer. I've reverted the changes in recorder, since they are redundant with the `rosbag2_cpp` ones so this should solve your second point regarding the `create_subscription` in the critical zone.

Do we need a similar thread-safe interface for the reader class? I have trouble imagining a use-case where you would read the from multiple threads with the current API (in particular the `has_next()`, `read_next()` wouldn't work nicely).

rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <unordered_map>
 #include <vector>
+#include <mutex>
```
Karsten1987, line 22: alpha sort

jarledo:
Any updates on this? Our system suffers from the bug described in #486, making it almost impossible to record our topics. Would be great to have this merged.

Karsten1987:
Thanks for the reminder, kicking off CI on this:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7780/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12768/)

Karsten1987 (approved):

Karsten1987:
@Stapelzeiger thanks for the contribution

Karsten1987:
@adamdbrw fyi

almaslov:
Hello! It would be great to backport this fix to Foxy. Can I do it myself according to [the Guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#guidelines-for-backporting-prs) ?

emersonknapp:
Yes, by all means. If the cherry-pick works without issue then it is an easy backport.

PR #493, opened 2020-08-07
add storage_config_uri

Karsten1987:
first step towards #437 

sometimes I'd wish our code base was less complex, with less test code. I had to change 34 files to introduce one parameter to `open()` ;-)

I guess I'll leave this PR as is, with the new functionality throwing one use.

PR #497, opened 2020-08-07
read yaml config file

Karsten1987:
implements to read a config file (`.yaml`) tailored for sqlite3. As for now, I've only implemented to read pragmas in the style of:

```
read:
  pragmas: ["schema_version"]
write:
  pragmas: ["synchronous = NORMAL", "journal_mode = WAL"]
```

I am struggling a bit to write tests for this as I have a hard time to introspect if these pragmas are actually affective. I can't really test for wrong pragmas either as [according to the sqlite documentation](https://www.sqlite.org/pragma.html#pragma_synchronous) "No error messages are generated if an unknown pragma is issued."

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -33,6 +33,19 @@
 #include "rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.hpp"
 #include "rosbag2_storage_default_plugins/sqlite/sqlite_exception.hpp"
 
+#ifdef _WIN32
+// This is necessary because of a bug in yaml-cpp's cmake
+#define YAML_CPP_DLL
+// This is necessary because yaml-cpp does not always use dllimport/dllexport consistently
+# pragma warning(push)
+# pragma warning(disable:4251)
+# pragma warning(disable:4275)
+#endif
+#include "yaml-cpp/yaml.h"
+#ifdef _WIN32
+# pragma warning(pop)
+#endif
+
 #include "../logging.hpp"
 
 namespace
@@ -75,9 +88,29 @@ void SqliteStorage::open(
   const rosbag2_storage::StorageOptions & storage_options,
   rosbag2_storage::storage_interfaces::IOFlag io_flag)
 {
+  std::vector<std::string> pragmas;
   if (!storage_options.storage_config_uri.empty()) {
-    fprintf(stderr, "going to open config file: %s\n", storage_options.storage_config_uri.c_str());
-    throw std::runtime_error("storage specific config file is not yet implemented.");
+    try {
+      auto key =
+        io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY ? "read" : "write";
+      YAML::Node yaml_file = YAML::LoadFile(storage_options.storage_config_uri);
+      pragmas = yaml_file[key]["pragmas"].as<std::vector<std::string>>();
+    } catch (const YAML::Exception & ex) {
+      throw std::runtime_error(std::string("Exception on parsing info file: ") + ex.what());
+    }
+    // poor developer's sqlinjection prevention ;-)
+    std::string invalid_characters = {"';\""};
+    auto throw_on_invalid_character = [](const auto & pragmas, const auto & invalid_characters) {
+        for (const auto & pragma_string : pragmas) {
+          auto pos = pragma_string.find_first_of(invalid_characters);
+          if (pos != std::string::npos) {
+            throw std::runtime_error(
+                    std::string("Invalid characters in sqlite3 config file: ") +
+                    pragma_string[pos]);
+          }
+        }
+      };
+    throw_on_invalid_character(pragmas, invalid_characters);
   }
 
   if (is_read_write(io_flag)) {
@@ -99,7 +132,7 @@ void SqliteStorage::open(
   }
 
   try {
-    database_ = std::make_unique<SqliteWrapper>(relative_path_, io_flag);
+    database_ = std::make_unique<SqliteWrapper>(relative_path_, io_flag, std::move(pragmas));
   } catch (const SqliteException & e) {
     throw std::runtime_error("Failed to setup storage. Error: " + std::string(e.what()));
   }
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -80,7 +80,9 @@ inline std::vector<std::string> parse_pragmas(
       YAML::Node yaml_file = YAML::LoadFile(storage_config_uri);
       pragmas = yaml_file[key]["pragmas"].as<std::vector<std::string>>();
     } catch (const YAML::Exception & ex) {
-      throw std::runtime_error(std::string("Exception on parsing info file: ") + ex.what());
+      throw std::runtime_error(
+              std::string("Exception on parsing sqlite3 config file: ") +
+              ex.what());
     }
     // poor developer's sqlinjection prevention ;-)
     std::string invalid_characters = {"';\""};
@@ -90,7 +92,9 @@ inline std::vector<std::string> parse_pragmas(
           if (pos != std::string::npos) {
             throw std::runtime_error(
                     std::string("Invalid characters in sqlite3 config file: ") +
-                    pragma_string[pos]);
+                    pragma_string[pos] +
+                    ". Avoid following characters: " +
+                    invalid_characters);
           }
         }
       };
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -23,6 +23,7 @@
 #include <fstream>
 #include <memory>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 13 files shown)


Karsten1987:
this is rebased on top of the latest changes in #493

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -75,9 +88,29 @@ void SqliteStorage::open(
   const rosbag2_storage::StorageOptions & storage_options,
   rosbag2_storage::storage_interfaces::IOFlag io_flag)
 {
+  std::vector<std::string> pragmas;
   if (!storage_options.storage_config_uri.empty()) {
-    fprintf(stderr, "going to open config file: %s\n", storage_options.storage_config_uri.c_str());
-    throw std::runtime_error("storage specific config file is not yet implemented.");
+    try {
```
emersonknapp, line 93: nitpick: this might be a nice separate function to keep the length of this `open` down

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -56,8 +58,33 @@ SqliteWrapper::SqliteWrapper(
         rc << "): " << sqlite3_errstr(rc);
       throw SqliteException{errmsg.str()};
     }
-    prepare_statement("PRAGMA journal_mode = WAL;")->execute_and_reset();
-    prepare_statement("PRAGMA synchronous = NORMAL;")->execute_and_reset();
+
+    auto journal_mode_set = std::find_if(
+      pragmas.begin(), pragmas.end(),
+      [](const auto & pragma) {
+        if (pragma.rfind("journal_mode", 0) == 0) {
```
emersonknapp, line 65: maybe
```suggestion
        return (pragma.rfind("journal_mode", 0) == 0);
```

same below - personally find it more readable

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -33,6 +33,19 @@
 #include "rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.hpp"
 #include "rosbag2_storage_default_plugins/sqlite/sqlite_exception.hpp"
 
+#ifdef _WIN32
```
emersonknapp, line 36: nitpick: i know i have also copied this - is there a simple way for us to dedupe this block?

emersonknapp (approved):
This looks good - a few stylistic notes if you want to take them

adamdbrw:
In terms of testing - perhaps querying if the pragma is set would suffice? 

`PRAGMA pragma_name;`

should return the current value of the specific pragma setting.

adamdbrw:
@Karsten1987 I would like to suggest some changes as well as discuss overall design. I don't have access rights to comment or push changes to the branch. How do we proceed? 

- Are you going to apply pending review suggestions by @emersonknapp? It would be helpful to resolve them in any way.
- I can open a PR to this branch with my suggestions and fixes. Would that PR be the right place to discuss design?
- Would you like to rebase the parent branch before we proceed?

Meanwhile, I will work on it locally

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
Karsten1987: I feel like that's a thing which should be handled upstream. Not sure how to properly address this within the context of rosbag2.

Karsten1987:
@adamdbrw you should have write access now to this repo. Feel free to simply push ontop of this branch. I've rebased and hopefully addressed Emerson's comments.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
emersonknapp: Ah, yeah, I see that we're using it in 3 different packages that each independently depend on `yaml_cpp_vendor` (`rosbag2_storage`, `rosbag2_transport`, and now `rosbag2_storage_default_plugins`). I might have suggested we have a common `rosbag2_cpp/yaml.hpp` that had this preprocessor patch, but rosbag2_cpp doesn't depend on yaml now.

Anyways, definitely not going to block anything on this. How big of a patch do you think upstream needs to fix this, do you have a sense of that?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -56,6 +69,37 @@ bool is_read_write(const rosbag2_storage::storage_interfaces::IOFlag io_flag)
...
+    } catch (const YAML::Exception & ex) {
+      throw std::runtime_error(std::string("Exception on parsing info file: ") + ex.what());
+    }
+    // poor developer's sqlinjection prevention ;-)
+    std::string invalid_characters = {"';\""};
+    auto throw_on_invalid_character = [](const auto & pragmas, const auto & invalid_characters) {
+        for (const auto & pragma_string : pragmas) {
+          auto pos = pragma_string.find_first_of(invalid_characters);
+          if (pos != std::string::npos) {
+            throw std::runtime_error(
```
emersonknapp, line 91: it might be useful to print the list of disallowed characters along with the whole pragma string, so it's more clear what they did wrong.

emersonknapp (approved):
lgtm

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
adamdbrw: I added this check in the PR to this branch #545

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12899/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7853/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12844/)

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12902/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7856/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10619/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12851/)

adamdbrw:
As I am documenting the use of storage configuration parameter, I am struggling to find a good example of a useful `PRAGMA` setting for `read:` section in the yaml.
Do we have an example of how this could be useful, considering we can't do much that affects the db and we don't actually have a mechanism to query (read) pragma settings and return them to the user (even for information purposes)?

If not, I would change the `--storage-config-file` option to be only available to `ros2 bag record` and simplify yaml section from `write/pragmas` to `pragmas`.

pijaro:
I think that `--storage-config-file` is semantically correct for both `play` and `record` so we should keep it to simplify the cli interface, even if one scenario is currently lacking utility.

emersonknapp:
> I think that `--storage-config-file` is semantically correct for both `play` and `record` so we should keep it to simplify the cli interface, even if one scenario is currently lacking utility.

I agree with this, it's more flexible to allow for future use cases

adamdbrw:
I had to force push to fix DCO issues
- Adam Dąbrowski as author vs Adam Dabrowski (ą -> a) in signoff. I have been using plain "a".
- Tomoya.Fujita vs tomoya in author/signoff mismatch (not sure how this one happened either).

Karsten1987:
> Tomoya.Fujita vs tomoya in author/signoff mismatch (not sure how this one happened either).

This change seems unrelated to your PR and should not happen. I don't appreciate changing history of committers not participating in this PR.

adamdbrw:
@Karsten1987 I assumed I was only reversing what happened as a side effect during a rebase I performed, since the original change was merged in already. Is there a suggestion how to handle this at this moment?

emersonknapp:
You should be able to solve your problem with a proper rebase on master with no merges.

One way to get the history quite clean, if you don't care about any intermediate history on this PR, is to just `git rebase --mixed origin/master` and make a single new commit that contains all changes.

adamdbrw:
@emersonknapp  thanks for the tip, I used your approach but also cherry-picked @Karsten1987 commits so we have contributions nicely separated. Now, this should be much better - I preserved authorship of commits and at the same time merged all mine into what looks to me like a proper, targeted change. I am quite pleased, but let me know if this looks good to you as well.

I am looking into Windows build unstability - it seems that the valid scenario throws an exception.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -76,6 +77,88 @@ SqliteWrapper::~SqliteWrapper()
...
+  // sqlite pragmas are assigned with either '= value' or '(value)' syntax, depending on pragma
+  const std::string pragma_assign = "=";
+  const std::string pragma_bracket = "(";
+
+  // Apply default pragmas if not overridden by user setting
+  {
+    // Used to check whether db is readable
+    const std::string schema = "schema_version";
+
+    typedef std::map<std::string, std::string> pragmas_map;
```
Karsten1987, line 93: do these pragmas have to ordered alphabetically? Or could we use a `unordered_map` here?

Karsten1987: absolute nitpick: I like to suffix these sorts of typedef's with `_t`, e.g. `pragmas_map_t`

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -76,6 +77,88 @@ SqliteWrapper::~SqliteWrapper()
...
+    pragmas_map default_pragmas = {
+      {schema, ""}
+    };
+    if (io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE) {
+      const pragmas_map write_default_pragmas = {
+        {"journal_mode", pragma_assign + "WAL"},
+        {"synchronous", pragma_assign + "NORMAL"}
```
Karsten1987, line 100: I am not sure how we decided on the default values in the end, but that's the non-optimized pragmas, right? Have we hereby decided that these are the default values?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -76,6 +77,88 @@ SqliteWrapper::~SqliteWrapper()
...
+    }
+  }
+
+  for (auto & pragma : pragmas) {
+    if (pragma.empty()) {
+      continue;
+    }
+
+    // Clean the setting from trailing characters
+    const std::string trailing_set("; \t\f\v\n\r");
```
Karsten1987, line 124: I feel like that functionality should be concerning the parser, not the application of the pragmas. I would say at this point, we should assume a syntactical correct pragma setting.

Karsten1987 (commented):
This looks principally good to me. Just a few outstanding comments inline.

Running CI on this:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12991/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7943/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10718/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12971/)

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
adamdbrw: If I recall correctly, we decided to change the defaults in a separate PR (along with changing the default split size to something else than 0 and including information about possible consequences)

pijaro:
I've inspected windows issues with `rosbag2_storage_defatult_plugins` tests. It looks like there is a problem with removing a temp dir when the storage throws an error in the ` throws_on_ivalid_pragma_in_config_file` test (btw, typo in 'invalid' 😄 ). Temp dir can’t be removed because of win32 ERROR_SHARING_VIOLATION: “The process cannot access the file because it is being used by another process.”. You can confirm that by checking return value of ` SHFileOperation(…)` in `temporary_directory_fixture.hpp` from `rosbag2_test_common` package.

Dirty workaround is to change tests order (move the test whith `EXPECT_THROW` to the end) - but still, some folders will remain in windows temp location.

adamdbrw:
This should also pass Windows tests now

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -76,6 +77,79 @@ SqliteWrapper::~SqliteWrapper()
...
+      if (found_value_assignment == 0) {
+        // Incorrect syntax, starts with = or (
+        std::stringstream errmsg;
+        errmsg << "Incorrect storage setting syntax: " << pragma;
+        throw SqliteException{errmsg.str()};
+      }
+      // Strip value assignment part, trim trailing whitespaces before = or (
+      pragma_name = pragma.substr(0, found_value_assignment);
+      const std::string trailing_set(" \t\f\v\n\r");
+      pragma_name = pragma_name.substr(0, pragma_name.find_last_not_of(trailing_set) + 1);
```
Karsten1987, line 139: again I think this should happen in the `parse_pragmas` function, rather than the application here.

Karsten1987 (commented):
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13000/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7952/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10727/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12981/)

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
adamdbrw: Here, besides some double checking (which could indeed happen in parser), we are also creating a read-value statement from a set-value one (`PRAGMA name; `from `PRAGMA name = VALUE;`).

We could parse pragma statements into name and value in the parser already, I could move the code there.

adamdbrw: This should be fixed with the last commit

adamdbrw:
Karsten1987: With all review remarks applied, are we good to merge this now?

Karsten1987:
This looks good to me. Thanks for iterating with me. Just fyi, I can't formerly approve this PR as it's actually still opened under my name, but I just believe that @emersonknapp's approval still counts ;-)

CI for completness:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13045/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7993/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10769/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13027/)

PR #502, opened 2020-08-10
resolve memory leak for serialized message

Karsten1987:
addressed #499 

as stated in https://github.com/ros2/rosbag2/issues/499#issuecomment-671629881, the change to `rclcpp::SerializedMessage` introduced this regression in which we transfer ownership, the smart pointer however doesn't cleanup the transferred memory. Specifically this change here:
https://github.com/ros2/rosbag2/pull/386/files#diff-1a63843318260f13eda07d98c90c9c66L107-L116

PR #503, opened 2020-08-11
Fix #499 memory leak

### LINKED ISSUE #499: rosbag2 memory leak
## Description
I was running `my_test_plugin` with a debug trace as described in #498 overnight; when I returned in the morning the node was killed by the kernel for consuming all of the memory in the system.

image

### Expected Behavior
No memory leaks.

## System
 - OS: U20.04
 - ROS 2 Distro: Foxy
 - Version: master (f1d1145de6762)
 - RMW: fastrtps (Default Settings)

## Additional context
I was attempting to track down why lidar and image topics stop (and apparently restart) recording. Here is the final output of the trace from #498:
image

dawonn-haval:
Inspired from https://github.com/ros2/rosbag2/pull/502.

That version seg-faulted on my system, this version does not.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -166,8 +166,22 @@ Recorder::create_subscription(
     qos,
     [this, topic_name](std::shared_ptr<rclcpp::SerializedMessage> message) {
       auto bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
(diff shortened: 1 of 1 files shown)


Karsten1987:
Can this be closed in favor of #499? This seems pretty much like a duplicate here.

dawonn:
I think you mean closed in favor of #502?

Yes, your solution looks better. I'll try and give it a test this afternoon.

Karsten1987:
closing in favor of #502

PR #506, opened 2020-08-14
Compress bag files in separate threads

### LINKED ISSUE #274: Compress files & messages aynchronously or in a separate thread
Original request  from #152:
> Compression is applied to each message asynchronously in parallel by a thread from a thread pool (with command line option to specify number of compression threads).

-----

Compression currently occurs in a single threaded workflow. To avoid potentially dropping messages from the message queue, compression should occur asynchronously or in a separate thread.

pjreed:
This offloads all compression into separate threads.  Doing so helps
to prevent rosbag2 from dropping messages due to the main thread
being busy, and it also helps to improve performance by spreading
the work across multiple CPU cores.  It uses a producer/consumer
model with a fixed number of pre-allocated threads that consume
messages or files off of an incoming queue.

This adds two new command line options:
- --compression-queue-size
   - The number of messages or files that can be sitting in the queue
     waiting for a thread to consume them.  If the queue is full, older
     messages or files will be discarded; this would lead to either
     messages being discarded or files not being compressed.
     The default value is 1, which should be sufficient if the system
     is capable of keeping up with the work load.
- --compression-threads
   - The number of threads that can be compressing data at once.  The
     default is 0, and values less than 1 will be interpreted to mean
     the number of concurrent threads supported by the current
     hardware.

Closes #274.

Distribution Statement A; OPSEC #4584

Signed-off-by: P. J. Reed <preed@swri.org>


rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
```diff
@@ -122,13 +127,15 @@ rosbag2_transport_record(PyObject * Py_UNUSED(self), PyObject * args, PyObject *
   bool include_hidden_topics = false;
   if (
     !PyArg_ParseTupleAndKeywords(
-      args, kwargs, "ssssss|bbKKKKObO", const_cast<char **>(kwlist),
+      args, kwargs, "ssssssKK|bbKKKKObO", const_cast<char **>(kwlist),
```
piraka9011, line 130: why not optional?

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -61,7 +61,7 @@ class SequentialCompressionWriterTest : public Test
 TEST_F(SequentialCompressionWriterTest, open_throws_on_bad_compression_format)
 {
   rosbag2_compression::CompressionOptions compression_options{
-    "bad_format", rosbag2_compression::CompressionMode::FILE};
+    "bad_format", rosbag2_compression::CompressionMode::FILE, 1, 4};
```
piraka9011, line 64: nit: consider naming these (ex. `default_compression_queue_size`)

rosbag2_compression/src/rosbag2_compression/zstd_compressor.cpp
```diff
@@ -26,6 +26,15 @@
 
 namespace rosbag2_compression
 {
+ZstdCompressor::ZstdCompressor()
+{
+  zstd_context_ = ZSTD_createCCtx();
```
piraka9011, line 31: I didn't know about this before but dumping here for reference (might be good as a comment too :) )

[From the manual](https://facebook.github.io/zstd/zstd_manual.html):

> When compressing many times, it is recommended to allocate a context just once, and re-use it for each successive compression operation.
> This will make workload friendlier for system's memory.
> Note : re-using context is just a speed / resource optimization. It doesn't change the compression ratio, which remains identical.
> Note 2 : In multi-threaded environments, use one different context per thread for parallel execution.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -304,18 +416,25 @@ void SequentialCompressionWriter::write(
...
+  } else {
+    std::lock_guard<std::recursive_mutex> lock(storage_mutex_);
+    storage_->write(converted_message);
   }
-
-  storage_->write(converted_message);
 }
 
-bool SequentialCompressionWriter::should_split_bagfile() const
+bool SequentialCompressionWriter::should_split_bagfile()
```
piraka9011, line 432: Why remove the `const`?

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
pjreed: Good point, I'll add consts for those.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
pjreed: Since multiple threads could be accessing `storage_` at once now, it needs to be protected by a mutex; this method needs to lock that mutex, so it has to be able to modify a class member.

rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp
pjreed: Ah, that was just an oversight, I'll fix that.

emersonknapp:
High level - should the compression work queue be independently configurable from the writer queue? It would make sense to me that compression would work on the preexisting writer queue and would need to complete before submitting that queue to the storage impl.

pjreed:
> High level - should the compression work queue be independently configurable from the writer queue?

Hmm.  I don't think I have a strong opinion one way or another on this; I can't think of any situations where you would need to configure them to be different values, but I also don't think it hurts anything to keep them separate.  Although, is the writer queue size currently configurable?  I just looked over the list of command line arguments and don't see anything that obviously maps to that.

pjreed:
@emersonknapp Just poking this again, do you think anything else needs to be done?

pjreed:
Looks like another recent merge has introduce some conflicts with these changes, I'll see if I can get it sorted out...

emersonknapp:
Sorry about that @pjreed i've been in and out of office - bad hand-off of responsibilities on my part. Change https://github.com/ros2/rosbag2/pull/526 went in to largely deduplicate the SequentialWriter vs SequentialCompressionWriter code - which will be better for all caching logic that goes in, since it was just written in two places before. That probably throws a wrench in this PR as it was written though. This could be good though to end up with something a little cleaner - the CompressionWriter can handle spinning up the thread and dispatching compression jobs to it, but not have to take care of just about anything else. 

The new structure with `get_writable_message` is not asynchronous right now, though. Maybe that should be dispatched to a thread from the SequentialWriter side, then, in which case compression may not need to know it is running in a thread? Perhaps message converters would also benefit from being put off of the transport thread, though they are less overhead than compression.

pjreed:
@emersonknapp Ok, I think this is ready to go now.

I spent a little while thinking about it, and I'm not sure that it makes a lot of sense to make the base SequentialWriter multi-threaded.  It doesn't really provide it any benefit, and if anything, the added contention from having multiple threads trying to write at once could actually decrease performance.  Even for subclasses that use threads, the way they handle offloading work into threads can vary from case to case; in the SequentialCompressionWriter's case, for example, it's actually got two different ways that it can use threads, since in per-message mode every message is compressed and written in a separate thread, while in per-file mode, the method to write messages to files is still single-threaded, but files are handed over to the compressor threads when it checks to see when it's time to split bag files.

ros2bag/ros2bag/verb/record.py
```diff
@@ -82,6 +82,16 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '--compression-format', type=str, default='', choices=['zstd'],
             help='Specify the compression format/algorithm. Default is none.'
         )
+        parser.add_argument(
```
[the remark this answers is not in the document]
pjreed: For the queue size, if your CPU is capable of keeping up with the load, it should never have more than one message waiting on it; if any threads are available, one will immediately take a message when it's pushed to the queue, so the only time more than one message could be in the queue is if every thread is busy.  If the message load is fairly constant, that means the queue is going to consistently keep growing, so you will end up dropping messages no matter how large the queue is.  If message load is bursty, having a larger queue size could help to deal with bursts without losing anything, although there's not a good way to determine what an appropriate queue size is without profiling the specific system in question.

For the number of threads, what is optimal will also vary depending on the host system.  Since compression is largely CPU-bound, the number of concurrent threads supported by the underlying hardware is a good starting point, although fewer than that will probably be sufficient on a system with a lot of cores.

Still, it might be interesting to use the performance tests to demonstrate the general effectiveness here, so I'll have to take a look at that.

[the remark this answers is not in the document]
pjreed: Ok, here's some fun statistics.  First, I modified the `writer_benchmark` so that it can take options to enable compression and adjust the number of threads & queue size.  (@emersonknapp , let me know if you think that'd be more appropriate for a different PR -- I went ahead and included it here since it's for the purpose of benchmarking multi-threaded compression anyway)  I ran tests with no compression, compression with 1 thread, and compression with 16 threads, all using this set of options:
```bash
  -p frequency:=100 \
  -p size:=1000000 \
  -p instances:=16 \
  -p max_cache_size:=100
```

And I profiled them all using Intel VTune to compare them.  The first thing that I noticed is that because there's just a single thread pushing messages into `SequentialWriter::write`, that is a bottleneck in every case.  When no compression is enabled, the writing to storage is done on the same thread, and the latency there causes it to miss a lot of messages; my test reported 13,555 missed messages and only 15.2812% were successfully recorded.
no-compression-summary
no-compression-tree
Neither of the tests with compression enabled dropped any messages, probably because the actual compression & writing are all offloaded to other threads, even when it's set to only use 1 thread.  We can also see that the load of the single compression thread roughly matches the load of the main thread that's pushing to it.
1-thread-summary
1-thread-tree
When 16 threads are running, the program doesn't actually finish any faster, since it still seems like it's bound by that main writer thread, but the load among the compressor threads looks like it's much more evenly balanced across cores.
16-thread-summary
16-thread-tree

Based on the 16-thread profile, it looks like zstd's compression is fast enough that I'm not sure we can really overload it without also having multiple threads pushing messages at once, which would probably require a more significant redesign of `writer_benchmark`, although that probably also means that just a few compression threads would be more than sufficient on anything but the most heavily taxed systems.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -92,33 +93,128 @@ void SequentialCompressionWriter::setup_compression()
...
+    "setup_compressor_threads: Starting %lu threads",
+    compression_options_.compression_threads);
+  if (compression_options_.compression_threads < 1) {
+    // This should have already been set to something reasonable prior to this
+    // point, but we'll double-check just to make sure.
+    auto hardware_threads = std::thread::hardware_concurrency();
+    compression_options_.compression_threads = hardware_threads > 0 ? hardware_threads : 1;
+  }
+  compression_is_running_ = true;
+  auto compress_fn = [&]()
```
emersonknapp, line 112: Stylistically - i think I would prefer if this compression function was not specified inside `setup_compressor_threads` - it's a little long and makes the method harder to read. Nitpick though, take or leave it.

rosbag2_compression/include/rosbag2_compression/sequential_compression_writer.hpp
```diff
@@ -104,9 +144,27 @@ class ROSBAG2_COMPRESSION_PUBLIC SequentialCompressionWriter
...
+  virtual void stop_compressor_threads();
+
 private:
   std::unique_ptr<rosbag2_compression::BaseCompressorInterface> compressor_{};
   std::unique_ptr<rosbag2_compression::CompressionFactory> compression_factory_{};
+  std::queue<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> compressor_message_queue_;
+  std::queue<std::string> compressor_file_queue_;
+  std::mutex compressor_mutex_;
+  std::vector<std::thread> compression_threads_;
+  bool compression_is_running_{false};
```
emersonknapp, line 165: should this maybe be `atomic_bool` - is there any danger of race condition in its usage? not sure exactly, probably not because it's only read by the worker threads, not written

emersonknapp (approved):
This LGTM. Thanks for the performance benchmark investigation

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
pjreed: I don't have a preference either way, so I'll go ahead and move it out.

rosbag2_compression/include/rosbag2_compression/sequential_compression_writer.hpp
pjreed: I don't think it's a problem, but it can't hurt to change it, so I'll do so.

emersonknapp:
Running this CI job
Gist: https://gist.githubusercontent.com/emersonknapp/86f84d73973ab681f5e23fe69e236da6/raw/e11f2391e32fb748418534862ada3b47c0871e82/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_compression rosbag2_performance rosbag2_transport rosbag2
TEST args: --packages-select ros2bag rosbag2_compression rosbag2_performance rosbag2_transport
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7836/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10598/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12824/)

pjreed:
It looks like another merge caused a conflict, but it should be resolved now.

pjreed:
Just pinging @emersonknapp about this again.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -74,8 +75,54 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+
+    if (message) {
+      compress_message(*compressor, message);
+
+      {
+        // Now that the message is compressed, it can be written to file using the
+        // normal method.
+        std::lock_guard<std::recursive_mutex> storage_lock(
+          storage_mutex_);
+        SequentialWriter::write(message);
```
murbanga, line 115: Sorry for suddenly jumping in here (as I am very new to project, and still getting familiar with code base):
As far as i understand, by introducing multiple threads concurrently racing for messages, would change messages sequential order in the bag.
I would even say, it became non-deterministic.
How important is it?

(please feel free to ignore me)

murbanga: wow, thanks for the clarification. if messages are sorted by timestamp at replay, it is not very important indeed.

emersonknapp: It is true that the order that messages are received is not necessarily the order in which they were sent. The behavior now is to replay them according to their timestamp - note that this timestamp is created at the time of receipt. So, by design rosbag2 will playback messages in the order they were received.

>  I'd say it's still deterministic

This isn't true. Deterministic in this point would mean that the same inputs will always cause the same order of writing to disk. Depending on how the compression threads are scheduled and given priority by the kernel, they may complete in different times and therefore have a different write queue order on subsequent runs.

It's not important though, given that playback is sorted by timestamp - just wanted to make the distinction.

> I don't think there's a need to worry about storage backends that don't exist

I should also note that we are evaluating another storage backend now https://github.com/ros2/rosbag2/pull/556. This shouldn't block us from moving forward, but it's something to note.

emersonknapp:
Left some comments on the discussion about write-order/determinism. Not blocking merging this, just adding context. I am going to try and resolve these merge conflicts now and see if I can get this merged this evening. Thanks for the patience and the important feature.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/c901a3ec7225a38da9d08b156fca658e/raw/e9efac2d5f3ca1f383e469e9578f0b328a88a70e/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_compression rosbag2_performance_writer_benchmarking rosbag2_transport rosbag2
TEST args: --packages-select ros2bag rosbag2_compression rosbag2_performance_writer_benchmarking rosbag2_transport rosbag2
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13221/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8151/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10936/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13238/)

emersonknapp:
Rerun with OSX warning fix

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13223/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8153/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10938/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13242/)

emersonknapp:
I think I've fixed the synchronization logic that the OSX test wast throwing on - we were not using the same mutex for calls to the condition variable, which is what you want to do if you are synchronizing access to a shared resource (in this case the message/file queue)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13224/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8154/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10939/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13244/)

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
chaoflow: > It's true that this will mean the order the messages are received is not necessarily the order in which they will be written to the file. I'd say it's still deterministic, it's just not something that a human would be able to realistically determine. wink
> 
> With the current default SQLite-based storage, this doesn't actually matter. When messages are read from the bag files, they are ordered based on message timestamp (https://github.com/ros2/rosbag2/blob/master/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp#L367), which is independent of the order they were written to the file. As an interesting side note, due to factors like network latency, there is no guarantee that messages from multiple different nodes will be delivered to rosbag2 in the order that they were created, so the playback order could be different from the order they were received even if everything is single-threaded.
> 
> Of course, this isn't necessarily true of any other storage backends, but I don't think there's a need to worry about storage backends that don't exist. ROS1's bag format notably played back messages in the order they were written in the file, which, in my experience, was mostly undesirable -- in practice, I almost always needed to run a script on my bag files to rewrite them so that they messages were in timestamp order.

@pjreed A follow-up to the order of messages. In your description of ROS1, are you referring to the Python or C++ implementation?

We looked into ROS1 Python [rosbag.bag](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag/src/rosbag/bag.py) and its C++ counterpart rosbag_storage [view](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag_storage/src/view.cpp) and came to the conclusion that both will merge messages from different chunks sorted by index time, but only the C++ code does sort messages within a chunk by index time, while the Python code did not:

- https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag/src/player.cpp#L184
- https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag/src/player.cpp#L315
- https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag_storage/src/view.cpp#L124
- https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag_storage/src/view.cpp#L320
- https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosbag_storage/src/view.cpp#L301

We created an MR to fix the Python code. It would be great, if you could take a look and verify our findings.

https://github.com/ros/ros_comm/pull/2089

rosbag2_performance/rosbag2_performance_writer_benchmarking/src/writer_benchmark.cpp
```diff
@@ -187,6 +209,16 @@ void WriterBenchmark::create_producers(const ProducerConfig & config)
...
+    rosbag2_compression::CompressionOptions compression_options{
+      compression_format_, rosbag2_compression::CompressionMode::MESSAGE,
+      compression_queue_size_, compression_threads_};
+
+    writer_ = std::make_unique<rosbag2_compression::SequentialCompressionWriter>(
+      compression_options);
+  } else {
+    writer_ = std::make_shared<rosbag2_cpp::writers::SequentialWriter>();
+  }
   writer_ = std::make_shared<rosbag2_cpp::writers::SequentialWriter>();
```
adamdbrw, line 222: The writer is overwritten here, meaning compression is never applied

rosbag2_performance/rosbag2_performance_writer_benchmarking/include/rosbag2_performance_writer_benchmarking/writer_benchmark.hpp
```diff
@@ -46,6 +47,11 @@ class WriterBenchmark : public rclcpp::Node
   std::vector<std::thread> producer_threads_;
   std::vector<std::unique_ptr<ByteProducer>> producers_;
   std::vector<std::shared_ptr<ByteMessageQueue>> queues_;
+
+  std::string compression_format_;
+  rosbag2_compression::CompressionMode compression_mode_;
```
adamdbrw, line 52: this seems unused

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
pjreed: Right, I was specifically thinking of the Python implementation, since that's what gets used when you run `rosbag play`. I reviewed that PR and that looks like a really good change to me, thanks!

Issue #505, opened 2020-08-14
rosbag2_py build failure in clang+libcxx nightly builds


emersonknapp:
## Description
Context https://github.com/ros2/rosbag2/pull/308#issuecomment-674079102

## Additional context
Introduced in https://github.com/ros2/rosbag2/pull/308

mabelzhang:
Just adding a note to summarize progress so far.
We got past the compiler failure by adding #pragma lines (my code is in `fix_issue_505_linux`). But now we have test failures:
https://ci.ros2.org/job/ci_linux/11953
```
>       reader.open(storage_options, converter_options)
E       RuntimeError: No storage could be initialized. Abort

../test/test_sequential_reader.py:30: RuntimeError

----------------------------- Captured stderr call -----------------------------
[ERROR] [1597438853.854357985] [rosbag2_storage]: Unable to load instance of read write interface: Could not create instance of type rosbag2_storage_plugins::SqliteStorage
[ERROR] [1597438853.854392941] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.
```

I am not sure what is causing the error. Seems a C++ exception?

A workaround, if we want to remove the `AMENT_IGNORE` file quickly, is to disable to test for Clang for now, which I've already done in the branch here and tested it makes CI green ( https://ci.ros2.org/job/ci_linux/11957/ ):
https://github.com/ros2/rosbag2/blob/fix_issue_505_linux/rosbag2_py/CMakeLists.txt#L61

Now that I look at it again, maybe it shouldn't have the GNUCXX part, then the line would be just:
```
if(BUILD_TESTING AND NOT (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
```

Issue #504, opened 2020-08-14
rosbag2_py test errors in Windows Debug nightlies


nuclearsandwich:
## Description

The rosbag2_py package doesn't seem to be building correctly in Windows debug and this is causing failures in test.
See https://ci.ros2.org/view/nightly/job/nightly_win_deb/1714/testReport/ which has two blank test failures attributed to rosbag2_py. Looking at the full log reveals excerpts like this one:<details>
```
1: =================================== ERRORS ====================================
1: ________________________ ERROR collecting test session ________________________
1: C:\Python38\lib\site-packages\pluggy\hooks.py:286: in __call__
1:     return self._hookexec(self, self.get_hookimpls(), kwargs)
1: C:\Python38\lib\site-packages\pluggy\manager.py:93: in _hookexec
1:     return self._inner_hookexec(hook, methods, kwargs)
1: C:\Python38\lib\site-packages\pluggy\manager.py:84: in <lambda>
1:     self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
1: C:\Python38\lib\site-packages\_pytest\python.py:195: in pytest_collect_file
1:     module = ihook.pytest_pycollect_makemodule(
1: C:\Python38\lib\site-packages\pluggy\hooks.py:286: in __call__
1:     return self._hookexec(self, self.get_hookimpls(), kwargs)
1: C:\Python38\lib\site-packages\pluggy\manager.py:93: in _hookexec
1:     return self._inner_hookexec(hook, methods, kwargs)
1: C:\Python38\lib\site-packages\pluggy\manager.py:84: in <lambda>
1:     self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
1: C:\ci\ws\install\Lib\site-packages\launch_testing\pytest\hooks.py:161: in pytest_pycollect_makemodule
1:     entrypoint = find_launch_test_entrypoint(path)
1: C:\ci\ws\install\Lib\site-packages\launch_testing\pytest\hooks.py:155: in find_launch_test_entrypoint
1:     return getattr(path.pyimport(), 'generate_test_description', None)
1: C:\Python38\lib\site-packages\py\_path\local.py:704: in pyimport
1:     __import__(modname)
1: C:\Python38\lib\site-packages\_pytest\assertion\rewrite.py:170: in exec_module
1:     exec(co, module.__dict__)
1: ..\test\test_sequential_reader.py:19: in <module>
1:     from rclpy.serialization import deserialize_message
[...]

PR #507, opened 2020-08-14
Revert "rosbag2_py reader and writer"

mabelzhang:
Reverts ros2/rosbag2#308 due to Nightly failures on Debug mode in Windows https://github.com/ros2/rosbag2/issues/504 and Linux with Clang https://github.com/ros2/rosbag2/issues/505

We have a solution for Linux with Clang that passes the compiler but the tests are failing.
Still need to look into Windows.


jacobperron:
What about disabling the package with an ignore file while we work to fix CI for the affected platforms?

mabelzhang:
Closing in favor of adding `AMENT_IGNORE`

emersonknapp:
I would not revert based on Clang since it's not a tier 1 platform, I'd instead just make sure to prioritize looking for a fix - though I see you closed this so I guess we're good to go :+1:

clalancette:
> I would not revert based on Clang since it's not a tier 1 platform, I'd instead just make sure to prioritize looking for a fix - though I see you closed this so I guess we're good to go +1

That's true, but I think it is also causing problems on Windows Debug: https://ci.ros2.org/view/nightly/job/nightly_win_deb/1714/#showFailuresLink

In any case, I like the idea to AMENT_IGNORE it for now, then deal with it next week.

PR #509, opened 2020-08-14
AMENT_IGNORE rosbag2_py for now

mabelzhang:
Adding `AMENT_IGNORE` to `rosbag2_py`, due to Nightly failures on Debug mode in Windows #504 and Linux with Clang #505.
We have a solution for Linux with Clang that passes the compiler but the tests are failing.
Still need to look into Windows.

We opted to add `AMENT_IGNORE` instead of reverting #308 because given that there were a large number of lines added in that PR, we thought this would make it easier to review the patch, which we will make to fix the errors.

CI is green, with Clang and build type Debug, running on meta package `rosbag2`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11961/)

(Ignore all the commits from my old fork. Will clean up the commit comments in the squashed commit when merging.)


clalancette:
The PR job is failing because you'll also need to remove this from the rosbag2 `exec_depend`: https://github.com/ros2/rosbag2/blob/038bae248e2f97e03010da7765bfbe9da0d8e0b9/rosbag2/package.xml#L16

mabelzhang:
Are you referring to the Rpr__rosbag2__ubuntu_focal_amd64? That seems to normally fail. Some of the recently merged PRs pass everything except for that one.

mabelzhang:
Looks just yellow. I think it's because some of the end to end tests are flaky.

clalancette:
> Are you referring to the Rpr__rosbag2__ubuntu_focal_amd64? That seems to normally fail. Some of the recently merged PRs pass everything except for that one.

Yeah, looking at an earlier PR job, it looks like there were 4 failures, while this one has 8.  But still, I don't see how this particular PR would cause that, so I'm still going to approve.  @emersonknapp Do you have anything to add here?  Otherwise we'll go ahead and merge.

PR #511, opened 2020-08-18
WIP broad strokes of a restructuring for the compression code

emersonknapp:
Related to https://github.com/ros2/rosbag2/pull/493

This PR is a code architecture proposal - it does not yet build, but I would like feedback before putting in the effort to complete it.

1. Create a new package `rosbag2_compression_default_plugins` - following the pattern of sqlite storage plugin
1. Move all `zstd` logic to `rosbag2_compression_default_plugins`
1. Remove `SequentialCompressionReader` and `SequentialCompressionWriter` entirely - integrate compression logic into `SequentialReader`/`SequentialWriter`, there was very little difference in code and they were diverging quickly. Treat the compressor plugin the same as the converter / storage plugins. 
    * `rosbag2_compression` now just contains the abstract interfaces for compression - `compression_factory`, `compression_options`, `base_compressor_interface`, `base_decompressor_interface`


thomas-moulard:
This globally LGTM. Personally I would rather have one package per compression plugin and make the compression default plugin a metapackage so that defining the default is just pulling a dependency to A or B?

emersonknapp:
> Personally I would rather have one package per compression plugin and make the compression default plugin a metapackage so that defining the default is just pulling a dependency to A or B?

I agree with that approach - and I think it would make sense for the sqlite storage plugin, too (but that'd outside the scope of this review)

Karsten1987:
I can second Thomas' comment. In retrospect I also wanted to have the sqlite3 plugin a package by itself, but we were copycatting what rviz did with their default plugins.

I am not sure I am agreeing with your third point. I still like the idea to have a modular reader and writer architecture. So we can easily swap out one reader for another one. What about simply inheriting from `SequentialWriter` and override the functions appropriately?

emersonknapp:
> What about simply inheriting from `SequentialWriter` and override the functions appropriately?

In practice there is very little different code, and within how the SequentialWriter is structured it makes more sense to integrate within the existing logic.

I've been thinking about it and I can't see a logical distinction between the Converter plugins and the Compression plugins - it feels like they are both "some extra thing that operates on messages or bag files before final storage". In the case of the converter plugin it only operates on messages, but a message compression plugin operates identically to the converter, in terms of how it fits into the flow of logic within the Reader/Writer.

maybe instead of exposing Converter/Compressor directly to the writer, we have an abstract Transform interface that both type conversion and compression could implement. The Reader/Writer would then have an arbitrary sequence of `MessageTransformer`s and `BagTransformer`s.

That might be over-architecting, though?

Karsten1987:
yeah, you might be right that the converter plugins are really no different from the compression. With that being said, your proposal looks good to me.

I think a general transformer interface might be tricky, because it's not clear if the combination of these actually work, e.g. first compress a message and then convert might be unfeasible.

emersonknapp:
> I think a general transformer interface might be tricky, because it's not clear if the combination of these actually work, e.g. first compress a message and then convert might be unfeasible.

Yeah, that's true. I can't think of other interfaces right now that would fit the category, so for now it makes sense to just have the two specific things, Converter and Compressor.

I'm trying to plan how to stage this without too massive of a PR. Maybe by factoring out the compressor plugin first, then migrating in the SequentialCompressionReader/Writer logic.

Should the zstd Compressor/Decompressor be separate plugins, or a single plugin with compression and decompression interfaces? Maybe it's possible to have compression schemes that are unidirectional at runtime?

emersonknapp:
Closing this as much code has changed since it was opened. I see the next steps as follows:
* pull the zstd compression implementation out into a separate package
* make that implementation into a plugin (which means a compression interface vs implementation)
* consolidate the Sequential[Compression][Reader/Writer] classes to that the compressor is a component attached to the writer, rather than being an override. This should make the double buffering storage writes and multi-threaded compression much easier to manage, rather than more difficult

PR #462, opened 2020-07-15
use a single temp dir for the test class

dirk-thomas:
Fixes #454: https://ci.ros2.org/view/nightly/job/nightly_win_rel/1613/testReport/ros2bag.test/test_record_qos_profiles/test_record_qos_profiles/

[Build Status](https://ci.ros2.org/job/ci_windows/11375/)

The diff ignoring whitespace changes is much easier to read: https://github.com/ros2/rosbag2/pull/462/files?w=1

PR #525, opened 2020-09-17
[ros2bag test_record] Gets rid of time.sleep and move to using command.wait_for_output

### LINKED ISSUE #454: ros2bag QOS test uses several sleeps on windows

jaisontj:
Tries to fix #454 

Looks to me that the original problem described at #454 was probably addressed by #466 
However, before #466, there was #462 which moved from using different temp directory per test to using one temp directory for all the tests - this PR also introduced the `time.sleep` within each test and also one within the cleanup block (which deletes the temp directory)

In this PR:
- I've moved from using `time.sleep` within each tests to using `bag_command.wait_for_output` which essentially waits for the given timeout number or seconds for the expected output. This waiting needs to be done one way or another to give the process enough time to execute and print whatever it needs to.
- I've also improved on the test conditions. For eg: `test_qos_simple` and `test_incomplete_qos_profile` were only testing if the `bag_command.output` was NOT of the format `[ERROR] [ros2bag]`. This causes a bug where the test will pass even if bag_command.output is empty ~ which can happen if the process is exited immediately i.e replace `time.sleep(3)` with a `pass`

### DIFF
--- a/ros2bag/test/test_record_qos_profiles.py
+++ b/ros2bag/test/test_record_qos_profiles.py
@@ -36,7 +36,6 @@ import pytest
 PROFILE_PATH = Path(__file__).parent / 'resources'
 TEST_NODE = 'ros2bag_record_qos_profile_test_node'
 TEST_NAMESPACE = 'ros2bag_record_qos_profile'
-ERROR_STRING = r'\[ERROR] \[ros2bag]:'
 
 
 @pytest.mark.rostest
@@ -84,12 +83,13 @@ class TestRos2BagRecord(unittest.TestCase):
         expected_string_regex = re.compile(
             r'\[rosbag2_storage]: Opened database .* for READ_WRITE')
         with self.launch_bag_command(arguments=arguments) as bag_command:
-            condition_result = bag_command.wait_for_output(
+            bag_command.wait_for_output(
                 condition=lambda output: expected_string_regex.search(output) is not None,
                 timeout=5)
-            assert condition_result, print('ros2bag CLI did not produce the expected output')
         bag_command.wait_for_shutdown(timeout=5)
         assert bag_command.terminated
+        matches = expected_string_regex.search(bag_command.output)
+        assert matches, print('ros2bag CLI did not produce the expected output')
 
     def test_incomplete_qos_profile(self):
         profile_path = PROFILE_PATH / 'incomplete_qos_profile.yaml'
@@ -99,12 +99,13 @@ class TestRos2BagRecord(unittest.TestCase):
         expected_string_regex = re.compile(
             r'\[rosbag2_storage]: Opened database .* for READ_WRITE')
         with self.launch_bag_command(arguments=arguments) as bag_command:
-            condition_result = bag_command.wait_for_output(
+            bag_command.wait_for_output(
                 condition=lambda output: expected_string_regex.search(output) is not None,
                 timeout=5)
-            assert condition_result, print('ros2bag CLI did not produce the expected output')
         bag_command.wait_for_shutdown(timeout=5)
         assert bag_command.terminated
+        matches = expected_string_regex.search(bag_command.output)
+        assert matches, print('ros2bag CLI did not produce the expected output')
 
     def test_incomplete_qos_duration(self):
         profile_path = PROFILE_PATH / 'incomplete_qos_duration.yaml'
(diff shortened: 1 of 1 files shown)


jaisontj:
Each of the tests can also be reduced to:

```python
 def test_nonexistent_qos_profile(self):
        profile_path = PROFILE_PATH / 'foobar.yaml'
        output_path = Path(self.tmpdir.name) / 'ros2bag_test_nonexistent'
        arguments = ['record', '-a', '--qos-profile-overrides-path', profile_path.as_posix(),
                     '--output', output_path.as_posix()]
        expected_string_regex = re.compile(
            r'ros2 bag record: error: argument --qos-profile-overrides-path: can\'t open')
        with self.launch_bag_command(arguments=arguments) as bag_command:
            condition_result = bag_command.wait_for_output(
                condition=lambda output: expected_string_regex.search(output) is not None,
                timeout=5)
            assert condition_result, print('ros2bag CLI did not produce the expected output')
        bag_command.wait_for_shutdown(timeout=5)
        assert bag_command.terminated
        assert bag_command.exit_code != launch_testing.asserts.EXIT_OK
```

However, this can have an edge case where `wait_for_command` gets called after the expected string is printed out.

ros2bag/test/test_record_qos_profiles.py
```diff
@@ -81,38 +80,47 @@ def test_qos_simple(self):
         output_path = Path(self.tmpdir.name) / 'ros2bag_test_basic'
         arguments = ['record', '-a', '--qos-profile-overrides-path', profile_path.as_posix(),
                      '--output', output_path.as_posix()]
+        expected_string_regex = re.compile(
+            r'\[rosbag2_storage]: Opened database .* for READ_WRITE')
         with self.launch_bag_command(arguments=arguments) as bag_command:
-            time.sleep(3)
+            bag_command.wait_for_output(
+                condition=lambda output: expected_string_regex.search(output) is not None,
+                timeout=5)
```
dabonnie, line 88: Defining the timeout would be nice since it's reused (no magic numbers).

ros2bag/test/test_record_qos_profiles.py
```diff
@@ -81,38 +80,47 @@ def test_qos_simple(self):
...
         bag_command.wait_for_shutdown(timeout=5)
         assert bag_command.terminated
-        expected_string_regex = re.compile(ERROR_STRING)
         matches = expected_string_regex.search(bag_command.output)
-        assert not matches, print('ros2bag CLI did not produce the expected output')
+        assert matches, print('ros2bag CLI did not produce the expected output')
```
dabonnie, line 92: In each case can you also print the error output?

dabonnie (changes requested):
Were you able to run the tests repeatedly by any chance?

jaisontj:
> Were you able to run the tests repeatedly by any chance?

I've been testing on windows using: `colcon test --packages-select ros2bag --retest-until-fail 25 --merge-install`

jaisontj:
Looks like the build failure on `Rpr__rosbag2__ubuntu_focal_amd64` is unrelated to this PR as the associated tests have passed: http://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/206/console#console-section-5

jaisontj:
Running on CI: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12377/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7346/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10089/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12289/)

jaisontj:
CI On windows with --retest-until-fail 25

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12294/)

jaisontj:
Do note that I was unable to get rid  of the `time.sleep` [here](https://github.com/ros2/rosbag2/blob/6e9d96659bb571d12a0d5b9ebdb293a884bed5e0/ros2bag/test/test_record_qos_profiles.py#L79) as this hack seems to be required.

dabonnie (approved):

emersonknapp (approved):
LGTM!

jikawa-az (approved):

PR #527, opened 2020-09-29
Remove some code duplication between SequentialWriter and SequentialCompressionWriter

jaisontj:
[Edited] Major changes:
- `SequentialCompressionWriter` now inherits from `SequentialWriter`
- New method `get_writeable_message` inside of `SequentialWriter` to be called to get the message to be written to storage or cache

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -193,33 +193,15 @@ void SequentialCompressionWriter::compress_message(
   compressor_->compress_serialized_bag_message(message.get());
 }
 
-void SequentialCompressionWriter::write(
+std::shared_ptr<rosbag2_storage::SerializedBagMessage>
+SequentialCompressionWriter::get_writeable_message(
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> message)
 {
-  if (!storage_) {
-    throw std::runtime_error{"Bag is not open. Call open() before writing."};
-  }
-
-  // Update the message count for the Topic.
-  ++topics_names_to_info_.at(message->topic_name).message_count;
-
-  if (should_split_bagfile()) {
-    split_bagfile();
-  }
-
-  const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>{
-    std::chrono::nanoseconds(message->time_stamp)};
-  metadata_.starting_time = std::min(metadata_.starting_time, message_timestamp);
-
-  const auto duration = message_timestamp - metadata_.starting_time;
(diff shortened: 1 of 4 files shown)


emersonknapp (approved):
This LGTM - but it looks like it will need another pass to consolidate the logic in `write`, so that the caching logic is not duplicated. In this case, it seems like the SequentialCompressionWriter isn't doing caching at all, that's how badly it's diverged.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -103,46 +99,7 @@ void SequentialCompressionWriter::open(
...
-    max_bagfile_size_ < storage_->get_minimum_split_file_size())
-  {
-    std::stringstream error;
-    error << "Invalid bag splitting size given. Please provide a value greater than " <<
-      storage_->get_minimum_split_file_size() << ". Specified value of " <<
-      storage_options.max_bagfile_size;
-    throw std::runtime_error{error.str()};
-  }
-
+  prepare_to_open(storage_options, converter_options);
```
emersonknapp, line 102: can this method be deleted by overriding `prepare_to_open`, calling the superclass, then calling `setup_compression`?

jaisontj: I don't particularly see the difference except in that `setup_compression` is maybe part of the `prepare` stage of this? I don't have a problem implementing this but am interested to know why you think this is better.

dabonnie: +1 to @emersonknapp's comment. This leaves the superclass alone and implements the compression specific details in the relevant place.

dabonnie (changes requested):
Can you please add unit tests for new methods?

jaisontj:
> Can you please add unit tests for new methods?

The new methods are non public and used internally as helper methods. Do we usually write unit tests for those?

emersonknapp:
No, generally you unit test the public interface, not the implementation details (private methods). You can of course unit test protected methods by making a test-specific subclass of the given class, but that depends on how much you consider it an interface surface vs an implementation detail.

dabonnie (approved):
Thanks for iterating, this is much cleaner!

jaisontj:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12594/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7560/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10305/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12518/)

Issue #464, opened 2020-07-17
Cache based on actual memory size (instead of message count)


emersonknapp:
## Description

And provide a nonzero default value. Remove the message count option.

Consider how this might live side-by-side with "cache by time" (maybe "whichever comes first")

## Related Issues

Highlighted by https://github.com/ros2/rosbag2/issues/435

## Completion Criteria

Writing cache clears when it has reached the provided size in bytes.

## Implementation Notes / Suggestions

Serilaized message has the buffer size, so this should be simple addition.

## Testing Notes / Suggestions

Add tests please.

emersonknapp:
Note this is related to https://github.com/ros2/rosbag2/issues/424

PR #530, opened 2020-10-02
SequentialWriter to cache by message size instead of message count

jaisontj:
This is part 1 of 2 PRs addressing  #464 
- The SequentialWriter cache now caches by message size and not by message count

As for:
> Consider how this might live side-by-side with "cache by time" (maybe "whichever comes first")

It can work the same way as split bag file by duration and split bag file by size in that, it will be "whichever comes first"

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -118,8 +118,10 @@ protected:
   std::chrono::seconds max_bagfile_duration;
 
   // Intermediate cache to write multiple messages into the storage.
-  // `max_cache_size` is the amount of messages to hold in storage before writing to disk.
+  // `max_cache_size` is the number of bytes of messages to hold in storage
+  // before writing to disk.
   uint64_t max_cache_size_;
+  uint64_t current_cache_size;
   std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
 
   // Used to track topic -> message count
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -121,7 +121,7 @@ protected:
   // `max_cache_size` is the number of bytes of messages to hold in storage
   // before writing to disk.
   uint64_t max_cache_size_;
-  uint64_t current_cache_size;
+  uint64_t current_cache_size_;
   std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> cache_;
 
   // Used to track topic -> message count
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -89,8 +89,8 @@ void SequentialWriter::open(
   max_bagfile_size_ = storage_options.max_bagfile_size;
   max_bagfile_duration = std::chrono::seconds(storage_options.max_bagfile_duration);
   max_cache_size_ = storage_options.max_cache_size;
-
   cache_.reserve(max_cache_size_);
+  current_cache_size = 0u;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
@@ -249,11 +249,13 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
     storage_->write(converted_msg);
   } else {
     cache_.push_back(converted_msg);
-    if (cache_.size() >= max_cache_size_) {
+    current_cache_size += converted_msg->serialized_data->buffer_length;
+    if (current_cache_size >= max_cache_size_) {
       storage_->write(cache_);
       // reset cache
       cache_.clear();
       cache_.reserve(max_cache_size_);
+      current_cache_size = 0u;
     }
   }
 }
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -90,7 +90,7 @@ void SequentialWriter::open(
   max_bagfile_duration = std::chrono::seconds(storage_options.max_bagfile_duration);
   max_cache_size_ = storage_options.max_cache_size;
   cache_.reserve(max_cache_size_);
-  current_cache_size = 0u;
+  current_cache_size_ = 0u;
 
   if (converter_options.output_serialization_format !=
     converter_options.input_serialization_format)
@@ -249,13 +249,13 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
     storage_->write(converted_msg);
   } else {
     cache_.push_back(converted_msg);
-    current_cache_size += converted_msg->serialized_data->buffer_length;
-    if (current_cache_size >= max_cache_size_) {
+    current_cache_size_ += converted_msg->serialized_data->buffer_length;
+    if (current_cache_size_ >= max_cache_size_) {
       storage_->write(cache_);
       // reset cache
       cache_.clear();
       cache_.reserve(max_cache_size_);
-      current_cache_size = 0u;
+      current_cache_size_ = 0u;
     }
   }
 }
--- a/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
@@ -266,15 +266,14 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
 }
 
 TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
-  const size_t counter = 1000;
+  const uint64_t counter = 1000;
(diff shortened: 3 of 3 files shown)


rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -118,8 +118,10 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   std::chrono::seconds max_bagfile_duration;
 
   // Intermediate cache to write multiple messages into the storage.
-  // `max_cache_size` is the amount of messages to hold in storage before writing to disk.
+  // `max_cache_size` is the number of bytes of messages to hold in storage
+  // before writing to disk.
   uint64_t max_cache_size_;
+  uint64_t current_cache_size;
```
dabonnie, line 124:
```suggestion
  uint64_t current_cache_size_;
```

jaisontj: Good catch, thank you!

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -249,11 +249,13 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
     storage_->write(converted_msg);
```
jaisontj, line 249: @emersonknapp we discussed cleaning this up better, however, I think this initial 0 check is required cause the cache size is set to 0 in this scenario (cause max_cache_size is 0) and so we need to handle it separately. 

Let me know what you think.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -314,8 +319,13 @@ TEST_F(SequentialWriterTest, do_not_use_cache_if_cache_size_is_zero) {
 
   std::string rmw_format = "rmw_format";
 
+  std::string msg_content = "Hello";
```
jaisontj, line 322: This test ensures that cache is not used if cache size is 0 and previously did not care about the `serialized_data` in the message. However, I thought it is better if we do test with a close to real world scenario where the message has some data as well.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -117,7 +118,7 @@ TEST_F(RecordFixture, record_end_to_end_test) {
   wrong_message->string_value = "wrong_content";
 
   auto process_handle = start_execution(
-    "ros2 bag record --output " + root_bag_path_.string() + " /test_topic");
+    "ros2 bag record --max-cache-size 0 --output " + root_bag_path_.string() + " /test_topic");
```
dabonnie, line 121: Isn't 0 the [default](https://github.com/ros2/rosbag2/blob/d460c1f837a0a3030c26c45c9f563ee966ddb568/ros2bag/ros2bag/verb/record.py#L70)?

jaisontj: Oops, I should add this to the comment and the title also. 

0 default = dont use cache and write to disk always, which isnt great for performance. So as per the issue #464 I've also changed the default to 1MB

ros2bag/ros2bag/verb/record.py
```diff
@@ -67,9 +67,11 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
                   'the bag will split at whichever threshold is reached first.'
         )
         parser.add_argument(
-            '--max-cache-size', type=int, default=0,
-            help='maximum amount of messages to hold in cache before writing to disk. '
-                 'Default it is zero, writing every message directly to disk.'
+            '--max-cache-size', type=int, default=1024*1024,
```
dabonnie, line 70: Does the default need to be changed [here](https://github.com/ros2/rosbag2/blob/19946e100dc19ae631966cd61fe1bb1d88c7080d/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp#L40) as well? There a few other places when searching for `max_cache_size`.

Karsten1987:
Thanks for picking this up. I haven't had time to fully look into this (will hopefully get to it tomorrow), but as a first glance I just want to challenge the fact that _number_ of messages is not useful. I could see that there's a use case of splitting bag files with a specific message count, e.g. recording a single image topic with the result of having 100 images per bag file.

Could we support both? I would see that each condition results in its own predicate which determines when to split. That way we could also easily support splitting by time at some point in the future.

jaisontj:
> Thanks for picking this up. I haven't had time to fully look into this (will hopefully get to it tomorrow), but as a first glance I just want to challenge the fact that _number_ of messages is not useful. I could see that there's a use case of splitting bag files with a specific message count, e.g. recording a single image topic with the result of having 100 images per bag file.
> 
> Could we support both? I would see that each condition results in its own predicate which determines when to split. That way we could also easily support splitting by time at some point in the future.

Hi @Karsten1987, this PR does not change the bag splitting functionality, rather it addresses the way messages are stored in cache before being written to a bag. Previously, it used the number of messages to hold in cache whereas bag splitting happens based on bag size or duration. I didnt see any support for split by message count (I could be wrong though). 

Let me know if I've understood you correctly.

Karsten1987:
> I haven't had time to fully look into this

That's what I meant with this 🤭 
I am sorry, I really should have looked better at this. Please ignore my comment.

jaisontj:
> > I haven't had time to fully look into this
> 
> That's what I meant with this 🤭
> I am sorry, I really should have looked better at this. Please ignore my comment.

Haha, its all good :)

ros2bag/ros2bag/verb/record.py
jaisontj: Does the data structure encapsulating options care about the default value that the command line arg should use? In this case from the comments on the structure it seems like default values are defined inside it. I am not sure about this honestly.

emersonknapp:
Just noting that this touches the same code as https://github.com/ros2/rosbag2/pull/506 for full visibility

ros2bag/ros2bag/verb/record.py
dabonnie: Perhaps yes for consistency. Otherwise, when someone inspects the code that may not have the same context as us, how does one infer the intended default unless stepping through the git blame process?

Also would note the new default in the argparse help string.

jaisontj: Was just looking into the  code and I agree with you. In that case, I am going to split this PR into 2. Leave this one to just move from using number of messages in cache to message size and another with changes to the default one.

> Also would note the new default in the argparse help string.

I have updated the help string [here](https://github.com/ros2/rosbag2/pull/530/files/0469bfbd5f31f8c161ec1bd1fec70f7d1268d67f#diff-c3b61a120c8901c73b21f179439f4010R72)? Or do you mean something else?

dabonnie: > I have updated the help string here? Or do you mean something else?

You're right: sorry, my mistake!

dabonnie (approved):
Seems reasonable with green CI

emersonknapp (approved):
LGTM, nice and minimal!

jaisontj:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12656/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7619/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10369/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12589/)

jaisontj:
Looks like there was a change in master which caused the CI to fail. Pushed again after a rebase with master. Running CI again: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12657/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7620/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10370/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12592/)

jaisontj:
The build run tried testing `rosbag2_performance_writer_benchmarking` which was not built. 

Here is the correct one with:
**build_args:** `--packages-up-to rosbag2 ros2bag rosbag2_performance_writer_benchmarking rosbag2_py`
**test_args:** `--packages-select-regex rosbag2 ros2bag`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12666/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7629/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10379/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12601/)

jikawa-az (approved):

jaisontj:
New CI Build after fixing the warning on windows: 

**build_args:** `--packages-up-to rosbag2 ros2bag rosbag2_performance_writer_benchmarking rosbag2_py`
**test_args:** `--packages-select-regex rosbag2 ros2bag`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12677/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7640/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12620/)

emersonknapp:
Instability on OSX and aarch64 is related to xmllint, which is failing on all packages currently in the nightly build - so it is definitely not related to this PR

PR #533, opened 2020-10-07
Change default cache size for sequential_writer to a non zero value

jaisontj:
**MERGE AFTER PR #530**

Currently, the default cache size for the `sequential_writer` is 0 which leads to a write to disk on every `write` call. This PR aims at changing that default value to 1MB [or any other value greater than 0]

Addresses #464 

_**A note on changing the default value**: This PR focuses on changing the default cache size when rosbag2 is invoked using the CLI. As for situations where a user might directly use the library, there is no mention of default but a note that a value of 0 will force all writes to disk._

For quick reference, these are all the files where `max_cache_size` is used/set:
- [rosbag2_py/_storage.cpp](https://github.com/ros2/rosbag2/blob/5150a736417ff6f0103f7e713dbb5001cad4949b/rosbag2_py/src/rosbag2_py/_storage.cpp#L47)
- [rosbag2_transport/rosbag2_transport_python.cpp](https://github.com/ros2/rosbag2/blob/19946e100dc19ae631966cd61fe1bb1d88c7080d/rosbag2_transport/src/rosbag2_transport/rosbag2_transport_python.cpp#L120)
- [rosbag2_cpp/storage_options.hpp](https://github.com/ros2/rosbag2/blob/19946e100dc19ae631966cd61fe1bb1d88c7080d/rosbag2_cpp/include/rosbag2_cpp/storage_options.hpp#L40)
- [rosbag2_cpp/reader.cpp](https://github.com/ros2/rosbag2/blob/546eb7c5182591a8c7f296c98ef3c06bb9aaccae/rosbag2_cpp/src/rosbag2_cpp/reader.cpp#L44)
- [ros2bag/verb/record.py](https://github.com/ros2/rosbag2/blob/d460c1f837a0a3030c26c45c9f563ee966ddb568/ros2bag/ros2bag/verb/record.py#L70)

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reader.cpp
@@ -40,8 +40,6 @@ void Reader::open(const std::string & uri)
   rosbag2_cpp::StorageOptions storage_options;
   storage_options.uri = uri;
   storage_options.storage_id = "sqlite3";
-  storage_options.max_bagfile_size = 0;  // default
-  storage_options.max_cache_size = 0;  // default
 
   rosbag2_cpp::ConverterOptions converter_options{};
   return open(storage_options, converter_options);
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
@@ -67,6 +67,7 @@ TEST_F(RecordFixture, record_end_to_end_test_with_zstd_file_compression) {
   cmd << "ros2 bag record" <<
     " --compression-mode file" <<
     " --compression-format zstd" <<
+    " --max-cache-size 0" <<
     " --output " << root_bag_path_.string() <<
     " " << topic_name;
 
@@ -117,7 +118,7 @@ TEST_F(RecordFixture, record_end_to_end_test) {
   wrong_message->string_value = "wrong_content";
 
   auto process_handle = start_execution(
-    "ros2 bag record --output " + root_bag_path_.string() + " /test_topic");
+    "ros2 bag record --max-cache-size 0 --output " + root_bag_path_.string() + " /test_topic");
   wait_for_db();
 
   pub_man_.add_publisher("/test_topic", message, expected_test_messages);
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -70,7 +70,8 @@ class RecordVerb(VerbExtension):
             '--max-cache-size', type=int, default=1024*1024,
             help='maximum size (in bytes) of messages to hold in cache before writing to disk. '
                  'Default is 1 mebibyte, everytime the cache size equals or exceeds 1MB, '
-                 'it will be written to disk'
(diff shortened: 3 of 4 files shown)


rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -67,6 +67,7 @@ TEST_F(RecordFixture, record_end_to_end_test_with_zstd_file_compression) {
   cmd << "ros2 bag record" <<
     " --compression-mode file" <<
     " --compression-format zstd" <<
+    " --max-cache-size 0" <<
```
jaisontj, line 70: These tests assume the cache size to be 0 [disk write always] and therefore this flag is now required for them to pass.

rosbag2_cpp/src/rosbag2_cpp/reader.cpp
```diff
@@ -40,8 +40,6 @@ void Reader::open(const std::string & uri)
   rosbag2_cpp::StorageOptions storage_options;
   storage_options.uri = uri;
   storage_options.storage_id = "sqlite3";
-  storage_options.max_bagfile_size = 0;  // default
```
jaisontj, line 43: I believe this assignment in unnecessary and keeps the code clean and consistent 

[Note: if we want to keep this then storage_options.max_bagfile_duration = 0, is something we should add to this.]

emersonknapp: Yes - I think you are right. It might also make sense to move the "sqlite3" default value to the struct definition as well. No need to do that right now though

emersonknapp (approved):
I assume this depends on https://github.com/ros2/rosbag2/pull/533 - you should probably mark it as such in the PR description so that we remember to merge this only after the other.

jaisontj:
> I assume this depends on #533 - you should probably mark it as such in the PR description so that we remember to merge this only after the other.

You're right. Done.

emersonknapp:
This would change the default cache to one million _messages_ if it was merged before #533 - I think that is probably an unreasonable amount, therefore I'd call it a dependency. E.g. you are storing an IMU message, this could easily make the default cache end up something like >300MB

jaisontj:
> This would change the default cache to one million _messages_ if it was merged before #533 - I think that is probably an unreasonable amount, therefore I'd call it a dependency. E.g. you are storing an IMU message, this could easily make the default cache end up something like >300MB

True, agreed. That makes sense.

dabonnie (approved):

jaisontj:
Running CI
**build_args:** `--packages-up-to rosbag2 ros2bag rosbag2_performance_writer_benchmarking rosbag2_py`
**test_args:** `--packages-select-regex rosbag2 ros2bag`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12683/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7647/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10399/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12627/)

jaisontj:
Looks like xmllint is failing on all packages currently in the nightly build - so the instability is not related to this PR.

dabonnie (approved):

PR #534, opened 2020-10-07
Removed duplicated code in record

jaisontj:
There is some duplicated code in rosbag2/record.py that can be cleaned up.

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -93,8 +93,12 @@ class RecordVerb(VerbExtension):
         self._subparser = parser
 
     def main(self, *, args):  # noqa: D102
+        # both all and topics cannot be true
         if args.all and args.topics:
             return print_error('Invalid choice: Can not specify topics and -a at the same time.')
+        # both all and topics cannot be false
+        if not(args.all or (args.topics and len(args.topics) > 0)):
+            return self._subparser.print_help()
 
         uri = args.output or datetime.datetime.now().strftime('rosbag2_%Y_%m_%d-%H_%M_%S')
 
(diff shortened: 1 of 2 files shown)


emersonknapp (approved):
Nice!

ros2bag/ros2bag/verb/record.py
```diff
@@ -114,30 +114,7 @@ def main(self, *, args):  # noqa: D102
             except (InvalidQoSProfileException, ValueError) as e:
                 return print_error(str(e))
```
dabonnie, line 116: Nit: add comment assuming that both are not true given the check on line 94.

dabonnie (approved):

ros2bag/ros2bag/verb/record.py
jaisontj: Was able to get rid of the if condition above record call entirely by placing the check at the top of the function.

ros2bag/ros2bag/verb/record.py
```diff
@@ -91,8 +91,12 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         self._subparser = parser
 
     def main(self, *, args):  # noqa: D102
+        # both all and topics cannot be true
         if args.all and args.topics:
             return print_error('Invalid choice: Can not specify topics and -a at the same time.')
+        # both all and topics cannot be false
+        if not(args.all or (args.topics and len(args.topics) > 0)):
+            return self._subparser.print_help()
```
dabonnie, line 99: Nit

```suggestion
            return print_error('Invalid choice: Must specify topic(s) or -a')
```

This way the error is clear instead of generally printing the help function.

dabonnie (approved):

jaisontj:
The instability looks unrelated to the PR and related to the failing xmllint tests that are happening in the nightly as well.

Issue #537, opened 2020-10-08
New warnings in Windows packaging jobs


nuclearsandwich:
## Description

New warnings when building on Windows in RelWithDebInfo configuration appeared in the 2020-10-06 nightly packaging jobs.

```
Building Custom Rule C:/ci/ws/src/ros2/rosbag2/rosbag2_py/CMakeLists.txt
  _reader.cpp
LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification [C:\ci\ws\build\rosbag2_py\_reader.vcxproj]
     Creating library C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_reader.lib and object C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_reader.exp
  Generating code
  Finished generating code
  _reader.vcxproj -> C:\ci\ws\build\rosbag2_py\RelWithDebInfo\_reader.cp38-win_amd64.pyd
  Building Custom Rule C:/ci/ws/src/ros2/rosbag2/rosbag2_py/CMakeLists.txt
  _storage.cpp
LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification [C:\ci\ws\build\rosbag2_py\_storage.vcxproj]
     Creating library C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_storage.lib and object C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_storage.exp
  Generating code
  Finished generating code
  _storage.vcxproj -> C:\ci\ws\build\rosbag2_py\RelWithDebInfo\_storage.cp38-win_amd64.pyd
  Building Custom Rule C:/ci/ws/src/ros2/rosbag2/rosbag2_py/CMakeLists.txt
  _writer.cpp
LINK : warning LNK4075: ignoring '/INCREMENTAL' due to '/LTCG' specification [C:\ci\ws\build\rosbag2_py\_writer.vcxproj]
     Creating library C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_writer.lib and object C:/ci/ws/build/rosbag2_py/RelWithDebInfo/_writer.exp
  Generating code
  Finished generating code
```

## To Reproduce

Build a packaging job with https://ci.ros2.org/job/ci_packaging_windows/
You can also see the error present in the nightly packaging jobs beginning with [packaging_windows #1895](https://ci.ros2.org/view/packaging/job/packaging_windows/1895/)

[...]

PR #538, opened 2020-10-09
 Workaround pybind11 bug on Windows when CMAKE_BUILD_TYPE=RelWithDebInfo

ivanpauno:
Another workaround for a pybind11 bug ...

I should start sending patches upstream (for this and https://github.com/ros2/rosbag2/issues/504).
When that gets included in a new pybind11 release, we can update our vendor package and get rid of these workarounds (I will open an issue to track that).

Though it failed in a packaging build, only `-DCMAKE_BUILD_TYPE=RelWithDebInfo` is needed to reproduce the issue:

* Windows RelWithDebInfo [Build Status](http://ci.ros2.org/job/ci_windows/12618/)

~TODO: This still doesn't pass, I have to check more closely in a Windows VM.~ [CI is passing now](https://github.com/ros2/rosbag2/pull/538#issuecomment-706382379).

Fixes https://github.com/ros2/rosbag2/issues/537.

### DIFF
--- a/rosbag2_py/CMakeLists.txt
+++ b/rosbag2_py/CMakeLists.txt
@@ -54,13 +54,13 @@ function(clean_windows_flags target)
   # - https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=vs-2019
 
   if(MSVC AND "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
-    get_target_property(target_link_flags ${target} LINK_FLAGS)
-    string(REPLACE "-LTCG" "" target_link_flags ${target_link_flags})
-    set_target_properties(${target} PROPERTIES LINK_FLAGS ${target_link_flags})
+    get_target_property(target_link_libraries ${target} LINK_LIBRARIES)
+    string(REPLACE "-LTCG" "" target_link_libraries ${target_link_libraries})
+    set_target_properties(${target} PROPERTIES LINK_LIBRARIES ${target_link_libraries})
 
-    get_target_property(target_compile_flags ${target} COMPILE_FLAGS)
-    string(REPLACE "/GL" "" target_compile_flags ${target_compile_flags})
-    set_target_properties(${target} PROPERTIES COMPILE_FLAGS ${target_compile_flags})
+    get_target_property(target_compile_options ${target} COMPILE_OPTIONS)
+    string(REPLACE "/GL" "" target_compile_options ${target_compile_options})
+    set_target_properties(${target} PROPERTIES COMPILE_OPTIONS ${target_compile_options})
   endif()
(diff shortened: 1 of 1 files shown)


ivanpauno:
This should pass now:
* Windows RelWithDebInfo [Build Status](http://ci.ros2.org/job/ci_windows/12629/)

ivanpauno:
This patch is terrible bad but:

- `Debug` and `RelWithDebInfo` build types set `/INCREMENTAL` link flag on Windows (see [here](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-MSVC.cmake#L292)).
- pybind11 is forcing `/LTCG` and `/GL` when build type is `RelWithDebInfo`, it shouldn't be doing that (see link in the diff).

Without a proper fix in `pybind11` I don't see another way to fix this.
Except ignoring the warning, but in that case MSVC will use "non standard" optimization options for a `RelWithDebInfo` build.

nuclearsandwich:
I'm not really in a position to evaluate the fitness here but I appreciate the head's up. Hopefully it's something we can push upstream in short order.

ivanpauno:
I removed the `ros2/tooling-reviewers`review request by accident, I've added it back.

jacobperron:
It would be good to fix this upstream, @ivanpauno are you able open a PR and reference it in this change so that we can revert it after an upstream release?

rosbag2_py/CMakeLists.txt
```diff
@@ -40,6 +40,30 @@ if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(PYTHON_LIBRARIES "${PYTHON_DEBUG_LIBRARIES}")
 endif()
 
+function(clean_windows_flags target)
+  # More hacks to avoid pybind11 issues, similar to (but not the same)
```
jacobperron, line 44: It'd be nice to have a TODO here to remove after a fix upstream.

ivanpauno: Done https://github.com/ros2/rosbag2/pull/538/commits/354c3abc8b35ef23286e565fc0c9ce77d3d7d1d7.

jacobperron (approved):
Thanks, this LGTM.

Looks like this will be fixed in the next pybind11 release (2.6.0). I'll keep my eyes open for it and update our vendor package when it's out.

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12755/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7713/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10465/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12698/)

Issue #395, opened 2020-04-28
DB3 bag is irrecoverable if metadata.yaml doesn't exist


dsharma2701:
## Description
ros2 db3 bags fails to load or get info on if the metadata yaml is missing

### Expected Behavior
Expecting it to have the ability to generate it offline in case of any corruption during saving the bag

### Actual Behavior
If accompanying metadata yaml doesn't exist with the db3, it fails to load/get info on.
[ERROR] [1588037968.719797386] [rosbag2_storage]: Could not open '/root/Downloads/jobq_bags.db3' with 'sqlite3'. Error: Failed to setup storage. Error: Error when processing SQL statement. SQLite error (11): database disk image is malformed

## To Reproduce
** Steps to reproduce the behavior, e.g.
1. Use any ros2 bag and rename metadata yaml to tmp
2. Run ros2 bag info <bag-name> 

## System (please complete the following information)
 - OS: Ubuntu Bionic
 - ROS 2 Distro: dashing
 - Version: 

## Additional context
Would love to find a way to recover the bag as its 50 GB worth of data.

emersonknapp:
can you please include copy/pasted output of the commands, in "Actual Behavior"?

Karsten1987:
@dsharma2701 thanks for bringing this up. I've thought about this a couple of times, especially in the case that if a `ctrl-c` command fails to stop a recording - which is does unfortunately quite reliably in the end to end tests on windows - the recorded data is still available in the `.db3`, yet _unusable_ due to the missing metadata file.

That is indeed a problem, and we should have something like `rosbag reindex` which is there in ROS1. I honestly don't think it's that hard to achieve, given that almost all information is available in the database. There is this function available, which we luckily didn't remove from the interface: https://github.com/ros2/rosbag2/blob/33cf70bbd3c166e381f40a6cbf3f7fcce244d2bf/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp#L72

That would extract the metadata for the stored messages. I think components like which compression algorithm was used is not part of it, but could then be added manually.

I think this could be a good starting point for a verb like `ros2 bag reindex <path/to/database.db3>` or alternatively `ros2 bag reindex <path/to/manual.yaml>` where the manual yaml file specifies some preliminary data, such as the relative file paths and the `reindex` function would then kind of fill in the blanks.

jhdcs:
[...]

PR #539, opened 2020-10-15
Introduce ros2 bag reindex

jhdcs:
This pull request addresses issue #395.

A new verb, `reindex` has been added, which attempts to reconstruct a `metadata.yaml` file from the sqlite3 databases found within the bag.

Functionality allowing different forms of `reindex` for bag types has been added as well.

Readme has been updated to indicate new verb, as well as how to use it.

Unit tests for `reindex` have been added to `rosbag2_tests`.


rosbag2_cpp/CMakeLists.txt
```diff
@@ -33,6 +33,14 @@ if(NOT DISABLE_SANITIZERS AND CMAKE_COMPILER_IS_GNUCXX)
   endif()
 endif()
 
+# Set Boost options
+set(Boost_USE_STATIC_LIBS OFF)
+set(Boost_USE_MULTITHREADED OFF)
+set(Boost_USE_STATIC_RUNTIME OFF)
```
Karsten1987, line 39: We currently don't have any dependencies on boost, that is the buildfarm is not provisioned with it and I would love to avoid boost if possible.
Is there are way we can workaround it?

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,73 @@
...
+class BaseReindexerInterface;
+}  // namespace reindexer_interfaces
+
+/**
+ * The reindexer attempts to construct a metadata file based on information contained in the bag.
+ */
+class ROSBAG2_CPP_PUBLIC Reindexer final
+{
+public:
+  explicit Reindexer(std::unique_ptr<reindexer_interfaces::BaseReindexerInterface> reindexer_impl);
```
Karsten1987, line 52: While I appreciate that you followed the style of this package, I am wondering whether it makes sense to apply here. The `pimpl` pattern for the readers and writers are meant to allow more flexibility and interchangeability for various implementations. In terms of reindexing, can you imagine there being more than one way to reindex?
I personally don't think so and would see that the code base can become a bit smaller by removing the interface class etc.

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,265 @@
...
+  std::vector<std::string> output;
+  for (auto & p_ : boost::filesystem::directory_iterator(base_folder)) {
+    // We are ONLY interested in database files
+    if (p_.path().extension() != ".db3") {
+      continue;
+    }
+
+    output.emplace_back(p_.path().c_str());
+    std::cout << "Found path: " << p_.path().c_str() << "\n";
+  }
```
Karsten1987, line 127: is there any way we can get rid of boost here? I think the iterator is the only place where you currently rely on boost.
Maybe we could come up with a solution with rcpputils.

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,265 @@
...
+  reset();
+}
+
+void SequentialReindexer::reset()
+{
+  if (storage_) {
+    storage_.reset();
+  }
+}
+
```
Karsten1987, line 90:
```suggestion
```

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,265 @@
...
+    topics_metadata_.push_back(topic_information.topic_metadata);
+  }
+}
+
+void SequentialReindexer::init_metadata(const std::vector<std::string> & files)
+{
+  metadata_ = rosbag2_storage::BagMetadata{};
+
+  // This reindexer will only work on SQLite files, so this can't change
+  metadata_.storage_identifier = "sqlite3";
```
Karsten1987, line 165: that has to be more generic. What happens if I try to reindex a ros1 bagfile? Or some custom storage?

Karsten1987 (commented):
Thanks for the PR, a couple of remarks though: 

* The reindexer has to be more generic. I found a couple of places where `.db3` and `sqlite3` is hardcoded. Does it make more sense to have that reindex functionality live within the sqlite3 plugin then?
* As mentioned inline, while I appreciate that you follow the style, I don't think we have to introduce interfaces for the reindexer. I don't think we need that flexibility on that level.
* Boost is currently kind of a show-stopper. I think we have to come up with a rcpputils solution which can iterate over files wo/ boost.

rosbag2_cpp/CMakeLists.txt
Karsten1987: Maybe that could help? https://stackoverflow.com/a/612112/4583130

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
jhdcs: Ah, I see. I fundamentally misunderstood the purpose of the interface, then. I thought that the interface was so that you could re-implement the `reindexer` in the language of your choice.

This makes my life SO much easier!

rosbag2_cpp/CMakeLists.txt
jhdcs: The reason I went with boost is actually that post... I'm not thrilled about using it either, but it was the only cross-platform way I could think of doing a directory search.

If it isn't too icky, I could try to implement the per-platform directory searches, and use a preprocessing directive to conditionally compile whichever one works for the OS that we're on...

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
jhdcs: Yeah, that's the only place I use it. I hate having to use a sledgehammer on a thumbtack, lol.

Scanning through `rcpputils::filesystem_helper` to see if there's a more... _Elegant_ solution.

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
jhdcs: Yeah, I thought `sequential_reindexer` was supposed to be the reindexer that worked on the `sqlite` files. Not sure why. I guess I'm derpy like that.

jhdcs:
Made a few comments on your suggestions, thank you for those!

I did have one question that's been nagging me, though: Since it's theoretically possible to extract all the data that a `metadata.yaml` file provides, why is the file needed? Is there another format that can't store all the information needed to read it within itself?

Karsten1987:
> I did have one question that's been nagging me, though: Since it's theoretically possible to extract all the data that a metadata.yaml file provides, why is the file needed?

The idea here is that the actual data payload can be huge compared to the metadata/index and distributed over multiple files. With a dedicated metadata file, we have all that data at hand without having to iterate over the payload.

jhdcs:
Ok. Well, the good news is I've gotten the Linux/Unix reindexer working again. The bad news is I don't really have a way of testing the Windows version right now, since ROS 2 refuses to build itself on my Windows VM...

Anyhoo, now trying to figure out the best way to transplant the reindexer into the `sqlite3` plugin. If I'm reading this correctly, the `sqlite3` plugin is a `StorageFactoryImpl`, correct? I thought that was a bit lower-level than what the reindexer does... Is there another spot that I'm missing?

Otherwise, I'll try to wire up a new `reindex` function in the `StorageFactoryImpl`, and propagate the changes along. Fingers crossed...

Karsten1987:
if you want to make your life a bit easier, I guess you could directly plant that reindex logic in a python script within ros2bag. There's a python API for sqlite3 and all it needed would be a bit of logic to fail gracefully when that script is not available for other storage plugins than sqlite3.

jhdcs:
That... Actually might fix a LOT of problems - We wouldn't need to worry about two code paths for finding out the files within a folder in either Windows or Linux, for example. And I've already updated the `get_metadata` function to return QOS profiles, so as long as I can call it from Python, we're gold.

On a slightly different subject; as you've probably noticed, I realized I was using the wrong license blurb at the top of the reindex files I've added. However, the copyright lint doesn't seem to like what my employers wanted. Is there something I can do to fix this?

Karsten1987:
> On a slightly different subject; as you've probably noticed, I realized I was using the wrong license blurb at the top of the reindex files I've added. However, the copyright lint doesn't seem to like what my employers wanted. Is there something I can do to fix this?

I would recommend you'd use the regular license text and change the copyright to reflect your company name. That should pass the linters - as you can see Open Robotics, Amazon and Bosch have contributions as such. For all other legal hints, I would put them in a separate doc block below the copyright header. If that's okay with your employer.

Karsten1987:
@jhdcs you might want to have a look at https://github.com/ros2/rcutils/pull/249. It's still in a draft but might be the PR you're looking for the iterate through a directory.

jhdcs:
Got the python version _probably_ working, albeit unlinted. I need to write unit tests for it, which means I need to find some test databases. Should I just copy them over from the other test packages?

Also, I'm not sure how to get this to work with compression yet...

jhdcs:
I've gotten the reindex operation working for uncompressed and file-compressed SQLite bags, now. Should I squash the commits? There appears to be a commit somewhere in the rebase that's confusing DCO.

jhdcs:
Figured out why the test is failing - in order to support the `zstd` compression scheme, a python package had to be installed via pip. The particular package is called "zstandard".

If we don't want to add new python packages to the workspace, I can comment out the areas of the code that reference it. However, reindexing operations will not be able to support `zstd` compressed bags.

jhdcs:
Reindexing for message-compressed files has been implemented. This code is ready for review.

However, we will need to figure out how we want to handle the inclusion of the `zstandard` package.

jhdcs:
May I please have another review on this? I'm not sure why the Rpr__rosbag2__ubuntu_focal_amd64 check is failing.

jhdcs:
Okay, got some time to look into it. The Rpr__rosbag2__ubuntu_focal_amd64 check is failing because the  `zstandard` python package is not included. Which means that the file can't be opened/recompressed, and therefore the test can't find the new `metadata.yaml` file, and the whole thing falls down.

Do you have any recommendations for solving this? Without that `zstandard` package, we're locked out of working with compressed files...

rosbag2_cpp/CMakeLists.txt
```diff
@@ -57,17 +57,19 @@ add_library(${PROJECT_NAME} SHARED
...
-  pluginlib
-  rcpputils
-  rcutils
-  rosbag2_storage
-  rosidl_runtime_c
-  rosidl_runtime_cpp
-  rosidl_typesupport_cpp
-  rosidl_typesupport_introspection_cpp
+  PUBLIC
+    ament_index_cpp
```
emersonknapp, line 61: nit: don't add extra indent, removes unrelated whitespace changes

rosbag2_cpp/CMakeLists.txt
```diff
@@ -57,17 +57,19 @@ add_library(${PROJECT_NAME} SHARED
...
+    rcpputils
+    rcutils
+    rosbag2_storage
+    rosidl_runtime_c
+    rosidl_runtime_cpp
+    rosidl_typesupport_cpp
+    rosidl_typesupport_introspection_cpp
 )
 
+
```
emersonknapp, line 72: nit: remove unrelated whitespace change

rosbag2_cpp/package.xml
```diff
@@ -10,7 +10,10 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>libboost-filesystem-dev</build_depend>
```
emersonknapp, line 13: this may be left from an old version of `master`? You'll probably want to revert all the changes from `rosbag2_cpp`, `rosbag2_storage_default_plugins`, `rosbag2_tests`, and `rosbag2_transport` - they don't seem related to your PR.

Additionally, I've noticed you checked in a `__pycache__` directory in `rosbag2_py` - that should be removed as well.

ros2bag/ros2bag/verb/reindex.py
```diff
@@ -0,0 +1,65 @@
...
+class ReindexVerb(VerbExtension):
+    """Generate metadata from a bag."""
+
+    def add_arguments(self, parser, cli_name):  # noqa: D102
+        parser.add_argument(
+            'bag_file', type=check_path_exists, help='Bag file to reindex')
+        parser.add_argument(
+            '-s', '--storage-id', default='sqlite3',
+            help="storage identifier to be used, defaults to 'sqlite3'")
+        parser.add_argument(
```
emersonknapp, line 37: Both `compression-format` and `compression-mode` choices shouldn't be duplicated. Perhaps they should be factored out of `record.py` and this file into a common place - or you could at least pull the values out of `record.py`'s `add_argument` calls into a global level variable for that module, and `from ros2bag.verb.record import COMPRESSION_MODES` - something like that. The duplication will inevitable get desynchronized in the future

ros2bag/ros2bag/reindexer/reindex_sqlite.py
```diff
@@ -0,0 +1,178 @@
...
+    decompressed_wal.unlink(True)
+
+    return return_val
+
+
+def get_metadata(db_file: pathlib.Path,
+                 compression_fmt: Literal['', 'zstd'],
+                 compression_mode: Literal['', 'none', 'file', 'message']) -> DBMetadata:
+    # Handle compression
+    if compression_fmt != '':
```
emersonknapp, line 110:
```suggestion
    if compression_fmt:
```

handles both `None` and `''`. I would recommend using `None` rather than empty-string for optional values.

Also - this block would be much more readable as

```
if compression_fmt in ['none', 'message']:
  # open normally
elif compression_fmt == 'file':
  # open compressed
else:
  raise ValueError('unknown compression mode')
```

The "by process of elimintation !=" pattern is a lot less readable, and includes some duplicated lines.

ros2bag/ros2bag/reindexer/reindex_base.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Literal, Optional
+
+from ros2bag.api import print_error
+
+from . import reindex_sqlite
+
+
+def reindex(uri: str,
+            storage_id: str,
+            compression_fmt: Literal['', 'zstd'],
```
emersonknapp, line 32: I don't think this is really a `Literal` value - it'll require shotgun surgery if a new compression format is added. You should probably take an `Optional[str]` (for both fmt and mode) - the underlying logic can decide whether or not it can handle this value - overkill and brittle to handle it in the type system.

ros2bag/ros2bag/reindexer/bag_metadata.py
```diff
@@ -0,0 +1,304 @@
...
+    return {'topic_metadata': metadata_topics, 'message_count': topic['tm_message_count']}
+
+
+class RosbagYamlDumper(yaml.Dumper):
+
+    def increase_indent(self, flow=False, indentless=False):
+        return super(RosbagYamlDumper, self).increase_indent(flow, False)
+
+
+class MetadataWriter:
```
emersonknapp, line 64: Considering we already have a `MetadataWriter` in C++, this duplicated code will be prone to rot if the metadata format changes at all - this should be a call into rosbag2_storage `MetadataIO` for the logic in this file

emersonknapp (changes requested):
Hey there - sorry I have not been involved on this PR to date!

I don't think we want to install the python-based `zstandard` package - because it may not match up with the version being used by the C++ code. Another reason is that we may support other compression implementations, and we don't want equivalent python bindings for every one of them. The real solution around this is probably to add an endpoint to the `rosbag2_py` API to decompress bagfiles, having it choose the correct compression implementation under the hood to do so. 

I think as a first step with this feature it would be acceptable to support reindexing uncompressed bags and print out an error message when trying to work on a compressed bag - we could have a followup ticket for `Enable reindex on compressed bagfiles`

If I am reading correctly, this doesn't need to actually worry about compressed messages, right? Since the type data is available in the sqlite table uncompressed.

jhdcs:
If you are concerned about duplicating functions that are within C++, should I just go back to the C++ version of the code I had originally pushed? Perhaps with a little more logic on the Python side to grab the relative file paths, and thus avoid having to use Boost?

emersonknapp:
I'm concerned with code duplication in general. I think when Karsten suggested a Python script, I don't think he intended to suggest that we rewrite existing logic in Python - just that a Python top level script would be easier to manage than fully-C++ logic.

We definitely do not want to introduce a boost dependency, we intentionally removed it from the ros2 core and rosbag2 in most places. For path manipulation, the C++ stdlib handles a lot of that now, and there are a good number of shims for any missing functions in https://github.com/ros2/rcpputils

I personally think that expanding the `rosbag2_py` API to expose metadata operations would be a desirable feature in and of itself, and could be a standalone PR that would enable this feature

ros2bag/ros2bag/reindexer/bag_metadata.py
danthony06: I don't think I've seen ROS code in the past calling C++ libraries from within Python code. Is there precedent for this, or is this something that will change going forward?

danthony06 (commented):
I think that overall the changes look good, and it would be nice to get this functionality merged into future releases.

jhdcs:
> I'm concerned with code duplication in general. I think when Karsten suggested a Python script, I don't think he intended to suggest that we rewrite existing logic in Python - just that a Python top level script would be easier to manage than fully-C++ logic.
> 
> We definitely do not want to introduce a boost dependency, we intentionally removed it from the ros2 core and rosbag2 in most places. For path manipulation, the C++ stdlib handles a lot of that now, and there are a good number of shims for any missing functions in https://github.com/ros2/rcpputils
> 
> I personally think that expanding the `rosbag2_py` API to expose metadata operations would be a desirable feature in and of itself, and could be a standalone PR that would enable this feature

I could revert all of the python changes and go back to the C++ version. However, that still leaves us with the problem that I can't query what files are in a directory. `rccputils` doesn't have that functionality in its `filesystem_helper.hpp` file, and `std::filesystem` doesn't exist in C++14. It does in C++17, but I'm not sure if or when `ros2` will be switching to that version.

The only other possibility I can think of that doesn't add a new dependency is to figure out how to have Python send a list of strings to C++, and have it converted into a `std::vector`. That way Python can query what files are in the directory, and give it to C++ for further processing. But I have no idea how to do that...

clalancette:
> I could revert all of the python changes and go back to the C++ version. However, that still leaves us with the problem that I can't query what files are in a directory. `rccputils` doesn't have that functionality in its `filesystem_helper.hpp` file, and `std::filesystem` doesn't exist in C++14. It does in C++17, but I'm not sure if or when `ros2` will be switching to that version.

We're likely to switch to C++17 for Galactic; see https://github.com/ros-infrastructure/rep/pull/291 .  In the meantime, you could also add the functionality to `rcpputils::fs` in a `std::filesystem` compliant way.

jhdcs:
> > I could revert all of the python changes and go back to the C++ version. However, that still leaves us with the problem that I can't query what files are in a directory. `rccputils` doesn't have that functionality in its `filesystem_helper.hpp` file, and `std::filesystem` doesn't exist in C++14. It does in C++17, but I'm not sure if or when `ros2` will be switching to that version.
> 
> We're likely to switch to C++17 for Galactic; see [ros-infrastructure/rep#291](https://github.com/ros-infrastructure/rep/pull/291) . In the meantime, you could also add the functionality to `rcpputils::fs` in a `std::filesystem` compliant way.

@emersonknapp If C++17 is likely to happen in Galactic, are we allowed to use its features in this pull request? If not, I can look into trying to figure out a way to replicate the functionality for iterating over directories in `rcpputils::fs`.

emersonknapp:
> @emersonknapp If C++17 is likely to happen in Galactic, are we allowed to use its features in this pull request?

I'd like to wait on the final decision before upgrading the target C++ version for rosbag2. Once it is finalized then it would be fine - the limitation being that it would be difficult/impossible to backport any C++17 code to Foxy. 

If we do decide to upgrade language versions, I'd prefer we did that CMakeLists upgrade as a standalone PR.

> That way Python can query what files are in the directory, and give it to C++ for further processing. But I have no idea how to do that...

Just wanted to address this - pybind11 (used for rosbag2_py layer) can do a lot, and passing lists is quite straightforward - see https://pybind11.readthedocs.io/en/stable/advanced/cast/overview.html - you can simply declare a function that takes a vector argument and pybind will automatically handle conversions to the bound function when called from python.

clalancette:
> I'd like to wait on the final decision before upgrading the target C++ version for rosbag2. Once it is finalized then it would be fine - the limitation being that it would be difficult/impossible to backport any C++17 code to Foxy.
> 
> If we do decide to upgrade language versions, I'd prefer we did that CMakeLists upgrade as a standalone PR.

As an FYI, we've now merged https://github.com/ros-infrastructure/rep/pull/291 , so Galactic switching to C++17 is now official.  How and when you switch is entirely up to you, but it is an option for the master branch now.

jhdcs:
I've been looking into adding compression for the reindexer, and I seem to be bumping into a wall. I can't find a clean way to introduce compression into the reindexer that does not duplicate a lot of code.

I took a look at the PR that created the compression functionality, and noticed this comment: https://github.com/ros2/rosbag2/pull/258#pullrequestreview-344234933. Do you think that now would be a good time to look at refactoring the compression functionality? If we changed it so that it could be called by the reader/writer/reindexer, rather than the compressing/decompressing versions being a copy of the base classes, I would be able to implement compression for the reindexer much more easily.

emersonknapp:
Yes - I am currently in the process of refactoring the compression code pretty seriously. Like I said in an earlier comment on this review, though, I think it's perfectly acceptable to partially implement this feature (not supporting compressed bags for a first pass). 

As a note - I am finding this PR difficult to review due to the sheer size of it - and would prefer smaller PRs that implement individual atomic functionalities.

For example - if you are introducing a `SequentialReindexer` class that can do the reindexing functionality - then a PR introducing just that class, with tests and a usable C++ API would be a great PR that I'll be much more likely to be able to review effectively. I'm looking for PRs that are self-contained, demonstrably correct, and do one usable thing well, even if it doesn't do everything you would want it to do. That's something that we could really make meaningful progress on.

ros-discourse:

https://discourse.ros.org/t/tooling-wg-breakout-meeting-2021-01-22-rosbag2-reindexing/18489/3

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,346 @@
...
+      // We are ONLY interested in database files
+      if (temp_path.extension().string() != ".db3") {
+        continue;
+      }
+
+      output.emplace_back(temp_path);
+    }
+    closedir(dirp);
+  }
+  #endif
```
Karsten1987, line 200: We've recently added code which supersedes this: https://github.com/ros2/rcutils/pull/323

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,346 @@
...
+#ifdef WIN32
+// Import windows filesystem functionality
+#include <windows.h>
+#include <tchar.h>
+#include <stdio.h>
+#else
+// We're on a UNIX system. Import their filesystem stuff instead
+#include <dirent.h>
+#endif
+
```
Karsten1987, line 49:
```suggestion
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,49 @@
...
+: reindexer_impl(std::move(reindexer_impl))
+{}
+
+Reindexer::~Reindexer() {}
+
+void Reindexer::reindex(const rosbag2_storage::StorageOptions & storage_options)
+{
+  reindexer_impl->reindex(storage_options);
+}
+
```
Karsten1987, line 47:
```suggestion
```

rosbag2_cpp/package.xml
```diff
@@ -11,6 +11,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <depend>ament_index_cpp</depend>
+  <depend>libboost-filesystem</depend>
```
Karsten1987, line 14:
```suggestion
```

rosbag2_cpp/include/rosbag2_cpp/reindexers/sequential_reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  void init_metadata(
+    const std::vector<rcpputils::fs::path> & files,
+    const rosbag2_storage::StorageOptions & storage_options);
+
+  // Attempts to harvest metadata from all bag files, and aggregates the result
+  void aggregate_metadata(
+    const std::vector<rcpputils::fs::path> & files,
+    const rosbag2_storage::StorageOptions & storage_options);
+
+  // Compairson function for std::sort with our filepath convention
```
Karsten1987, line 111:
```suggestion
  // Comparison function for std::sort with our filepath convention
```

rosbag2_cpp/include/rosbag2_cpp/reindexers/sequential_reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  SequentialReindexer(
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
+
+  virtual ~SequentialReindexer();
+
```
Karsten1987, line 72:
```suggestion
```

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+namespace rosbag2_cpp
+{
+namespace reindexers
+{
+namespace details
```
Karsten1987, line 55: I feel like the following two functions should be accessible independently from the reindexer. Aren't we having similar code in the reader already? I think we're introducing code duplication here.
Especially, given that version differentiation, I believe it's important to keep code as such to a minimum.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -430,19 +430,19 @@ rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
 
   auto statement = database_->prepare_statement(
     "SELECT name, type, serialization_format, COUNT(messages.id), MIN(messages.timestamp), "
-    "MAX(messages.timestamp) "
+    "MAX(messages.timestamp), offered_qos_profiles "
```
Karsten1987, line 433: why are we needing this change here for the reindexing?

Karsten1987 (commented):
First off, thanks a lot for keeping this PR alive, I know it's not trivial. A few things here:

* I would try to keep the change to a minimal set. That is, if possible, I would prefer if we'd started without compression and end-to-end integration to fix the internal architecture of reindexing. Similarly, I was wondering what's the purpose of introducing an interface for the indexer. Don't get me wrong here, there might be a legit reason for it, I just want to rule out that we don't introduce unnecessary complexity. That is, I believe a reindexing of the database will always yield the same result, I don't really see what's a sequential reindexing doing differently from a different method of reindexing.
* We've currently introduced a set of functions within `rcutils` (linked inline) which replaces much of the custom logic to iterate through files in a directory. I strongly advertise this. 
* I feel there's a lot of code around file handling, rosbag versioning and such which could be reduced or reorganized to be reused across the rosbag2 functionality. I believe there's code within the sequential_indexer wich can be reused within the sequential reader and writer.
* I don't really like that `rosbag2_transport` is involved in this. I know it's currently our only entry-point from `ros2bag` to `rosbag2_cpp`, but would it make sense to rather go through `rosbag2_py` instead? I don't think it's a hard requirement here for me, but I think we should generally rethink the architecture on how to deal with python and cpp, now that we have `rosbag2_py`

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
jhdcs: This is because the original metadata file includes the QOS profiles of the topics, so I have to query them from the database. I did some searching around, and I don't think this change affects anything else in the project, though if you have an alternate way to query the QOS profiles from the database, that might be preferable.

rosbag2_cpp/src/rosbag2_cpp/reindexers/sequential_reindexer.cpp
jhdcs: Is there a spot that you think this would fit better in?

Also, looking back on the code, not sure why I have the version parameter there. Huh. That should probably come out.

jhdcs:
Closing in favor of #641

PR #543, opened 2020-10-20
Use ZSTD's streaming interface for compressing files

### LINKED ISSUE #295: Use streaming for compression
## Description

The current implementation of compression with Zstd loads a file into memory and compresses in place. A better solution would be to use Zstd's [`compressStream`](https://github.com/facebook/zstd/blob/dev/examples/streaming_compression.c) feature which reads data in buffers. This enables rosbag2 to generate and compress large bag files on memory constrained systems. 

## Completion Criteria

* Refactor `rosbag2_compression::zstd_de/compressor` to use zstd's `ZSTD_de/compressStream` API. 

* Previous unit tests should pass.

pjreed:
This uses ZSTD's [streaming compression](https://facebook.github.io/zstd/zstd_manual.html#Chapter7) interface when compressing files, which will significantly reduce memory usage since it no longer needs to read in the entire file at once.

One significant caveat to note about this is that the previous version of the code had a check to verify that the size of the decompressed output matched the size recorded in the compression header for the file.  ZSTD's streaming interface does not record the decompressed size in its frame header (it cannot, since it does not know the size ahead of time); this means that compressed bag files recorded with the streaming interface cannot be played by older versions of rosbag2 since they will be missing that bit of information in their headers, and that will throw an exception.

Fixes #295.

Distribution Statement A; OPSEC #4584


pjreed:
Seems like a test failed after merging in the latest master, I'll investigate...

pjreed:
I'm actually not sure why that test failed -- I don't *think* this change would have affected it -- and everything passes for me in my environment.  @emersonknapp , any ideas?

emersonknapp:
The output suggests that the failing test is happening with CycloneDDS - are you also using cyclone locally, or running with FastDDS?

emersonknapp:
BTW if you update your branch - the test problem should be resolved - let me know when you feel ready for a review here

emersonknapp (commented):
Thanks for your patience on this - the code looks fine. I'd like it if some of the common setup and shutdown could be deduplicated, perhaps as some utility function that sets up the ifstream/ofstream for you. But that's minor, I won't block the review on that, pursue it if you like.

I think you're fine to take this out of draft (and it will need a rebase) - and let's make sure the tests are passing, then we can go ahead and merge.

pjreed:
I would also like to see the code around the compressor/decompressor setup/shutdown look a little cleaner, but I've been staring at it a while and they're just different enough that I haven't thought of a good way to do so.

Anyway, I've merged master into this and all of the tests pass for me; I was previously using FastRTPS, but switched to CycloneDDS, and everything still builds & runs fine for me.  Not sure why the CI build would have a problem with it...

emersonknapp:
> Not sure why the CI build would have a problem with it...

The important one works - the `Rpr__rosbag2__ubuntu_focal_amd64` is nice in theory but unreliable in practice in my experience, for a couple reasons

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -67,6 +67,8 @@ void SequentialCompressionWriter::compression_thread_fn()
   auto compressor = compression_factory_->create_compressor(
     compression_options_.compression_format);
 
+  std::mutex mutex;
+  std::unique_lock<std::mutex> lock(mutex);
```
emersonknapp, line 71:
```suggestion

```

These lines are dangling from some past state, I think

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -104,7 +106,8 @@ void SequentialCompressionWriter::compression_thread_fn()
       {
         // Now that the message is compressed, it can be written to file using the
         // normal method.
-        std::lock_guard<std::recursive_mutex> storage_lock(storage_mutex_);
+        std::lock_guard<std::recursive_mutex> storage_lock(
+          storage_mutex_);
```
emersonknapp, line 110:
```suggestion
        std::lock_guard<std::recursive_mutex> storage_lock(storage_mutex_);
```

revert unrelated formatting change

emersonknapp (approved):
LGTM minus dangling unrelated changes - once that's fixed I'll kick off the ci.ros2.org build

pjreed:
Tests still build & run for me without any failures; is there a way to make just that one run again?

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/6320545b1692e39a982c128f44c559bc/raw/6eb426aa03b3283fa5051a2398d609df9a203d0c/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7554
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13553/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8447/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11276/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13618/)

PR #544, opened 2020-10-22
Updates for pybind11 v2.6.0

jacobperron:
Depends on https://github.com/ros2/pybind11_vendor/pull/8

This PR reverts #538, removing a workaround for Windows that should no longer be needed (284134e).

pybind11 v2.6.0 introduced a CMake warning if `PYTHON_VERSION` is set:

https://github.com/pybind/pybind11/blob/4d037439ba0c7aa9add3c58b4754b634d55203df/tools/pybind11Tools.cmake/#L15-L19

IIUC, we can just quiet it by setting `PYBIND11_PYTHON_VERSION` to the version of Python we are using (2204e4e)


clalancette:
We updated to pybind11 2.9.1 a while ago, so this is no longer relevant.  Closing this out.

PR #545, opened 2020-10-23
Sqlite3 storage config fixes

adamdbrw:
- updated handling and application of pragmas for sqlite storage
- added checks whether pragma setting is a correct keyword and applied to db
- tested with example configuration files

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
@@ -70,8 +70,8 @@ std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_rese
     if (no_result || sqlite3_column_type(statement_, 0) == SQLITE_NULL) {
       // No result or result is null means that no such pragma exists
       std::stringstream errmsg;
-      errmsg << "Statement returned empty value while result was expected, check syntax: " <<
-        sqlite3_sql(statement_);
+      errmsg << "Statement returned empty value while result was expected: \'" <<
+        sqlite3_sql(statement_) << "\'";
 
       throw SqliteException{errmsg.str()};
     }
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
@@ -52,7 +52,8 @@ SqliteStatementWrapper::~SqliteStatementWrapper()
   }
 }
 
-std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset()
+std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_reset(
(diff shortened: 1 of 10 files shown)


adamdbrw:
Some changes that I consider improvements. What I am working on and would see added directly to the sqlite3_storage_config branch once this is merged:
- Tests (including some example configuration files with both good and invalid settings). I added a few to the performance testing package since at least the optimized/non-optimized settings are very relevant there.
- documentation of settings
Where would you say are the best places to add these?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
```diff
@@ -65,6 +65,28 @@ std::shared_ptr<SqliteStatementWrapper> SqliteStatementWrapper::execute_and_rese
...
+  check_value = true;
+  bool no_result = return_code == SQLITE_DONE || sqlite3_column_count(statement_) == 0;
+
+  if (no_result || sqlite3_column_type(statement_, 0) == SQLITE_NULL) {
+    // No result or result is null means that no such pragma exists
+    check_value = false;
+  }
+  return reset();
+}
+
```
Karsten1987, line 89: I think this should be a function overlay of `execute_and_reset`. This makes the behavior clearer, but also avoids code duplication.

Karsten1987 (approved):
Looks good to me. Expect that one overlay remark.

Question tho: Is there anyway to have a basic unit tests to see the behavior if some wrong yaml configuration is used?

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_statement_wrapper.cpp
adamdbrw: Will make this one an overload of the other function (you meant overload, right? I am not familiar with the term overlay in this context). 

As for testing, the first thing that comes to mind is to look for warnings in the output. Currently I added WARN logs, e.g. ones that show up when the pragma is an unrecognized setting: 

[WARN] [1603730319.797767516] [rosbag2_storage]: PRAGMA setting not recognized by sqlite, not applied: bad_pragma_name
[WARN] [1603730319.797800819] [rosbag2_storage]: PRAGMA setting not recognized by sqlite, not applied: worse_pragma_name

Would that be ok?

Karsten1987: yes, overload!

Warning parsing seems very tedious for me and most likely indicates a misleading API. Is there a chance we can throw when pragmas are wrongly set?

adamdbrw: Of course we can - and I agree this is better than ignoring and warning. I will implement that!

adamdbrw: since the change is actually only adding an optional check, I added an optional parameter instead of an overload

adamdbrw:
Added a couple of simple tests to check sqlite3 valid/invalid config. @Karsten1987, @emersonknapp, should I merge this now?

Karsten1987 (approved):

Issue #436, opened 2020-06-10
Improving performance through asynchronous storage writes


adamdbrw:
## Description

With the current implementation, callbacks are not processed until storage finishes writing. This degrades performance with sufficient load as messages are lost when the subscriber queue gets full. Also, I/O use can oscillate between 100% and 0% for large messages resulting in inefficient use of a bottleneck resource (disk write bandwidth)

A solution is to implement a consumer-producer scheme, where callbacks are put in a shared queue for storage writer to process.

The transaction should be opened right when the first write comes and writing should be continuous. When transaction is committed, a buffer is to be used to accumulate messages that arrived in the meanwhile.

## Related Issues

This issue comes from work on performance testing of rosbag2: #435 

## Completion Criteria

- implementation is reviewed and pushed out
- debugging shows that callbacks are processed with the desired frequency, without waiting for disk writes.
- I/O and system tools such as iotop show smoothed I/O profile

## Testing Notes / Suggestions

- Use performance test packages (coming soon, see #435) to compare performance before and after the change (it is expected to improve for heavier loads).
- Run rosbag2_tests to verify that no new issues were introduced

adamdbrw:
Attachment: 
A working hacky solution patch I used to test different settings (and an example input file). Of course proper implementation should be very different, but this could be useful to someone.
[config_reading_hacky.zip](https://github.com/ros2/rosbag2/files/4781602/config_reading_hacky.zip)

Karsten1987:
I think a double buffering approach could be a good solution to this. Saying instead of having only one vector, we feature two (or many) cache instances and write the one being filled asynchronously to disk.

emersonknapp:
That sounds like a good approach - I don't see why we would need more than two, though.

* Buffer0 receives messages (transaction is open and storage thread is writing)
* Buffer0 full - swap message callback pointer to Buffer1 - commit Buffer0 transaction
* Buffer1 receives messages
* Buffer0 transaction commit complete - open transaction for Buffer 1
* Buffer1 full - swap message callback pointer to Buffer0 - commit Buffer1 transaction
* etc...

Like so? I suppose we might need more buffers if the "transaction commit" takes longer than filling the next buffer - but in that case we're just writing too slowly and are going to probably drop data anyways, right?

PR #546, opened 2020-10-26
Sqlite storage double buffering

adamdbrw:
Ongoing work on implementation for #436


adamdbrw:
Current issues:

- Role of max_cache_size parameter is confused atm. The vector storing messages should likely be removed now.
- Decision whether buffering is storage implementation specific. 
- When to open transaction. Also, transaction size/cache size should not be specified by number of messages (this is rather poor way to parametrize for the user)

emersonknapp:
Note that cache size is now in bytes rather than number of messages, as of https://github.com/ros2/rosbag2/pull/530

pijaro:
Yes, we are aware that `max_cache_size` is now a size in bytes. Currently we use `max_cache_size` as a limit for each buffer (proposed 2 buffers configuration) whereas transaction size is dynamic but `<= max_cache_size`, depending on disk write speed (if the disk is throttled with writing data, then `transaction_size ~= max_cache_size`). 

We will update draft soon.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -241,15 +249,10 @@ void SequentialWriter::write(std::shared_ptr<rosbag2_storage::SerializedBagMessa
...
-  } else {
-    cache_.push_back(converted_msg);
-    current_cache_size_ += converted_msg->serialized_data->buffer_length;
-    if (current_cache_size_ >= storage_options_.max_cache_size) {
-      reset_cache();
-    }
+
+  if (!buffer_layer_->push(converted_msg)) {
+    // Undo message count because it was dropped by buffer layer
+    --topics_names_to_info_.at(message->topic_name).message_count;
```
adamdbrw, line 255: only add count when successfully pushed to the cache (like it was before)

adamdbrw:
This is ready for review now. We have some optimistic results from performance testing, to be presented and discussed on Tooling WG (attached).

[performance_500Mbs_disk.zip](https://github.com/ros2/rosbag2/files/5494986/performance_500Mbs_disk.zip)

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -155,12 +156,19 @@ void SequentialCompressionWriter::compress_last_file()
...
+  buffer_layer_->close();
+
   storage_options_.uri = format_storage_uri(
     base_folder_,
     metadata_.relative_file_paths.size());
 
   storage_ = storage_factory_->open_read_write(storage_options_);
 
+  // Update storage in buffer layer and restart consumer thread
+  buffer_layer_->set_storage(storage_);
```
Karsten1987, line 169: I am somehow having a problem with this chain of dependency. I don't have a real argument for this, but I would feel better about it if the storage would not belong to the buffer.
Can we have the consumer/producer threads operate on the buffer, but not belong to them?

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
+#include "rosbag2_storage/storage_options.hpp"
+#include "rosbag2_cpp/visibility_control.hpp"
+
+namespace rosbag2_cpp
+{
+namespace writers
+{
+
+class ROSBAG2_CPP_PUBLIC BufferLayer
```
Karsten1987, line 33: bikeshedding here, but could we rename this to `CircularBuffer` or something alike? What's the `Layer` leading to here?

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
```diff
@@ -0,0 +1,131 @@
...
+  ~BufferLayer();
+
+  // Flush data into storage, and reset cache
+  void reset_cache();
+  // Push data into primary buffer
+  bool push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg);
+  // Flush buffers and reset them, shut down consumer thread
+  void close();
+  // Starts consumer thread if one is not working
+  void start_consumer();
```
Karsten1987, line 48: I would appreciate if we could separate the storage in the buffer from the execution model. That way we can have the buffer being a single task class whereas the threading can be handled in various ways independently of the buffer.

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
```diff
@@ -0,0 +1,131 @@
...
+namespace rosbag2_cpp
+{
+namespace writers
+{
+
+class ROSBAG2_CPP_PUBLIC BufferLayer
+{
+public:
+  BufferLayer(
+    std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage,
```
Karsten1987, line 37: somewhat related to the comment with the threading, but I would appreciate if we could separate the storage from the buffer. I feel that should be a separation of concerns here. The buffer really only buffers incoming data and should not have a (strong) dependency to the storage backend.
Could we maybe add an API which takes a lamdba function or some sort of callback which is getting called with a filled buffer to be written?

The reason why I am arguing for this is have introduce a bit of flexibility on what to do with this buffer or have a simply way to bypass this in total.

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  // Write secondary buffer data to a storage
+  void exec_consuming();
+
+  // Storage handler
+  std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage_;
+
+  // Double buffers
+  std::shared_ptr<BagMessageBuffer> primary_buffer_;
+  std::shared_ptr<BagMessageBuffer> secondary_buffer_;
```
Karsten1987, line 110: would it make sense to pack this in an `std::array` or `std::vector`? This would allow us to increase the amount of buffers dynamically. We could think about a storage plugin which allows a concurrent write access. That way we could technically feature a triple buffer approach in which two (or more) buffers can be written to disk.

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  // Total number of dropped messages
+  uint32_t elements_dropped_ = {0u};
```
Karsten1987, line 113: might be worth a follow up PR, but instead of a total number of messages, can we have a more detailed statistic  here at which time and which topic we lost messages?

rosbag2_cpp/src/rosbag2_cpp/writers/buffer_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  std::shared_ptr<rosbag2_storage::storage_interfaces::ReadWriteInterface> storage,
+  const uint64_t & max_buffer_size)
+: storage_(storage)
+{
+  // Set buffers
+  primary_buffer_ = std::shared_ptr<BagMessageBuffer>(
+    new BagMessageBuffer(max_buffer_size));
+  secondary_buffer_ =
+    std::shared_ptr<BagMessageBuffer>(
+    new BagMessageBuffer(max_buffer_size));
```
Karsten1987, line 37: consider `std::make`

rosbag2_cpp/src/rosbag2_cpp/writers/buffer_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  // Run consumer thread. We need this only when max cache size is not zero. Otherwise direct
+  // storage write call is used.
+  if (max_buffer_size > 0u) {
```
Karsten1987, line 41: That's pretty much the reason why I feel we should exclude this logic from the buffer. The implementation of the butter should not be concerned whether it is actually being used or not.

rosbag2_cpp/src/rosbag2_cpp/writers/buffer_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  if (!pushed) {
+    elements_dropped_++;
+  }
+  return pushed;
+}
+
+void BufferLayer::exec_consuming()
+{
+  bool exit_flag = false;
+  while (true) {
```
Karsten1987, line 73: sorry but no! What about `while (!exit_flag)`? That would also ease the `return` logic at the end of the while loop.

rosbag2_cpp/src/rosbag2_cpp/writers/buffer_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+      std::unique_lock<std::mutex> lock(buffer_mutex_);
+      buffers_condition_var_.wait(lock);
+    }
+
+    // Swap buffers
+    {
+      std::lock_guard<std::mutex> lock(buffer_mutex_);
+      std::swap(primary_buffer_, secondary_buffer_);
+    }
+
```
Karsten1987, line 84:
```suggestion
```

rosbag2_cpp/src/rosbag2_cpp/writers/buffer_layer.cpp
```diff
@@ -0,0 +1,153 @@
...
+  return pushed;
+}
+
+void BufferLayer::exec_consuming()
+{
+  bool exit_flag = false;
+  while (true) {
+    if (!is_stop_issued_) {
+      std::unique_lock<std::mutex> lock(buffer_mutex_);
+      buffers_condition_var_.wait(lock);
```
Karsten1987, line 76: this needs a bit of a comment and/or a renaming of the condition variable.

The condition variable indicates that the buffer for incoming messages is currently not being used and can be written to disk, is that right? If that comment is correct, shouldn't the buffer swap be handled in the same lock? If not, the lock is being released here - a context switch could happen and a new messages appears in the incoming buffer - where at the same time we below swap the buffer.

Karsten1987 (commented):
just a first pass through.

Karsten1987:
A question which occurred to me during the first review is whether we could reuse the shared consumer/producer queues we're currently featuring in the rosbag2_transport package: https://github.com/cameron314/readerwriterqueue
This seems pretty much exactly what we wanted plus it's lock-free.

adamdbrw:
> A question which occurred to me during the first review is whether we could reuse the shared consumer/producer queues we're currently featuring in the rosbag2_transport package: https://github.com/cameron314/readerwriterqueue
> This seems pretty much exactly what we wanted plus it's lock-free.

It is certainly an useful data structure, but it is not a data structure we want:
- Different purpose and semantics: It implements circular buffers, while what we need are memory-constrained buffers (e.g. limited by bytes-size, not elements-count). We also always want a full message in, not a count of bytes. 
- Unnecessary functionalities: It handles memory allocation for the elements in blocks, which we don't need since we use handle already allocated memory through shared pointers (it would also slow things down). Also, IMO lock free might be more important if you are only handling memory, but when you can associate each lock with a disk write operation, the priority of optimizing it out drops hard.

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
adamdbrw: I think that the idea was that this is the layer between writer and storage that is responsible for buffering (buffers themselves plus threading), thus the name. Is that correct, @pijaro?

If we separate threading from data structure:

CircularBuffer would be misleading, since this data structure is not a circular buffer.
Perhaps: CacheBuffer or MemoryConstrainedBuffer?

adamdbrw: Please take a look at current names of classes and let me know if they are ok

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
adamdbrw: This is naturally possible. When implementing such storage, one could expand cache_consumer to use a thread set/pool, and implement switching logic for more buffers in message_cache.

rosbag2_cpp/include/rosbag2_cpp/writers/buffer_layer.hpp
adamdbrw: I don't see any problem with that - we have access to all we need in the place where we count it

adamdbrw:
@Karsten1987 @pijaro 
Take a look at my changes - should be more divided into separate concerns now.

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+void CacheConsumer::exec_consuming()
+{
+  bool exit_flag = false;
+  while (!exit_flag) {
+    if (!is_stop_issued_) {
+      // wait for any data in producer buffer to swap it with consumer buffer
+      message_cache_->swap_when_allowed();
```
Karsten1987, line 74: I feel this funciton name is misleading, can we rename this to `wait_for_buffer` or something, indicating that this call is blocking?

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // consume all the data from consumer buffer
+    auto consumer_buffer = message_cache_->consumer_buffer();
+    storage_->write(consumer_buffer->data());
```
Karsten1987, line 79: I feel that could be a dynamic callback, then it's truly independent from the storage.

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,91 @@
...
+    consumer_buffer->clear();
+
+    {
+      std::lock_guard<std::mutex> consumer_lock(consumer_mutex_);
+      if (is_stop_issued_) {exit_flag = true;}
```
Karsten1987, line 84: why can't we make this the condition for the while loop? `while (!is_stop_issued)` and make that variable an atomic bool?

rosbag2_cpp/src/rosbag2_cpp/writers/cache/message_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  return pushed;
+}
+
+void MessageCache::allow_swap()
+{
+  swap_ready_.notify_one();
+}
+
+void MessageCache::swap_when_allowed()
```
Karsten1987, line 56: consider renaming this `wait_for_swap`

rosbag2_cpp/src/rosbag2_cpp/writers/cache/message_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+  bool pushed = false;
+  {
+    std::lock_guard<std::mutex> buffer_lock(buffer_mutex_);
+    pushed = primary_buffer_->push(msg);
+  }
+
+  allow_swap();
+
+  if (!pushed) {
+    elements_dropped_++;
```
Karsten1987, line 46: we should log at least on which topic (should be part of `msg`) is getting lost.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -265,41 +265,6 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
   }
 }
 
-TEST_F(SequentialWriterTest, only_write_after_cache_is_full) {
```
Karsten1987, line 268: why is that test deleted?

rosbag2_cpp/src/rosbag2_cpp/writers/cache/message_cache_buffer.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <memory>
+#include <vector>
+
+#include "rosbag2_cpp/writers/cache/message_cache_buffer.hpp"
+
+namespace rosbag2_cpp
+{
+namespace writers
+{
```
Karsten1987, line 23: consider removing this namespace

rosbag2_cpp/src/rosbag2_cpp/writers/cache/message_cache.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+std::shared_ptr<MessageCacheBuffer> MessageCache::consumer_buffer()
+{
+  return secondary_buffer_;
+}
+
+void MessageCache::log_dropped()
+{
+  if (elements_dropped_ > 0) {
+    ROSBAG2_CPP_LOG_WARN_STREAM(
```
Karsten1987, line 71: consider this a follow-up task, but I would appreciate if we could log these information not only to console output, but maybe in a separate file next to the metdata.yaml. One might to process this data offline

Karsten1987 (commented):
I like it a lot. Mostly nitpicks, except that I would love if you could re-organize the cache functionality in their own namespace/folder structure and consider passing in a generic callback, which allows you to make the cache independent from the writer.

One thing I require though before merging, I feel that cache functionality is complex enough that it actually requires its own set of unit tests. Please add these as suited.

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,91 @@
...
+  close();
+}
+
+void CacheConsumer::close()
+{
+  {
+    std::lock_guard<std::mutex> consumer_lock(consumer_mutex_);
+    ROSBAG2_CPP_LOG_INFO_STREAM(
+      "Writing remaining messages from cache to the bag. It may take a while");
+    is_stop_issued_ = true;
```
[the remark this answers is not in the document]
adamdbrw: indeed, since is_stop_issued is already atomic

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
adamdbrw: The cache semantic changed - we no longer want to only start writing when the cache is full (we want to keep writing at all times if there is anything in buffers to optimize I/O). Hence the test doesn't make sense.

However, what makes perfect sense is to replace the test with some new tests for cache!

adamdbrw:
First batch of changes to address review comments. You can give your feedback while I work on the remaining issues.
Coming next:
- fixing flushing of messages & loop control for consumer thread
- unit tests

rosbag2_cpp/src/rosbag2_cpp/writers/cache/message_cache.cpp
adamdbrw: We could log the messages dropped per topic in metadata.yaml - it seems to be a good place since we have per topic stats of recorded messages there.

I thought about moving counting elsewhere (outside of cache class)

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  message_cache_->allow_swap();
```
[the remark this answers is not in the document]
adamdbrw: Good observation. Not at all, it seems - only in non-stressed situations. I think flushing guarantee got lost in first refactor when I removed one bool&lock in the consumer loop. 
I will address that in the incoming change set to this PR

rosbag2_cpp/src/rosbag2_cpp/writers/cache/cache_consumer.cpp
adamdbrw: I will solve this along with the flushing issue you pointed out in next set of changes to this PR tomorrow

rosbag2_cpp/include/rosbag2_cpp/cache/cache_consumer.hpp
```diff
@@ -0,0 +1,67 @@
...
+// This class is responsible for constantly writing the cache to storage
+class ROSBAG2_CPP_PUBLIC CacheConsumer
+{
+public:
+  typedef std::function<void (const std::vector<MessageCacheBuffer::buffer_element_t> &)>
+    consume_callback_function_t;
+
+  CacheConsumer(
+    std::shared_ptr<MessageCache> message_cache,
+    consume_callback_function_t consume_callback);
```
Karsten1987, line 40: sweet 👍

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include "rosbag2_cpp/visibility_control.hpp"
+
+#include "rosbag2_storage/serialized_bag_message.hpp"
+
+namespace rosbag2_cpp
+{
+namespace cache
+{
+// This class is responsible for implementing message cache, using cache buffers
+// and providing synchronisation API for producer-consumer pattern
```
Karsten1987, line 35:
```suggestion
// and providing synchronization API for producer-consumer pattern
```

'murica ;-)

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+  // Push data into primary buffer
+  bool push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg);
+
+  // Summarize dropped/remaining messages
+  void log_dropped();
+
+  // Notify of data in the primary buffer
+  void allow_swap();
+
+  // Consumer API: wait until there is data to consume and swap
```
Karsten1987, line 52: while I understand that "swap" indicates that a buffer is good to go for being consumed, I feel that swap might not be as easily understood. Can we add some more doxygen here?

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,74 @@
...
+  // Consumer API: get current buffer to consume
+  std::shared_ptr<MessageCacheBuffer> consumer_buffer();
+
+private:
+  // Double buffers
+  std::shared_ptr<MessageCacheBuffer> primary_buffer_;
+  std::shared_ptr<MessageCacheBuffer> secondary_buffer_;
+
+  // Dropped messages per topic. Used for printing in alphabetic order
+  std::map<std::string, uint32_t> messages_dropped_per_topic_;
```
Karsten1987, line 64: controversial, but would a `std::unordered_map` make more sense here and call `std::sort` upon printing? I feel that hashing might be faster than traversing a RB tree.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_buffer.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace cache
+{
+// This class implements a single buffer for message cache. Note that it could
+// be reused as a template with any class that has ->byte_size() - like interface
+class ROSBAG2_CPP_PUBLIC MessageCacheBuffer
+{
+public:
+  explicit MessageCacheBuffer(const uint64_t max_cache_size);
+
+  typedef std::shared_ptr<const rosbag2_storage::SerializedBagMessage> buffer_element_t;
```
Karsten1987, line 36:
```suggestion
  using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
```

rosbag2_cpp/include/rosbag2_cpp/cache/cache_consumer.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace rosbag2_cpp
+{
+namespace cache
+{
+// This class is responsible for constantly writing the cache to storage
+class ROSBAG2_CPP_PUBLIC CacheConsumer
+{
+public:
+  typedef std::function<void (const std::vector<MessageCacheBuffer::buffer_element_t> &)>
+    consume_callback_function_t;
```
Karsten1987, line 36:
```suggestion
  using  consume_callback_function_t = std::function<void (const std::vector<MessageCacheBuffer::buffer_element_t> &)>;
```

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -111,16 +113,20 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_;
   std::unique_ptr<Converter> converter_;
 
+  bool use_cache_;
+  std::shared_ptr<rosbag2_cpp::cache::MessageCache> message_cache_;
+  std::unique_ptr<rosbag2_cpp::cache::CacheConsumer> cache_consumer_;
+
+  void switch_to_next_storage();
```
Karsten1987, line 120:
```suggestion
  void switch_to_next_storage();
  
```

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,85 @@
...
+  if (consumer_thread_.joinable()) {
+    consumer_thread_.join();
+  }
+}
+
+void CacheConsumer::change_consume_callback(
+  CacheConsumer::consume_callback_function_t consume_callback)
+{
+  std::lock_guard<std::mutex> consumer_lock(consumer_mutex_);
+  consume_callback_ = consume_callback;
```
Karsten1987, line 56: If I understand this correctly, this would technically mean that a single buffer instance can be executed from two different callbacks, is that right? 
Wouldn't it make sense to indicate that the consumer thread should be finished its current execution, so that the new callback is getting executed on an empty buffer?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+  allow_swap();
+
+  if (!pushed) {
+    messages_dropped_per_topic_[msg->topic_name]++;
+  }
+  return pushed;
+}
+
+void MessageCache::allow_swap()
```
Karsten1987, line 51: sorry, don't want to be picky here, but we might as well rename this function to `notify_swap` to be on pair with `wait_for_swap`. 
Even though in general, I am still not too happy about the expression "swap" as I don't think it indicates fully what's going to happen. Does "wait_for_buffer` and `notify_buffer_swap` make sense?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
```diff
@@ -0,0 +1,103 @@
...
+bool MessageCache::push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg)
+{
+  // While pushing, we keep track of inserted and dropped messages as well
+  bool pushed = false;
+  {
+    std::lock_guard<std::mutex> buffer_lock(buffer_mutex_);
+    pushed = primary_buffer_->push(msg);
+  }
+
+  allow_swap();
```
Karsten1987, line 43: Can elaborate this for me once more? If I allow for disc-write the buffer after each push to the buffer, how can be buffer ever be full? Shouldn't we allow the swap only when `primary_buffer_->push(msg)` returns `false`, indicating that the primary buffer is full?

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
adamdbrw: makes sense because of proportion of expected value increments

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
adamdbrw: As far as I understand:
Only one callback can be executed by the consumer thread at one time (since only consumer thread executes callbacks).

What could (probably very rarely) happen is that the callback is changed after the buffer is acquired from the cache, but before it is passed to the "old" callback as would happen normally. 

This would cause part of cache to already go into the new bag, which would mean that:
- the former bag could turn out a bit smaller than suggested split size
- 

A bad alternative is to implement a locking mechanism - but change_consume_callback would block the writer.
A much more reasonable alternative is to set the next callback as a member variable and synchronize that - only switching in exec_consuming between loop iterations.

Shall I implement that?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
adamdbrw: I am opting for wait_for_buffer() and notify_buffer_consumer(), would that do?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
adamdbrw: The key here is the **max**_cache_size 
How can buffer ever be full?
- The consumer thread always consumes the whole consumer buffer, which can have 1 or more messages
- Consuming involves storage writing, which could take considerable time (and the time can spike)
- During this time, producer buffer can fill to some degree (up to max cache)
- Cache being full is a kind of pathology here - since we start dropping messages right when this happens
This implementation is centered around optimizing I/O - we want to guarantee that consumer thread is always busy when there is any amount of data to be written to disc. This might not be very important in average case, but when the throughput is near the disk write limit, this approach makes a difference.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
Karsten1987: sounds good.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
emersonknapp: > Consuming involves storage writing, which could take considerable time (and the time can spike)

I think this is the key point - while a write is going on that's when we're filling up the other buffer - the more traffic there is, the more will go into the buffer and the longer the write will take, so high throughput scenarios actually have less frequent but much bigger write transactions

Karsten1987: understood. I would love to see a similar text as in your answer being part of the documentation for these cache classes. I think that makes a lot of sense to document the intend here.

adamdbrw:
I am still working on resolve the issue that apparently causes one test to fail on metadata file missing (weird). Maybe it is the timing it takes to flush. I am investigating.

Another remaining issue is what test would be the most useful here. I would imagine adding a couple of tests that verify if all messages were written successfully through count in yaml file (which now should be the real number of messages in the bag).
Any ideas? @pijaro?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
```diff
@@ -0,0 +1,134 @@
...
+void MessageCache::wait_for_buffer()
+{
+  std::unique_lock<std::mutex> lock(cache_mutex_);
+  if (!flushing_) {
+    // Required condition check to protect against spurious wakeups
+    cache_condition_var_.wait(
+      lock, [this] {
+        return primary_buffer_can_be_swapped_ || flushing_;
+      });
+    primary_buffer_can_be_swapped_ = false;
```
Karsten1987, line 80: what is variable for? I see that it's set to false here, and set to true in the notify function. But were is it actually being used?

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,86 @@
...
+  bool exit_flag = false;
+  bool flushing = false;
+  while (!exit_flag) {
+    // Invariant at loop start: consumer buffer is empty
+
+    // swap producer buffer with consumer buffer
+    message_cache_->wait_for_buffer();
+
+    // make sure to use consistent callback for each iteration
+    auto callback_for_this_loop = consume_callback_;
```
Karsten1987, line 73: 👍

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    // consume all the data from consumer buffer
+    auto consumer_buffer = message_cache_->consumer_buffer();
+    callback_for_this_loop(consumer_buffer->data());
+    consumer_buffer->clear();
+
+    if (flushing) {exit_flag = true;}  // this was the final run
+    if (is_stop_issued_) {flushing = true;}  // run one final time to flush
```
Karsten1987, line 81: I still wonder why we can't deal with a single `is_stop_issued_` flag here. But that's no show-stopper.

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
Karsten1987: > actually, I believe it would be enough to use copy assignment in the consumer loop so that the callback used within is consistent.

That should do from what I understand.

Karsten1987:
>  I would imagine adding a couple of tests that verify if all messages were written successfully through count in yaml file (which now should be the real number of messages in the bag).

While that's nice on a system integration level, I'd be happy to mock the callback with a simple sleep and assert that the callback is called with the right amount of messages. More like unit testing the cache and its concurrent behavior rather than a huge integration test. I am afraid that we might introduce some flakiness here again, whereas with a simple mock-up we can limit the amount of places a timing can have a bad influence.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
adamdbrw: in the predicate for the conditional wait (two lines above the false assignment). This is due to the case of "spurious wake-ups" e.g. proper use of conditional variables should include checking whether the wake up was indeed due to intended logic change

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
adamdbrw: In short - once the stop is issued, we need another execution (since there is no guarantee that the consumer is waiting on conditional when the last notify from producer push comes)

pijaro: Indeed, we need to do exactly one extra loop after stop is issued to finish current buffer writing and flush second one. Single flag brakes the while loop on the `while` condition just after current buffer is done what results in second buffer being left unflushed.

adamdbrw:
The issue was with not restoring the state properly after flushing was started, this is fixed thanks to @pijaro now

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
adamdbrw: Added 2 tests

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,110 @@
...
+* accepted and buffer switching can be done unconditionally on consumer demand.
+*
+* The cache holds infomation about dropped messages (per topic). These are
+* messages that were pushed to the cache when it was full. Such situation signals
+* performance issues, most likely with the CacheConsumer consumer callback.
+*/
+class ROSBAG2_CPP_PUBLIC MessageCache
+{
+public:
+  MessageCache(uint64_t max_buffer_size);
```
Karsten1987, line 56:
```suggestion
  explicit MessageCache(uint64_t max_buffer_size);
```

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "rosbag2_cpp/cache/message_cache_buffer.hpp"
+#include "rosbag2_cpp/visibility_control.hpp"
+
+#include "rosbag2_storage/serialized_bag_message.hpp"
+
+namespace rosbag2_cpp
+{
+namespace cache
+{
+/* This class is responsible for implementing message cache, using two cache
```
Karsten1987, line 35:
```suggestion
/**
 * This class is responsible for implementing message cache, using two cache
```

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -0,0 +1,110 @@
...
+* performance issues, most likely with the CacheConsumer consumer callback.
+*/
+class ROSBAG2_CPP_PUBLIC MessageCache
+{
+public:
+  MessageCache(uint64_t max_buffer_size);
+
+  ~MessageCache();
+
+  // Puts msg into primary buffer. With full cache, msg is ignored and counted as lost
```
Karsten1987, line 60:
```suggestion
  /// Puts msg into primary buffer. With full cache, msg is ignored and counted as lost
```

Karsten1987: same below for Doxygen to pick it up.

Karsten1987 (commented):
Looks good to me. Let' see what CI has to say:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13154/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8085/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10869/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13159/)

adamdbrw:
Quite some warnings on Windows, but pretty straightforward to address. We will handle them tomorrow.

adamdbrw:
@Karsten1987 @mjeronimo, could you re-trigger Windows CI job?

mjeronimo:
Restarting Windows build
[Build Status](https://ci.ros2.org/job/ci_windows/13168/)

adamdbrw:
There are 9 warnings left. They don't look like they are associated with the PR changes though.. or am I wrong?

Karsten1987 (approved):
The same windows warnings are currently seen on the nightlies. So yes, there are unrelated.

@mjeronimo I'll leave it up to you to take a pass on this as well.

mjeronimo:
I've also confirmed that all of the CMakeLists.txt warnings are also in the nightlies. However, there is one new warning related to the use of the std::accumulate template function from the \<numeric\> header: 

`'=': conversion from 'size_t' to '_Ty', possible loss of data`

Which, I believe, is referring to the usage here:

https://github.com/ros2/rosbag2/blob/sqlite_storage_double_buffering/rosbag2_cpp/test/rosbag2_cpp/test_message_cache.cpp#L54

The sizes of size_t and uint32_t are different (size_t == 8 bytes, while uint32_t == 4 bytes). I think you can get rid of this one with some slight modifications to the code to use the appropriate types for the accumulation.

adamdbrw:
@mjeronimo: final warning should be fixed now

adamdbrw:
Performance of master vs this PR:

image(4)

Note that both old and new default for cache (--max-cache-size parameter), namely 1MB and 100MB. For this test setup, with 100MB to 500MB of messages being written each second, a cache of 1MB is pathological. Note that Recorded Messages % reaches 100% for each row when this PR is combined with #568.

mjeronimo:
Restarting Windows build to confirm removal of warning: 
[Build Status](https://ci.ros2.org/job/ci_windows/13187/)

adamdbrw:
Seems like the numeric warning is gone and only cmake (unrelated) ones remain. Are we good to merge this?

mjeronimo:
Yes, just confirmed with the latest build that all CMake warnings are also in the nightly build and that the warning is gone. Merging.

emersonknapp:
Woohoo! Thanks everyone for pushing this through - it looks great, this is a huge improvement for rosbag2.

y-okumura-isp:
Are there any plans to backport it to Foxy?

emersonknapp:
> Are there any plans to backport it to Foxy?

It's not certain yet - the code has diverged significantly in API-breaking ways from the Foxy release that will make it hard to backport this on its own. We're having an active conversation now to potentially backport all of the latest rosbag2 code to Foxy, but this is in its early stages.

PR #556, opened 2020-11-04
Add leveldb storage plugin support

Barry-Xu-2018:


ros-discourse:

https://discourse.ros.org/t/tooling-proposal-rosbag2-leveldb-plugin/17072/4

Barry-Xu-2018:
`Rpr__rosbag2__ubuntu_focal_amd64` failed since `libleveldb-dev` isn't installed.   
Whether do I need to add `libleveldb-dev` to 
https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml in advance ?

clalancette:
> `Rpr__rosbag2__ubuntu_focal_amd64` failed since `libleveldb-dev` isn't installed.
> Whether do I need to add `libleveldb-dev` to
> https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml in advance ?

There is already a rule for it here: https://github.com/ros/rosdistro/blob/6c1c5064478fcc6cfcaf159fe84806d9ab7c2cd5/rosdep/base.yaml#L1702 .  You just need to reference that key in the `package.xml`, and the buildfarm should pull it in.

Relatedly, the "leveldb_vendor" package looks problematic.  If it is truly just a CMake wrapper around leveldb, it should be called `leveldb_cmake_module`; that way it is clear it is not "vendoring" anything.  However, you should also consider how this code is going to be built on macOS and Windows.  If there are homebrew/chocolatey packages available, respectively, then `leveldb_cmake_module` is the way to go.  If there are not packages available, then you either need to think about creating packages for those platforms or truly vendoring leveldb.

Barry-Xu-2018:
@clalancette 

> There is already a rule for it here: https://github.com/ros/rosdistro/blob/6c1c5064478fcc6cfcaf159fe84806d9ab7c2cd5/rosdep/base.yaml#L1702 . You just need to reference that key in the package.xml, and the buildfarm should pull it in.

> Relatedly, the "leveldb_vendor" package looks problematic. If it is truly just a CMake wrapper around leveldb, it should be called leveldb_cmake_module; that way it is clear it is not "vendoring" anything. However, you should also consider how this code is going to be built on macOS and Windows. If there are homebrew/chocolatey packages available, respectively, then leveldb_cmake_module is the way to go. If there are not packages available, then you either need to think about creating packages for those platforms or truly vendoring leveldb.

Yeah. It's my fault.   
Thank you for your guidance.  I will recheck it.

emersonknapp:
My top level comment here is - is it necessary to check this in as part of the rosbag2 repository? I think this would better be suited to a separate repository, providing leveldb as an alternative storage plugin. I'm happy to do a quick pass review here, but don't think it's necessarily a good idea to add this plugin to this repository.

Barry-Xu-2018:
@emersonknapp 

> My top level comment here is - is it necessary to check this in as part of the rosbag2 repository? I think this would better be suited to a separate repository, providing leveldb as an alternative storage plugin. I'm happy to do a quick pass review here, but don't think it's necessarily a good idea to add this plugin to this repository.

Could you tell the reason why you think this should be removed from rosbag2 repository ?  
Do you means leveldb storage plugin should not be as one of default storage plugin ?

fujitatomoya:
@emersonknapp @clalancette thanks for checking this at 1st 👍 

> My top level comment here is - is it necessary to check this in as part of the rosbag2 repository?

I think that this is something we could discuss. I see that default plugin is sqlite3, default implementation is already there and why more implementation needs to be checked in this repository? if user wants to do that, they can do that by their own requirement, that is the `plugin`. (this repository will not have all of the implementation.) we totally understand this 😄 

but taking a look at the performance result https://discourse.ros.org/t/tooling-proposal-rosbag2-leveldb-plugin/17072/3, we believe that there would be some area we could discuss to check leveldb implementation in this repository.

I think that justifications are following,

- leveldb could be tier-1 plugin with sqlite3. (sqlite3 is the default, I think default it not the same with tier-1 plugin.)
- performance improvement. (we are not pushing this, but we probably would want to re-think the default plugin.)
- light-weight memory foot print and scalability.
- simply more options for user and use cases.

and we understand that more codes, more maintenance cost. and that is also something we can discuss!

fujitatomoya:
@Barry-Xu-2018 

I will be doing review, but i found the following. could you check?

```
root@5c7a784cba7f:~/ros2_ws/rosbag2_leveldb# ros2 bag play -s leveldb rosbag2_2020_11_08-19_47_51
[INFO] [1604832544.222166464] [rosbag2_storage]: Opened database 'rosbag2_2020_11_08-19_47_51/rosbag2_2020_11_08-19_47_51_0' for READ_ONLY.

root@5c7a784cba7f:~/ros2_ws/rosbag2_leveldb# ls rosbag2_2020_11_08-19_47_51
metadata.yaml  rosbag2_2020_11_08-19_47_51_0
```

if using leveldb storage as backend, there will be one more directory on top. in this case, `ros2 bag play -s leveldb <bag directory>` does not work as expected. instead, it needs to issue `ros2 bag play -s leveldb <bag directory>/<levebdb directory>`. I think this needs to be fixed as user experience, this is extra cost to operate.

Barry-Xu-2018:
@fujitatomoya 

> I will be doing review, but i found the following. could you check?
> ...
> if using leveldb storage as backend, there will be one more directory on top. in this case, ros2 bag play -s leveldb <bag directory> does not work as expected. instead, it needs to issue ros2 bag play -s leveldb <bag directory>/<levebdb directory>. I think this needs to be fixed as user experience, this is extra cost to operate.

Yes. This is problem. 
For leveldb storage plugin, the user need to specify which "database file" is used.  
This behavior is different from sqlite storage plugin.  
I will fix it.

Barry-Xu-2018:
@fujitatomoya 

8acfe6a fix issue reported by you.   
Please try it.

leveldb_vendor/CHANGELOG.rst
```diff
@@ -0,0 +1,5 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package leveldb_vendor
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+0.0.1 (2020-10-30)
+------------------
```
Karsten1987, line 5: please refrain from adding changelog files by hand. The release tool will take care of it.

leveldb_vendor/package.xml
```diff
@@ -0,0 +1,21 @@
...
+  <description>leveldb vendor package, providing a dependency for leveldb.</description>
+  <maintainer email="barry.xu@sony.com">Barry Xu</maintainer>
+  <license>Apache License 2.0</license>  <!-- the contents of this package are Apache 2.0 -->
+  <license>BSD</license>  <!-- leveldb is BSD -->
+
+  <url type="website">https://github.com/google/leveldb</url>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>snappy</depend>
```
Karsten1987, line 15: is `snappy` available on all platforms or does this have to be vendor`ed as well?

Karsten1987:
> My top level comment here is - is it necessary to check this in as part of the rosbag2 repository?

I think that's a fair remark, but I think I would be more concerned if `leveldb` were a relational database - which it is not.
From their readme: `This is not a SQL database. It does not have a relational data model, it does not support SQL queries, and it has no support for indexes.`

If that's no limitation and we can provide a working solution with the current interface, I am in favor of adding it as a Tier1 plugin. Obviously it has to compile and pass tests on all platforms. That being said, I'd appreciate if we could enhance the test suits to reflect these changes and test for all storage plugins.

@Barry-Xu-2018, @fujitatomoya can assume you will take over the maintainership of these packages?

fujitatomoya:
@Karsten1987 

> can assume you will take over the maintainership of these packages?

by saying `these packages`, is that supposed to mean `leveldb related packages`? or `rosbag2`. if the former, we will do that to work with you 👍 if the latter, i cannot just say `yes` right now, but it can be discussed.

leveldb_vendor/CHANGELOG.rst
Barry-Xu-2018: > please refrain from adding changelog files by hand. The release tool will take care of it.  

Okay.  
I will remove this file.

leveldb_vendor/package.xml
Barry-Xu-2018: Yeah.  I am considering this.

Barry-Xu-2018:
@Karsten1987

> I think that's a fair remark, but I think I would be more concerned if leveldb were a relational database - which it is not.
From their readme: This is not a SQL database. It does not have a relational data model, it does not support SQL queries, and it has no support for indexes.
> 
> If that's no limitation and we can provide a working solution with the current interface, I am in favor of adding it as a Tier1 plugin.

Yeah. leveldb isn't SQL database. But leveldb can satisfy requirement of saving and reading data for rosbag based on current interface and provide better performance.  So we want to add it.  

> can assume you will take over the maintainership of these packages?

As fujitatomoya said, we can maintain leveldb storage plugin (including test codes) and related packages (such as leveldb_vendor).

Karsten1987:
Yes, I am talking here exclusively about the leveldb related packages. I just want to highlight that while it's great to have more functionality, these things always come with a cost of maintenance. But I'll take it from your comment that you guys are happily doing so 👍

Barry-Xu-2018:
The check on the window platform has been done.

Barry-Xu-2018:
Notice:
The database file of sqilte3 is a file. But leveldb is a directory.  
So some information output by rosbag2 (such as ros2 bag info) need to be updated.  
```bash
D:\dev\ros2_rolling>ros2 bag info -s leveldb mydb

Files:             mydb_0
                   mydb_1
                   mydb_2
                   mydb_3
                   mydb_4
```
Besides, leveldb depend on this PR (ros2/rcutils#306).  
Otherwise, rosbag cannot show correct database size for leveldb.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/leveldb/leveldb_wrapper.hpp
```diff
@@ -0,0 +1,172 @@
...
+  ((((x) & 0xff00000000000000ull) >> 56) \
+  | (((x) & 0x00ff000000000000ull) >> 40) \
+  | (((x) & 0x0000ff0000000000ull) >> 24) \
+  | (((x) & 0x000000ff00000000ull) >> 8) \
+  | (((x) & 0x00000000ff000000ull) << 8) \
+  | (((x) & 0x0000000000ff0000ull) << 24) \
+  | (((x) & 0x000000000000ff00ull) << 40) \
+  | (((x) & 0x00000000000000ffull) << 56))
+
+class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC LeveldbWrapper
```
fujitatomoya, line 46: having this wrapper to absorb the difference of database file structure against sqlite3. just a comment to confirm that we have the same understanding.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/leveldb/leveldb_storage.cpp
```diff
@@ -0,0 +1,387 @@
...
+{
+  if (topic_ldb_map_.find(topic.name) == topic_ldb_map_.end()) {
+    // The "/" in topic name will be replaced by '_' for directory name
+    // e.g. /ns1/topic => _ns1_topic
+    std::string dir_name = topic.name;
+    std::replace(dir_name.begin(), dir_name.end(), '/', '_');
+    std::shared_ptr<class LeveldbWrapper> ldb_wrapper =
+      std::make_shared<class LeveldbWrapper>(relative_path_, topic.name, dir_name, true);
+
+
```
fujitatomoya, line 249: we can delete this line.

fujitatomoya (changes requested):

leveldb_vendor/README.md
```diff
@@ -0,0 +1,3 @@
+# leveldb_vendor
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yeah.  
I refer to `https://github.com/ros2/rosbag2/tree/master/zstd_vendor`. In this directory, there is a README.md.  
Actually, it seems that there is no unique rule for *_vendor.

leveldb_vendor/leveldb_cmakefile.diff
```diff
@@ -0,0 +1,13 @@
+diff -Nur a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt	2019-05-04 04:24:26.000000000 +0800
++++ b/CMakeLists.txt	2020-11-18 09:30:19.275999814 +0800
+@@ -94,6 +94,9 @@
+   add_compile_options(-fvisibility=hidden)
+ endif(BUILD_SHARED_LIBS)
+ 
++# Must be included before CMAKE_INSTALL_INCLUDEDIR is used.
++include(GNUInstallDirs)
```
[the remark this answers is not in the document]
Barry-Xu-2018: Yes. 
Master branch has fixed this problem. But currently the latest version (1.22) doesn't include this fixing.  
After the new version is released, we can remove this.

rosbag2_storage_default_plugins/CMakeLists.txt
```diff
@@ -28,19 +28,32 @@ find_package(rosbag2_storage REQUIRED)
 find_package(sqlite3_vendor REQUIRED)
 find_package(SQLite3 REQUIRED)  # provided by sqlite3_vendor
 find_package(yaml_cpp_vendor REQUIRED)
+find_package(leveldb_vendor REQUIRED)
```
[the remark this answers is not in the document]
Barry-Xu-2018: > but I am curious that is there any way to add vendor plugin w/o changing any mainline code? if vendor implementation can be plugin, it should be easy to do that w/o knowing rosbag2 framework part. @emersonknapp is that why you mentioned about re-organizing / re-constructing the repository, correct?

Agree.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/leveldb/leveldb_storage.cpp
```diff
@@ -0,0 +1,387 @@
...
+    case rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY:
+      return "READ_ONLY";
+    case rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE:
+      return "READ_WRITE";
+    case rosbag2_storage::storage_interfaces::IOFlag::APPEND:
+      return "APPEND";
+    default:
+      return "UNKNOWN";
+  }
+}
```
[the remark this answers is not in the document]
Barry-Xu-2018: After re-organizing rosbag2, I will do this.

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/leveldb/leveldb_wrapper.hpp
Barry-Xu-2018: Yeah. This wrapper wrap the detailed operation (open, read, etc) of leveldb.

fujitatomoya:
@emersonknapp 
CC: @Karsten1987 @Barry-Xu-2018 

in last ROS2 tooling WG, you mentioned about re-construct the repository structure, divide them into framework part and vendor implementation. it that in progress or any activities? if not, is that possible to merge this into master before re-construction?

emersonknapp:
> If that's no limitation and we can provide a working solution with the current interface, I am in favor of adding it as a Tier1 plugin.

Ok - I saw Karsten's comment here. I still would personally prefer that the plugin was in a separate repository, but we can move forward with this PR for now as-is. I will review it today

clalancette:
>  I still would personally prefer that the plugin was in a separate repository, but we can move forward with this PR for now as-is. I will review it today

I will say that I would also greatly prefer this to be in a separate repository.  We are well north of 300 packages in the core ROS 2, and it is getting worse all the time.  This slows everything down, from `colcon` operations (the more packages, the longer it takes to build the dependency tree), to builds (more code to build), to tests (more tests to run), to development itself (there are just more tests that can fail and cause the build to go yellow/red).

The whole point of having a plugin based architecture is to enable hosting this stuff externally, in my opinion.  So I'd heavily favor putting this in a separate repository.  If someday we want to *replace* sqlite3 with leveldb as the default, we can consider that separately.

In the end, I'll leave it up to the tooling working group/@Karsten1987 to decide, but I wanted to leave my thoughts here.

emersonknapp:
> The whole point of having a plugin based architecture is to enable hosting this stuff externally, in my opinion. So I'd heavily favor putting this in a separate repository. If someday we want to replace sqlite3 with leveldb as the default, we can consider that separately.

This is my point exactly, I completely agree. I think it's easier to transfer ownership of leveldb to ros-tooling or ros2/rosbag2 in the future than it would be to extract it later.

Unfortunately I don't think @Karsten1987 is available right now to weigh in, I think he's on holiday.

Again, it would be trivial to port the existing plugin to ros2/rosbag2 later, but difficult to remove it, so in the interest of making a decision that doesn't block future decisions, let's move forward with having leveldb live in a different repository for now. It can be bloom-released into rolling and everything.

fujitatomoya:
@clalancette @emersonknapp 

we don't want to make trouble here, so if that's the decision, we are good to go with it.

@Barry-Xu-2018 

let's keep this in our repository until the time comes. let's discuss online today.

Barry-Xu-2018:
FYI.  

I have moved codes to my repository [rosbag2_storage_leveldb_plugin](https://github.com/Barry-Xu-2018/rosbag2_storage_leveldb_plugin) and add test codes.

emersonknapp:
Based on the above decision to host leveldb storage in a separate repository, I will close this pull request for now.

PR #568, opened 2020-11-24
SQLite storage optimized by default

adamdbrw:
With optimized settings applied, performance of storage writing is significantly increased. This reduces the problem of messages being lost when there is a high throughput of recorded data.

image(2)

Note that before this PR optimization is already accessible to users through `--storage-config-file settings`. This PR makes storage optimization the default. 

Includes an option to use former settings: `--resilient-storage-writing`

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,13 @@ read:
 write:
   pragmas: <list of pragma settings for read/write>
 ```
-Please refer to [documentation of pragmas](https://www.sqlite.org/pragma.html).
+
+By default, SQLite settings are significantly optimized for performance.
+This might have consequences of bag data being corrupted after an application or system-level crash.
+If increased crash-caused corruption resistance is necessary, use `--resilient-storage-writing` flag.
+
 Settings are fully exposed to the user and should be applied with understanding.
+Please refer to [documentation of pragmas](https://www.sqlite.org/pragma.html).
 
 An example configuration file could look like this:
 
--- a/README.md
+++ b/README.md
@@ -128,7 +128,8 @@ write:
 
 By default, SQLite settings are significantly optimized for performance.
 This might have consequences of bag data being corrupted after an application or system-level crash.
-If increased crash-caused corruption resistance is necessary, use `--resilient-storage-writing` flag.
+This consideration only applies to current bagfile in case bag splitting is on (through `--max-bag-*` parameters).
+If increased crash-caused corruption resistance is necessary, use `resilient` option for `--storage-preset-profile` setting.
 
 Settings are fully exposed to the user and should be applied with understanding.
 Please refer to [documentation of pragmas](https://www.sqlite.org/pragma.html).
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -103,10 +103,12 @@ class RecordVerb(VerbExtension):
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.'
         )
         parser.add_argument(
-            '--resilient-storage-writing', action='store_true',
-            help='Indicate preference for avoiding data corruption in case of crashes, '
-                 'at the cost of performance. Setting this flag disables optimization '
-                 'settings for storage (the defaut). When set, this flag will override '
+            '--storage-preset-profile', type=str, default='none', choices=['none', 'resilient'],
+            help='Select a configuration preset for storage.'
+                 'resilient (sqlite3):'
+                 'indicate preference for avoiding data corruption in case of crashes,'
+                 'at the cost of performance. Setting this flag disables optimization settings '
+                 'for storage (the defaut). This flag settings can still be overriden by '
                  'corresponding settings in the config passed with --storage-config-file.'
         )
         parser.add_argument(
@@ -178,7 +180,7 @@ class RecordVerb(VerbExtension):
             topics=args.topics,
             include_hidden_topics=args.include_hidden_topics,
             qos_profile_overrides=qos_profile_overrides,
-            resilient_storage_writing=args.resilient_storage_writing,
+            storage_preset_profile=args.storage_preset_profile,
             storage_config_file=storage_config_file)
 
         if os.path.isdir(uri) and not os.listdir(uri):
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -91,6 +91,13 @@ class RecordVerb(VerbExtension):
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.'
         )
         parser.add_argument(
+            '--resilient-storage-writing', action='store_true',
+            help='Indicate preference for avoiding data corruption in case of crashes, '
+                 'at the cost of performance. Setting this flag disables '
+                 'optimization settings for storage. When set, this flag can override '
+                 'settings in config passed with --storage-config-file.'
+        )
+        parser.add_argument(
             '--storage-config-file', type=FileType('r'),
             help='Path to a yaml file defining storage specific configurations. '
                  'For the default storage plugin settings are specified through syntax:'
@@ -154,6 +161,7 @@ class RecordVerb(VerbExtension):
             topics=args.topics,
             include_hidden_topics=args.include_hidden_topics,
             qos_profile_overrides=qos_profile_overrides,
+            resilient_storage_writing=args.resilient_storage_writing,
             storage_config_file=storage_config_file)
 
         if os.path.isdir(uri) and not os.listdir(uri):
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -93,9 +93,9 @@ class RecordVerb(VerbExtension):
         parser.add_argument(
             '--resilient-storage-writing', action='store_true',
             help='Indicate preference for avoiding data corruption in case of crashes, '
-                 'at the cost of performance. Setting this flag disables '
-                 'optimization settings for storage. When set, this flag can override '
-                 'settings in config passed with --storage-config-file.'
+                 'at the cost of performance. Setting this flag disables optimization '
+                 'settings for storage (the defaut). When set, this flag will override '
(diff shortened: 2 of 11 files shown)


adamdbrw:
@mjeronimo and/or @Karsten1987, could you review this one?

ros2bag/ros2bag/verb/record.py
```diff
@@ -90,6 +90,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '--qos-profile-overrides-path', type=FileType('r'),
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.'
         )
+        parser.add_argument(
+            '--resilient-storage-writing', action='store_true',
+            help='Indicate preference for avoiding data corruption in case of crashes, '
```
mjeronimo, line 95: Could clarify here that --resilient-storage-writing overrides only corresponding flags; additional flags from the config file remain in the pragmas list. I believe this is what is meant by "can override," but perhaps we can make that sentence a bit more explicit. For example, "When set, this flag will override corresponding settings in the config passed with --storage-config-file."

mjeronimo: Related, I'd like to see a test for the above behavior to ensure that overriding of params in a config file by using --resilient-storage-writing is occurring as expected.

adamdbrw: Applied - take a look at the unit test. I preferred to extend the sqlite storage API to enable querying of current settings (which seems useful anyway) because testing otherwise would be awkward.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -322,6 +322,28 @@ TEST_F(StorageTestFixture, loads_config_file) {
...
+    journal_setting + ", " + synchronous_setting + ", " + not_overriden_setting + "]\n";
+  const auto writable_storage = std::make_unique<rosbag2_storage_plugins::SqliteStorage>();
+  auto options = make_storage_options_with_config(to_be_overriden_yaml, kPluginID);
+  options.resilient_storage_writing = true;
+  writable_storage->open(options, rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE);
+
+  // resilient_storage_writing should replace "some_value" with "wal
+  EXPECT_EQ(writable_storage->get_storage_setting("journal_mode"), "wal");
+
+  // resilient_storage_writing should replace "another_value" with "2"
```
mjeronimo, line 340: Comment does not match code. The comment mentions setting "another_value" to "2", while the code is checking against "1".

adamdbrw: The correct value for default is NORMAL (returned as 1), so the comment is mistaken (commented before I checked) and the code is right.

Fixed

https://www.sqlite.org/pragma.html#pragma_synchronous

mjeronimo (approved):

mjeronimo:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13181/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8112/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10897/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13195/)

emersonknapp (changes requested):
I'm back from long holiday weekend - sorry for the slow response. My main comment is the structural concern of making the storage settings profiles more extensible.

ros2bag/ros2bag/verb/record.py
```diff
@@ -90,6 +90,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '--qos-profile-overrides-path', type=FileType('r'),
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.'
         )
+        parser.add_argument(
+            '--resilient-storage-writing', action='store_true',
```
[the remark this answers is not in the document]
adamdbrw: @emersonknapp that would likely be better in the long run, if there are sensible presets other than optimized/safer. Is this a blocker for this PR? The primary goal of this PR is to change the default optimization for the storage.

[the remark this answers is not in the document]
adamdbrw: @emersonknapp I am fine with either, please let me know if I should put it into this PR or a separate one.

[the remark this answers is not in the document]
emersonknapp: I would prefer to do this change now - the `resiliant_storage_writing` value is the sort of thing that seems likely to linger even though it is an overly narrow API. Calling is a `preset_config` probably also does a good job of clarifying that the `storage-config-file` overrides these values.

If you will do the followup change, though, I won't block this PR.

README.md
```diff
@@ -125,8 +125,13 @@ read:
 write:
   pragmas: <list of pragma settings for read/write>
 ```
-Please refer to [documentation of pragmas](https://www.sqlite.org/pragma.html).
+
+By default, SQLite settings are significantly optimized for performance.
+This might have consequences of bag data being corrupted after an application or system-level crash.
+If increased crash-caused corruption resistance is necessary, use `--resilient-storage-writing` flag.
```
emersonknapp, line 131: Should note that `any values specified in the `--storage-config-file` take precedence over default or preset values.

README.md
```diff
@@ -125,8 +125,13 @@ read:
 write:
   pragmas: <list of pragma settings for read/write>
 ```
-Please refer to [documentation of pragmas](https://www.sqlite.org/pragma.html).
+
+By default, SQLite settings are significantly optimized for performance.
+This might have consequences of bag data being corrupted after an application or system-level crash.
```
emersonknapp, line 130: It would be good to clarify that only the current bagfile has the possibility to be corrupted in this case, any already-split files are safe once finalized, so the maximum data loss is bounded by the maximum bag (time or size)

ros2bag/ros2bag/verb/record.py
```diff
@@ -90,6 +90,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            help='Indicate preference for avoiding data corruption in case of crashes, '
+                 'at the cost of performance. Setting this flag disables optimization '
+                 'settings for storage (the defaut). When set, this flag will override '
+                 'corresponding settings in the config passed with --storage-config-file.'
```
emersonknapp, line 98: I think that this should not have top precedence, logically. It should modify the baseline settings, anything the user passes _explicitly_ (the `storage-config-file`) should always be respected.

emersonknapp (changes requested):

adamdbrw:
Fixing the merge - went a bit sideways

ros2bag/ros2bag/verb/record.py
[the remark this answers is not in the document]
adamdbrw: @emersonknapp I made an attempt to implement it, but there are some drawbacks to each solution I came up with. There is currently no ros2_storage interface exposed to the python (except for some structs) and, since this API requires plugin-specific return value, one would need to instantiate storage to return these profiles. Which would be the only case when this is done, but the storage remains not-open and is discarded after the call. 

Another solution would be to have a separate object hierarchy (and a factory) just for the profiles.

Yet another would be to simply keep a dictionary of plugin name - preset profiles at the top level, and return based on which plugin is meant to be used (as in the argument).

Each solution has drawbacks and I would love to discuss that with you. Did you have a particular design in mind? We could also set up a call if that would make things easier, let me know.

emersonknapp:
Running for evidence - we can override my comments for now if @mjeronimo you'd rather move forward this way and have us address that stuff later

Gist: https://gist.githubusercontent.com/emersonknapp/c32bfc13fc1341e9bdaa088b4ef23bf1/raw/34a8fb38e8af97a7e3f68337d715b0a0c1411b52/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_storage rosbag2_storage_default_plugins rosbag2_transport
TEST args: --packages-select ros2bag rosbag2_storage rosbag2_storage_default_plugins rosbag2_transport
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7389
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13345/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8273/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13397/)

adamdbrw:
@emersonknapp I addressed the API issue - should remain the same now and we are free to add new presets (or change the way we acquire presets from a simple static list as it is currently). 

Also, config file is now overriding the preset as you suggested. I also added explanation that the crash-caused corruption risk only applies to current bagfile if splitting is on.

Looking into Windows issues now.

emersonknapp (approved):
LGTM! Thanks for the changes. Running CI now

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/b8abd49ac4aa32d70feee3fa3b3b06b3/raw/34a8fb38e8af97a7e3f68337d715b0a0c1411b52/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_storage rosbag2_storage_default_plugins rosbag2_transport rosbag2_tests
TEST args: --packages-select ros2bag rosbag2_storage rosbag2_storage_default_plugins rosbag2_transport rosbag2_tests
Job: ci_launcherci_launcher ran: https://ci.ros2.org/job/ci_launcher/7421
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8305/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11108/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13433/)

PR #563, opened 2020-11-18
Add back rosbag2_cpp::StorageOptions as deprecated

jacobperron:
The struct was removed in https://github.com/ros2/rosbag2/pull/493, but in order
to avoid a hard-break for users coming from Foxy I've added it back with a deprecation
warning.

PR #569, opened 2020-11-24
Add missing distro in github CI

agutenkunst:
Seems to be missing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Test rosbag2
 on:
-  # pull_request:
+  pull_request:
   push:
     branches:
       - master
@@ -29,6 +29,7 @@ jobs:
           sqlite3_vendor
           rosbag2_test_common
           rosbag2_tests
+        target-ros2-distro: foxy
     - uses: actions/upload-artifact@v1
       with:
         name: colcon-logs


dabonnie:
Could you please uncomment [line 3](https://github.com/ros2/rosbag2/pull/569/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88L3)? This way the workflow should run and we can see the result in the checks here.

agutenkunst:
Ok I assumed that master was developed to be compiled with foxy but I am not able to comprehend the current development practices. See https://github.com/ros2/rosbag2/pull/563#issuecomment-733153338 . Feel free to close if master is not supposed to be compiled with foxy or adapt to your needs.

Karsten1987:
The rosbag2 master branch is targeting all other ros2 core master branches (rcl, rclcpp, etc.). We have a GH action on the foxy branch which is targeting ros2 foxy:
https://github.com/ros2/rosbag2/blob/foxy/.github/workflows/test.yml#L16

I'll close that PR as I think there's nothing to do here, but thanks anyway for the contribution. Please re-open it or continue commenting on the closed PR if you feel there's still something wrong.

emersonknapp:
That's interesting - I am wondering how it was passing without that argument - it passed on the CR where I merged it.

The confusion that we're seeing is that `master` branch of `rosbag2` targets the `master` version of the ROS 2 core. Because the ROS 2 core has move on beyond Foxy in its development, building against a Foxy release is not guaranteed to work. This CI build could be useful against Rolling, but would be most guaranteed "correct" against a full source build of the ROS 2 core. 
If, however, you are a robot team building a robot against a stable distribution of ros2, you would want to build against e.g. Foxy in your application build, so that no new changes break you as you build your product - I think that's where the differences of expectation come in.

Hopefully that helps clarify the situation a little bit.

This action-ros-ci build was currently disabled on pull requests because it was not working reliably, so we were not using it to evaluate contributions.

PR #572, opened 2020-11-25
Enable sanitizers only if code actually can run with them

sloretz:
This uses check_cxx_source_runs instead of check_cxx_source_compiles so that sanitizers are disabled in an environment that they cannot run in. I think this will disable the sanitizers on the buildfarm because they're run in docker containers without the ptrace capability.

This should fix the test `test_ros2_message` failing in Foxy CI, meaning it should also be backported to Foxy.

http://build.ros2.org/job/Fci__nightly-debug_ubuntu_focal_amd64/210/testReport/(root)/projectroot/test_ros2_message/

```
...
[----------] Global test environment tear-down
[==========] 11 tests from 1 test case ran. (18 ms total)
[  PASSED  ] 11 tests.
==27433==LeakSanitizer has encountered a fatal error.
==27433==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==27433==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
-- run_test.py: return code 23
```


emersonknapp (approved):

emersonknapp:
this lgtm - pending a CI run that shows it works right.

Is there any CI environment that is running these sanitizers? Or, can we enable PTRACE in the build containers with `--cap-add=SYS_PTRACE`? Would be nice to know we're getting some value from the sanitizers in an enforceable way

jacobperron:
Friendly ping. Looks like this PR has been forgotten about.

emersonknapp:
I'm fine with just merging this - though it would be nice to see some CI run that "proves" it works as intended. @sloretz any thoughts on that?

Issue #528, opened 2020-10-02
test_record__rmw_fastrtps_cpp failing


tfoote:
## Description
The nightly job has started failing for test_record using rmw_fastrtps_cpp

https://ci.ros2.org/job/nightly_linux_repeated/2063/testReport/(root)/projectroot/test_record__rmw_fastrtps_cpp/

### Expected Behavior
Nightly CI passes

### Actual Behavior
The CI failed with a segfault. 

```
[==========] Running 7 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 7 tests from RecordIntegrationTestFixture
[ RUN      ] RecordIntegrationTestFixture.published_messages_from_multiple_topics_are_recorded
[NON-XML-CHAR-0x1B][37;1m2020-10-01 13:26:57.035 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mPARTICIPANT[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mType with the same name already exists:rcl_interfaces::srv::dds_::GetParameters_Request_[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mregisterType[NON-XML-CHAR-0x1B][m
[NON-XML-CHAR-0x1B][37;1m2020-10-01 13:26:57.035 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mPARTICIPANT[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mType with the same name already exists:rcl_interfaces::srv::dds_::GetParameters_Response_[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mregisterType[NON-XML-CHAR-0x1B][m
[INFO] [1601558817.041625662] [rosbag2_transport]: Listening for topics...
[INFO] [1601558817.045021523] [rosbag2_transport]: Subscribed to topic '/string_topic'
[INFO] [1601558817.146217994] [rosbag2_transport]: Subscribed to topic '/array_topic'
[INFO] [1601558817.146253032] [rosbag2_transport]: All requested topics are subscribed. Stopping discovery...
[       OK ] RecordIntegrationTestFixture.published_messages_from_multiple_topics_are_recorded (231 ms)
[...]

PR #576, opened 2020-12-03
Stabilize test_record by reducing copies of executors and messages

emersonknapp:
Fixes https://github.com/ros2/rosbag2/issues/528

Tested locally with 

```
stress -c 16
# and
colcon test --packages-select rosbag2_transport --retest-until-fail 10
```

As well as 1400 successful sequential runs (closed at that point, not failed) of `test_record__rmw_fastrtps_cpp`

Introduces 2 changes which both caused issues:

### Executor copies (wait_for.hpp)

Every time we called spin_some, a new executor was constructed. This would happen potentially a few thousand times per test. Given that the global context was not shutdown during this time, it would accumulate `on_shutdown_callbacks_`, one from each executor that was created. When this list was very large we would occasionally get `std::bad_function_call` for one of the callbacks in the shutdown.

With this change, the global `Context` has only 2 items in `on_shutdown_callbacks_`, and I am not able to reproduce `bad_function_call` any more.

After this change I still see a segfault, usually after 100-500 iterations of (for example) `RecordIntegrationTestFixture.records_sensor_data`. I'm still investigating this, but it is a separate error.

### STL container copies

The `MockSequentialWriter` used in `test_record` would pass its internal containers back to viewers by value. This created  many copies of the `shared_ptr<SerializedBagMessage>`s. Upon running repeatedly, this copy operation would occasionally end in a segfault. Since outside the `MockSequentialWriter`, the values are only observed, we are able to avoid that situation by never copying the values and returning const references to the existing structures.

### Conclusion

Though it's a little unclear _exactly_ what caused either issue at the cause end, both changes are best-practices more correct code and avoid the situations that brought about the problems, which improves performance and stability of these tests.

Recommended followups:
* find out why `rclcpp::Context` throws `bad_function_call` on shutdown when `on_shutdown_callbacks_` is very large. A simple repro case should be easy enough to construct

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/wait_for.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/wait_for.hpp
@@ -26,11 +26,13 @@ bool spin_and_wait_for(Timeout timeout, const Node & node, Condition condition)
 {
   using clock = std::chrono::system_clock;
   auto start = clock::now();
+  rclcpp::executors::SingleThreadedExecutor exec;
+  exec.add_node(node);
   while (!condition()) {
     if ((clock::now() - start) > timeout) {
       return false;
     }
-    rclcpp::spin_some(node);
+    exec.spin_some();
   }
   return true;
 }
--- a/rosbag2_transport/test/rosbag2_transport/mock_sequential_writer.hpp
+++ b/rosbag2_transport/test/rosbag2_transport/mock_sequential_writer.hpp
@@ -51,17 +51,17 @@ public:
     messages_per_topic_[message->topic_name] += 1;
   }
 
-  std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> get_messages()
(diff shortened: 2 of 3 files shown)


thomas-moulard (approved):

jaisontj (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/d47520e678e57235041ab30236612b47/raw/bce3c825c328d028c5604e18798bf07c217c9976/ros2.repos
Job: ci_launcher

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13220/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8150/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10935/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13236/)

jacobperron:
@emersonknapp Thanks for the fix!

PR #579, opened 2020-12-05
Patch zstd 1.4.4 to include cmake_minimum_version bump to 2.8.12

### LINKED ISSUE #573: cmake 3.19.1 deprecated a policy for cmake versions < 2.8.12
With a recent update to cmake, a policy which whose behavior was split in 2.8.12 has now been deprecated (see https://cmake.org/cmake/help/latest/policy/CMP0022.html for information). This causes a regular warning for windows builds on ci.ros2.org because those jobs are always downloading the latest version of cmake.

I submitted a PR to `zstd`, which was accepted and merged to update the cmake version. However, I'm not sure when it will get released. When it becomes available, `zstd_vendor` may want to update it's release version.

Upstream PR
https://github.com/facebook/zstd/pull/2399

emersonknapp:
Fixes #573

This eliminates CMake warnings on the Windows build (see https://ci.ros2.org/job/ci_windows/13244/cmake/#issuesContent)

We can remove the patch when a new zstd version gets released

### DIFF
--- a/zstd_vendor/CMakeLists.txt
+++ b/zstd_vendor/CMakeLists.txt
@@ -39,7 +39,9 @@ macro(build_zstd)
       -DZSTD_BUILD_STATIC=OFF
       -DZSTD_BUILD_SHARED=ON
       -DZSTD_BUILD_PROGRAMS=OFF
-      ${extra_cmake_args})
+      ${extra_cmake_args}
+    # Note: zstd v1.4.6 will include the following fix. When that is released, upgrade and remove this patch.
+    PATCH_COMMAND patch -p1 -d . < ${CMAKE_CURRENT_SOURCE_DIR}/cmake_minimum_required_2.8.12.patch)
 
   install(
     DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}_install/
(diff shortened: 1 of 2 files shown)


emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13225/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8155/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10940/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13253/)

clalancette (approved):

zmichaels11 (approved):

zstd_vendor/CMakeLists.txt
```diff
@@ -39,7 +39,9 @@ macro(build_zstd)
       -DZSTD_BUILD_STATIC=OFF
       -DZSTD_BUILD_SHARED=ON
       -DZSTD_BUILD_PROGRAMS=OFF
-      ${extra_cmake_args})
+      ${extra_cmake_args}
+    # Note: zstd v1.4.6 will include the following fix. When that is released, upgrade and remove this patch.
+    PATCH_COMMAND patch -p1 -d . < ${CMAKE_CURRENT_SOURCE_DIR}/cmake_minimum_required_2.8.12.patch)
```
ivanpauno, line 44: should something like in https://github.com/ros2/rviz/blob/2206b9257716c05dc4d5eaa7ffae12e434f151ca/rviz_ogre_vendor/CMakeLists.txt#L157 be done here?

when setting up a windows machine for debugging an issue, I needed to install patch from choco but the places that were using `find_package(Patch)` were working

emersonknapp: https://github.com/ros2/rosbag2/issues/632 is the planned solution, which should also work well without this dependency

ivanpauno: Sounds good, we will need something like that before the Galactic release as the current build from source instructions doesn't seem to be working on Windows

Issue #602, opened 2021-01-19
Record crashes with locked database


adamdbrw:
## Description

The issue with multi-threaded writing to db causes crashes due to a race condition. The root seems to be similar to what has been reported in #595, but it is a more general issue - occurs on one machine as well and does not require specific conditions as described in #595.

### Expected Behavior

Recording copes well with multiple topics being discovered while there is also data being written to database. No crash occurs regardless of how much data is being written and how many topics are being discovered.

### Actual Behavior

An exception is raised when recording. This occurs (seemingly) in a deterministic way when the number of topics is high enough (e.g. 500).

```
[ros2-1] [INFO] [1611056622.945250784] [rosbag2_transport]: Subscribed to topic '/rosbag2_performance_benchmarking_node/benchmarking_small_67'
[ros2-1] [INFO] [1611056622.948577997] [rosbag2_transport]: Subscribed to topic '/rosbag2_performance_benchmarking_node/benchmarking_small_171'
[ros2-1] [INFO] [1611056622.951884593] [rosbag2_transport]: Subscribed to topic '/rosbag2_performance_benchmarking_node/benchmarking_small_71'
[ros2-1] terminate called after throwing an instance of 'rosbag2_storage_plugins::SqliteException'
[ros2-1]   what():  Error when processing SQL statement. SQLite error (5): database is locked
```

## To Reproduce

The easiest way to reproduce is to use the rosbag2_performance_benchmarking package with https://github.com/ros2/rosbag2/pull/594. Follow the README of the package.
- build with `--cmake_args -DBUILD_ROSBAG2_BENCHMARKS=1`
- run with `ros2 launch rosbag2_performance_benchmarking benchmark_launch.py benchmark:=`ros2 pkg prefix rosbag2_performance_benchmarking`/share/rosbag2_performance_benchmarking/config/benchmarks/test.yaml producers:=`ros2 pkg prefix rosbag2_performance_benchmarking`/share/rosbag2_performance_benchmarking/config/producers/mixed_110Mbs.yaml`

You can manipulate the mixed_110Mbs.yaml file to easily adjust the number of publishers.

## System (please complete the following information)
 - Ubuntu 20
 - Master branch, #c48471e (current when raising the issue)

## Additional context

I also have a PR that fixes this issue: #603. I believe setting sqlite3 to FULLMUTEX as proposed in #595 is a bit too much (since the usual vectors of messages would then lock inside for each single write), although the performance hit is not very substantial either way since the I/O dominates.

[...]

PR #603, opened 2021-01-19
Mutex protection for db writing and stl collections in writer & storage

adamdbrw:
Adresses the https://github.com/ros2/rosbag2/issues/602 issue.

The issue is with that the current locking mechanism on top level (mutexes in writer.cpp in rosbag2_cpp, added with https://github.com/ros2/rosbag2/pull/491) is valid only with assumption that all writes of data and all creation/removal of topics are synchronous. This is not the case since double buffering has been introduced in the writer. This also would not be the case if another, non-sequential writer implementation was added, and would be unnecessary for another, parallel-enabled storage implementation.

This PR however does not remove mutexing on the rosbag2_cpp level, only adds storage implementation level locks - lets discuss/test whether remaining (intermittent) layers are multithreading-ready.

### DIFF
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -17,6 +17,7 @@
 
 #include <atomic>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -108,6 +109,7 @@ private:
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
   rosbag2_storage::StorageFilter storage_filter_ {};
+  std::mutex database_write_mutex_;
 };
 
 }  // namespace rosbag2_storage_plugins
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -17,7 +17,6 @@
 
 #include <atomic>
 #include <memory>
-#include <mutex>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -109,7 +108,6 @@ private:
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
   rosbag2_storage::StorageFilter storage_filter_ {};
-  std::mutex database_write_mutex_;
 };
 
 }  // namespace rosbag2_storage_plugins
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -109,6 +109,10 @@ private:
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
   rosbag2_storage::StorageFilter storage_filter_ {};
+
+  // This mutex is necessary to protect:
+  // a) database access (this could also be done with FULLMUTEX), but see b)
+  // b) topics_ collection - since we could be writing and reading it at the same time
   std::mutex database_write_mutex_;
 };
 
--- a/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
+++ b/rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
@@ -22,6 +22,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "rcpputils/thread_safety_annotations.hpp"
 #include "rcutils/types.h"
 #include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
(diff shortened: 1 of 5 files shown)


adamdbrw:
The failing test (due to a timeout) for reading looks unrelated and the same test seems to be failing on master as well

mjeronimo:
My understanding so far:
* The mutex usage at the rosbag2 writer level are only valid in the case that the back-end implementation is synchronous (it basically assumes a synchronous implementation). 
* Currently, the sqlite3 back-end opens sqlite3 databases using SQLITE_OPEN_NOMUTEX, which means that separate threads are allowed to use SQLite at the same time, as long as each thread is using a different database connection. However, since we are seeing this crash, I'd assume that it is not true that each thread is using a different database connection.
* The proposed solution keeps the above while introducing another mutex at the Writer level of the sqlite3 back-end, used for any database write operations, effectively compensating for multiple threads using the same database connection. 

IMO,
* The mutex at the rosbag2 writer level should be removed and the burden of thread safety should be on the back-end implementation. 
* Regarding the sqlite3 back-end, if performance does not seem to be an issue, why not push the responsibility further by opening the databases with SQLITE_OPEN_FULLMUTEX and not having a mutex at the Writer level? Then, the responsibility is only in sqlite3 (versus what would be in three places). Vectors of messages seemed to be a potential concern, but it isn't clear if there is a performance issue in practice (because of the I/O dominating).

adamdbrw:
Ok this is ready to be merged. I tested that no locking occurs as well with version (in a case that was apparently deterministically causing it before this PR). @mjeronimo.

adamdbrw:
@csli8192 this fix does not include discovery mechanism changes which you earlier suggested. If they are still necessary for your use case, I believe is best to merge them in separately.

mjeronimo:
@adamdbrw Thanks for the detailed explanation. I now understand that the intermediate data structures also have to be protected and that it's not enough to just use FULLMUTEX.

mjeronimo (approved):

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -92,6 +92,7 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   void fill_topics_and_types();
   void activate_transaction();
   void commit_transaction();
+  void write_locked(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message);
```
emersonknapp, line 95: Where is this function used? Since it's private I don't see how it can connect to any existing code, so I'm confused

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
```diff
@@ -42,7 +42,7 @@ SqliteWrapper::SqliteWrapper(
   if (io_flag == rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY) {
     int rc = sqlite3_open_v2(
       uri.c_str(), &db_ptr,
-      SQLITE_OPEN_READONLY | SQLITE_OPEN_NOMUTEX, nullptr);
+      SQLITE_OPEN_READONLY | SQLITE_OPEN_FULLMUTEX, nullptr);
```
emersonknapp, line 45: Is there a regression test to verify that this changes any measurable behavior?

adamdbrw:
We have given it a deeper thought and ran extensive tests. The race condition was still apparently there (in reality no, a separate concurrency problem), and the version 1 of this changeset was the correct one. Why?

SqliteStorage has unordered_map structure holding the topics. It is not a thread safe structure. As it is now, create_topic can be run from one thread and initiate `.emplace`. Meanwhile, the buffer writer thread will attempt to find a topic for the message it is trying to insert, running `.find` on the same structure. As you likely know, this threading scenario is not supported in stl containers.

In particular, for unordered_map, a likely case is rehashing. "If rehashing occurs due to the insertion, all iterators are invalidated. Otherwise iterators are not affected". 

Thus a rare case which we observed in extensive testing with over a thousand of topics (thanks @pijaro). 

Anyway, bottom line is that any stl collections that can be accessed outside of supported model (only 1 thread writes while 0 read, or any number of threads read) should be locked. We don't need to dig deep into details such as rehashing during find - we just need to lock them. And with locking on a member function level, we don't need FULLMUTEX anymore - it is completely redundant.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_wrapper.cpp
adamdbrw: This is not there anymore - but yes, a good catch since it wasn't needed in READ_ONLY

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -92,6 +93,7 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
...
```
emersonknapp, line 96:
```suggestion
  void write_locked(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message) REQUIRES(write_mutex_);
```

emersonknapp (approved):
This looks good - I'm going to (optionally) ask for Clang thread safety annotations, as they will help validate and clarify the logic a little bit.

e.g. https://github.com/ros2/rcpputils/blob/master/include/rcpputils/thread_safety_annotations.hpp#L133

I'm not sure if you need this across the class - but if so, you could guard the two class members that you seem to be synchronizing access to

```
std::unordered_map<std::string, int> topics_ GUARDED_BY(database_write_mutex_);
std::shared_ptr<SqliteWrapper> database_ GUARDED_BY(database_write_mutex_);
```

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
adamdbrw: Done - thank you for suggestions! I didn't know we had such annotations. 
I also rebased this PR on the new master.

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/ca5fc9b7640834b2ec160af39e38a1c2/raw/c00c5de4bbf391319d623fde04f97a2a0200aa1b/ros2.repos
BUILD args: --packages-up-to rosbag2_storage_default_plugins rosbag2_tests
TEST args: --packages-select rosbag2_storage_default_plugins rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7502
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13481/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8389/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11202/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13546/)

adamdbrw:
All looks green, shall we merge this? :]

Issue #559, opened 2020-11-11
Support for regex in defining recorded topics


TimoKettunen:
## Description
rosbag2 command should have an option -e to allow defining recording topics that match regular expression.
ROS(1) has this feature, ROS2 should have parity with ROS, See http://wiki.ros.org/rosbag/Commandline.
## Related Issues
None

## Completion Criteria
giving command 
```
$ ros2 bag record -e "<regex>" 
```
records all topics that match the regex

emersonknapp:
This sounds like a neat feature!

adamdbrw:
Agreed, at least a couple of times I needed this as well

adamdbrw:
@TimoKettunen, @emersonknapp  turns out this would be very useful for our benchmarking package, so I submitted a PR draft. Hopefully this feature will soon be available.

adamdbrw:
This is resolved with #604

Issue #560, opened 2020-11-11
Support for excluding topics


TimoKettunen:
## Description
rosbag2 command should have an option -x to allow excluding recording topics that match regular expression.
ROS(1) has this feature, ROS2 should have parity with ROS, See http://wiki.ros.org/rosbag/Commandline.
## Related Issues
None

## Completion Criteria
giving command 
```
$ ros2 bag record -a -x "<regex>" 
```
records all topics except those that match the regex

## Testing Notes / Suggestions

emersonknapp:
Yes, confirmed that this is a feature gap. PRs welcome, not sure when we'll have bandwidth to take this on at the moment. It shouldn't be too difficult, probably a good first issue for someone.

adamdbrw:
This is resolved with #604

PR #604, opened 2021-01-19
Recorder --regex and --exclude options

adamdbrw:
Resolves issues #559 and #560. This is a minimal implementation that brings the regex filtering feature to rosbag2.

Implements ros2 bag record -e / --regex and -x / --exclude options to specify how topics are recorded.
- the --regex option is exclusive with -all and specifying topics
- the --exclude option can be applied together with either --regex or --all, but is exclusive with explicit topic list

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -30,9 +30,13 @@ class RecordVerb(VerbExtension):
     def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '-a', '--all', action='store_true',
-            help='recording all topics, required if no topics are listed explicitly.')
+            help='recording all topics, required if no topics '
+            'are listed explicitly or through a regex')
         parser.add_argument(
             'topics', nargs='*', default=None, help='topics to be recorded')
+        parser.add_argument(
+            '-x', '--regex', default='', help='recording only topics '
+            'matching provided regular expression')
         parser.add_argument(
             '-o', '--output',
             help='destination of the bagfile to create, \
@@ -122,11 +126,11 @@ class RecordVerb(VerbExtension):
 
     def main(self, *, args):  # noqa: D102
         # both all and topics cannot be true
-        if args.all and args.topics:
-            return print_error('Invalid choice: Can not specify topics and -a at the same time.')
-        # both all and topics cannot be false
-        if not(args.all or (args.topics and len(args.topics) > 0)):
-            return print_error('Invalid choice: Must specify topic(s) or -a')
+        if (args.all and (args.topics or args.regex)) or (args.topics and args.regex):
+            return print_error('Must specify only one option out of topics, regex or -a')
+        # one out of "all", "topics" and "regex" must be true
+        if not(args.all or (args.topics and len(args.topics) > 0) or (args.regex)):
+            return print_error('Invalid choice: Must specify topic(s), regex or -a')
 
         uri = args.output or datetime.datetime.now().strftime('rosbag2_%Y_%m_%d-%H_%M_%S')
 
@@ -178,6 +182,7 @@ class RecordVerb(VerbExtension):
             max_bagfile_duration=args.max_bag_duration,
             max_cache_size=args.max_cache_size,
             topics=args.topics,
+            regex=args.regex,
             include_hidden_topics=args.include_hidden_topics,
             qos_profile_overrides=qos_profile_overrides,
             storage_preset_profile=args.storage_preset_profile,
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -35,7 +35,7 @@ class RecordVerb(VerbExtension):
         parser.add_argument(
             'topics', nargs='*', default=None, help='topics to be recorded')
         parser.add_argument(
-            '-x', '--regex', default='', help='recording only topics '
+            '-e', '--regex', default='', help='recording only topics '
             'matching provided regular expression')
         parser.add_argument(
             '-o', '--output',
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -131,16 +131,17 @@ class RecordVerb(VerbExtension):
     def main(self, *, args):  # noqa: D102
         # both all and topics cannot be true
         if (args.all and (args.topics or args.regex)) or (args.topics and args.regex):
-            return print_error('Must specify only one option out of topics, regex or -a')
+            return print_error('Must specify only one option out of topics, --regex or --all')
         # one out of "all", "topics" and "regex" must be true
         if not(args.all or (args.topics and len(args.topics) > 0) or (args.regex)):
-            return print_error('Invalid choice: Must specify topic(s), regex or -a')
+            return print_error('Invalid choice: Must specify topic(s), --regex or --all')
 
         if args.topics and args.exclude:
-            return print_error('Exclude argument only works with -a or -e')
+            return print_error('--exclude argument cannot be used when specifying a list '
+                               'of topics explicitly')
 
         if args.exclude and not(args.regex or args.all):
-            return print_error('Exclude argument requires either -a or -e')
+            return print_error('--exclude argument requires either --all or --regex')
 
         uri = args.output or datetime.datetime.now().strftime('rosbag2_%Y_%m_%d-%H_%M_%S')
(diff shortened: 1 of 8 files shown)


mjeronimo (approved):

adamdbrw:
@mjeronimo can I ask for a review and if this is alright, to run jobs?

mjeronimo:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13487/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8394/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11208/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13552/)

ros2bag/ros2bag/verb/record.py
```diff
@@ -122,11 +130,17 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
 
     def main(self, *, args):  # noqa: D102
         # both all and topics cannot be true
-        if args.all and args.topics:
-            return print_error('Invalid choice: Can not specify topics and -a at the same time.')
-        # both all and topics cannot be false
-        if not(args.all or (args.topics and len(args.topics) > 0)):
-            return print_error('Invalid choice: Must specify topic(s) or -a')
+        if (args.all and (args.topics or args.regex)) or (args.topics and args.regex):
+            return print_error('Must specify only one option out of topics, regex or -a')
```
emersonknapp, line 134: In help messages, please use long-form versions for clarity - power users will always find or know the short form

```suggestion
            return print_error('Must specify only one option out of topics, --regex or --all')
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -122,11 +130,17 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+        # one out of "all", "topics" and "regex" must be true
+        if not(args.all or (args.topics and len(args.topics) > 0) or (args.regex)):
+            return print_error('Invalid choice: Must specify topic(s), regex or -a')
```
emersonknapp, line 137:
```suggestion
            return print_error('Invalid choice: Must specify topic(s), --regex or --all')
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -122,11 +130,17 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+
+        if args.topics and args.exclude:
+            return print_error('Exclude argument only works with -a or -e')
```
emersonknapp, line 140:
```suggestion
            return print_error('--exclude argument cannot be used when specifying a list of topics explicitly')
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -122,11 +130,17 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+
+        if args.exclude and not(args.regex or args.all):
+            return print_error('Exclude argument requires either -a or -e')
```
emersonknapp, line 143:
```suggestion
            return print_error('--exclude argument requires either --all or --regex')
```

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -564,7 +564,7 @@ TEST_F(RecordFixture, record_fails_if_both_all_and_topic_list_is_specified) {
   auto error_output = internal::GetCapturedStderr();
 
   EXPECT_THAT(exit_code, Eq(EXIT_FAILURE));
-  EXPECT_THAT(error_output, HasSubstr("Can not specify topics and -a at the same time."));
+  EXPECT_THAT(error_output, HasSubstr("Must specify only one option out of topics, regex or -a"));
```
emersonknapp, line 567: Checking the string exactly isn't very useful, it's just a ["change detector"](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) - this should probably just check that the error output isn't empty

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -574,7 +574,7 @@ TEST_F(RecordFixture, record_fails_if_neither_all_nor_topic_list_are_specified)
   auto output = internal::GetCapturedStderr();
 
   EXPECT_THAT(exit_code, Eq(EXIT_FAILURE));
-  EXPECT_THAT(output, HasSubstr("Invalid choice: Must specify topic(s) or -a"));
+  EXPECT_THAT(output, HasSubstr("Invalid choice: Must specify topic(s), regex or -a"));
```
emersonknapp, line 577: see prior comment on "change detector"

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -79,36 +77,53 @@ void Recorder::record(const RecordOptions & record_options)
...
 std::unordered_map<std::string, std::string>
-Recorder::get_requested_or_available_topics(
-  const std::vector<std::string> & requested_topics,
-  bool include_hidden_topics)
+Recorder::get_requested_or_available_topics(const RecordOptions & record_options)
 {
-  return requested_topics.empty() ?
-         node_->get_all_topics_with_types(include_hidden_topics) :
-         node_->get_topics_with_types(requested_topics);
+  auto tt = record_options.topics.empty() ?
```
emersonknapp, line 102: Please avoid one or two character variable names that aren't loop variables (and even then it doesn't hurt to name those better) - even in relatively short functions these makes for difficult reading. Maybe `desired_topics`?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -79,36 +77,53 @@ void Recorder::record(const RecordOptions & record_options)
...
+    return tt;
+  }
+
+  std::unordered_map<std::string, std::string> filtered_by_regex;
+
+  std::regex topic_regex(record_options.regex);
+  std::regex exclude_regex(record_options.exclude);
+  bool take = record_options.all;
+  for (const auto & kv : tt) {
+    if (!record_options.regex.empty()) {
```
emersonknapp, line 116: It does not seem necessary to check for empty first - `regex_search` will return false if there is nothing in there to be found, and the logic will be a little more clear here without the extra blocks

emersonknapp (changes requested):
Functionally this looks good, but I have some nitpicks on style and language that I'd like to address before merging

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
adamdbrw: I have changed both to only check for whether the string is empty. There are a few other tests that also check the exact string, but I believe it would be incorrect to include them in this PR since it otherwise does not change related functionalities.

emersonknapp: I agree that they shouldn't be changed as part of this review, but it's best to avoid copying bad patterns

emersonknapp (approved):
Looks great - thanks for the updates

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/20bbe7e473b4159fb5f88662b552963c/raw/656e51e4a43294405eae946a887b2d838489c554/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_tests rosbag2_transport
TEST args: --packages-select ros2bag rosbag2_tests rosbag2_transport
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7512
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13492/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8399/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13557/)

purentap:
is this still usable? cant seem rosbag2 cant seem to recognize -x flag for regex

MichaelOrlov:
@purentap Yes `exclude` option is currently supported for the recorder. See https://github.com/ros2/rosbag2/blob/rolling/ros2bag/ros2bag/verb/record.py#L73-L76 for details.

PR #616, opened 2021-01-26
Fix deadlock race condition on compression shutdown

emersonknapp:
When running the test `test_sequential_compression_writer` repeatedly, the test sometimes hangs on shutdown, due to a gap in the threading synchronization between the compressor threads and the main thread.

The order of operations is:
* Worker thread N: `compressor_thread_fn` reads `while(compression_is_running_)` to be `true`, so it enters the while block
* Main thread: `stop_compressor_threads` runs `compression_is_running_=false; compressor_condition_.notify_all()`, notifying any _currently waiting_ compression threads to wake up, which lets them exit after finding no work to do
    * main thread then calls `.join()` on all the threads
* Worker thread N: `compressor_thread_fn`, having entered the while loop successfully, initializes some variables, acquires the `compressor_queue_mutex_`, and calls `compressor_condition_.wait()`
* Because the compressor_thread has called `wait` _after_ the final `notify_all`, it will never receive a notification, and the `.join()` call waits forever for this waiting thread to exit

This PR fixes the problem by synchronizing the checks of both exit conditions so that the main thread can't interleave the notifications in a way that causes the deadlock.

NOTE: I am not sure how to add a repeatable regression test for this - the way I can reproduce it is by running the following, which reliably makes it happen quite quickly (within a second). After my change it'll run indefinitely without failure.

```
build/rosbag2_compression/test_sequential_compression_writer --gtest_repeat=-1 --gtest_throw_on_failure
```

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -170,10 +170,8 @@ void SequentialCompressionWriter::stop_compressor_threads()
     {
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
       compression_is_running_ = false;
-      // Normally you are better off not holding the mutex when calling notify -
-      // but the behavior is still correct, and we need it to synchronize these two variables
-      compressor_condition_.notify_all();
     }
+    compressor_condition_.notify_all();
     for (auto & thread : compression_threads_) {
       thread.join();
     }
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -73,11 +73,18 @@ void SequentialCompressionWriter::compression_thread_fn()
             "Cannot compress message; Writer is not open!"};
   }
 
-  while (compression_is_running_) {
+  while (true) {
     std::shared_ptr<rosbag2_storage::SerializedBagMessage> message;
     std::string file;
     {
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
+      // The mutex needs to be held for checking both exit conditions (shutdown & work available)
+      // If it is not, the main thread could set both compression_is_running_ and notify_all()
+      // on shutdown, after we read running_, and before we call wait().
+      // In this case this thread would have missed the final notification and wait forever.
+      if (!compression_is_running_) {
+        break;
+      }
       compressor_condition_.wait(lock);
       if (!compressor_message_queue_.empty()) {
         message = compressor_message_queue_.front();
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -77,11 +77,9 @@ void SequentialCompressionWriter::compression_thread_fn()
     std::shared_ptr<rosbag2_storage::SerializedBagMessage> message;
     std::string file;
     {
+      // This mutex synchronizes both the condition and the running_ boolean, so it has to be
+      // held when dealing with either/both
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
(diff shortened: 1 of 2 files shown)


zmichaels11 (approved):

jikawa-az (approved):

zmichaels11 (commented):
Do you need to wrap sections where `compression_is_running_` is assigned with a lock on the same mutex?

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -73,11 +73,18 @@ void SequentialCompressionWriter::compression_thread_fn()
...
+  while (true) {
     std::shared_ptr<rosbag2_storage::SerializedBagMessage> message;
     std::string file;
     {
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
+      // The mutex needs to be held for checking both exit conditions (shutdown & work available)
+      // If it is not, the main thread could set both compression_is_running_ and notify_all()
+      // on shutdown, after we read running_, and before we call wait().
+      // In this case this thread would have missed the final notification and wait forever.
+      if (!compression_is_running_) {
```
mm318, line 85: I think this boolean check should be in the condition variable `wait()` predicate.

prajakta-gokhale (approved):

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -73,11 +73,18 @@ void SequentialCompressionWriter::compression_thread_fn()
             "Cannot compress message; Writer is not open!"};
   }
 
-  while (compression_is_running_) {
+  while (true) {
     std::shared_ptr<rosbag2_storage::SerializedBagMessage> message;
     std::string file;
     {
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
```
david-prody, line 80: Shouldn't the main thread also grab this lock.  You are using it to gate the critical section check on the compression_is_running_ flag.

This current change looks like it makes the window smaller for the deadlock but does not remove it completely.

emersonknapp: Yes - you must be right. I've added the lock to `stop_compressor_threads` as well

david-prody (approved):

david-prody (approved):

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -73,11 +73,16 @@ void SequentialCompressionWriter::compression_thread_fn()
...
     std::shared_ptr<rosbag2_storage::SerializedBagMessage> message;
     std::string file;
     {
+      // This mutex synchronizes both the condition and the running_ boolean, so it has to be
+      // held when dealing with either/both
       std::unique_lock<std::mutex> lock(compressor_queue_mutex_);
+      if (!compression_is_running_) {
+        break;
+      }
       compressor_condition_.wait(lock);
```
adamdbrw, line 86: Because of spurious wake-ups, it is not the best practice to wait without a condition.

emersonknapp: Spurious wakeups are handled by this code - both queues are checked and if there is no work to do, we go to the next iteration of the while loop and wait again.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/1d3e63e26efdc9f1596e750454471f19/raw/198dbe7ad4255058e89eefccdfb923da7ff52b7e/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7528
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13510/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8416/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11233/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13576/)

PR #620, opened 2021-01-27
Regex and exclude fix for rosbag recorder

pijaro:
There is a bug in rosbag recorder when using `--regex` parameter. Currently when `--regex` parameter is set and `--exclude` parameter is not (which is by default), none of the topics are marked to subscribe.

This is caused because of misused `regex_search` method which returns always `true` when passed with "empty regex". Because of that, every topic is always dropped as soon as it is discovered.

This PR changes `regex_search` with `regex_match` which is more suitable for this case. Additional test case is also included.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -114,7 +114,9 @@ Recorder::get_requested_or_available_topics(const RecordOptions & record_options
   bool take = record_options.all;
   for (const auto & kv : unfiltered_topics) {
     // regex_match returns false for 'empty' regex
-    if (!record_options.regex.empty()) {take = std::regex_match(kv.first, topic_regex);}
+    if (!record_options.regex.empty()) {
+      take = std::regex_match(kv.first, topic_regex);
+    }
     if (take) {
       take = !std::regex_match(kv.first, exclude_regex);
     }
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -113,9 +113,10 @@ Recorder::get_requested_or_available_topics(const RecordOptions & record_options
   std::regex exclude_regex(record_options.exclude);
   bool take = record_options.all;
   for (const auto & kv : unfiltered_topics) {
-    take = std::regex_search(kv.first, topic_regex);
+    // regex_match returns false for 'empty' regex
+    if (!record_options.regex.empty()) {take = std::regex_match(kv.first, topic_regex);}
     if (take) {
-      take = !std::regex_search(kv.first, exclude_regex);
+      take = !std::regex_match(kv.first, exclude_regex);
     }
     if (take) {
       filtered_by_regex.insert(kv);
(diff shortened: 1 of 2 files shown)


adamdbrw:
Indeed I meant to use regex_match logic there, good catch!

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -113,9 +113,10 @@ Recorder::get_requested_or_available_topics(const RecordOptions & record_options
   std::regex exclude_regex(record_options.exclude);
   bool take = record_options.all;
   for (const auto & kv : unfiltered_topics) {
-    take = std::regex_search(kv.first, topic_regex);
+    // regex_match returns false for 'empty' regex
+    if (!record_options.regex.empty()) {take = std::regex_match(kv.first, topic_regex);}
```
zmichaels11, line 117: Can you expand this onto a separate line please?

zmichaels11 (approved):

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/a67fff37631acc9e22752cdafef26818/raw/ba38a7af06694891bd32974dbd9e78777c3f7955/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7541
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13531/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8431/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11254/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13596/)

rosbag2_transport/src/rosbag2_transport/recorder.cpp
pijaro: Done :+1:

adamdbrw:
What can be a bit confusing for some people (certainly was for me) is that colcon semantics of regex is different than in rosbag (and now ros2 bag). Colcon will match "rosba" to all "rosbag2" packages, so it needs to match as prefix ("osba" won't match anything), but does not need to match the whole package name string.

In rosbag (and ros2 bag) only full matches are filtered ("/to" won't match "/topic", you need "/to(.)*").

emersonknapp:
> What can be a bit confusing for some people (certainly was for me) is that colcon semantics of regex is different than in rosbag (and now ros2 bag). Colcon will match "rosba" to all "rosbag2" packages, so it needs to match as prefix ("osba" won't match anything), but does not need to match the whole package name string.
> 
> In rosbag (and ros2 bag) only full matches are filtered ("/to" won't match "/topic", you need "/to(.)*").

This is unrelated to this particular fix - but it's a useful point. It's not too late to change the behavior for the feature, do you think it should work the same as colcon? If so, please consider opening a feature request ticket, I'm not opposed to having it work that way

adamdbrw:
> This is unrelated to this particular fix - but it's a useful point. It's not too late to change the behavior for the feature, do you think it should work the same as colcon? If so, please consider opening a feature request ticket, I'm not opposed to having it work that way

I don't have a strong opinion about it - either way someone will be slightly confused for a while (ros1 users in the other case)

Issue #618, opened 2021-01-27
Start playing messages after timestamp


Kaju-Bubanja:
## Description

A [full fledged player](https://github.com/ros2/rosbag2/issues/484) would be the optimal solution, but already being able to only start a bag after a certain timestamp can be useful. I propose to add a command line option --start_time timestamp and then to skip all messages until after that timestamp.

## Related Issues

This feature is a subset of the idea for a [full fledged player](https://github.com/ros2/rosbag2/issues/484).
This feature would be really useful in tandem with the [pause feature](https://github.com/ros2/rosbag2/pull/443)

## Completion Criteria

Only messages after the passed timestamp are published.

## Implementation Notes / Suggestions

The implementation seems really simple, basically only one line, it might not be the most efficient, because the not used messages still get loaded, but it's the most simple change that came to my mind, namely adding the if statement to the Player::play_messages_until_queue_empty function:

```
while (message_queue_.try_dequeue(message) && rclcpp::ok()) {
      if(options.start_time < message.message->time_stamp) {
          std::this_thread::sleep_until(
                  start_time_ + std::chrono::duration_cast<std::chrono::nanoseconds>(
                          1.0 / rate * message.time_since_start));
          if (rclcpp::ok()) {
              auto publisher_iter = publishers_.find(message.message->topic_name);
              if (publisher_iter != publishers_.end()) {
                  publisher_iter->second->publish(message.message->serialized_data);
              }
          }
      }
  }
```

## Testing Notes / Suggestions

A simple bag with some message play it with a starting time and check that all published messages have timestamps after the specified timestamp.

I am kinda stuck on how to add the option to the play_options.hpp file. I added following to play.py in rosbag2/ros2bag/verb/play.py:

```
parser.add_argument(
            '--start_time', type=check_positive_float, default=1.0,
            help='start timestamp from which to play the bag')
```

## Questions

Now there seems to be some importing stuff going on in rosbag2_transport/rosbag2_transport/\_\_init\_\_.py, namely: 
`rosbag2_transport_py = import_c_library('._rosbag2_transport_py', package='rosbag2_transport')`

but I haven't been able to find the module ._rosbag2_transport_py, I guess this is some autogenerated file by configure_python.cmake? How would I go about adding the start_time option to the PlayOptions struct? Also where is the cpp file for the PlayOptions? Also some options in play.py are not available in play_options.hpp for example uri. If somebody could help me shed some light on the process of how to add a config option that would be amazing.

[...]

Issue #55, opened 2018-10-29
Pause/resume `ros2 bag play` with keyboard controls


Karsten1987:
# Description

Like in ROS 1, there should be a chance to control the replaying of a bag file.

`spacebar` should toggle the start/pause replaying.
`s` should solely spin the bag once.

Generally, as for now, there is no status feedback given when replaying a bag file. As a first step, there should be a message printed, whether the bag file is currently publishing data or not.

## Completion criteria

* [ ] `ros2 bag play` uses `keyboard_handler` to receive key callbacks
* [ ] `spacebar` toggles pause/resume state for playback
* [ ] `s` does `play_next`
* [ ] `ros2 bag play` prints a help message when playback starts, explaining these controls

## Related issues

#696 

## Implementation Notes

* May need to bloom-release `keyboard_handler` into Rolling - it is not yet in use.
* May need to expose keybindings to Python layer via Pybind - it makes sense for the CLI to define which keys do what, rather than hardcoding on the C++ end

## Testing Notes

* May be difficult to do key-based end-to-end testing. It should be OK to mock the key callbacks

weihunko:
Hi @Karsten1987 , while troubleshooting some perception issues with the recorded bag file, we would hope to have the functionalities to go back and forward a fixed amount of time, maybe 100ms or 1 second. 

For forwarding one step, we can use the `spin_once` function, but is it possible to go back one step or a fix amount of time?

Karsten1987:
I don't think I would temper with the `spin_once` functionality for going back and forth, more like with the "current" time step. 
See https://github.com/ros2/rosbag2/blob/56e440caca4bed3ea4cdde795a24a1bc60cdc355/rosbag2_transport/src/rosbag2_transport/player.cpp#L160

I imagine the overall behavior being the same, whether you'd like to step forward or backward. You'd simply add a `delta` to the current time and let the bag file play as usual.
Regardless of how you'd realize this though, there's the [issue of time (simulated) time](https://github.com/ros2/rosbag2/issues/99), which comes into play here as well, I assume.

weihunko:
@Karsten1987 From 
https://github.com/ros2/rosbag2/blob/56e440caca4bed3ea4cdde795a24a1bc60cdc355/rosbag2_transport/src/rosbag2_transport/player.cpp#L160

 It looks like `start_time_` is only used to control the publishing rate? Does it also used to determine which messages to publish according to their timestamps?

Kaju-Bubanja:
[...]

PR #622, opened 2021-01-28
Feature/start stop time

Kaju-Bubanja:
Fixes [#618](https://github.com/ros2/rosbag2/issues/618). This feature is not tested and not in a state to be merged. But I still put it out here, in case anybody else wants this hack. It works well for me so maybe others find it useful. I also do not plan to getting it into a mergable state. If somebody else feels like polishing it feel free or if this is inappropriate to have here just close the PR.

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -22,6 +22,7 @@ from ros2bag.api import print_error
 from ros2bag.verb import VerbExtension
 from ros2cli.node import NODE_NAME_PREFIX
 import yaml
+import sys
 
 
 class PlayVerb(VerbExtension):
@@ -41,6 +42,12 @@ class PlayVerb(VerbExtension):
         parser.add_argument(
             '-r', '--rate', type=check_positive_float, default=1.0,
             help='rate at which to play back messages. Valid range > 0.0.')
+        parser.add_argument(
+            '--start_time', type=int, default=1,
+            help='start timestamp from which to start playing the bag')
+        parser.add_argument(
+            '--stop_time', type=int, default=sys.maxsize,
+            help='stop timestamp after which to stop playing the bag')
         parser.add_argument(
             '--topics', type=str, default='', nargs='+',
             help='topics to replay, separated by space. If none specified, all topics will be '
@@ -91,6 +98,8 @@ class PlayVerb(VerbExtension):
             node_prefix=NODE_NAME_PREFIX,
             read_ahead_queue_size=args.read_ahead_queue_size,
             rate=args.rate,
+            start_time=args.start_time,
+            stop_time=args.stop_time,
             topics=args.topics,
             qos_profile_overrides=qos_profile_overrides,
             loop=args.loop,
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -31,6 +31,8 @@ public:
   size_t read_ahead_queue_size;
   std::string node_prefix = "";
   float rate = 1.0;
+  long start_time = 1;
+  long stop_time = 1;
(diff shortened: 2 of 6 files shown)


Karsten1987:
please target `master` and not `foxy` as your base branch for this PR.

Kaju-Bubanja:
Any reason for that? Isn't master behind foxy? I just see a way bigger diff to master, I also based it of off foxy. I think somebody seeing this huge div is going to be discouraged to use the feature, compared to the small diff to foxy.

emersonknapp:
> Any reason for that? Isn't master behind foxy?

`master` is the latest development branch, and is about 7 months of development ahead of the distro release branch `foxy`, which has only received backported bugfixes since the Foxy release in June 2020.

The practice is for new development to happen on the master branch, and for fixes or non-API-breaking new features to then be selectively backported to released ROS distributions.

Kaju-Bubanja:
Ah, makes sense. Is that true for all ros packages? Because I'm quite sure I saw some where master was pointing to ros1 and foxy was way ahead of master. There also might have been a ros2 branch, I'm not completely sure anymore. Is it possible that master is always latest, except if there is a ros2 branch? Than master is latest ros1 dev and ros2 is latest ros2 dev?

emersonknapp:
> Ah, makes sense. Is that true for all ros packages? Because I'm quite sure I saw some where master was pointing to ros1 and foxy was way ahead of master. There also might have been a ros2 branch, I'm not completely sure anymore. Is it possible that master is always latest, except if there is a ros2 branch? Than master is latest ros1 dev and ros2 is latest ros2 dev?

It's generally true for most packages - there are always exceptions. In the case of packages that support both ros1 and ros2, `master` was originally for ROS1, so the `ros2` branch tracks latest developments for ROS2 - while branches named after specific distributions are release branches. Sometimes that naming looks like `DISTRO-devel`, e.g. `kinetic-devel`, instead of just `kinetic`.

Kaju-Bubanja:
Is there a way to speed up the seeking to the right message? Because right now it takes quite some time for the reader to go to the start position, because it has to read each message. But I saw no other method which I could use to skip ahead or just skip reading the messages.

emersonknapp:
You'll also want to rebase your changes on master so that we can review the feature rather than all the differences between this branch and foxy :)

As for speeding up seeking - the SQLite database allows for random access. But what I think I'd like to see is a targeted API proposal (without implementation!) for the Player to expose the desired behavior.

Kaju-Bubanja:
Already looks like the rebase worked. Cool feature of git

Kaju-Bubanja:
Hmmm unfortunately looks like the rebase broke this feature, since I can't built rosbag2_storage in master as described here: https://github.com/ros2/rosbag2/issues/650. Isn't master always at least buildable?

emersonknapp:
> Isn't master always at least buildable?

Yes - the master branch builds. You can see that on https://github.com/ros2/rosbag2/actions?query=event:schedule (plus or minus an occasional flaky test failure) and https://ci.ros2.org/view/nightly/job/nightly_linux_release/ (which builds and tests all of https://github.com/ros2/ros2/blob/master/ros2.repos

Even the check on this PR built https://github.com/ros2/rosbag2/pull/622/checks?check_run_id=1885762308 (the tests failed though - it's at least the linters as a start). This suggests that something is incorrect in your local setup.

emersonknapp:
Closing - this feature has been implemented

kosuke55:
With which PR was this feature implemented?

Kaju-Bubanja:
https://github.com/ros2/rosbag2/issues/696

emersonknapp:
More specifically as the work on https://github.com/ros2/rosbag2/issues/55 - there was a string of PRs involved to build the infrastructure, but pause/resume was enabled in https://github.com/ros2/rosbag2/pull/704

Issue #598, opened 2021-01-07
Make zstd into a plugin to allow for alternate compression implementations


emersonknapp:
## Description

Right now `rosbag2_compression` hardcodes the `zstd` implementation. This is intended to be a plugin architecture like `rosbag2_storage` and `rosbag2_cpp/converter_interfaces`

## Related Issues

N/A

## Completion Criteria

* Create new package `rosbag2_compression_zstd` 
* remove all `zstd` from `rosbag2_compression`, making it an abstract plugin interface

## Testing Notes / Suggestions

Existing compression tests should verify that compression continues working unchanged.

# Work log

The implementation for this feature is being split into multiple parts for easier review

* [ ] Make zstd implementation into a pluginlib plugin, within the rosbag2_compression package (https://github.com/ros2/rosbag2/pull/624)
* [ ] Create a new package `rosbag2_compression_zstd`
* [ ] Move zstd implementation into `rosbag2_compression_zstd` - moving around tests where necessary to

PR #624, opened 2021-01-29
Make compressor implementations into a plugin via pluginlib

emersonknapp:
Part of #598

This is incremental progress - the next step will be to create `rosbag2_compression_zstd` package and move the zstd implementation there.

Key point: Use the `shared_ptr` implementation of `pluginlib::ClassLoader` instead of either `unmanagedInstance` or `uniqueInstance`. The unique version has a custom deleter, meaning we would have to pass around a `pluginlib::UniquePtr` - requiring all uses to include the pluginlib headers, which seems a bit restrictive. The unmanaged instance means that pluginlib can't warn us if we destroy the loader before the instance - this can lead to segfaults on the `SequentialReader/Writer` destructors if the member variables `compressor_` and `compression_factory_` aren't declared in the right order (which they aren't in the current code). This makes the shared library the most desirable, as it allows pluginlib to clean up after itself while letting the rest of the code.

### DIFF
--- a/rosbag2_compression/include/rosbag2_compression/compression_factory.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_factory.hpp
@@ -53,7 +53,7 @@ public:
    * \return A unique pointer to the newly created compressor.
    * \throw invalid_argument If the compression format does not exist.
    */
-  virtual std::unique_ptr<rosbag2_compression::BaseCompressorInterface>
+  virtual std::shared_ptr<rosbag2_compression::BaseCompressorInterface>
   create_compressor(const std::string & compression_format);
 
   /**
@@ -63,7 +63,7 @@ public:
    * \return A unique pointer to the newly created decompressor.
    * \throw invalid_argument If the compression format does not exist.
    */
-  virtual std::unique_ptr<rosbag2_compression::BaseDecompressorInterface>
+  virtual std::shared_ptr<rosbag2_compression::BaseDecompressorInterface>
   create_decompressor(const std::string & compression_format);
 
 private:
--- a/rosbag2_compression/include/rosbag2_compression/compression_factory.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_factory.hpp
@@ -50,7 +50,7 @@ public:
    * Create a compressor based on the specified compression format.
    *
    * \param compression_format The compression format as a string.
-   * \return A unique pointer to the newly created compressor.
+   * \return A shared pointer to the newly created compressor.
    * \throw invalid_argument If the compression format does not exist.
    */
   virtual std::shared_ptr<rosbag2_compression::BaseCompressorInterface>
@@ -60,7 +60,7 @@ public:
    * Create a decompressor based on the specified compression format.
    *
    * \param compression_format The compression format as a string.
-   * \return A unique pointer to the newly created decompressor.
+   * \return A shared pointer to the newly created decompressor.
    * \throw invalid_argument If the compression format does not exist.
    */
   virtual std::shared_ptr<rosbag2_compression::BaseDecompressorInterface>
--- a/rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
+++ b/rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
@@ -56,7 +56,7 @@ get_class_loader()
 }
 
 template<typename InterfaceT>
-std::unique_ptr<InterfaceT>
+std::shared_ptr<InterfaceT>
 get_interface_instance(
   std::shared_ptr<pluginlib::ClassLoader<InterfaceT>> class_loader,
   const std::string & compression_format)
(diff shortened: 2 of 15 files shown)


rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
```diff
@@ -20,67 +20,116 @@
...
+{};
 
-/// Verify whether two case-insensitive chars are equal
-bool compare_char(const char c1, const char c2)
+template<>
+struct CompressionTraits<BaseCompressorInterface>
 {
-  return c1 == c2 || std::tolower(c1) == std::tolower(c2);
-}
+  static constexpr const char * name = "rosbag2_compression::BaseCompressorInterface";
```
zmichaels11, line 41: is there a difference between `constexpr const char * name` and `constexpr const char name[]`?

rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
```diff
@@ -20,67 +20,116 @@
...
-  return (str1.size() == str2.size()) &&
-         std::equal(str1.begin(), str1.end(), str2.begin(), &compare_char);
+  static constexpr const char * name = "rosbag2_compression::BaseDecompressorInterface";
+};
+
+template<typename InterfaceT>
+std::shared_ptr<pluginlib::ClassLoader<InterfaceT>>
+get_class_loader()
+{
+  const char * lookup_name = CompressionTraits<InterfaceT>::name;
```
zmichaels11, line 54: Can probably be `const auto` or maybe `std::string` since `ClassLoader` ctor is implicitly converting to a std::string.

rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
```diff
@@ -20,67 +20,116 @@
...
-      return std::make_unique<rosbag2_compression::ZstdCompressor>();
-    } else {
-      std::stringstream errmsg;
-      errmsg << "Compression format \"" << compression_format << "\" is not supported.";
-      ROSBAG2_COMPRESSION_LOG_ERROR_STREAM(errmsg.str());
-      throw std::invalid_argument{errmsg.str()};
+    auto instance = get_interface_instance(compressor_class_loader_, compression_format);
+    if (instance == nullptr) {
+      ROSBAG2_COMPRESSION_LOG_ERROR_STREAM(
+        "Could not load/open plugin for compression format '" << compression_format << "'");
```
zmichaels11, line 113: I think it would read cleaner if `return nullptr` here instead of sharing the same exit point regardless if `instance == nullptr`.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -66,12 +65,7 @@ void SequentialCompressionWriter::compression_thread_fn()
   // Every thread needs to have its own compression context for thread safety.
   auto compressor = compression_factory_->create_compressor(
     compression_options_.compression_format);
-
-
-  if (!compressor) {
-    throw std::runtime_error{
-            "Cannot compress message; Writer is not open!"};
-  }
+  rcpputils::check_true(compressor != nullptr, "Could not create compressor.");
```
zmichaels11, line 68: Update header to state that this now throws `rcpputils::InvalidStateException` instead of `std::runtime_error`?

zmichaels11 (approved):

rosbag2_compression/include/rosbag2_compression/compression_factory.hpp
```diff
@@ -53,7 +53,7 @@ class ROSBAG2_COMPRESSION_PUBLIC CompressionFactory
    * \return A unique pointer to the newly created compressor.
```
prajakta-gokhale, line 53: Update docstrings to match function.

rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
```diff
@@ -20,67 +20,116 @@
...
+std::shared_ptr<InterfaceT>
+get_interface_instance(
+  std::shared_ptr<pluginlib::ClassLoader<InterfaceT>> class_loader,
+  const std::string & compression_format)
 {
+  const auto & registered_classes = class_loader->getDeclaredClasses();
+  auto class_iter = std::find(
+    registered_classes.begin(), registered_classes.end(), compression_format);
+  if (class_iter == registered_classes.end()) {
+    ROSBAG2_COMPRESSION_LOG_DEBUG_STREAM(
```
prajakta-gokhale, line 68: Should this be logged at higher than DEBUG?

prajakta-gokhale (approved):
Left a couple nits, mostly looks good!

rosbag2_compression/src/rosbag2_compression/compression_factory_impl.hpp
emersonknapp: yes - but I don't understand what it is. It builds as `char * name`, but not as `char name[]`, like so

```
/usr/bin/ld: librosbag2_compression.so: undefined reference to `rosbag2_compression::CompressionTraits<rosbag2_compression::BaseCompressorInterface>::name'
```

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/5de8ab86a0473449296fbc19dcf5e60f/raw/c1a41f3fe0dbd1cf61ab98919c12003f79f58659/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7558
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13557/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8451/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11280/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13622/)

PR #634, opened 2021-02-04
Performance benchmarking improvements

pijaro:
A small pack of few quality of life improvements and one fix for performance benchmarking:

- wait 5 secs between tests (to let disks "cool down" a bit and flush their internal caches),
- removing bagfiles now leaves an informative file about deleted file size + additional file about total bag size (both in MB),
- (FIX) updated `ros2 bag record` to properly use new `--regex` syntax ( `-a` which is used currently causing sometimes over 100% recorded output when unnecessary topics like ` /rosout` are recorded),
- `report_gen.py` tool additionally returns "min" and "max" percents of recorded messages, not only average.

### DIFF
--- a/rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
+++ b/rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
@@ -51,6 +51,7 @@ import pathlib
 import shutil
 import signal
 import sys
+import time
 
 from ament_index_python import get_package_share_directory
 
@@ -196,8 +197,20 @@ def _producer_node_exited(event, context):
     # Handle clearing bag files
     if not node_params['preserve_bags']:
         db_files = pathlib.Path.cwd().joinpath(node_params['db_folder']).glob('*.db3')
+        total_size = 0
         for f in db_files:
+            filesize = f.stat().st_size
             f.unlink()
+            # Replace removed storage file with empty file informing about size
+            pathlib.Path.cwd().joinpath(
+                node_params['db_folder'],
+                str(int(filesize/1000000)) + 'MB_' + f.name).touch()
+            total_size += filesize
+
+        # Additional info about total size
+        pathlib.Path.cwd().joinpath(
+            node_params['db_folder'],
+            str(int(total_size/1000000)) + 'MB_TOTAL').touch()
 
     # If we have non empty rosbag PID, then we need to kill it (end-to-end transport case)
     if _rosbag_pid is not None and transport:
@@ -217,6 +230,10 @@ def _producer_node_exited(event, context):
 
     # Bump up producer index, so the launch sequence can continue
     _producer_idx += 1
+
+    # Give disks some time to flush their internal cache before starting next experiment
+    time.sleep(5)
+
     return [
         launch.actions.LogInfo(
             msg='---------------------------'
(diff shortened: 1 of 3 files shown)


rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
```diff
@@ -217,6 +230,10 @@ def _producer_node_exited(event, context):
 
     # Bump up producer index, so the launch sequence can continue
     _producer_idx += 1
+
+    # Give disks some time to flush their internal cache before starting next experiment
+    time.sleep(5)
```
emersonknapp, line 235: Is this really the right way to do this? Is there any way to check it programmatically? I'm always suspicious of arbitrary waits.

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
```diff
@@ -196,8 +197,20 @@ def _producer_node_exited(event, context):
...
+            # Replace removed storage file with empty file informing about size
+            pathlib.Path.cwd().joinpath(
+                node_params['db_folder'],
+                str(int(filesize/1000000)) + 'MB_' + f.name).touch()
+            total_size += filesize
+
+        # Additional info about total size
+        pathlib.Path.cwd().joinpath(
+            node_params['db_folder'],
+            str(int(total_size/1000000)) + 'MB_TOTAL').touch()
```
emersonknapp, line 213: Am I reading this correctly that you're writing the total size as a file name? It seems much more appropriate to write this information _into_ a file

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
adamdbrw: I don't know of any portable system API to check whether there is something in the disk cache. I am not a fan of arbitrary waits either, but I don't see an alternative atm.

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
adamdbrw: agreed, should be a quick change. @pijaro

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
emersonknapp: Do we have any reason to wait 5 seconds vs any other time? Not going to block this review, but I want to make sure we're at least thinking about hard problems

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
pijaro: Indeed, I will write this into `yaml` file.

pijaro: Ok, it is now saved in single `bagfiles_info.yaml` file.

rosbag2_performance/rosbag2_performance_benchmarking/launch/benchmark_launch.py
pijaro: Well, actually that was set based on our intuition. There is surely a place for improvement here, but - what Adam said - that would require some way of inspecting hardware cache.

adamdbrw:
@emersonknapp is this good to be merged?

emersonknapp (approved):
LGTM

PR #635, opened 2021-02-04
Fix a crash in test_message_cache

clalancette:
Fix a crash in test_message_cache.
    
The cache consumer requires that you call close()
before calling change_consumer_callback().  The test
forgot the close() call, so was running into a race
where changing out the callback was not atomic.
Remember to call close() in the test, which should fix
some occasional failures in CI.

This should fix the occasional crashes we are seeing, like https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/2240/testReport/junit/(root)/projectroot/test_message_cache/ .

While this change makes sense to me, and passes the unit tests, I am hesitant just because the previous code was so weird.  I'm not sure if I'm missing something.  @adamdbrw , @Karsten1987 , @mjeronimo , @emersonknapp review here is welcome.

### DIFF
--- a/rosbag2_cpp/test/rosbag2_cpp/test_message_cache.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_message_cache.cpp
@@ -131,6 +131,7 @@ TEST_F(MessageCacheTest, message_cache_changing_callback) {
   uint32_t counter = 0;
   for (uint32_t i = 0; i < message_count; ++i) {
     if (counter >= message_count / 2) {
+      mock_cache_consumer->close();
       mock_cache_consumer->change_consume_callback(cb2);
     }
     auto msg = make_test_msg();


adamdbrw:
Analyzing the code I can say that this only occurs in the test due to the fact that the close() is not called as it should before the change of callback.

I am not sure about the change - I will take some time to analyze it. If it works properly, I am for it since it takes away the requirement of close().

Note that the actual code of caching in rosbag2 was tested to the extreme in our benchmarks (including plenty of splitting and huge volumes of data and/or topics) and we had no crashes.

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -53,11 +55,8 @@ void CacheConsumer::close()
 void CacheConsumer::change_consume_callback(
   CacheConsumer::consume_callback_function_t consume_callback)
 {
+  std::lock_guard<std::mutex> lg(consumer_mutex_);
```
adamdbrw, line 58: one worry I have is that this could affect performance as it effectively stops the thread which is used to process the subscriber queue for possibly quite long. In stress cases (at the edge of I/O capability to write), the cache buffer being flushed can be as much as 500 MB of data if --max-cache-size is set to such value (which is reasonable with fast drives). Then we need to wait for a 500 MB data write. I suspect this could cause message loss due to subscription queue (size 10 for example) not being consumed and having to overwrite messages.

I could test this PR with our benchmarks to see if it makes such a big difference.

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -70,12 +69,12 @@ void CacheConsumer::exec_consuming()
     // swap producer buffer with consumer buffer
     message_cache_->wait_for_buffer();
 
-    // make sure to use consistent callback for each iteration
-    auto callback_for_this_loop = consume_callback_;
-
     // consume all the data from consumer buffer
```
adamdbrw, line 72: the point for the callback_for_this_loop was performance. We changed the current callback in crucial producer (subscription processing) thread and moved on, instead of waiting for the whole buffer flush to finish.

clalancette: I see what you are saying.  In the case where this thread is currently writing I/O, this could block the `change_consume_callback` for a while until that is finished.

What do you think if I changed this to:

```
consume_callback_function_t callback_for_this_loop;
{
std::lock_guard<std::mutex> lg(consumer_mutex_);
callback_for_this_loop = consume_callback_
}
auto consumer_buffer = message_cache_->consumer_buffer();
callback_for_this_loop(consumer_buffer->data());
```

That way we only hold up `change_consume_callback` for a short while (just enough to copy the `std::function`), but we still retain the property that we don't need to call `close`.

adamdbrw: This looks good! Now I am trying to remember why this section:

```
if (!consumer_thread_.joinable()) {
    is_stop_issued_ = false;
    consumer_thread_ = std::thread(&CacheConsumer::exec_consuming, this);
  }
```
was there. Plus I will stress-test the change with our benchmarks, which I can do tomorrow, to see if all is ok.

clalancette: OK, great.  I'll make that change and wait to hear back from you how it does in the benchmarks.  Thanks for your support.

clalancette: All right, changed as discussed in 6e137ce .

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -53,11 +55,8 @@ void CacheConsumer::close()
 void CacheConsumer::change_consume_callback(
   CacheConsumer::consume_callback_function_t consume_callback)
 {
+  std::lock_guard<std::mutex> lg(consumer_mutex_);
   consume_callback_ = consume_callback;
-  if (!consumer_thread_.joinable()) {
```
adamdbrw, line 57: This is a required section. The cache consumer is closed when splitting bags - which means that we await the consumer thread to finish its work (flush the cache) and join it. We also enter the flushing state which is controlled by the flag.

If we don't restart the thread for the new bag/callback, we have no thread to do the work. Predictably, without this section the PR benchmarks results in all the messages being lost in subsequent bags (apart from bag 0).

This has to stay and it is a necessary condition. What could be improved is certainly some clarification around it.

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
adamdbrw: I explain why this section is necessary and also ran benchmarks of this PR:
1) without the section (PR as is) - failure to record anything post 1st bag
2) with the section restored - all correct and looking good

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
clalancette: Oh, I see.  Yeah, that makes sense.

However, the previous code was buggy as well in that if you called `change_consumer_callback` without a `close`, you would "leak" the thread.

So I still think this class needs to be refactored.  One way to do it would be to make it totally RAII, where the constructor creates the thread, the destructor deletes the thread, and nothing else ever kills off the thread.  In that case, all `close` would do is to do a `message_cache->finalize()`, followed by `message_cache->notify_flushing_done`.  What do you think about that?

adamdbrw: Calling message_cache->finalize() and then message_cache->notify_flushing_done would be a logic error - flushing is not done until the consumer thread exits the loop. Closing a bag (whether it is the final one or not) should ensure the flush.

clalancette: I'm open to ideas.  At the end of the day, I'm not a huge fan of APIs that are easy to misuse, and this one seems to fit that description.  Do you have any thoughts on how we could make this nicer?

adamdbrw: With this PR merged, I don't think it is will be as easy to misuse anymore. Which part do you think is still easy to misuse?
Refactoring this, I would like to make sure #640 is solved - there is no reason not to use the cache while the consumer is finalizing, but some additional logic is needed.

clalancette: > Refactoring this, I would like to make sure #640 is solved - there is no reason not to use the cache while the consumer is finalizing, but some additional logic is needed.

All right.  I'm mostly interested in fixing the occasional crash we see in CI, so instead of getting bogged down with #640 I think we can instead fix the test to properly call `close`.  I'm going to change this PR to do that for now instead.

I'll leave it to you whether you want to refactor this as part of #640.  Thanks for the input so far.

adamdbrw:
This looks good to me, thank you Chris! @mjeronimo  could you take a look and merge this? It fixes a non-deterministic test crash.

mjeronimo (approved):

mjeronimo:

clalancette:
The Rpr job is failing, but that doesn't seem unique to this PR; other PRs seem to have the same problem.  Probably something to be looked into, but I don't think it should hold this one up.  I'm going to run CI on this just to ensure the other platforms are happy as well:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8524/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11359/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13708/)

PR #641, opened 2021-02-08
Introducing Reindexer core

### LINKED ISSUE #395: DB3 bag is irrecoverable if metadata.yaml doesn't exist

jhdcs:
Replaces #539 
Part of #395

After some discussions with the Tooling Working Group, I decided to do a major refactoring in order to implement all of their changes. This PR is going to be for the core reindexing functionality - it does not contain any Python API. That will come in a future PR, once this is polished up to an acceptable level.

This does not currently have the ability to reindex compressed bags, though I have added in test bags and target files for when such functionality becomes available.


rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+// a) the member is not accessible from the outside
+// b) there are no inline functions.
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2_cpp
+{
+namespace reindexers
```
emersonknapp, line 56: I don't think we need a nested namespace for this, `rosbag2_cpp::Reindexer` seems fine.

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+{
+
+class ROSBAG2_CPP_PUBLIC Reindexer
+{
+public:
+  Reindexer(
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
```
emersonknapp, line 65: I would argue: Conversion is not part of the Reindexing API. I don't think this parameter should be here. Since we explicitly plan to not initialize a converter, there is no need for the user to be able to customize it.

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2_cpp
+{
+namespace reindexers
+{
+
+class ROSBAG2_CPP_PUBLIC Reindexer
```
emersonknapp, line 59: I know it hasn't been held to very well in this codebase, but can you please add a docstring to this class and its public member functions indicating their purpose and usage?

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+#endif
+
+
+namespace rosbag2_cpp
+{
+namespace reindexers
+{
+namespace details
+{
+std::vector<rcpputils::fs::path> resolve_relative_paths(
```
emersonknapp, line 60: this is a copy-paste, and it's not being used. remove

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+      continue;
+    }
+    file = (base_path / path).string();
+  }
+
+  return relative_files;
+}
+}  // namespace details
+
+std::string strip_parent_path(const rcpputils::fs::path & relative_path)
```
emersonknapp, line 88: this is a copy-paste, and it's only used once - i think you should just call `.filename().string()` inline

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  void fill_topics_metadata();
+
+  void reset();
+
+  void finalize_metadata();
+
+protected:
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory_{};
+  std::unique_ptr<Converter> converter_{};
```
emersonknapp, line 83: IIUC, the reindexer doesn't need a converter - why does it have this member?

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io_{};
+  rosbag2_storage::BagMetadata metadata_{};
+  std::vector<rosbag2_storage::TopicMetadata> topics_metadata_{};
+  std::vector<rcpputils::fs::path> file_paths_{};  // List of database files.
```
emersonknapp, line 87:
```suggestion
  std::vector<rcpputils::fs::path> discovered_bagfile_paths_{}; 
```

If you need a comment to tell what a member is for, maybe it's not a good name (and bags are not guaranteed to be "databases" specifically)

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  // Index of file to read from
```
emersonknapp, line 88: Is this comment adding value?

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  std::vector<rcpputils::fs::path>::iterator current_file_iterator_{};
+
+private:
+  rcpputils::fs::path base_folder_;   // The folder that the bag files are in
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_{};
+  // std::unique_ptr<rosbag2_cpp::Reader> bagfile_reader_;
```
emersonknapp, line 94: remove unused code

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  std::vector<rcpputils::fs::path> get_database_files(const rcpputils::fs::path & base_folder);
+
+  // // Open a single bag FILE (not the whole bag, just an internal file) for processing
+  // void open(const rcpputils::fs::path & bag_file);
```
emersonknapp, line 99: remove unused code

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+  std::vector<rcpputils::fs::path> file_paths_{};  // List of database files.
+  // Index of file to read from
+  std::vector<rcpputils::fs::path>::iterator current_file_iterator_{};
+
+private:
+  rcpputils::fs::path base_folder_;   // The folder that the bag files are in
+  std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_{};
+  // std::unique_ptr<rosbag2_cpp::Reader> bagfile_reader_;
+
+  std::vector<rcpputils::fs::path> get_database_files(const rcpputils::fs::path & base_folder);
```
emersonknapp, line 96:
```suggestion
  std::vector<rcpputils::fs::path> get_bag_files(const rcpputils::fs::path & base_folder);
```

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+    const std::vector<rcpputils::fs::path> & files,
+    const rosbag2_storage::StorageOptions & storage_options);
+
+  // Attempts to harvest metadata from all bag files, and aggregates the result
+  void aggregate_metadata(
+    const std::vector<rcpputils::fs::path> & files,
+    const rosbag2_storage::StorageOptions & storage_options);
+
+  // Compairson function for std::sort with our filepath convention
+  static bool comp_rel_file(
```
emersonknapp, line 112:
```suggestion
  static bool compare_relative_file(
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+  converter_(nullptr),
+  metadata_io_(std::move(metadata_io)),
+  converter_factory_(std::move(converter_factory)) {}
+
+Reindexer::~Reindexer()
+{
+  reset();
+}
+
+void Reindexer::reset()
```
emersonknapp, line 107: not used? if so, why part of API?

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+    std::make_shared<SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
+
+  virtual ~Reindexer();
+
+
+  void reindex(const rosbag2_storage::StorageOptions & storage_options);
+
+  void fill_topics_metadata();
```
emersonknapp, line 75: why is this part of the API? Is it just a copy-paste from sequentialreader?

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  void reset();
+
+  void finalize_metadata();
```
emersonknapp, line 79: why is this part of the public API? it seems like an implementation detail

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,124 @@
...
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory =
+    std::make_shared<SerializationFormatConverterFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
+
+  virtual ~Reindexer();
+
+
+  void reindex(const rosbag2_storage::StorageOptions & storage_options);
```
emersonknapp, line 73: It seems to me that this should probably be the full extent of the public API - definitely add a docstring to describe what this does. Considering it returns no values and is non-const, we should describe exactly what side effects are expected on calling this and what can happen otherwise (does it throw exceptions?)

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+    ROSBAG2_CPP_LOG_INFO_STREAM("Preparing to open...");
+    bag_reader->open(temp_so, blank_converter_options);
+    ROSBAG2_CPP_LOG_INFO_STREAM("Opened");
+    // Get metadata from the bag file
+    auto temp_metadata = bag_reader->get_metadata();
+
+    if (temp_metadata.starting_time < metadata_.starting_time) {
+      metadata_.starting_time = temp_metadata.starting_time;
+    }
+    // ROSBAG2_CPP_LOG_INFO_STREAM(
```
emersonknapp, line 237: remove unused code

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+    // Add the topic metadata
+    for (const auto & topic : temp_metadata.topics_with_message_count) {
+      auto found_topic = std::find_if(
+        metadata_.topics_with_message_count.begin(),
+        metadata_.topics_with_message_count.end(),
+        [&topic](const rosbag2_storage::TopicInformation & agg_topic)
+        {return topic.topic_metadata.name == agg_topic.topic_metadata.name;});
+      if (found_topic == metadata_.topics_with_message_count.end()) {
+        // It's a new topic. Add it.
+        metadata_.topics_with_message_count.emplace_back(topic);
```
emersonknapp, line 254: just a thought - you could use an `std::map` for the interim while aggregating data so that you don't have to do as much `find`ing, then convert to a vector at the end

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+        if (topic.topic_metadata.type != "") {
+          found_topic->topic_metadata.type = topic.topic_metadata.type;
+        }
+      }
+    }
+
+    ROSBAG2_CPP_LOG_INFO("Closing database");
+    bag_reader->reset();   // Close the reader's storage
+  }
+  bag_reader.reset();   // Delete the reader
```
emersonknapp, line 275: This call is redundant - by using a smart pointer you have already guaranteed that the object will be deleted when the scope exits

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,320 @@
...
+  // Create initial metadata
+  init_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed init_metadata");
+
+  // Collect all metadata from database files
+  aggregate_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed aggregate_metadata");
+
+  // Perform final touch-up
+  finalize_metadata();
```
emersonknapp, line 299: you're already aggregating duration and message counts in `aggregate_metadata` - i don't see why a separate vaguely-named function is needed to add up the sizes?

rosbag2_cpp/test/rosbag2_cpp/test_reindexer.cpp
```diff
@@ -0,0 +1,129 @@
...
+  so.uri = bag_dir;
+  so.storage_id = "sqlite3";
+
+  reindexer->reindex(so);
+
+  // We should have a metadata file at this point
+  auto generated_file = rcpputils::fs::path(bag_dir + "/metadata.yaml");
+  EXPECT_TRUE(generated_file.exists());
+
+  // We made it here. Excellent. Truly excellent.
```
emersonknapp, line 77: I'm glad you're having fun, but let's tone down the comments across the board to "only if necessary to understand the code (and sometimes maybe the code needs clarifying rather than comments), and only as much information as needed" - the flavor commentary doesn't always translate well across all readers and distracts from the point at hand

rosbag2_cpp/test/rosbag2_cpp/test_reindexer.cpp
```diff
@@ -0,0 +1,129 @@
...
+  // Compare starting times
+  // Disabled for now, since it may not be possible to 100% recreate
+  //   original starting time from metadata
+  // EXPECT_EQ(generated_metadata.starting_time, target_metadata.starting_time);
+
+  // Compare durations
+  // Disabled for now, since I'm not sure how duraction is created, and if it's correct (jhdcs)
+  // EXPECT_EQ(generated_metadata.duration, target_metadata.duration);
+
+  // Compare message count
```
emersonknapp, line 104: It seems to me that the code here is self-documenting - that's the point of the googletest expectation macro syntax. Double check if you need most of these comments.

emersonknapp (changes requested):

.gitignore
```diff
@@ -5,3 +5,5 @@ cmake-build-release/
 build/
 venv/
 **/.pytest_cache/
+**/__pycache__/
+*.db3-*
```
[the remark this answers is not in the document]
jhdcs: Not quite. There's a `-` at the end of the `.db3-` (line 9). This is to exclude the `.db3-shm` and `.db3-wal` files that get generated during the testing process.

If you prefer, though, I could change line 9 into two lines:
`.db3-shm`
`.db3-wal`
I just did it this way since it was less typing.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: Good thinking! Implementing it now...

.gitignore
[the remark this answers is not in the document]
emersonknapp: Looking at the diff - you have checked in at least one each of `.db3-shm` and `.db3-wal` files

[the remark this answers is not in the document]
jhdcs: Good catch! Hopefully they're removed now...

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,128 @@
...
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2_cpp
+{
+
+/*************************************************************************************************
+ * Base class for a Reindexer
```
emersonknapp, line 58: Is this a "Base class for a Reindexer"? Is the intent that this is an abstract interface that is primarily intended to be overridden by concrete implementations? Or is this a concrete utility "Tool to reconstruct bag metadata files when lost or corrupted."

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Sort relative file path by database number
+  std::sort(
+    output.begin(), output.end(),
+    [](rcpputils::fs::path a, rcpputils::fs::path b) {return compare_relative_file(a, b);});
+
+  return output;
+}
+
+void Reindexer::fill_topics_metadata()
```
emersonknapp, line 123: Function no longer used, remove dead code

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+}
+
+void Reindexer::init_metadata(
+  const std::vector<rcpputils::fs::path> & files,
+  const rosbag2_storage::StorageOptions & storage_options)
+{
+  metadata_ = rosbag2_storage::BagMetadata{};
+
+  // This reindexer will only work on SQLite files, so this can't change
```
emersonknapp, line 139: I don't think this comment makes sense. We expect that all bags use the same storage implementation - that's all.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+void Reindexer::reindex(const rosbag2_storage::StorageOptions & storage_options)
+{
+  ROSBAG2_CPP_LOG_INFO("Beginning Reindex Operation.");
+
+  // Identify all bag files
+  base_folder_ = storage_options.uri;
+  auto files = get_bag_files(base_folder_);
+  std::cout << "Finished getting database files\n";
```
emersonknapp, line 235:
```suggestion
  ROSBAG2_CPP_LOG_DEBUG("Finished getting database files");
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  if (files.empty()) {
+    throw std::runtime_error("No database files found for reindexing. Abort");
+  }
+
+  // Create initial metadata
+  init_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed init_metadata");
```
emersonknapp, line 242: This doesn't seem INFO-worthy, it's an implementation detail as written. Maybe DEBUG. Or change the text to be something a user without knowledge of the implementation would benefit from.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Collect all metadata from database files
+  aggregate_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed aggregate_metadata");
+
+  ROSBAG2_CPP_LOG_INFO_STREAM("Base folder located at: " + base_folder_.string());
+  metadata_io_->write_metadata(base_folder_.string(), metadata_);
+  ROSBAG2_CPP_LOG_INFO("Reindexing operation completed.");
```
emersonknapp, line 250:
```suggestion
  ROSBAG2_CPP_LOG_INFO("Reindexing complete.");
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Create initial metadata
+  init_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed init_metadata");
+
+  // Collect all metadata from database files
+  aggregate_metadata(files, storage_options);
+  ROSBAG2_CPP_LOG_INFO_STREAM("Completed aggregate_metadata");
+
+  ROSBAG2_CPP_LOG_INFO_STREAM("Base folder located at: " + base_folder_.string());
```
emersonknapp, line 248:
```suggestion

```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  // Convert the topic map into topic metadata
+  for (auto & topic : temp_topic_info) {
+    metadata_.topics_with_message_count.emplace_back(topic.second);
+  }
+}
+
+void Reindexer::reindex(const rosbag2_storage::StorageOptions & storage_options)
+{
+  ROSBAG2_CPP_LOG_INFO("Beginning Reindex Operation.");
```
emersonknapp, line 230:
```suggestion
  ROSBAG2_CPP_LOG_INFO_STEAM("Reindexing bag in directory " << base_folder_);
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io)
+: storage_factory_(std::move(storage_factory)),
+  metadata_io_(std::move(metadata_io)) {}
+
+Reindexer::~Reindexer() {}
+
+bool Reindexer::compare_relative_file(
+  const rcpputils::fs::path & first_path, const rcpputils::fs::path & second_path)
+{
+  std::regex regex_rule(".*_(\\d+)\\.db3", std::regex_constants::ECMAScript);
```
emersonknapp, line 68: This is hardcoded to db3 files? Let's have the storage implementation give us the expected extension. 

Or for a looser requirement, we could generalize the regex to 
```
.*_(\d+)\.([a-z0-9+])
```

So that we capture that it's "name_{N}.{EXT}"

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  auto first_path_string = first_path.string();
+  auto second_path_string = second_path.string();
+
+  auto first_regex_good = std::regex_match(first_path_string, first_match, regex_rule);
+  auto second_regex_good = std::regex_match(second_path_string, second_match, regex_rule);
+
+  // Make sure the paths have regex matches
+  if (!first_regex_good || !second_regex_good) {
+    throw std::runtime_error("Malformed relative file name. Expected numerical identifier.");
```
emersonknapp, line 81: "expected numerical identifier" sounds unclear as an end user. Maybe just something like 
```
"Path " << first_path << " didn't meet expected naming convention " << regex_rule
```
That way you just give all the info

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  u_int32_t first_db_num = std::stoul(first_match.str(1), nullptr, 10);
+  u_int32_t second_db_num = std::stoul(second_match.str(1), nullptr, 10);
+
+  return first_db_num < second_db_num;
+}
+
+std::vector<rcpputils::fs::path> Reindexer::get_bag_files(
```
emersonknapp, line 90: nitpick: in C++ I generally like to avoid returning vectors by value - the API preference for me is

```
void get_bag_files(path base_folder, vector<path>& output) {}
```

In this case it probably doesn't matter, but that allows the caller to control resource allocation and reuse.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  const rcpputils::fs::path & base_folder)
+{
+  // Look in the uri directory to see what database files are there
+  std::vector<rcpputils::fs::path> output;
+
+  auto allocator = rcutils_get_default_allocator();
+  auto dir_iter = rcutils_dir_iter_start(base_folder.string().c_str(), allocator);
```
emersonknapp, line 97: it's too bad that this isn't exposed in `rcpputils` yet! I opened a feature request https://github.com/ros2/rcpputils/issues/125 for it now after realizing that. This comment is just a comment, does not need to be addressed.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  if (dir_iter == nullptr) {
+    throw std::runtime_error("Empty directory.");
+  }
+
+  // Get all file names in directory
+  do {
+    auto found_file = rcpputils::fs::path(dir_iter->entry_name);
+
+    // We are ONLY interested in database files
+    if (found_file.extension().string() == ".db3") {
```
emersonknapp, line 109: as per earlier comment - let's avoid hardcoding the extension - let's have the storage implementation provide it

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+      auto full_path = base_folder / found_file;
+      output.emplace_back(full_path);
+    }
+  } while (rcutils_dir_iter_next(dir_iter));
+
+  // Sort relative file path by database number
+  std::sort(
```
emersonknapp, line 116: Given that bagfiles are "NAME_N.EXT" - wouldn't a lexicographical sort be sufficient without a custom comparator?

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+void Reindexer::aggregate_metadata(
+  const std::vector<rcpputils::fs::path> & files,
+  const rosbag2_storage::StorageOptions & storage_options)
+{
+  std::map<std::string, rosbag2_storage::TopicInformation> temp_topic_info;
+
+  // In order to most accurately reconstruct the metadata, we need to
+  // visit each of the contained relative database files in the bag,
```
emersonknapp, line 158:
```suggestion
  // visit each of the contained relative files in the bag,
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  // open them, slurp up the info, and stuff it into the master
+  // metadata object.
```
emersonknapp, line 160:
```suggestion
  // open them, read the info and write it into an aggregated metadata object.
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  auto metadata_io_default = std::make_unique<rosbag2_storage::MetadataIo>();
```
emersonknapp, line 161: nitpick: it might be nice to change the `metadata_io_` to be optional in the sequentialreader, based on observing this use case. Then we wouldn't have to construct a new one for every fille.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  auto bag_reader = std::make_unique<rosbag2_cpp::readers::SequentialReader>(
+    std::move(storage_factory_), converter_factory_, std::move(metadata_io_default));
```
emersonknapp, line 163: This move of the storage_factory is interesting. Ownership is never transferred back to the Reindexer - so if you call `reindex` twice, you will run into an error on the second call, meaning that the API as presented is broken.

This start to look to me like "should the reindexer even be a class? Or is it just a function?"

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+      storage_options.storage_id,  // storage_id
+      storage_options.max_bagfile_size,  // max_bagfile_size
+      storage_options.max_bagfile_duration,  // max_bagfile_duration
+      storage_options.max_cache_size,  // max_cache_size
+      storage_options.storage_config_uri  // storage_config_uri
+    };
+
+    // We aren't actually interested in reading messages, so use a blank converter option
+    rosbag2_cpp::ConverterOptions blank_converter_options {};
+    ROSBAG2_CPP_LOG_INFO_STREAM("Preparing to open...");
```
emersonknapp, line 182: Preparing to open what? Be careful with INFO - this is output that the user observes - the vast majority of which do not know the implementation code

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,252 @@
...
+  // open them, slurp up the info, and stuff it into the master
+  // metadata object.
+  auto metadata_io_default = std::make_unique<rosbag2_storage::MetadataIo>();
+  auto bag_reader = std::make_unique<rosbag2_cpp::readers::SequentialReader>(
+    std::move(storage_factory_), converter_factory_, std::move(metadata_io_default));
+  ROSBAG2_CPP_LOG_INFO_STREAM("Extracting metadata from database(s)");
+  for (const auto & f_ : files) {
+    ROSBAG2_CPP_LOG_INFO_STREAM("Extracting from file: " + f_.string());
+
+    metadata_.bag_size = f_.file_size();
```
emersonknapp, line 168:
```suggestion
    metadata_.bag_size += f_.file_size();
```

I think you ran into a preexisting bug (i have yet to confirm) that the file size always simply reports the last file's size? If so,  we shouldn't propagate incorrect behavior here.

emersonknapp (changes requested):
The API is looking pretty good. A few categories of comments:
* usage of INFO should be very clear to the uninitiated
* if it's marked "nitpick" i'm not going to block on it, but it might be nice
* I don't think it's ok to hardcode the sqlite implementation

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: It was an old comment that I kept overlooking. It's gone now.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: How do I get the storage implementation give the extension? I just want to make sure that we don't pick up something like a `.db3-wal` file, and mis-collect it as a relative file...

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: I would not mind converting the Reindexer to a purely functional implementation. It's basically a one-and-done operation, so having an object made out of it could be considered a sort of "waste of space".

Would you like me to make the changes? If so, where should these functions live? A new namespace named something like `reindex`? Or just have them sort of float around in `rosbag2_cpp`?

emersonknapp: Up to you - either make sure the class API is consistent, or refactor into being a single public function `rosbag2_cpp::reindex_bag` (with private implementation in an unnamed namespace in the .cpp). Either solution works for me. The class makes the most sense if you forsee a need for reuse - e.g. do we ever want to call the same reindexer multiple times, or is it fine to just construct new reindexer for every reindexing operation (I'm leaning to the latter)

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
emersonknapp: I think a new pure-virtual API in https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_io_interface.hpp would make sense - `virtual std::string get_default_bagfile_extension() = 0;`

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: When I was originally writing the reindexer, I had actually considered doing it all as one function. It's just that the other operations were all classes, so I wanted to fit in with how you guys were doing things. But I think that having the reindexer as just a function makes a lot of sense - it's only really getting run once per bag, and doesn't need to remember any information from run to run. There's no real "state", in that regard.

So I think that it's certainly doable. I'll finish up getting the storage API to be able to pass the expected file extension (it's looking like the storage is a protected member variable of the reader, so I may have to make a new  function for it; unless you know of a better way to query that info?), and then I'll get started on this.

emersonknapp: You have the `StorageFactory` - so you could construct your own storage instance just to call this function

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: Unfortunately, I'm running into a problem getting this to work. I can create the API just fine, but actually using it in the Reindexer is another story.

The Reindexer uses a Reader instance in order to access the files, rather than using the storage directly. This creates a couple problems: The storage within the Reader is `protected` which means that I can't get to it from my Reader instance. Trying to get around this problem by creating a function that queries the storage for its extension makes me run into another "interesting" issue: until the reader opens a file, `storage_` doesn't exist!

Which means, in order to determine which files I'm allowed to open, I have to open a file...

The only way I can think of to get around this problem would be to create a function that queries the `rosbag2_storage::StorageOptions` that would be used to open the file, and query its storage_id. Then I construct the extension from that. But that opens a slight risk of code duplication, and I'm not sure how I'd be able to make it extendable to other bag types.

Any ideas?

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: Unfortunately, I still run into the same problem: In order for the storage factory to create a storage instance, it needs a proper `storage_options` URI, complete with extension. Otherwise, I get errors like:
    Could not open `/rosbag2_cpp/test/rosbag2_cpp/reindex_test_bags/multiple_files` with `sqlite3`. Error: Failed to setup storage. Error: Could not read-only open database. SQLite error (10): disk I/O error

emersonknapp: Let's just go ahead and ignore the suffix for now then - I don't want to block on getting this 100% perfect - let's just avoid hardcoding `sqlite3` stuff

jhdcs:
I _think_ I've addressed all the concerns you've raised. If I missed one, please let me know!

I think I will hold off on turning the reindexer into just a function, rather than a class, until after it's current version is proven to work and is merged in. I still have to write the CLI for this, after all.

emersonknapp:
Can you go up and mark all the resolved comments as "resolved"? That helps note what's been intentionally addressed (vs unintentionally forgotten)

jhdcs:
> Can you go up and mark all the resolved comments as "resolved"? That helps note what's been intentionally addressed (vs unintentionally forgotten)

Oh, right. Sure, I'll get on that.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: Possibly, but I wasn't positive, so I erred on the side of caution.

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,128 @@
...
+// b) there are no inline functions.
+#ifdef _WIN32
+# pragma warning(push)
+# pragma warning(disable:4251)
+#endif
+
+namespace rosbag2_cpp
+{
+
+/*************************************************************************************************
```
emersonknapp, line 57: Still too many stars - this should still be

```
/**
 *
 */
```

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,128 @@
...
+public:
+  Reindexer(
+    std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory =
+    std::make_unique<rosbag2_storage::StorageFactory>(),
+    std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io =
+    std::make_unique<rosbag2_storage::MetadataIo>());
+
+  virtual ~Reindexer();
+
+  // Uses the supplied storage options to reindex a bag defined by the storage options URI/
```
emersonknapp, line 84:
```suggestion
  /// Use the supplied storage options to reindex a bag defined by the storage options URI.
```

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -0,0 +1,128 @@
...
+  /*
+   * The passed-in storage options should contain the best-guess information of the bag's
+   * original creation parameters.
+  */
```
emersonknapp, line 88:
```suggestion
  /**
   * \param storage_options Provides best-guess parameters for the bag's original settings. 
   */
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+#ifdef WIN32
+// Import windows filesystem functionality
+#include <windows.h>
+#include <tchar.h>
+#include <stdio.h>
+#else
+// We're on a UNIX system. Import their filesystem stuff instead
+#include <dirent.h>
+#endif
```
emersonknapp, line 52:
```suggestion
```

You don't still need this because you got it from `rcpputils`/`rcutils`, right?

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,277 @@
...
+Reindexer::Reindexer(
+  std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
+  std::unique_ptr<rosbag2_storage::MetadataIo> metadata_io)
+: storage_factory_(std::move(storage_factory)),
+  metadata_io_(std::move(metadata_io))
+{
+  regex_bag_pattern_ = R"(.+_(\d+)\.([a-zA-Z0-9])+)";
+}
+
+Reindexer::~Reindexer() {}
```
emersonknapp, line 66: nit: you could leave this as `virtual ~Reindexer() = default` in the header if you wanted to.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+// Determines which path should be placed first in a vector ordered by file number.
```
emersonknapp, line 68:
```suggestion
/// Determine which path should be placed first in a vector ordered by file number.
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,277 @@
...
+    throw std::runtime_error(error_text.c_str());
+  }
+
+  u_int32_t first_db_num = std::stoul(first_match.str(1), nullptr, 10);
+  u_int32_t second_db_num = std::stoul(second_match.str(1), nullptr, 10);
+
+  return first_db_num < second_db_num;
+}
+
+// Retrieves bag storage files from the bag directory.
```
emersonknapp, line 108:
```suggestion
/// Retrieve bag storage files from the bag directory.
```

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * @param output: A vector to save the discovered files inside of
```
emersonknapp, line 110: nit: If you're gonna list one param, you might as well list them all

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_info_interface.hpp
```diff
@@ -51,6 +51,12 @@ class ROSBAG2_STORAGE_PUBLIC BaseInfoInterface
    * \returns the identifier.
    */
   virtual std::string get_storage_identifier() const = 0;
+
+  /**
+   * Returns the extension of the bagfile.
+   * \returns the bagfile's extension
+   */
+  virtual std::string get_storage_extension() const = 0;
```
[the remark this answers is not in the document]
emersonknapp: Rm?

emersonknapp (commented):
Looks good other than a few minor formatting nits, and to remove the `file_extension` stuff. Considering it's not being used it doesn't fit in this PR. 

It could be a useful thing, but would need to be a separate PR with a clear use case.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: Right. I just forgot to remove it.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
jhdcs: I don't know that it will be necessary. I was able to move the initialization of the `Reader` out of the `aggregate_metadata` function, so I didn't have to keep recreating `metadata_io_default` objects. I _did_ have to keep creating `StorageOptions` objects for each new URI, but I don't think there's a pleasant way around that.

emersonknapp (approved):
All right! This is looking good to me. Thanks for going through the revisions. Let's get going on the UI!

emersonknapp:
Note: we'll need to run ci.ros2.org Jenkins before merging - but it's down for maintenance at the moment. I'll see if it's up later today and try again.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13747/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8616/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11458/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13821/)

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -86,8 +86,8 @@ bool Reindexer::compare_relative_file(
     throw std::runtime_error(error_text.c_str());
   }
 
-  u_int32_t first_db_num = std::stoul(first_match.str(1), nullptr, 10);
-  u_int32_t second_db_num = std::stoul(second_match.str(1), nullptr, 10);
+  auto first_db_num = std::stoul(first_match.str(1), nullptr, 10);
```
emersonknapp, line 89: I think that auto is fine here - but as a note, the portable C++ type is called `uint32_t` (only one underscore)

jhdcs: Ah. That explains a lot.

I _THOUGHT_ that type looked odd, but my editor seemed to like it...

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13752/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8620/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11463/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13828/)

emersonknapp:
I'm concerned about this OSX test - it seems like it failed both times. Do you have access to an OSX machine to try that out?

jhdcs:
I do not, unfortunately. And the error isn't very descriptive, is it?

jhdcs:
I just found this in the logs:
`[ERROR] [1614384152.306481264] [rosbag2_storage]: Could not open '/Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/jhdcs/rosbag2/rosbag2_cpp/test/rosbag2_cpp/reindex_test_bags/multiple_files/multiple_files_0.db3' with 'sqlite3'. Error: Failed to setup storage. Error: Error when processing SQL statement. SQLite error (14): unable to open database file
[ERROR] [1614384152.306521279] [rosbag2_storage]: Could not load/open plugin with storage id 'sqlite3'.`

So it looks like it's trying to open the right file, but something with OSX doesn't like the format? Am I understanding that right?

jhdcs:
I found a location that looks like it should have been logging to the error stream, but was actually logging to the debug stream. It seems to be for the case that the storage factory is unable to find a registered class with the passed storage id (In this case, `sqlite3`). Since I'm not seeing any other errors in the OSX logs, I think this might be where the error occurs.

Would you be able to run the OSX tests one more time, and see if that error does show up? And if it does, what would cause `sqlite3` to not be a registered class in OSX, but only for the `reindex` case? I don't see other tests failing when opening `.db3` files, so it's got to be something special that OSX requires for the reindexer to do for registering classes...

emersonknapp:
* osx rebuild [Build Status](https://ci.ros2.org/job/ci_osx/11487/)

Based on that analysis (thanks for continuing to dig!) - I noticed that `rosbag2_storage_default_plugins` isn't explicitly a `test_depend` of `rosbag2_cpp` (https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/package.xml). It is marked as `exec_depend` though, which I believe is incorrect (rosbag2_cpp should be completely independent of storage implementation).

Now that this has been brought to our attention - I am realizing that tests that depend on plugins being installed should be run through `rosbag2_tests` anyways - we don't want to build implementation dependencies into the core library APIs. Can you move this concrete reindexer test over to that package? Any more specific unit tests are still good to live in `rosbag2_cpp`

jhdcs:
Sure, I can do that. Glad to be of help!

I'm still going to leave that change from my last commit in, though. I think that probably was a bug.

emersonknapp:
> I'm still going to leave that change from my last commit in, though. I think that probably was a bug.

That's fine - I believe I wrote that one as an error for compression plugins. Probably the plugin factories should get consolidated into a single bit of code, it's a bit duplicated right now.

jhdcs:
Moved the testing over. I'm noticing that the error "Requested storage id `sqlite3` does not exist" just before I open each database in the bag. But the test passes, so I guess the system recovers.

Trying to track down _WHY_ those messages pop up, and _WHY_ it recovers, though...

emersonknapp (approved):

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13933/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8775/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11637/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14019/)

emersonknapp:
OSX: [Build Status](https://ci.ros2.org/job/ci_osx/11662/)

emersonknapp:
Hey @jhdcs after testing it on the mackbook, I couldn't quite figure out what was wrong, but the database files in the `multiple_files` test were subtly misconfigured in some way that only showed up on OSX. I recreated the files with simple `ros2 topic pub` and `ros2 bag record`.

See https://github.com/ros2/rosbag2/pull/691 for my tweaks. The only important part are the actual `.db3` and the metadata file, feel free to just grab those directly and use them for the test (note there is only `_0`, `_1`, and `_2` - I didn't see a specific need to have 4).

I rebased on `master` there as well, you're in fine shape, the only conflict is in `rosbag2_tests/CMakeLists.txt` - the `if (fastrtps)` if-check was removed so all those tests moved out an indentation level.

LMK how you want to proceed - you can resolve it from here, or I'm happy to push changes to this branch, run CI, and get it merged.

jhdcs:
If you're willing to push the changes to the branch, that would be amazing! Please, go ahead! You are correct that the number of files in the bag doesn't really make a difference. I'm only trying to check that the reindexer tallies things properly when dealing with bag splitting. My only concern now is whether or not the other bagfiles I added are corrupt as well... I suppose we at least know what to look for, now.

What sort of corruption occurred? In order to create the original bagfile, I just used the talker/listener examples built in with ROS 2, run in a VM running Ubuntu 20.04. Perhaps sometime we should schedule a test where I create a bagfile, record **ALL** steps I took, and look for this corruption? As well as triple-check it can be run/reindexed on OSX... But that would be on a separate issue.

emersonknapp:
> What sort of corruption occurred?

I'm really not entirely sure. I was able to open the files with `sqlite3` CLI on OSX, it was just the C++ API that was unable to open the files. It's worth investigating further if there is some issue of recording on Ubuntu Focal and playing back on OSX - but I don't want to block progress here on that.

I'll get this through to merge today.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/64a7282df97ba795d6629acf27242136/raw/18f71a06ca711cf38f35718b361f917f8c1afc91/ros2.repos
BUILD args: --packages-up-to rosbag2_compression rosbag2_cpp rosbag2_storage rosbag2_storage_default_plugins rosbag2_tests
TEST args: --packages-select rosbag2_compression rosbag2_cpp rosbag2_storage rosbag2_storage_default_plugins rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7984
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14080/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8898/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11773/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14172/)

PR #643, opened 2021-02-09
Fix reported % become over 100% (#642)

### LINKED ISSUE #642: benchmark tool reports higher than 100% recorded percentage
## Description

`rosbag2_performance_benchmarking` reports higher than 100% recorded percentage.

### Expected Behavior
`<= 100%`

### Actual Behavior

When we run test with transport mode, we got the following report.

```
        Recorded messages for different caches and storage config:
                storage config: default:
                        cache [bytes]: 100,000,000: 100.04% recorded
```

## To Reproduce

Change `no_transport: False`.
https://github.com/ros2/rosbag2/blob/master/rosbag2_performance/rosbag2_performance_benchmarking/config/benchmarks/test.yaml#L8

And follow the steps in https://github.com/ros2/rosbag2/tree/master/rosbag2_performance/rosbag2_performance_benchmarking.

```
$ ros2 launch rosbag2_performance_benchmarking benchmark_launch.py benchmark:=`ros2 pkg prefix rosbag2_performance_benchmarking`/share/rosbag2_performance_benchmarking/config/benchmarks/test.yaml producers:=`ros2 pkg prefix rosbag2_performance_benchmarking`/share/rosbag2_performance_benchmarking/config/producers/mixed_110Mbs.yaml

$ scripts/report_gen.py -i <BENCHMARK_RESULT_DIR>
```

 - OS: ubuntu 20.04
 - ROS 2 Distro: rolling
 - Version: 7548a04c3b5839ed839ab3ae7c5d25f0d522807d

## Additional context

- `report_gen.py` looks `total_recorded_count` in `results.csv`
- `total_recorded_count` is from `message_count` in `metadata.yaml`
- `message_count` contains ALL topics including `/rosout` or `/parameter_events` and so on.

It looks OK to filter message counts in `result_utils.cpp`.

y-okumura-isp:
Fixed #642

Signed-off-by: y-okumura-isp <y-okumura@isp.co.jp>

### DIFF
--- a/rosbag2_performance/rosbag2_performance_benchmarking/src/result_utils.cpp
+++ b/rosbag2_performance/rosbag2_performance_benchmarking/src/result_utils.cpp
@@ -49,9 +49,19 @@ int get_message_count_from_metadata(const std::string & uri)
   int total_recorded_count = 0;
   std::string metadata_filename(rosbag2_storage::MetadataIo::metadata_filename);
   std::string metadata_path = uri + "/" + metadata_filename;
+  const std::string topic_prefix = "/rosbag2_performance_benchmarking_node";
+
   try {
     YAML::Node yaml_file = YAML::LoadFile(metadata_path);
-    total_recorded_count = yaml_file["rosbag2_bagfile_information"]["message_count"].as<int>();
+    YAML::Node topics = yaml_file["rosbag2_bagfile_information"]["topics_with_message_count"];
+    for (auto it = topics.begin(); it != topics.end(); it++) {
+      auto metadata = (*it)["topic_metadata"];
+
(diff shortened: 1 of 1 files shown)


pijaro:
There is already pending PR that fixes this problem, but in a different way: https://github.com/ros2/rosbag2/pull/634 

In proposed https://github.com/ros2/rosbag2/pull/634, the filtering is performed with `--regex` flag of `ros2 bag record`, which, in my opinion, is more suitable here (since we are no longer writing unnecessary messages and already existing API is used).

y-okumura-isp:
Thank you for comments.
I feel #634 is better way, so I close this PR.

PR #631, opened 2021-02-03
Zstd should not install internal headers

emersonknapp:
Some of them try include others that aren't installed. We don't use them. Avoid the situation entirely.

Specific case: `cover.h` includes `mem.h`, which isn't installed. This causes the ABI/API checkers on PR builds to fail because they can't find `mem.h`

PR #653, opened 2021-02-12
[foxy backport] Zstd should not install internal headers - some of them try include others that aren't installed. We don't use them. Avoid the situation (#631)

emersonknapp:
Backport #631 to Foxy


Karsten1987 (approved):
lgtm given the PR build runs through.
I think technically that's already a breaking change (people might link against the installed headers), but I feel confident to let that slip here.

emersonknapp:
> I think technically that's already a breaking change (people might link against the installed headers)

Yes, I think you're right, but I also feel confident that nobody knows about or uses these headers.

emersonknapp:

emersonknapp:
The failure is just the API/ABI checker failing on these Zstd headers - I'll merge this and get a Foxy patch build out for rosbag2 in hopes that we can remedy the situation

Issue #319, opened 2020-03-12
Remove rmw_fastrtps dependency from rosbag2 cdr converter plugin


rotu:
## Description
When running rosbag2, it does not use the expected RMW implementation, instead always using FastRTPS.

Instead of depending on `rmw_fastrtps_cpp`, depend on Fast-CDR directly to get the CDR implementation for rosbag2_converters_default_plugins

### Expected Behavior
If RMW_IMPLEMENTATION=rmw_cyclonedds_cpp, cdr_converter should use rmw_cyclonedds_cpp.

### Actual Behavior
If RMW_IMPLEMENTATION=rmw_cyclonedds_cpp, rosbag2_converter_default_plugins tests fail.

```
/opt/ros/master/build/rosbag2_converter_default_plugins/Testing/20200312-1905/Test.xml: 7 tests, 0 errors, 1 failure, 0 skipped
- test_cdr_converter
  <<< failure message
    -- run_test.py: invoking following command in '/opt/ros/master/build/rosbag2_converter_default_plugins':
     - /opt/ros/master/build/rosbag2_converter_default_plugins/test_cdr_converter --gtest_output=xml:/opt/ros/master/build/rosbag2_converter_default_plugins/test_cdr_converter.gtest.xml
    [ERROR] [1584039940.511109472] [rcl]: Expected RMW implementation identifier of 'rmw_cyclonedds_cpp' but instead found 'rmw_fastrtps_cpp', exiting with 102.
    -- run_test.py: return code 102
    -- run_test.py: generate result file '/opt/ros/master/build/rosbag2_converter_default_plugins/test_cdr_converter.gtest.xml' with failed test
    -- run_test.py: verify result file '/opt/ros/master/build/rosbag2_converter_default_plugins/test_cdr_converter.gtest.xml'
  >>>
/opt/ros/master/build/rosbag2_converter_default_plugins/test_cdr_converter.gtest.xml: 1 test, 0 errors, 1 failure, 0 skipped
- rosbag2_converter_default_plugins test_cdr_converter.gtest.missing_result
  <<< failure message
    The test did not generate a result file:
    
    [ERROR] [1584039940.511109472] [rcl]: Expected RMW implementation identifier of 'rmw_cyclonedds_cpp' but instead found 'rmw_fastrtps_cpp', exiting with 102.
  >>>
[...]

PR #662, opened 2021-02-18
Try to make PR builder pass

emersonknapp:
The PR builder hasn't been working since the default rmw implementation changed on the buildfarm.https://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/

Hypothesis: internally to the core, the default RMW implementation is set to `rmw_cyclonedds_cpp`. However, since rosbag2 installs `rmw_fastrtps_cpp` explicitly as a dependency, installing `rmw` doesn't force Cyclone to get installed, because the "at least one RMW implementation" requirement is met. Even though `rmw_fastrtps_cpp` is the only installed RMW implementation, the default value is still set to `rmw_cyclonedds_cpp`, it's not dynamically checked.

By adding this dependency explicitly, the PR builder should install the default and be able to pass tests when `RMW_IMPLEMENTATION` is not specified.

This won't be necessary when we remove the converter dependency on `rmw_fastrtps_cpp` (and maybe we should just do that instead of this - I am testing the above hypothesis via this PR)

### DIFF
--- a/rosbag2_tests/package.xml
+++ b/rosbag2_tests/package.xml
@@ -18,6 +18,11 @@
   <test_depend>ament_lint_common</test_depend>
   <test_depend>rclcpp</test_depend>
   <test_depend>rcpputils</test_depend>
+  <!--
+  TODO(emersonknapp): remove this once we aren't pulling in fastrtps as a dependency.
+  For now, since fastrtps is installed, cyclone is not automatically installed for PR builds
+  -->
+  <test_depend>rmw_cyclonedds_cpp</test_depend>
   <test_depend>ros2bag</test_depend>
   <test_depend>rosbag2_compression</test_depend>
   <test_depend>rosbag2_compression_zstd</test_depend>


clalancette:
Yeah, this is what I was getting at with https://github.com/ros2/rosbag2/issues/319#issuecomment-777829206 .  In point of fact, there are a number of ways to solve this:

1. Do what you are doing here, and add in the explicit dependency on `rmw_cyclonedds_cpp`.
2. Switch the dependency from `rmw_fastrtps_cpp` to just `fastcdr`.
3. Change [load_library](https://github.com/ros2/rmw_implementation/blob/dadb0ec69c5fafec15e5fb0fcc315e6a81d6f4fd/rmw_implementation/src/functions.cpp#L46) to fallback and try other available RMW implementations.

While 1 here will work, it is pretty icky in that you are introducing an artificial dependency, and it might not be correct in the future if we change the default RMW again.  The second option seems to be the best since you "only" need a CDR serializer/deserializer, though it is also probably the most amount of work.  Fixing item 3 might be interesting regardless of the issues here with PR jobs, but I haven't thought through all of the implications.

emersonknapp:
> While 1 here will work, it is pretty icky in that you are introducing an artificial dependency, 

Agreed that it's icky - I was just trying to figure out if we should do (1) it temporarily until I can get to a better solution for #319 via (2) , which might not be for a few more weeks.

Issue #652, opened 2021-02-12
Support zero-copy transport


albtam:
## Description

**Edit**

ROS 2 Foxy brought zero-copy APIs to publish and take messages. `rosbag2` should be able to support zero-copy transport.

## How Apex.AI supports zero-copy transport with rosbag2

Since RTI Connext DDS Micro does not provide an API to get the serialized data, our workflow to record and replay messages is different compared to the others DDS implementations. When a message is recorded, the following methods are called: [execute_subscription](https://github.com/ros2/rclcpp/blob/7ccd64c9c1d7b7c4ae996cb114157c3d4211cf3f/rclcpp/src/rclcpp/executor.cpp#L511) -> [rcl_take_serialized_msg](https://github.com/ros2/rclcpp/blob/2531787550084307a5c8144a68c0195fcdd6442a/rclcpp/src/rclcpp/subscription_base.cpp#L166) -> [rmw_take_serialized_message_with_info](https://github.com/ros2/rcl/blob/6d38be606ee3a16bd67b89830029c1201f07933b/rcl/src/rcl/subscription.c#L295)

Until now, this is the same workflow with or without Connext Micro.

What changes in Connext Micro's rmw implementation is that we actually take a non-serialized message (that could be a loaned message) -- by-passing the ROS 2 `take_loaned_message_*` API -- and explicitly serialize the message.

That's the reason why we are already able to use zero-copy. As you might have noticed when I am talking about zero-copy, I don't mean true zero-copy in the sense that the message still needs to be serialized but only the IPC mechanism that can avoid copying the messages to/from SHMEM.

As for the publisher side (rosbag player), the ideal case would be as Karsten explained to "[get] a large enough byte-buffer from the middleware without any further type information". Since this is currently not supported in RTI Connext DDS Micro, we actually borrow a loaned message and fill it with the deserialized message. (see https://github.com/ros2/rosbag2/pull/664) Publishing the message with the existing rosbag2 API (i.e. without borrowing a message) doesn't work because the actual message is [annotated in a way](https://community.rti.com/static/documentation/connext-micro/3.0.3/doc/html/usersmanual/zerocopy.html#getting-started) that forces the user to borrow a loan to publish the message.

## Related PR

https://github.com/ros2/rosbag2/pull/664

## Implementation Notes / Suggestions

After a second thought, I am not sure that what Apex.AI did can be reused as a generic rosbag2. I still opened a PR for reference.

What could make sense in the upstream implementation is:

[...]

PR #664, opened 2021-02-19
Allow a loaned message to be published

albtam:
Closes https://github.com/ros2/rosbag2/issues/652

I was not able to successfully build master (against ROS 2 Foxy) with the following commands:

```bash
$ docker pull osrf/ros:foxy-desktop
$ docker run --name=foxy -it osrf/ros:foxy-desktop bash
$ source /opt/ros/foxy/setup.bash
$  mkdir -p ~/rosbag_ws/src && cd ~/rosbag_ws/src
$ git clone https://github.com/ros2/rosbag2.git
$ cd ..
$ colcon build
--- stderr: rosbag2_storage                                                                         
/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp: In member function ‘virtual rosbag2_storage::BagMetadata rosbag2_storage::MetadataIo::read_metadata(const string&)’:
/rosbag2/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp:236:53: error: could not convert ‘& metadata.rosbag2_storage::BagMetadata::bag_size’ from ‘uint64_t*’ {aka ‘long unsigned int*’} to ‘rcutils_allocator_t’
  236 |       rcutils_calculate_directory_size(uri.c_str(), &metadata.bag_size, allocator))
      |                                                     ^~~~~~~~~~~~~~~~~~
      |                                                     |
      |                                                     uint64_t* {aka long unsigned int*}
```

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
@@ -35,7 +35,8 @@ GenericPublisher::GenericPublisher(
   const rosidl_message_type_support_t & type_support,
   const std::string & topic_name,
   const rclcpp::QoS & qos)
-: rclcpp::PublisherBase(node_base, topic_name, type_support, rosbag2_get_publisher_options(qos))
+: rclcpp::PublisherBase(node_base, topic_name, type_support, rosbag2_get_publisher_options(qos)),
+  type_support_{type_support}
 {}
 
 void GenericPublisher::publish(std::shared_ptr<rmw_serialized_message_t> message)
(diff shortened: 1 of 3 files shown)


emersonknapp:
> I was not able to successfully build master (against ROS 2 Foxy)

`master` is the latest development and is now Rolling/Galactic - it has diverged since the Foxy release

emersonknapp:
To clarify - we will need to make this upgrade against master, then we can selectively backport to released distros (e.g. Foxy) after

albtam:
Thanks @emersonknapp , pulling `osrf/ros2:nightly` instead solved my build problem.

emersonknapp:
Quick note - to pass the DCO check you'll need to sign your commits with `--signoff`. Given that it's just a single commit right now, you can achieve that with `git commit --amend --signoff`

rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
```diff
@@ -48,4 +49,34 @@ void GenericPublisher::publish(std::shared_ptr<rmw_serialized_message_t> message
   }
 }
 
+void GenericPublisher::publish_loaned_message(void * ros_message)
```
Karsten1987, line 52: I am having a bit of a problem with this signature. First off, I don't really like to publish a void pointer. That gives us really the least amount of type safety here. The least we should do is to verify that it's not null.
But then further, this function actually requires a full-fleshed ROS message, opposing to a serialized message like the rest of the publishing functions. What's the benefit here?

I would actually see that there should be a `SerializedLoanedMessage` and the appropriate set of RMW functions to loan a naive data buffer. While with this approach you might safe a copy in the initialization phase, you add another two complex steps (deserialize message to ROS2, copy ros2 into loaned message).

Karsten1987 (commented):
This code needs tests to reflect the newly introduced function behavior. Code looks okay so far, even though I am really not a fan of publishing a `void *` message. I don't feel too bad about it given that it's basically a private API, but I can see this being complicated when trying to bring this in as a public API in rclcpp.

albtam:
> friendly ping @albtam do you have any thoughts on Karsten's comments above?

I agree with Karsten's feedback. I'll try to address his comment soon but it's a quite busy time for me

rosbag2_transport/src/rosbag2_transport/generic_publisher.cpp
albtam: @Karsten1987 

I agree that a typed interface would be better.  I am just using the interface defined in rcl (https://github.com/ros2/rcl/blob/master/rcl/include/rcl/publisher.h#L407) though. This interface takes a `void * ros_message`. The `ros_message` is not dereferenced in `GenericPublisher::publish_loaned_message`, that's why I did not check for a nullptr. `rcl_publish_loaned_message` does check for nullptr.

What would you suggest here?

albtam:
@Karsten1987 @emersonknapp 

About the tests, we internally create a message with a DDS implementation that supports zero-copy. In other words, the test depends on a specific DDS implementation as well as the proposed changes.

AFAIK, the next default DDS implementation (Cyclone DDS) does not support zero-copy yet.  So I would propose to close this PR and reopen it when zero-copy becomes supported by the default middleware. Let me know if you agree.

To come back to the initial issue addressed, the problem was that the ROS 2 APIs allowed to use the normal `publish` API even if the publisher was able to loan messages. If this is not the desired behavior, it might be worth adding some logic in  `rclpcpp`  to forbid this use-case.

Karsten1987:
@albtam I'll go ahead and close your PR according to your last comment. Happy to re-address it once the setup of zero-copy is in place.

PR #670, opened 2021-03-05
RMW-implementation-searcher converter in rosbag2_cpp

### LINKED ISSUE #319: Remove rmw_fastrtps dependency from rosbag2 cdr converter plugin

emersonknapp:
Fixes #319 

## Overview

* Converter looks for a registered converter plugin first (existing logic unchanged)
* If no suitable plugin is found, fall back to looking for an RMW implementation that can do the job (`RMWImplementedConverter` - introduced by this PR)
    * This fallback is not implemented as a plugin - it searched based on the format name, and so does not have a format name at compile-time to register itself as a specific format plugin. 
* Removed `rosbag2_converter_default_plugins` because this implementation is a slightly more generic version of the "CDR converter plugin" that lived there. So, there are now no converter plugins installed by default. If a user has installed a converter  plugin, that will be preferred, but on a default system the `RMWImplementedConverter` will be called when type conversion is requested.


rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
```diff
@@ -0,0 +1,132 @@
...
+template<typename T>
+T get_function_from(const char * function_name, std::shared_ptr<rcpputils::SharedLibrary> library)
+{
+  if (!library->has_symbol(function_name)) {
+    std::stringstream ss;
+    ss << "Converter could not find expected symbol '" << function_name <<
+      "' in rmw implementation library " << library->get_library_path();
+    throw std::runtime_error{ss.str()};
+  }
+  T loaded_function = nullptr;
```
Karsten1987, line 40: why do I need this line? Why not directly `T loaded_function = (T)(library->get_symbol(function_name));`?

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
```diff
@@ -0,0 +1,132 @@
...
+  // Function expected to return a value because has_symbol was checked first.
+  loaded_function = (decltype(loaded_function))(library->get_symbol(function_name));
+  return loaded_function;
+}
+}  // namespace
+
+namespace rosbag2_cpp
+{
+
+class RMWImplementedConverterImpl
```
Karsten1987, line 50: I can't help it, but that name somewhat sounds complicated to me. Unfortunately, I also don't have a proposal for a better name :P

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
```diff
@@ -0,0 +1,132 @@
...
+      serialize_fn_ = &rmw_serialize;
+      return;
+    }
+
+    auto packages_with_prefixes = ament_index_cpp::get_resources("rmw_typesupport");
+
+    for (const auto & package_prefix_pair : packages_with_prefixes) {
+      ROSBAG2_CPP_LOG_INFO_STREAM("Found package " << package_prefix_pair.first);
+      const auto & pkg = package_prefix_pair.first;
+      if (pkg == "rmw_implementation") {
```
Karsten1987, line 67: why do we have to filter out `rmw_implementation`? Why does this package even return when querying `rmw_typesupport`?

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
```diff
@@ -0,0 +1,132 @@
...
+  explicit RMWImplementedConverterImpl(const std::string & format)
+  {
+    const std::string current_implementation_format{rmw_get_serialization_format()};
+    if (current_implementation_format == format) {
+      deserialize_fn_ = &rmw_deserialize;
+      serialize_fn_ = &rmw_serialize;
+      return;
+    }
+
+    auto packages_with_prefixes = ament_index_cpp::get_resources("rmw_typesupport");
```
Karsten1987, line 62: why do we have to filter for `rmw_typesupport`? Isn't there a query to return all registered rmw implementations just like in cmake? https://github.com/ros2/system_tests/blob/master/test_rclcpp/CMakeLists.txt#L40

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
```diff
@@ -0,0 +1,132 @@
...
+      library_ = std::make_shared<rcpputils::SharedLibrary>(libpath);
+
+      auto get_format_fn = get_function_from<decltype(&rmw_get_serialization_format)>(
+        "rmw_get_serialization_format", library_);
+      const char * impl_serialization_format = get_format_fn();
+      if (impl_serialization_format == format) {
+        deserialize_fn_ = get_function_from<decltype(deserialize_fn_)>(
+          "rmw_deserialize", library_);
+        serialize_fn_ = get_function_from<decltype(serialize_fn_)>(
+          "rmw_serialize", library_);
```
Karsten1987, line 86: might not be an immediate problem, but I wanted to mention it anyway.

For Fast-RTPS we were certain that the two functions (`rmw_deserialize, rmw_serialize`) could be called without any state context - that is `rmw_init`. However in this generic case here, can we always guarantee that I can call these functions without any prior calls to initialize the RMW?

emersonknapp:
@Karsten1987 thanks for the review - can I take it based on the smaller comments that you think this high level approach is all right? I'd like to make sure of that before diving into small implementation details. Regardless - I'll answer questions above

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
emersonknapp: extra declaration not needed, that line works fine

Karsten1987:
yeah, absolutely. Approach seems legit to me. From a user cli perspective, nothing actually changes here, right?
You still only specify a string-identifier for the converter and we look for a suited RMW implementation first, then fallback to something else? I am still a bit uncertain for the "something else" part though, but maybe you'll address this in a second PR?

emersonknapp:
> From a user cli perspective, nothing actually changes here, right?

Yes, user will not see any difference.

> You still only specify a string-identifier for the converter and we look for a suited RMW implementation first, then fallback to something else? I am still a bit uncertain for the "something else" part though, but maybe you'll address this in a second PR?

No, it's the other way around.
* Converter looks for a registered converter plugin first (the already existing logic)
* If no suitable plugin is found, fall back to looking for an RMW implementation that can do the job (introduced by this PR)
* I've removed `rosbag2_converter_default_plugins` because this implementation is a slightly more generic version of the "CDR converter plugin" we already had - so there are now now converter plugins installed by default - so unless a user has installed a converter plugin this fallback will be the default behavior

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
emersonknapp: There may not be such a guarantee. `rmw_init` creates an `rmw_context_t`, which you have to pass to certain API functions. For RMW functions that do not require an `rmw_context_t`, it's unclear to me whether `rmw_init` was required to have been called. It may not hurt to call it here and hold onto a context in the converter.

rosbag2_cpp/src/rosbag2_cpp/rmw_implemented_serialization_format_converter.cpp
emersonknapp: I looked at it, and there is only `ament_index_cpp`, `ament_index_python`, and `ament_index_cmake` - so unless I want to rewrite the ament index code in C, `rmw` and `rcl` are out. I don't really want to introduce the `rclcpp` dependency just for this one thing (assuming I created a `rclcpp::utilities::get_available_rmw_implementations`)... so given that situation, I'm inclined to just move this into a separate utility function here, out of the class constructor, with comment so that at least it is clear.

rosbag2_cpp/src/rosbag2_cpp/serialization_format_converter_factory_impl.hpp
```diff
@@ -93,32 +96,37 @@ class SerializationFormatConverterFactoryImpl
...
-    try {
-      return std::unique_ptr<SerializationFormatIface>(
-        class_loader->createUnmanagedInstance(converter_id));
-    } catch (const std::runtime_error & ex) {
-      (void) ex;  // Ignore, try to load converter instead
+      try {
+        return std::unique_ptr<SerializationFormatIface>(
+          class_loader->createUnmanagedInstance(converter_id));
+      } catch (const std::runtime_error & ex) {
+        (void) ex;  // Ignore, try to load converter instead
```
Karsten1987, line 109: I wonder if it makes sense to log your comment here.

Karsten1987 (approved):
looks good to me.
I'd propose to run CI with all DDS vendors as well as a single run with only cyclone as a default.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/7bbe063f0f1044d7a493519b39afafc1/raw/61a9f2756cc9080f46458845d4764d2b0f289d0b/ros2.repos
BUILD args: --packages-up-to rosbag2_cpp rosbag2_tests rosbag2_transport rosbag2_py rosbag2
TEST args: --packages-select rosbag2_cpp rosbag2_tests rosbag2_transport rosbag2_py rosbag2
Job: ci_launcher

PR #675, opened 2021-03-10
Design for rosbag2 handling of clock/simtime

emersonknapp:
Designs https://github.com/ros2/rosbag2/issues/99
Designs https://github.com/ros2/rosbag2/issues/55

Adds a design doc (in a new design folder) which lays out how time should be handled in rosbag2 for all anticipated playback cases.

### DIFF
--- a/docs/design/rosbag2_playback_time.md
+++ b/docs/design/rosbag2_playback_time.md
@@ -38,7 +38,7 @@ public:
    * playback_rate must be positive, 1.0 means real-time
    * starting_time provides an offset for all future times
    * clock_topic_publish_frequency:
-   *   - if > 0 - publishes `now()` to /clock at the specified frequency (in Hz)
+   *   - if > 0 - publishes `now()` to /clock at the specified frequency in Hz, rate determined by a system steady clock
    *   - if <= 0 - does not publish the /clock topic
    */
   Clock(float playback_rate, rclcpp::Time starting_time, float clock_topic_publish_frequency);
@@ -100,3 +100,16 @@ The `Player` must register a `JumpHandler` with the `Clock` - so that when a jum
 ## Open questions
 
 * Should time interpolation in the ROS clock be used for external time driver? If the publish rate is too slow, playback cannot reach any close fidelity if there is none. E.g. a topic publishing at 200Hz would need at least a 200Hz `/clock`, unless the `Clock` can interpolate between samples.
+
+## Implementation Staging
+
+This should not be implemented monolithically. Implementation should focus on small incremental PRs with solid testing that are easy to review.
+This is a proposed order of operations.
+
+* Implement `rclcpp::Clock::sleep_until` - which will allow us to handle all external time drivers (e.g. Gazebo)
+* Create `rosbag2_transport::PlayerClock` with unit tests, with rate setting - but without pause and jump
+* Expose "rosbag2 publishes to `/clock`" (the most requested feature), passing the `PlayerClock` to `Player`
+* Add pause/resume to the `PlayerClock`, controllable by Service on the Rosbag2 Node
+* Add time jump to the `PlayerClock` (and handlers to the `Player`), controllable by a Service on the Rosbag2 Node
+
+^ Note that the provided Services will allow for building GUI and CLI/keyboard controllers for this functionality, without having to build against rosbag2 directly
--- a/docs/design/rosbag2_playback_time.md
+++ b/docs/design/rosbag2_playback_time.md
@@ -1,0 +1,102 @@
+# Rosbag2 Playback Time Handling
+
+## Goal
+
+We need rosbag2 playback to do more than just playing back in real time according to the system clock.
+
+These situations must be handled
+1. No ROS-time, rosbag2 plays back messages without considering external time
+2. Rosbag2 drives ROStime by publishing to the /clock topic
+3. Rosbag2 is driven by an external ROStime source - most commonly the Gazebo simulator
+
+In the first two cases - rosbag2 needs to control the flow of time with the following operations
+* pause and resume
+* set a (forward) rate of playback - either faster or slower than real time
+* Jump back or forward to an arbitrary point in time
+
+In the final case, when rosbag2 is driven by an external source, it still must be able to handle the above operations, which may be performed by that external time controller. Therefore, a generic solution is required, so that  `rosbag2_transport::Player` is agnostic to the controller of the time.
+
+## Current Implementation Notes
+
+`rosbag2_transport::Player` currently uses `std::chrono::system_clock` to query time, and `std::this_thread::sleep_until` to wait between publishing messages, with explicit handling of playback rate.
+
+## Proposal
+
+Pass a `rclcpp::Clock` instance to `rosbag2_transport::Player`
+* Use `Clock::now` to query starting time
+* Use `Clock::sleep_until` between messages
+  * `rclcpp::Clock` does not yet implement `sleep_until` - as noted in https://github.com/ros2/rcl/issues/898
+  * If we are not able to get into `rclcpp` in time for the Galactic API freeze - it may need to temporarily go into a rosbag2-based subclass for the Galactic release.
+
+When an external source drives time, a plain `rclcpp::Clock` can be used, configured to ROStime as the source. For situations where rosbag2 drives time, a `rosbag2_transport::PlayerClock` class is created - which returns the correct time based on requirements (see API declaration below).
+
+```
+class PlayerClock : public rclcpp::Clock
+{
+public:
+  /**
+   * playback_rate must be positive, 1.0 means real-time
+   * starting_time provides an offset for all future times
+   * clock_topic_publish_frequency:
+   *   - if > 0 - publishes `now()` to /clock at the specified frequency (in Hz)
+   *   - if <= 0 - does not publish the /clock topic
+   */
+  Clock(float playback_rate, rclcpp::Time starting_time, float clock_topic_publish_frequency);
+
+  // rclcpp::Clock interface
+  /**
+   * based on an internal steady clock, calculates current "Player Time"
+   */
+  rclcpp::Time now() override;
+
+  /**
+   * Sleeps (non-busy-waiting) the current thread until the provided time is reached - according to this Clock!
+   *
+   * If time is paused, the requested time may never be reached
+   * `real_time_timeout` uses the internal steady clock, if the timeout elapses, return false
+   * If jump() is called, return false, allowing the caller to handle the new time
+   * Return true if the time has been reached
+   */
+  bool sleep_until(rclcpp::Time until, rclcpp::Duration real_time_timeout) override;
+
+  // Special playback interface
+
+  /**
+   * Pauses/resumes time.
+   * While paused, `now()` will repeatedly return the same time, until resumed
+   *
+   * Note: this could have been implemented as `set_rate(0)`, but this interface allows this clock to maintain the
+   * clock's rate internally, so that the caller does not have to save it in order to resume.
+   */
+  void set_paused(bool paused);
+
+  /**
+    * Changes the rate of playback. Must be positive (nonzero)
+    */
+  void set_rate(float rate);
+
+  /**
+    * Changes the current internally maintained offset so that next published time is different.
+    * This should trigger any JumpHandler callbacks that are registered with the clock (see "Time Jumps" section)
+    */
+  void jump(rclcpp::Time time);
+
+}
+```
+
+### Pause and Resume
+
+No special allowance needs to be made in the `Player` - time will still be provided, but it will not move forward, so the next message will not be published until time is resumed.
+
+### Setting the Rate
+
+No special allowance needs to be made in the `Player` - time will be provided at a different rate, and `sleep_until` will handle this.
+
+### Time Jumps
+
+The `Player` must register a `JumpHandler` with the `Clock` - so that when a jump occurs, the current message playback queue can be invalidated and re-enqueued according to the new starting time.
+
+
+## Open questions
+
+* Should time interpolation in the ROS clock be used for external time driver? If the publish rate is too slow, playback cannot reach any close fidelity if there is none. E.g. a topic publishing at 200Hz would need at least a 200Hz `/clock`, unless the `Clock` can interpolate between samples.
--- a/docs/design/rosbag2_playback_time.md
+++ b/docs/design/rosbag2_playback_time.md
@@ -158,6 +158,9 @@ No special allowance needs to be made in the `Player` - time will be provided at
 
 The `Player` must register a `JumpHandler` with the `Clock` - so that when a jump occurs, the current message playback queue can be invalidated and re-enqueued according to the new starting time.
 
+### Synchronizing rosbag2 playback
+
+As a consequence of this design, it is possible to synchronize the playback of multiple rosbags, by setting one to publish with `--clock`, and the others to listen with `--use-sim-time`. This feature is useful if for example bags were recorded for different topic selections, in the same time range.
 
 ## Implementation Staging
 
--- a/docs/design/rosbag2_playback_time.md
+++ b/docs/design/rosbag2_playback_time.md
@@ -44,90 +44,92 @@ Pass a `PlayerClock` instance to `rosbag2_transport::Player`
   * `rclcpp::Clock` does not yet implement `sleep_until` - as noted in https://github.com/ros2/rcl/issues/898
   * If we are not able to contribute this feature into `rclcpp` in time for the Galactic API freeze - it may need to temporarily go into a rosbag2-based subclass for the Galactic release.
 
+The following pseudocode suggests the high level function and API (but does not set it in stone)
+
 ```
-/**
- * Used to control the timing of bag playback.
- * This clock should be used to query times and sleep between message playing,
- * so that the complexity involved around time control and time sources
- * is encapsulated in this one place.
- * Internally, it may own an rclcpp::Clock, but does not override the class in order to implement
- * a slightly different API.
- */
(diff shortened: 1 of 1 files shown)


docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,115 @@
+# Rosbag2 Playback Time Handling
+
+## Goal
+
+We need rosbag2 playback to do more than just playing back in real time according to the system clock.
+
+These situations must be handled
+1. No ROS-time, rosbag2 plays back messages without considering external time
+2. Rosbag2 drives ROStime by publishing to the /clock topic
```
piraka9011, line 9: Does rosbag2 need to handle any conflicts in case two rosbags are playing or another node is publishing to this topic?

It doesn't seem like a common use case, but I think is something to consider.

emersonknapp: That's a good callout! I'll add a note for that - I think I'll just say "the user is responsible for making sure that no more than one time source is publishing to `/clock` at the same time, rosbag will provide no deconflicting". Does that sound all right?

I also just realized that setting one bag to control time and others to receive time externally, provides a way to synchronize playback between multiple separate bags! This was unintentional, but I'll call that out as well

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,115 @@
...
+   *
+   * Note: this could have been implemented as `set_rate(0)`, but this interface allows this clock to maintain the
+   * clock's rate internally, so that the caller does not have to save it in order to resume.
+   */
+  void set_paused(bool paused);
+
+  /**
+    * Changes the rate of playback. Must be positive (nonzero)
+    */
+  void set_rate(float rate);
```
piraka9011, line 76: Implementation detail, but should this `throw` if `rate < zero`?

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,115 @@
...
+No special allowance needs to be made in the `Player` - time will be provided at a different rate, and `sleep_until` will handle this.
+
+### Time Jumps
+
+The `Player` must register a `JumpHandler` with the `Clock` - so that when a jump occurs, the current message playback queue can be invalidated and re-enqueued according to the new starting time.
+
+
+## Open questions
+
+* Should time interpolation in the ROS clock be used for external time driver? If the publish rate is too slow, playback cannot reach any close fidelity if there is none. E.g. a topic publishing at 200Hz would need at least a 200Hz `/clock`, unless the `Clock` can interpolate between samples.
```
piraka9011, line 102: IIUC, I don't think we can reliably ensure that time interpolation will always work for external time drivers.
If this is the case, rosbag2 can either provide the 'best effort' rate and/or inform the user via some warning it was unable to publish at the requested rate.

docs/design/rosbag2_playback_time.md
emersonknapp: I assumed that's what i would do to enforce the "Must be positive nonzero" requirement

docs/design/rosbag2_playback_time.md
emersonknapp: I was thinking we could do something simple but reasonably effective to acieve interpolation on the receiving end, based on e.g. the last 2 samples plus a deadline timeout to handle when no new clock is received.

emersonknapp: Without any interpolation, e.g. "discrete time" - faster messages would still be published in the right order and at the right "rate", but they'd be bursty, being published in groups immediately after receiving each clock sample, rather than smoothly publishing through the gaps between clock samples. e.g.

```
# dots are arbitrary time tick, letters are when a message is published
# M is topic Message, C is /clock message

# real time (or with interpolation
M..M..M..M..M..M..M..M..M..M..M..
C.......C........C........C......

# without any interpolation if clock is slow
M.......MM.......MMM......MMM....
C.......C........C........C......
```

Karsten1987 (commented):
Thanks for kicking this off. I have two major questions I'd like to have sorted out before starting to implement this:

* How does the `/clock` publisher frequency influence the overall system if the messages have a potential higher frequency.
* What does pause/resume mean in terms of recording and playing back (are we pausing time or pausing the player/recorder).

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,115 @@
...
+```
+class PlayerClock : public rclcpp::Clock
+{
+public:
+  /**
+   * playback_rate must be positive, 1.0 means real-time
+   * starting_time provides an offset for all future times
+   * clock_topic_publish_frequency:
+   *   - if > 0 - publishes `now()` to /clock at the specified frequency in Hz, rate determined by a system steady clock
+   *   - if <= 0 - does not publish the /clock topic
```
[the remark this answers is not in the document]
emersonknapp: I've added a clarification to "starting_time" parameter to `PlayerClock` - I propose that we publish time as if we are exactly re-living the recorded session, so the published time matches with the recorded message timestamps, as provided originally by Gazebo or the wall clock

emersonknapp:
@Karsten1987  @piraka9011 I've added a lot of clarification, please let me know if there's anything else I should add!

emersonknapp:
@jaisontj @zmichaels11 @Karsten1987 @piraka9011 @mjeronimo friendly ping :) I need to get started implementing if we want to get any/all of this proposal into Galactic. I've just pushed a new revision that comes out of starting to implement and get a feel for the API. 

I plan to start start opening PRs approximately in line with the "Implementation Staging" notes in the design - so if there are any high level concerns, those would be good to hear ASAP.

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+While the above is technically true, I propose that we implement a simple extrapolation anyways for the case of subscribing to an external ROS Time Source:
+* Store internally the last 2 time samples (noting the rate that samples are received)
+* When `now()` is called, determine the rate from the stored time samples and extrapolate based on time
+* Put a watchdog on the clock to notice if new messages are not received in the expected period
+  * Deadline QoS on the `/clock` publisher is preferred for this, but we may not be able to enforce this for all ROS Time Sources
+  * Immediately "pause time" when a missing sample is noticed
+
+Possible error:
+  * ROS Time Source increases rate-of-time: playback may be slightly behind, and catches up by the next sample
+  * ROS Time Source decreases rate-of-time: playback may be slightly ahead based on extrapolated messages being published early, and
```
emersonknapp, line 165: finish sentence:
 
```suggestion
  * ROS Time Source decreases rate-of-time: playback may be slightly ahead based on extrapolated messages being published early, and will wait until the newly correct time to publish following messages - will not republish an already-published message
```

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Constructor.
+   *
+   * \param use_sim_time
+   *   If false, starts paused.
+   *     Use member functions to configure before calling set_paused(false)
+   *   If true, subscribe to /clock to provide time
+   *     All time control (rate, pause, jump) is disabled, and will not publish to /clock
+   */
+  explicit PlayerClock(bool use_sim_time);
```
emersonknapp, line 68: PlayerClock needs an API to enable/disable time extrapolation

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+    * \throws std::invalid_argument if rate is <= 0.0
+    */
+  void set_rate(float rate);
+  float get_rate() const;
+
+  /**
+   * Set the rate that the clock will be published. Defaults to 0.
+   * \param frequency If this is set to <= 0, then the clock will not be published
+   */
+  void set_clock_publish_frequency(float frequency);
```
david-prody, line 115: Is this frequency in Hertz?  Is that defined as a standard somewhere in ros?

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+  /**
+    * Change the current internally maintained offset so that next published time is different.
+    *
+    * This will trigger any registered JumpHandler callbacks.
+    * Call this with the first message timestamp for a bag before starting playback (otherwise this will return current wall time)
+    */
+  void jump(rclcpp::Time time);
+
+  rclcpp::JumpHandler::SharedPtr
+  create_jump_callback(
```
david-prody, line 127: It is unclear from the documentation if these callbacks will be executed if use_sim_time == true and the incoming sim time jumps back

docs/design/rosbag2_playback_time.md
emersonknapp: Yes. I will clarify this, to make sure that we're definitely not talking about kilohertz or millihertz :)

david-prody: Does this start publishing immediately when frequency is set to >0?   Does the client need to remember to call jump to set the initial time before calling this method?

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+
+# playback on use_sim_time with default behavior - when a new time sample is received a burst of backed up messages will be played
+M.......MM.......MMM......MMM....
+C.......C........C........C......
+```
+
+The next section "No Advanced Estimating Clock By Default" lays out that more advanced behavior is possible, though "these techniques will require making assumptions about the future behavior of the time abstraction. And in the case that playback or simulation is instantaneously paused, it will break any of these assumptions."
+
+While the above is technically true, I propose that we implement a simple extrapolation anyways for the case of subscribing to an external ROS Time Source:
+* Store internally the last 2 time samples (noting the rate that samples are received)
```
david-prody, line 157: With using only two samples to determine rate this could be a little noisy.  but thats probably fine.

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+```
+/**
+ * Used to control the timing of bag playback.
+ * This clock should be used to query times and sleep between message playing,
+ * so that the complexity involved around time control and time sources
+ * is encapsulated in this one place.
+ * Internally, it may own an rclcpp::Clock, but does not override the class in order to implement
+ * a slightly different API.
+ */
+class PlayerClock
```
Karsten1987, line 56: I'd generally be hesitant to put in larger amount of source code in a design doc. It will eventually be out of sync with the actual implementation.

Karsten1987: Could we rather create some sort of UML or simple diagram to visualize it?

docs/design/rosbag2_playback_time.md
Karsten1987: if it's hertz, I'd recommend using a `size_t` data type or something non decimal.

docs/design/rosbag2_playback_time.md
```diff
@@ -0,0 +1,200 @@
...
+# original system
+M..M..M..M..M..M..M..M..M..M..M..
+C.......C........C........C......
+
+# playback on use_sim_time with default behavior - when a new time sample is received a burst of backed up messages will be played
+M.......MM.......MMM......MMM....
+C.......C........C........C......
+```
+
+The next section "No Advanced Estimating Clock By Default" lays out that more advanced behavior is possible, though "these techniques will require making assumptions about the future behavior of the time abstraction. And in the case that playback or simulation is instantaneously paused, it will break any of these assumptions."
```
Karsten1987, line 154: I am bit confused by this sentence. Where is that section "No Advanced Estimating Clock By Default"? 

Also what exactly is meant by "future behavior of the time abstraction"?

Karsten1987 (approved):
This doc sounds pretty solid to me! Thanks for putting it together.

docs/design/rosbag2_playback_time.md
emersonknapp: It's hertz, but it could be fractional Hz, 2.5 is a valid frequency

docs/design/rosbag2_playback_time.md
emersonknapp: I was trying to leave all implementation out while suggesting an API - but you're right, it is likely to get out of date. UML still suggests an API... but would it seem less final that way?

docs/design/rosbag2_playback_time.md
emersonknapp: "No Advanced Estimating Clock By Default" is the name of the section in the linked design doc http://design.ros2.org/articles/clock_and_time.html#ros-time. The rest is a direct quote from that document - what they are trying to say is that "we don't know what the time source will do in the future", as in whether it will pause, speed up, slow down, etc. I'll add some language here to make that more clear

docs/design/rosbag2_playback_time.md
Karsten1987: it's not a deal breaker, but I'd rather highlight the key concepts here (maybe even just pseudo code) and leave out stuff like `virtual ~PlayerClock()`

docs/design/rosbag2_playback_time.md
Karsten1987: fair enough.

docs/design/rosbag2_playback_time.md
emersonknapp: That's fair - I will re-work it a little bit to be less prescriptive

emersonknapp:
Thanks @david-prody and @Karsten1987 for the extra comments! I've done another clarifying and pseudocode-ing pass. I also added in the concept of some "factory constructors" to differentiate the external-source vs time-control configuration modes of the clock.

The revisions have gotten small enough that I'll start opening PRs against this design, and will plan to merge this design in 48 hours if there are no major concerns.

Karsten1987:
This came up during the migration of the GenericPublisher to rclcpp, but I believe starting from Foxy there's two timestamps exposed in the message info for each incoming message. It's a time-sent and a time-received stamp, which we should probably leverage within rosbag2. I don't know the actual state of this, saying if all the RMWs are actually really setting these two fields, but if so, we could think about using them.
That being said, we most likely might have to expose this as a flag to the player - whether to use sent or received stamps - but I wanted to bring this up because it might be related to your design.

c.f. https://github.com/ros2/rclcpp/issues/1604

emersonknapp:
Thanks! That's useful information - we will probably want to use the "sent" timestamp for playback, to be more faithful to the original. That could be an update to this design - but it could also be an update to recording, if we choose only to store one of the timestamps, then the "sent" time might make more sense than the "received" tmie.

Karsten1987:
One more idea I wanted to vocalize here before I forget them. We can't certainly discuss this in more detail in the WG meeting.

What if I wanted to replay a bagfile in a sort of software-lock-step fashion, in which the system is synchronized and can be replayed deterministically. What this means to me is that we don't synchronize the system on a time source, but rather on a data driven source. Essentially what I mean here is that I'd like to have a way of replaying a message once at a time. Independent of time, simply sequentially. Imagine I have velocity commands recorded for a mobile robot and I want to replay each command given an external trigger - which is not time related but rather some other event from my system. Whenever that external signal comes, the next message will be replayed.

emersonknapp:
> I'd like to have a way of replaying a message once at a time

This sounds like the "play next message while paused" feature that @MichaelOrlov is working on implementing. We can extend this design to call it out specially. I think it can be implemented like follows:

```
return_code Player::play_next() {
  if (!clock->is_paused()) return failure;
  clock->jump(message_queue.front().time_stamp);
  return success;
}
```

If we keep the clock paused and jump ahead by message timestamps, the Player will be locked in with the trigger. Combined with topic filters, this would allow playing back message-by-message for subsets of topics, or just a single topic. 

Extending the implementation started in https://github.com/ros2/rosbag2/pull/704, I would expect it to look like

```
PlayerClock::sleep_until(ROSTime ros_until) {
  // Called by "Player::play_messages_until_queue_empty" from the Player thread
  Lock lock(mutex);
  // Called before the "jump", knows what time we need to sleep until
  SteadyTime steady_until = ros_to_steady(ros_until);
  // if this wakes up before "jump" is called, then the loop will just immediately call "sleep_until" again
  // if "jump" is called during this sleep, it notifies the variable so we wake up
  condition_variable.wait_until(lock, steady_until);
  // since the jump set us to ros_until explicitly, this returns true, so "play_messages_until_queue_empty" plays the next message
  return now() >= ros_until;
}


PlayerClock::jump(ROSTime time) {
  // called by "Player::play_next" from separate callback thread via Service, keyboard, GUI, whatever
  // update references
  snapshot(time);
  // wake up any current sleepers
  condition_variable.notify_all();
}
```

Based on the above, I think that with the current development approach this feature can be explained with a small design note, and implemented with a small PR.

Do you envision more cases that we will need to account for?

MichaelOrlov:
@emersonknapp I see one potential caveat. It looks like we are going to have "busy loop" when in pause. \
I judged based on the logic described in comment.
> // if this wakes up before "jump" is called, then the loop will just immediately call "sleep_until" again

It's going to bump processor core utilization close to 90-100% when in pause.

emersonknapp:
No - it won't, because this wakeup only happens when an event occurs. Normally, `sleep_until` will sleep for the amount of time between "now" and "until". It still sleeps this amount of time and then returns false, then is called again. It has non-busy wait built in to the loop.

However, I do see that it would loop at a speed relative to the amount of time until the next message, which isn't the best behavior. This is an implementation detail though, the design of this discussion doesn't suggest that this is a tight busy-loop, it's a controlled wakeup loop at a much slower rate.

MichaelOrlov:
We can actually check that we are in pause mode and call
```cpp
condition_variable.wait_for(lock, 5mSec)
```
in case if pause mode.

emersonknapp:
Yes, that would be a good start for the implementation.

PR #689, opened 2021-03-24
PlayerClock initial implementation - Player functionally unchanged

emersonknapp:
# Related Issues

* Design: #675
* Design extension: #700
* Part of #99
* Part of #696
* [X] Depends on #690 (fixes incorrect test)

# Description

* Adds new class `PlayerClock` with `now()` and `sleep_until()` with simple rate handling.
* Adds tests for class
* Removes time handling from `Player` in favor of configuring and using `PlayerClock`


rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Stores an exact time match between a system steady clock and the semantic play clock.
+ * This is created whenever a factor changes such that a new base reference is needed
+ * such as pause, resume, rate change, or jump
+ */
+struct TimeSync
+{
+  rosbag2_cpp::PlayerClock::PlayerTimePoint player_time;
+  rosbag2_cpp::PlayerClock::RealTimePoint real_time;
```
MichaelOrlov, line 33: After reviewing logic in this PR I come to conclusion that it will be much clear and easier to understand if these two variables would be named as:
```cpp
  rosbag2_cpp::PlayerClock::RecorderTimePoint recording_start_time;
  rosbag2_cpp::PlayerClock::RealTimePoint playback_start_time;
```
And also `PlayerTimePoint` would be named as `RecorderTimePoint` since it is time point in time when message was really recorder.

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(duration).count();
+}
+
+}  // namespace
+
+namespace rosbag2_cpp
+{
+
+class PlayerClockImpl
```
MichaelOrlov, line 47: @emersonknapp It looks like `PlayerClockImpl` class is redundant (at least in current implementation) since it is pure POD type and unlikely to be replaced. \
I think it would  be much simpler to bring all incapsulated variables directly to the `PlayerClock` class and create them on the stack since they are small in size.

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  // TODO(emersonknapp) - when we have methods that can change timeflow during a sleep,
+  // it will probably be better to use a condition_variable::wait_until
+  std::this_thread::sleep_until(real_until);
+  return now() >= until;
+}
+
+double PlayerClock::get_rate() const
+{
+  return impl_->rate;
```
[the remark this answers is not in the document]
emersonknapp: This PR leaves the functionality unchanged and therefore does not expose set_rate - that will be added in a following PR soon

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -103,7 +122,7 @@ void Player::play(const PlayOptions & options)
 
   wait_for_filled_queue(options);
 
-  play_messages_from_queue(options);
+  play_messages_from_queue();
 }
 
 void Player::wait_for_filled_queue(const PlayOptions & options) const
```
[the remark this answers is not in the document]
emersonknapp: This function does not need that information right now. The queue is filled elsewhere

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
emersonknapp: The "realtime" point is totally unrelated to the player or the recorder, this name change would be incorrect. "real" in this implementation refers to arbitrary system steady clock time, which is never revealed to users, it is only used internally to calculate offsets. 

These will not always be a "start" time. The plan is to create a new reference pair when a change occurs, such as set_rate or resume from pause. This allows us to avoid complexity of accumulating these offsets over time.

I disagree with Recorder naming - this is the "current time according to the player" which may be totally unrelated to the recording, if for example it's using simulated time from an external /clock publisher

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
emersonknapp: It will be expanded very soon in followup PRs - I am inclined to present only a pure interface in the headers and hide the implementation via Pimpl pattern, because I think we have exposed far too much in headers in the past, making backports impossible.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/56fa49315468fd07d734740da7975b4a/raw/b85678a7e3040cb2fef7dbccb3e6133c1431ee56/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2_tests ros2bag rosbag2_py
TEST args: --packages-select rosbag2_transport rosbag2_tests ros2bag rosbag2_py
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7996
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14092/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8910/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11785/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14187/)

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
emersonknapp: I renamed `RealTimePoint` to `SteadyTimePoint` and added some documenting comments - so that at least makes this more clear.

MichaelOrlov: @emersonknapp How about renaming `player_time` to the `recording_start_time`? \
 I've tried to think about it more in perspective for future development and it seems this variable always will keep `recording_start_time`. Even if we will advance time we will advance `real_time` and keep `player_time` with original "`recording_start_time`". \
I might missed some case which you might have in mind. Please correct me if you aware about such case or we can postpone this question till we will have some progress in feature development and it will be clear if it's going to changed or not. \
Or it could be alternative like `recording_base_time` and `playback_base_time`.

emersonknapp: This is the way I'm thinking about it - based on the following reasoning I believe we need two movable reference points. One reference point in ROS time (Player time / message time) and one reference point in Steady time.

In the following diagram you can see the flow of Steady Time (`S`), which always continues at the same rate, and the flow of ROS Time (`R`) which may change speed or pause, as two 1 dimensional timelines. The bag we are playing starts at `R0` and goes until `R3`. We start playback rate at 0.5, and we change the rate twice over the course of the playback.

time_ref

* At time `S1` we change the rate to `2.0`. At time `S2` we change the rate to `1.0`
* When we want to know `R PlayerClock::now()` - we know `S s_now = now_fn()`
* When we want to `void PlayerClock::sleep_until(R r_until)` 

To figure out the conversions between arbitrary points Rx<->Sx, Ry<->Sy, Rz<->Sz, I see the equations looking like the following

Screen Shot 2021-03-26 at 10 29 39 PM

Based on the above, it shows that in order to convert between the two time points we need both of the latest `Si` and `Ri` reference points - I'll call these `S_ref` and `R_ref`

Screen Shot 2021-03-26 at 10 29 58 PM

Whenever we change the rate (or pause, or jump) - we need to take a snapshot of the current `S_ref` and `R_ref` values to use in order to calculate the time conversion. The combination of these two values together contains the integration of all past rates, pauses, etc. I don't see how you could do this calculation accurately with only 1 datapoint, I think you need both. Please let me know if I'm missing something that lets us reduce these 2 reference numbers into a single reference number and a `const ros_starting_time`

emersonknapp: I've opened a PR adding a section to the design, with this information - https://github.com/ros2/rosbag2/pull/700

If you have a counterpoint to this logic, let's discuss it there. Until then, I plan to move forward with this PR so that we can start building the next features on top of it - if we revise the design it should be simple enough to modify the internals of the `PlayerClock` to match the updated design.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8950/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11831/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14244/)

jikawa-az (approved):
Thanks for the documentation, diagrams, and test cases!

rosbag2_cpp/src/rosbag2_cpp/player_clock.cpp
MichaelOrlov: @emersonknapp Thanks for the explanation. I will try to comprehend your approach and comment in #700 directly. \
And for sure feel free to move forward with this PR.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/6fb38f4384ee9db70d518f601711591a/raw/e3aa6275de0ca40c21d1a6860083e814c486933c/ros2.repos
BUILD args: --packages-up-to rosbag2_cpp rosbag2_transport rosbag2_py ros2bag rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_cpp rosbag2_transport rosbag2_py ros2bag rosbag2_tests rosbag2
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8062
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14179/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8976/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11861/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14280/)

david-prody (commented):
API looks good. my main question is how scenerio 3 from the design would be set up using this class.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -94,6 +93,26 @@ bool Player::is_storage_completely_loaded() const
...
+      rate = options.rate;
+    }
+    if (reader_->has_next()) {
+      auto message = reader_->read_next();
+      time_first_message = message->time_stamp;
+      // Could not peek, so need to enqueue this popped first message to be played.
+      message_queue_.enqueue(message);
+    } else {
+      // There are no messages in the storage at all
+      return;
```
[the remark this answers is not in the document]
emersonknapp: No, if the storage is empty here, it means that the bag on disk had no messages, playback will end immediately.

rosbag2_cpp/include/rosbag2_cpp/player_clock.hpp
```diff
@@ -0,0 +1,96 @@
...
+   * \param starting_time: provides an initial offset for managing time
+   *    This will likely be the timestamp of the first message in the bag
+   * \param rate: Rate of playback, a unit-free ratio. 1.0 is real-time
+   * \param now_fn: Function used to get the current steady time
+   *   defaults to std::chrono::steady_clock::now
+   *   Used to control for unit testing, or for specialized needs
+   * \throws std::runtime_error if rate is <= 0
+   */
+  ROSBAG2_CPP_PUBLIC
+  PlayerClock(
```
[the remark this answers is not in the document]
emersonknapp: I was thinking to add a separate static factory constructor for sim time, or have the sim time version be a subclass with a different constructor. This constructor will eventually take a 'clock_publish_frequency' argument to handle the first two cases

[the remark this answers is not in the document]
emersonknapp: I've gone ahead and clarified how this will be implemented, by making `PlayerClock` a pure virtual interface and providing this logic as `TimeControllerClock`.

rosbag2_transport/src/rosbag2_transport/player.cpp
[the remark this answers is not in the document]
emersonknapp: I've added a comment to that effect and added a `prepare_clock` function to separate interests a bit

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14182/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8979/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11864/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14284/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14215/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9015/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11908/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14318/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14216/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9016/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11908/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14318/)

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+bool TimeControllerClock::sleep_until(ROSTimePoint until)
+{
+  SteadyTimePoint steady_until;
+  {
+    std::lock_guard<std::mutex> lock(impl_->mutex);
+    steady_until = impl_->ros_to_steady(until);
+  }
+  // TODO(emersonknapp) - when we have methods that can change timeflow during a sleep,
+  // it will probably be better to use a condition_variable::wait_until
```
Karsten1987, line 98: why not addressing this TODO in this PR already? You can have a CV and wake it up with a small lambda function. If not, I'd suggest we open a ticket for this.

emersonknapp:
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/14324/)

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -0,0 +1,107 @@
...
+  std::this_thread::sleep_until(steady_until);
+  return now() >= until;
```
Karsten1987, line 100: what is the consequence for the Player if that sleep_until fails?

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include <gmock/gmock.h>
+
+#include "rosbag2_cpp/clocks/time_controller_clock.hpp"
+
+using namespace testing;  // NOLINT
+using SteadyTimePoint = rosbag2_cpp::PlayerClock::SteadyTimePoint;
+using ROSTimePoint = rosbag2_cpp::PlayerClock::ROSTimePoint;
+using NowFunction = rosbag2_cpp::PlayerClock::NowFunction;
+
```
Karsten1987, line 23:
```suggestion
```

rosbag2_storage/include/rosbag2_storage/serialized_bag_message.hpp
```diff
@@ -31,6 +31,8 @@ struct SerializedBagMessage
   std::string topic_name;
 };
 
+typedef std::shared_ptr<SerializedBagMessage> SerializedBagMessagePtr;
```
Karsten1987, line 34:
```suggestion
typedef std::shared_ptr<SerializedBagMessage> SerializedBagMessageSharedPtr;
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -94,6 +94,17 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play(const PlayOptions & options)
 {
+  if (reader_->has_next()) {
+    // Reader does not have "peek", so we must "pop" the first message to see its timestamp
+    auto message = reader_->read_next();
+    prepare_clock(options, message->time_stamp);
+    // Make sure that first message gets played by putting it into the play queue
+    message_queue_.enqueue(message);
```
Karsten1987, line 102: isn't this kind of contradicting with `load_storage_content`? Why are we explicitly load the first (and only the first) message in the queue?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -224,4 +214,13 @@ void Player::prepare_publishers(const PlayOptions & options)
   }
 }
 
+void Player::prepare_clock(const PlayOptions & options, rcutils_time_point_value_t starting_time)
+{
+  float rate = 1.0;
+  if (options.rate > 0.0) {
+    rate = options.rate;
+  }
```
Karsten1987, line 222: nitpick
```suggestion
  float rate = options.rate > 0.0 ? options.rate : 1.0;
```

Karsten1987 (commented):
Generally lgtm, but I am finding the typedefs a bit wild. I'd actually stick to their original types if we don't plan on evolving them further into their own structs.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: In order to know its timestamp - we were already doing this, I just moved it to a slightly different place.

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
emersonknapp: they sleep again - you may notice that Player runs `while (rclcpp::ok() && !sleep_until(message->timestamp)) {}`

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
emersonknapp: OK - i can address it here. I was interested in getting things moving, this is already large. I have the implementation in https://github.com/ros2/rosbag2/pull/704 - but I can pull it in here

rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * Virtual interface used to drive the timing of bag playback.
+ * This clock should be used to query times and sleep between message playing,
+ * so that the complexity involved around time control and time sources
+ * is encapsulated in this one place.
+ */
+class PlayerClock
+{
+public:
+  // Type representing the current time as according to the playback
+  typedef rcutils_time_point_value_t ROSTimePoint;
```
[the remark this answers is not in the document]
emersonknapp: I personally liked the shorter CamelCaseName than the longer C-style name. Perhaps I should use an `rclcpp::Time` instead?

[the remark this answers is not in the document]
Karsten1987: I think `rclcpp::Time` is the cleanest solution!

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
Karsten1987: that's fine if it's already addressed in a follow up PR. I just wanted to make sure we don't search for this `TODO` later on.

rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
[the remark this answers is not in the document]
emersonknapp: On second thought `rclcpp::Time` includes too much extra stuff, I prefer just passing the numerical value - I will change to `rcutils_time_point_value_t`

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: load_storage_content picks up at the next message and keeps loading - we've called this before starting that process so it has a strict ordering

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
Karsten1987: I may be off here, but consider this:

* dequeue message at time `t0+10`
* `while !sleep_until...`
* <external change of rate or playback, time now yields `t0+5`.

We should actually load & dequeue another message from the bag. Otherwise we'd sleep an additional 5 seconds skipping over existing messages in the bag until we hit `t0+10` again.

emersonknapp: That example doesn't parse as a possibility for me the way I am reading it, let me put it back to you the way I understood it:
* dequeue message, where `message->time_stamp == t0+10`
* `while !sleep_until...`
* External source calls `set_rate` - so we should be sleeping for less time than originally anticipated

What happens now is that when the external source calls `set_rate()`, the `condition_variable.notify_all` is called, so the player awakens immediately (sleep == `condition_variable.wait_until`), and will sleep again until new appropriate time

Or, are you talking about `jump` - which can move us to an arbitrary time, and therefore moves us back in time? This absolutely needs a "queue-busting" effect, as called out in the design documents via jump callbacks, which will also need to clear the most-recently-dequeued variable here. I would say that needs to be carefully considered during the implementation of `jump`.

emersonknapp: Note: `PlayerClock::now()` is monotonically increasing in all cases unless `jump` is called, therefore the "next messaage" always stays correct (unless jump is called) - and this call will have to trigger extra logic in the `Player` - it cannot operate in the `PlayerClock` alone. 

I believe `jump` will therefore also need new `Reader` APIs - or use the approach "start over from the beginning and dequeue messages up to the appropriate time", which is probably not the desired approach.

Karsten1987: my bad, I wasn't verbalizing my thought correctly. I was indeed referring to `jump` rather than `rate`. 
I'll take it from your comments that we currently don't consider a time jump and therefore don't cope with `now()` being potentially a time stamp in the past.

emersonknapp: > we currently don't consider a time jump and therefore don't cope with now() being potentially a time stamp in the past.

Yeah, that's correct. Handling that will definitely need some updates here, but I'm not sure precisely what they will look like yet

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -81,22 +83,18 @@ TimeControllerClock::TimeControllerClock(
...
-  {
-    std::lock_guard<std::mutex> lock(impl_->mutex);
-    steady_until = impl_->ros_to_steady(until);
-  }
-  // TODO(emersonknapp) - when we have methods that can change timeflow during a sleep,
-  // it will probably be better to use a condition_variable::wait_until
-  std::this_thread::sleep_until(steady_until);
+  std::unique_lock<std::mutex> lock(impl_->mutex);
+  steady_until = impl_->ros_to_steady(until);
+  impl_->cv.wait_until(lock, steady_until);
```
Karsten1987, line 97: 👍

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
emersonknapp: No - might as well. It's approximately the same amount of code and makes it TODO-free

Karsten1987 (approved):

emersonknapp:
@Karsten1987 thanks for the feedback - I think it put a nice finishing touch on this review. Now, fingers crossed for a green result :)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14225/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9024/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11916/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14329/)

emersonknapp:
Fixed float->double conversion warning

Windows: [Build Status](https://ci.ros2.org/job/ci_windows/14331/)

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -0,0 +1,108 @@
...
+  TimeControllerClockImpl() = default;
+  virtual ~TimeControllerClockImpl() = default;
+
+  template<typename T>
+  rcutils_duration_value_t duration_nanos(const T & duration)
+  {
+    return std::chrono::duration_cast<std::chrono::nanoseconds>(duration).count();
+  }
+
+  rcutils_time_point_value_t steady_to_ros(std::chrono::steady_clock::time_point steady_time)
```
MichaelOrlov, line 50: @emersonknapp It would be nice to add `REQUIRES(mutex) ` attribute to `steady_to_ros` function. \
https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#getting-started \
However I am not sure if it possible to do with `rclcpp` abstraction layer.

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -0,0 +1,108 @@
...
+  {
+    return reference.ros + static_cast<rcutils_duration_value_t>(
+      rate * duration_nanos(steady_time - reference.steady));
+  }
+
+  std::chrono::steady_clock::time_point ros_to_steady(rcutils_time_point_value_t ros_time)
+  {
```
MichaelOrlov, line 57: Similar. It would be nice to add `REQUIRES(mutex) ` attribute to `ros_to_steady` function.

PR #690, opened 2021-03-24
Fix and clarify logic in test_play filter test

emersonknapp:
The test was allowing the filter to be incorrect by not checking for the correct number of messages, while setting the filter in a way that was overridden by `play_options_`. This PR restructures the existing test to be more readable, clear in its intention, and correct in its final expectations.

This doesn't fix the API confusion - I want to do that but it will have to wait for the moment - in the meantime this change at least fixes the test. The `Reader.set_filter` call will always be overridden in `rosbag2_transport.play`, which uses `PlayOptions.topics_to_filter`.

Note to the reviewer - the _most important part_ of this change is `EXPECT_THAT(topics, SizeIs(0u))` on the filtered tests - it was `SizeIs(Ge(0u))` which was essentially not testing the filter

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
+++ b/rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
@@ -55,6 +55,7 @@ public:
 
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() override
   {
+    // filter_ was considered when incrementing num_read_ in has_next()
     return messages_[num_read_++];
   }
 
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -199,9 +199,7 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_filtered_topics)
 
   // Filter allows /topic2, blocks /topic1
   {
-    rosbag2_storage::StorageFilter storage_filter;
     play_options_.topics_to_filter = {"topic2"};
-    storage_filter.topics.push_back("topic2");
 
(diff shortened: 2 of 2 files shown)


jaisontj (approved):
Nit: We could do a `SizeIs(Ge(1u)` -> check for "Does atleast one message arrive"

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/532afc479a05ca8e823d55cd04b7ac9f/raw/324d2087b557a72928c09ff8c2166fbb3507c091/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2_tests
TEST args: --packages-select rosbag2_transport rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7995
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14091/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8909/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11784/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14185/)

emersonknapp:
Windows CI is currently broken in the infrastructure - see https://github.com/ros-infrastructure/ros2-cookbooks/pull/30 - I don't want to be blocked by that, we can fix any issues that arise later, if they do come up

clalancette:
> Windows CI is currently broken in the infrastructure - see [ros-infrastructure/ros2-cookbooks#30](https://github.com/ros-infrastructure/ros2-cookbooks/pull/30) - I don't want to be blocked by that, we can fix any issues that arise later, if they do come up

Just as a point of order; I might suggest not doing this.  If any platform is going to fail, it is going to be Windows.

PR #695, opened 2021-03-26
/clock publisher in Player

### LINKED ISSUE #99: play - publish /clock topic

emersonknapp:
Closes #99
Design: #675

Add a `rosgraph_msgs/Clock` publisher to the `Player` - that uses `PlayerClock::now()` to publish current time.

Expose this publish frequency to the CLI

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -190,6 +190,7 @@ PYBIND11_MODULE(_transport, m) {
     &PlayOptions::setTopicQoSProfileOverrides)
   .def_readwrite("loop", &PlayOptions::loop)
   .def_readwrite("topic_remapping_options", &PlayOptions::topic_remapping_options)
+  .def_readwrite("clock_publish_frequency", &PlayOptions::clock_publish_frequency)
   ;
 
   py::class_<RecordOptions>(m, "RecordOptions")
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -43,7 +43,7 @@ public:
 
   // Rate in Hz at which to publish to /clock.
   // 0 (or negative) means that no publisher will be created
-  double clock_publish_frequency = 0.f;
+  double clock_publish_frequency = 0.0;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -43,7 +43,7 @@ public:
 
   // Rate in Hz at which to publish to /clock.
   // 0 (or negative) means that no publisher will be created
-  float clock_publish_frequency = 0.f;
+  double clock_publish_frequency = 0.f;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -40,6 +40,10 @@ public:
   std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {};
   bool loop = false;
   std::vector<std::string> topic_remapping_options = {};
+
+  // Rate in Hz at which to publish to /clock.
+  // 0 (or negative) means that no publisher will be created
+  float clock_publish_frequency = 0.f;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -94,6 +94,12 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play(const PlayOptions & options)
 {
+  rclcpp::executors::SingleThreadedExecutor exec;
+  exec.add_node(transport_node_);
+  auto node_spin_thread = std::thread([&exec]() {
+    exec.spin();
+  });
+
   if (reader_->has_next()) {
     // Reader does not have "peek", so we must "pop" the first message to see its timestamp
     auto message = reader_->read_next();
@@ -222,6 +228,22 @@ void Player::prepare_clock(const PlayOptions & options, rcutils_time_point_value
 {
   double rate = options.rate > 0.0 ? options.rate : 1.0;
   clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(starting_time, rate);
+
+  // Create /clock publisher
+  if (options.clock_publish_frequency > 0.f) {
+    const auto publish_period = std::chrono::nanoseconds(
+      static_cast<uint64_t>(RCUTILS_S_TO_NS(1) / options.clock_publish_frequency));
+    // NOTE: PlayerClock does not own this publisher because rosbag2_cpp
+    // should not own transport-based functionality
+    clock_publisher_ = transport_node_->create_publisher<rosgraph_msgs::msg::Clock>(
+      "/clock", rclcpp::ClockQoS());
+    clock_publish_timer_ = transport_node_->create_wall_timer(
+      publish_period, [this]() {
+        auto msg = rosgraph_msgs::msg::Clock();
+        msg.clock = rclcpp::Time(clock_->now());
+        clock_publisher_->publish(msg);
+      });
+  }
 }
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -94,13 +94,6 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play(const PlayOptions & options)
 {
-  rclcpp::executors::SingleThreadedExecutor exec;
-  exec.add_node(transport_node_);
-  auto spin_thread = std::thread(
-    [&exec]() {
-      exec.spin();
-    });
-
   if (reader_->has_next()) {
     // Reader does not have "peek", so we must "pop" the first message to see its timestamp
     auto message = reader_->read_next();
@@ -122,8 +115,6 @@ void Player::play(const PlayOptions & options)
   wait_for_filled_queue(options);
 
   play_messages_from_queue();
-  exec.cancel();
-  spin_thread.join();
 }
 
 void Player::wait_for_filled_queue(const PlayOptions & options) const
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -96,9 +96,10 @@ void Player::play(const PlayOptions & options)
 {
   rclcpp::executors::SingleThreadedExecutor exec;
   exec.add_node(transport_node_);
-  auto spin_thread = std::thread([&exec]() {
-    exec.spin();
-  });
+  auto spin_thread = std::thread(
+    [&exec]() {
+      exec.spin();
+    });
 
   if (reader_->has_next()) {
     // Reader does not have "peek", so we must "pop" the first message to see its timestamp
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -96,7 +96,7 @@ void Player::play(const PlayOptions & options)
 {
   rclcpp::executors::SingleThreadedExecutor exec;
   exec.add_node(transport_node_);
-  auto node_spin_thread = std::thread([&exec]() {
+  auto spin_thread = std::thread([&exec]() {
     exec.spin();
   });
 
@@ -121,6 +121,8 @@ void Player::play(const PlayOptions & options)
   wait_for_filled_queue(options);
 
   play_messages_from_queue();
+  exec.cancel();
+  spin_thread.join();
 }
 
 void Player::wait_for_filled_queue(const PlayOptions & options) const
--- a/rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
@@ -1,0 +1,66 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include "rosbag2_transport/rosbag2_transport.hpp"
+#include "rosgraph_msgs/msg/clock.hpp"
+#include "test_msgs/message_fixtures.hpp"
+
+#include "rosbag2_play_test_fixture.hpp"
+
+using namespace ::testing;  // NOLINT
+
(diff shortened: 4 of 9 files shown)


emersonknapp:
I'm thinking about how to effectively test, but the feature works and is ready for review

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -40,6 +40,10 @@ struct PlayOptions
   std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {};
   bool loop = false;
   std::vector<std::string> topic_remapping_options = {};
+
+  // Rate in Hz at which to publish to /clock.
+  // 0 (or negative) means that no publisher will be created
+  float clock_publish_frequency = 0.f;
```
Karsten1987, line 46: just throwing this out here, but do you envision we could profit from `double` precision here?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -93,6 +93,12 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play(const PlayOptions & options)
 {
+  rclcpp::executors::SingleThreadedExecutor exec;
+  exec.add_node(transport_node_);
+  auto node_spin_thread = std::thread([&exec]() {
+    exec.spin();
+  });
```
Karsten1987, line 100: I will address this in #716 but I am not convinced that's the right approach. I feel we have to separate between the player as a standalone application in which we take care of node execution and a player as an API in which the node execution will be taken care of externally.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -217,6 +223,22 @@ void Player::prepare_clock(const PlayOptions & options, rcutils_time_point_value
 {
   double rate = options.rate > 0.0 ? options.rate : 1.0;
   clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(starting_time, rate);
```
Karsten1987, line 225: I know not related to the change, but why is the clock heap allocated?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -217,6 +223,22 @@ void Player::prepare_clock(const PlayOptions & options, rcutils_time_point_value
...
+
+  // Create /clock publisher
+  if (options.clock_publish_frequency > 0.f) {
+    const auto publish_period = std::chrono::nanoseconds(
+      static_cast<uint64_t>(RCUTILS_S_TO_NS(1) / options.clock_publish_frequency));
+    // NOTE: PlayerClock does not own this publisher because rosbag2_cpp
+    // should not own transport-based functionality
+    clock_publisher_ = transport_node_->create_publisher<rosgraph_msgs::msg::Clock>(
+      "/clock", rclcpp::ClockQoS());
```
Karsten1987, line 234: will there every be a use case where we want to namespace the clock topic?

Karsten1987 (commented):
this generally looks good to me. As mentioned in the comments, I don't think we can so easily just fire up an executor in the play function, but we can address this in a follow up PR.

Testing is indeed a bit tricky, but maybe a end-to-end test in which you test that you get at least a non-zero amount of `/clock` callbacks?

Karsten1987:
@emersonknapp can we officially put this up for review and run CI on it? I've also noticed a few `ROSBAG2_TRANSPORT_LOG...` macros on your branch. Does this branch have to be rebased?

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: I was thinking about deferred initialization - we don't have enough information to create the clock in the constructor, as it is right now. I could just overwrite a stack-allocated member here, alternatively. Thoughts?

emersonknapp:
>  Does this branch have to be rebased?

It looks like it!

I was just waiting to have a reasonable test - which it looks like I have now. I'll push it in a couple minutes.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: I'm not sure. For now I don't see the use case, but i think we could easily add this configurability later.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: Yeah - I wasn't really sure what to do here because currently there is no spin on the `transport_node_`, so my publisher timer doesn't run. If you're ok with moving forward like this, then re-doing it in #716 that would work for me.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/8785419a3ef05c024b54657b89f76d54/raw/d318e3e26f06cb8534b480ec3dffcf7dcb64e942/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_py rosbag2_transport rosbag2_tests rosbag2
TEST args: --packages-select ros2bag rosbag2_py rosbag2_transport rosbag2_tests rosbag2
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8143
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14276/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9065/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11960/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14376/)

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -40,6 +40,10 @@ struct PlayOptions
   std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {};
   bool loop = false;
   std::vector<std::string> topic_remapping_options = {};
+
+  // Rate in Hz at which to publish to /clock.
+  // 0 (or negative) means that no publisher will be created
+  double clock_publish_frequency = 0.f;
```
Karsten1987, line 46:
```suggestion
  double clock_publish_frequency = 0.0;
```

rosbag2_transport/src/rosbag2_transport/player.cpp
Karsten1987: I might address this in #716

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -0,0 +1,70 @@
...
+    "/clock", expected_clock_messages, rclcpp::ClockQoS());
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  // Play
+  rosbag2_transport::Rosbag2Transport rosbag2_transport(reader_, writer_);
+  rosbag2_transport.play(storage_options_, play_options_);
+  await_received_messages.get();
+
+  auto received_clock = sub_->get_received_messages<rosgraph_msgs::msg::Clock>("/clock");
+  EXPECT_THAT(received_clock, SizeIs(Ge(expected_clock_messages)));
```
Karsten1987, line 69: I'd kinda bet that sooner or later this will be flaky and the error tolerance has to be updated. What about instead asserting that we have more than 1 message and verify that the time interval between these messages indeed corresponds to the expected interval?

Karsten1987 (approved):
lgtm, I just had one remark towards the test. I am ok to move on with the rest.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14279/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9068/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11963/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14379/)

emersonknapp:
Not checking received times (weirdly slow on Windows) - instead checking the deltas between the contents of the clock messages, which should be more reliable. Also checking on playback rate being reflected in these values.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14280/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9069/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11964/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14380/)

emersonknapp:
Not sure exactly why this test works so poorly on OSX/Win - trying a OSX build locally to investigate

emersonknapp:
Quick directed test
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/14393/)
* OSX: [Build Status](https://ci.ros2.org/job/ci_osx/11975/)

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -222,6 +231,22 @@ void Player::prepare_clock(const PlayOptions & options, rcutils_time_point_value
 {
   double rate = options.rate > 0.0 ? options.rate : 1.0;
   clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(starting_time, rate);
+
+  // Create /clock publisher
+  if (options.clock_publish_frequency > 0.f) {
```
MichaelOrlov, line 236: It looks like that  `Player::prepare_publishers(const PlayOptions & options)` would be a more appropriate place for creating publisher for the `/clock` topic. \
It least I would expect to search for it there if I would try elaborate with source code.

MichaelOrlov (changes requested):

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,6 +190,7 @@ PYBIND11_MODULE(_transport, m) {
     &PlayOptions::setTopicQoSProfileOverrides)
   .def_readwrite("loop", &PlayOptions::loop)
   .def_readwrite("topic_remapping_options", &PlayOptions::topic_remapping_options)
+  .def_readwrite("clock_publish_frequency", &PlayOptions::clock_publish_frequency)
```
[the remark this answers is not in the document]
emersonknapp: The reason I prefer frequency is that we always deal in Hz - the domains we deal with do not require kHz, MHz, GHz, as the unit. Whereas when you specify a period, you _always_ have to very carefully specify the unit. I think it's harder to use in general because you have to read the documentation to understand what units you are working in.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -40,6 +40,10 @@ struct PlayOptions
   std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides = {};
   bool loop = false;
   std::vector<std::string> topic_remapping_options = {};
+
+  // Rate in Hz at which to publish to /clock.
+  // 0 (or negative) means that no publisher will be created
+  double clock_publish_frequency = 0.0;
```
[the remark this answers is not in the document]
emersonknapp: This wouldn't be appropriate for `PlayOptions` - because the default should be "do not publish /clock". This consideration should be at the CLI argument parsing level. Maybe something like:

```
parser.add_argument('--clock', nargs='?', const=40, default=0)
```

With the above, then just `--clock` would mean 40Hz. But if it is not provided, then 0Hz. Otherwise, if an argument is provided, then use that.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: I disagree - this is a pretty different feature. It doesn't help that there aren't docstrings on the functions, but `prepare_publishers` is, in my mind, explicitly for the playback publishers, which have metadata from the storage.

This is probably a matter of taste, I think I'm going to not take this note.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -115,6 +122,8 @@ void Player::play(const PlayOptions & options)
   wait_for_filled_queue(options);
 
   play_messages_from_queue();
+  exec.cancel();
+  spin_thread.join();
```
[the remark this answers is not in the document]
emersonknapp: That would be a `finally`, rather than duplicating code. And it's probably better to use a "Scope Exit Object", rather than adding a `try`-block

rosbag2_py/src/rosbag2_py/_transport.cpp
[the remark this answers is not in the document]
emersonknapp: I'll also add that - for me - thinking about frequency is easier. I want to say "publish this message 20 times per second. I don't want to have to do mental math and say ... ok that's how many milliseconds/nanoseconds/microsedons? is that 50? maybe, i'll pull up the calculator". Absent a psychological study proving that one is easier than the other for people across different native language groups to use, I'm going to skip this comment, though I do appreciate you thinking carefully about this. After all, it's `ros2 topic hz`, not `ros2 topic period`.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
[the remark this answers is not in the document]
emersonknapp: I've update the CLI to work this way

rosbag2_transport/src/rosbag2_transport/player.cpp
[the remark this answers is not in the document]
emersonknapp: I've resolved this by moving the spinner out to `Rosbag2Transport` wrapper layer, which owns the node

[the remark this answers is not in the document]
Karsten1987: as noted previously, #716 will have to address this again. More details once the PR is in a presentable state, but essentially I opted to make the `Player` and `Recorder` inherit from `rclcpp::Node`. That way we can easily take care of things like `NodeOptions`, their initialization and eventually also load them as components.

That has the consequence of necessarily spinning outside of the class.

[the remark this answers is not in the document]
emersonknapp: I like the `Player : public rclcpp::Node` concept, that is looking good. In that case, `spin` will be outside of `Player`, so it makes sense for me to do that here, too. It feels a little cleaner, anyways.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14294/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9083/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11980/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14398/)

rosbag2_transport/include/rosbag2_transport/play_options.hpp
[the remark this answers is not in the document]
MichaelOrlov: Thanks.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: @emersonknapp Can we come to some compromise solution? \
I would suggest to move code related to the creation of the publisher for the `/clock` topic to the separate function. \
May be name it something like `prepare_clock_publisher(clock_publish_frequency)`.

It's kind of separate operation and doesn't really directly relates to the `Player::prepare_clock(..)` to be inside of it.

emersonknapp: Since this is a matter of taste, I am going to move forward as-is - feel free to open a follow-up PR afterwards if it feels important.

Issue #696, opened 2021-03-26
[Epic] Rosbag2 playback time control and simulation time


emersonknapp:
## Description

Tracking issue for the group of features (planned for Galactic) around control of the flow of time during rosbag2 playback. See "Completion Criteria" for full feature list

## Design

* [x]  Initial Design: https://github.com/ros2/rosbag2/pull/675
* [x] Extension with time conversion logic #700 
* [x] Extension with `play_next` https://github.com/ros2/rosbag2/pull/708

## Feature List / Completion Criteria

Design at https://github.com/ros2/rosbag2/blob/master/docs/design/rosbag2_playback_time.md is fully implemented including all of the following features (names tagged where items are known to be work-in-progress):

* [x] `PlayerClock` integration into `Player` (@emersonknapp)
    * [X] (bugfix/unblock) Fix incorrect filter test: #690
    * [X] Initial `PlayerClock` implementation: #689
    * [x] Thread safety annotation followup #710
* [x] pause/resume time during playback #55 (@emersonknapp)
    * [x] impl: #704 
    * [x] fix behavior https://github.com/ros2/rosbag2/pull/730
    * [x] add error checking https://github.com/ros2/rosbag2/pull/731
* [x] change rate during playback (@emersonknapp )
    * [x] impl: #727
* [x] jump/seek to arbitrary time during playback (@MichaelOrlov) #821
    * [x] naive clock jump implementation for simpler setup https://github.com/ros2/rosbag2/pull/754 (@emersonknapp)
    * [x] jump callbacks #775 (@MichaelOrlov)
    * [x] https://github.com/ros2/rosbag2/issues/821
    * [x] storage seek API #824 (@lihui815)
        * [x] #836 
        * [x] https://github.com/ros2/rosbag2/pull/873
* [x] "Play next message" while paused (@MichaelOrlov) https://github.com/ros2/rosbag2/pull/762
* [x] pause/rate/play-next controllable by keyboard handler 
    * [x] #55 (@MichaelOrlov)
    * [x] Design document for unified keyboard handler https://github.com/ros2/rosbag2/pull/768
    * [x] Unified keyboard handler https://github.com/ros2/rosbag2/pull/735 (@MichaelOrlov)
    * [x] expose Rate to keyboard
* [x] pause/rate/play-next/jump controllable by external caller via Service (@emersonknapp )
    * [x] Add service definitions https://github.com/ros2/rosbag2/pull/728 
[...]

PR #704, opened 2021-03-31
Pause/resume PlayerClock

emersonknapp:
Part of #696

Add `set_paused` and `get_paused` functions to `PlayerClock` and expose to `Player`.

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
@@ -55,8 +55,10 @@ public:
     std::chrono::steady_clock::time_point steady;
   };
 
-  explicit TimeControllerClockImpl(PlayerClock::NowFunction now_fn)
-  : now_fn(now_fn)
+  explicit TimeControllerClockImpl(
+    PlayerClock::NowFunction now_fn, std::chrono::milliseconds sleep_time_while_paused)
+  : now_fn(now_fn),
+    sleep_time_while_paused(sleep_time_while_paused)
   {}
   virtual ~TimeControllerClockImpl() = default;
 
@@ -81,19 +83,29 @@ public:
     return reference.steady + std::chrono::nanoseconds(diff_nanos);
   }
 
+  void snapshot(rcutils_time_point_value_t ros_time)
+  RCPPUTILS_TSA_REQUIRES(state_mutex)
+  {
+    reference.ros = ros_time;
+    reference.steady = now_fn();
+  }
+
   const PlayerClock::NowFunction now_fn;
+  const std::chrono::milliseconds sleep_time_while_paused;
 
   std::mutex state_mutex;
   std::condition_variable cv RCPPUTILS_TSA_GUARDED_BY(state_mutex);
   double rate RCPPUTILS_TSA_GUARDED_BY(state_mutex) = 1.0;
+  bool paused RCPPUTILS_TSA_GUARDED_BY(state_mutex) = false;
   TimeReference reference RCPPUTILS_TSA_GUARDED_BY(state_mutex);
 };
 
 TimeControllerClock::TimeControllerClock(
   rcutils_time_point_value_t starting_time,
   double rate,
-  NowFunction now_fn)
-: impl_(std::make_unique<TimeControllerClockImpl>(now_fn))
+  NowFunction now_fn,
+  std::chrono::milliseconds sleep_time_while_paused)
+: impl_(std::make_unique<TimeControllerClockImpl>(now_fn, sleep_time_while_paused))
 {
   std::lock_guard<std::mutex> lock(impl_->state_mutex);
   impl_->reference.ros = starting_time;
@@ -114,8 +126,12 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
 {
   {
     TSAUniqueLock lock(impl_->state_mutex);
-    const auto steady_until = impl_->ros_to_steady(until);
-    impl_->cv.wait_until(lock, steady_until);
+    if (impl_->paused) {
+      impl_->cv.wait_for(lock, impl_->sleep_time_while_paused);
+    } else {
+      const auto steady_until = impl_->ros_to_steady(until);
+      impl_->cv.wait_until(lock, steady_until);
+    }
   }
   return now() >= until;
 }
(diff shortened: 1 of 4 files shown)


rosbag2_cpp/include/rosbag2_cpp/player_clock.hpp
```diff
@@ -87,6 +87,20 @@ class PlayerClock
   ROSBAG2_CPP_PUBLIC
   double get_rate() const;
 
+  /**
+   * Set the paused state of the clock.
+   * If currently in a `sleep_until`, this will wake the sleep to return false,
+   * so that the caller is not hanging forever.
+   */
+  ROSBAG2_CPP_PUBLIC
+  void set_paused(bool paused);
```
Karsten1987, line 96: sorry for the premature comment here, but `set_paused` reads kinda funky to me. 
What about simply having `void pause()` and `void resume()`? That sounds a bit more natural to me.

Similar, `get_paused` could be renamed to `is_paused()`.

emersonknapp: I'm fine with that API - thanks for the early feedback

MichaelOrlov:
@emersonknapp Could you please rebase your branch on top of master? \
I have dependent work in #718 and it's going to be some restructure after rebase. \
It would be better to rebase before writing unit tests and polishing `jump` implementation.

emersonknapp:
> @emersonknapp Could you please rebase your branch on top of master?

I read your mind and did it 2 minutes before you asked :)

MichaelOrlov:
@emersonknapp Looks good to me.

emersonknapp:
>  I will need to be able specify time point for snaphsot aka snapshot(time_point); when implementing jump method.

I've added a new `snapshot(ros_time)` signature for this purpose.

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -55,45 +55,72 @@ class TimeControllerClockImpl
...
     return reference.steady + std::chrono::nanoseconds(diff_nanos);
   }
 
+  rcutils_time_point_value_t ros_now() const
+  RCPPUTILS_TSA_REQUIRES(state_mutex)
+  {
+    if (paused) {
+      return reference.ros;
+    }
+    return steady_to_ros(now_fn());
```
zmichaels11, line 92: Can `now_fn` ever be null?

emersonknapp: No - it's `const std::function` by value - assigned in the constructor.

zmichaels11 (approved):

jikawa-az (approved):

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
zmichaels11: I looked into this for the case if null is passed in the constructor.
It looks like `std::function` cannot be assigned to null. It can be assigned to `nullptr`, where it then becomes "empty".
Calling an empty std::function will throw std::bad_function_call.

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
```diff
@@ -112,3 +115,51 @@ TEST_F(TimeControllerClockTest, slow_rate)
   return_time = some_time;
   EXPECT_EQ(pclock.now(), as_nanos(some_time) * playback_rate);
 }
+
+TEST_F(TimeControllerClockTest, basic_pause_resume)
```
prajakta-gokhale, line 119: I'm wondering if it's better to break this into multiple smaller tests. Generally with this pattern

```cpp
// Comment
...code...

// Comment
...code...

// Comment
...code...
```

it may be better written as separate tests that can run independently.

prajakta-gokhale (approved):

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
emersonknapp: OK - good idea

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/93a1b355caa64dbd15867e203c48dcba/raw/e448dfbf5b37d9fabc6532e4c69b5961c7741736/ros2.repos
BUILD args: --packages-up-to rosbag2 rosbag2_tests rosbag2_cpp rosbag2_transport rosbag2_py
TEST args: --packages-select rosbag2 rosbag2_tests rosbag2_cpp rosbag2_transport rosbag2_py
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8205
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14347/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9128/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12027/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14446/)

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
emersonknapp: Thanks for looking into it! Do you think it would be wise to add an `assert(now_fn != nullptr)` of some sort in the `TimeControllerClockImpl` constructor?

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -126,4 +157,34 @@ double TimeControllerClock::get_rate() const
...
+}
+
+void TimeControllerClock::resume()
+{
+  std::lock_guard<std::mutex> lock(impl_->state_mutex);
+  if (!impl_->paused) {
+    return;
+  }
+  // Note: needs to not be paused when taking snapshot, otherwise it will use last ros ref
+  impl_->paused = false;
```
david-prody, line 179: I dont think this works correctly.   Lets say i have check this clock 1 a second.  '

with no pauses i would expect to get: `1,2,3,4,5,....,12,13` as results.

if i put start pausing at 4 and lift the pause at 10 (6 second pause). i would expect the outcome from the clock to look like this: `1,2,3,4,4,4,4,4,4,4,6,7,8`

pretty sure the output i'm going to get is `1,2,3,4,4,4,4,4,4,4,11,12`

Lets just assume the inital set up `ref.ros = 1` ,` ref.steady = 1`.  if i call now at T=1 I get 1 + 1-1) = 1.
if i call now() at T=4. i get 1 + 4-1 = 4.  still ok.  

Than i hit pause.
ros_time = 4
steady_time = 4

now at t=5 i get 4 because the pause flag now() return 4, and will until i unpause

at T=10 i call resume
ref.ros = 10 //(ref.ros + 10 - ref.steady)
ref.steady = 10.   ref.ros should be 4 not 10.

emersonknapp: Thanks for this! Opened https://github.com/ros2/rosbag2/pull/730 to fix

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
emersonknapp: Good call - opened https://github.com/ros2/rosbag2/pull/731 to handle this

PR #705, opened 2021-04-01
Fix bad_function_call by replacing rclcpp::spin_some with SingleThreadedExecutor

emersonknapp:
Extension of https://github.com/ros2/rosbag2/pull/576 - same fix applied in more places. I have been seeing occasional test flakiness due to this cause. See details from #576 following

> Every time we called spin_some, a new executor was constructed. This would happen potentially a few thousand times per test. Given that the global context was not shutdown during this time, it would accumulate on_shutdown_callbacks_, one from each executor that was created. When this list was very large we would occasionally get std::bad_function_call for one of the callbacks in the shutdown.

> With this change, the global Context has only 2 items in on_shutdown_callbacks_, and I am not able to reproduce bad_function_call any more.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_record.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record.cpp
@@ -151,9 +151,14 @@ TEST_F(RecordIntegrationTestFixture, receives_latched_messages)
   // Publish messages before starting recording
   test_msgs::msg::Strings msg;
   msg.string_value = "Hello";
-  for (size_t i = 0; i < num_latched_messages; i++) {
-    publisher_transient_local->publish(msg);
-    rclcpp::spin_some(publisher_node);
+  {
+    // Scope the executor so that the node is removed from it before `spin_and_wait_for`
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.add_node(publisher_node);
+    for (size_t i = 0; i < num_latched_messages; i++) {
+      publisher_transient_local->publish(msg);
+      exec.spin_some();
+    }
   }
   start_recording({false, false, {topic}, "rmw_format", 100ms});
   auto & writer = static_cast<MockSequentialWriter &>(writer_->get_implementation_handle());
--- a/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp
@@ -79,8 +79,9 @@ public:
   {
     return async(
       std::launch::async, [this]() {
-        while (continue_spinning(expected_topics_with_size_)) {
-          rclcpp::spin_some(subscriber_node_);
+        rclcpp::executors::SingleThreadedExecutor exec;
+        while (rclcpp::ok() && continue_spinning(expected_topics_with_size_)) {
+          exec.spin_node_some(subscriber_node_);
         }
       });
   }
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -446,6 +446,7 @@ public:
     } else {
       EXPECT_NE(result, std::future_status::timeout);
     }
+    // Have to rclcpp::shutdown here to make the spin_subscriptions async thread exit
     transport.shutdown();
   }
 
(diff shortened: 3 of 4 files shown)


rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -151,9 +151,11 @@ TEST_F(RecordIntegrationTestFixture, receives_latched_messages)
   // Publish messages before starting recording
   test_msgs::msg::Strings msg;
   msg.string_value = "Hello";
+  rclcpp::executors::SingleThreadedExecutor exec;
+  exec.add_node(publisher_node);
```
clalancette, line 155: This is still throwing an exception in tests; in particular, the tests show:

```
unknown file: Failure
C++ exception with description "Node has already been added to an executor." thrown in the test body.
```

I think that is because we are adding the publisher_node to an executor here, but then down below `rosbag2_test_common::spin_and_wait_for` is also creating it's own executor and adding the node.

I think you can probably fix this by doing `exec.remove_node()` right before `rosbag2_test_common::spin_and_wait_for`.

clalancette (changes requested):
Overall, this is a good thing to fix (it's one of the persistent failures in CI).

There's a problem in the code that I pointed out inline; once that is fixed, hopefully the tests will successfully pass.

Separately, this isn't a great way for the `rclcpp::spin_some` API to behave.  This would be hard for others to debug.  I have a running list of APIs that I think are "hard" to use; I'll add this to see what we can do to make it less dangerous.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -151,9 +151,14 @@ TEST_F(RecordIntegrationTestFixture, receives_latched_messages)
...
-    publisher_transient_local->publish(msg);
-    rclcpp::spin_some(publisher_node);
+  {
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.add_node(publisher_node);
+    for (size_t i = 0; i < num_latched_messages; i++) {
+      publisher_transient_local->publish(msg);
+      exec.spin_some();
+    }
+    exec.remove_node(publisher_node);
```
clalancette, line 161: Oh, since you put the executor in a scope, this is probably redundant now.  But it doesn't hurt anything, so I'm fine if you want to keep it.

clalancette (approved):
This improves things, certainly.  The tests are still failing with a timeout in `test_play__rmw_cyclonedds_cpp`; I'm not sure if that is known or new from this PR.  I'm going to approve anyway and let you figure out whether this is an improvement regardless of that failing test.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/ac0d6eb824f96d7d6dd8306fdace023b/raw/82faf3109ee0d7fdb8bbc453154b6b7ebd0a5be2/ros2.repos
BUILD args: --packages-up-to rosbag2_test_common rosbag2_transport rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_test_common rosbag2_transport rosbag2_tests rosbag2
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8074
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8988/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11876/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14293/)

emersonknapp:
> This improves things, certainly. The tests are still failing with a timeout in `test_play__rmw_cyclonedds_cpp`; I'm not sure if that is known or new from this PR.

It was new from this PR - it took a bit of debugging, but the `std::async` thread from `spin_subscriptions` had no actual exit condition, and so was hanging the caller indefinitely. The reason that it worked before was that the `Executor` constructor would throw an exception once rclcpp had been shutdown, thereby killing the thread. I think this was a totally accidental flow control - I've fixed it now by extending the loop check to `while (rclcpp::ok() && other_condition)`

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14196/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8994/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11885/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14299/)

PR #714, opened 2021-04-06
use rclcpp::Node for generic pub/sub

Karsten1987:
There were some functions for topic expansion and filtering on the `rosbag2node` which I moved into a new class `TopicFilter`.

PR #715, opened 2021-04-06
use rclcpp logging macros

Karsten1987:
sits on top of #714 but essentially replaces all `ROSBAG2_TRANSPORT_LOG_*` with their respective `RCLCPP_*` equivalent. We introduced the custom macros back at the original time of writing as there weren't any `_STREAM` macros available in rclcpp at that time.


emersonknapp:
In theory I like the idea of not having custom logging macros.

I'm not sure I'm sold on the pattern of "borrowing" the logger from the `transport_node_` in code that runs outside the node - maybe the `Player` / `Recorder` should have their own loggers? I'm not 100% sure on that, just thinking out loud.

It's unfortunate that the logging "feels" more verbose this way - total characters mostly fewer, but more lines because of the extra argument.

Karsten1987:
interesting. I would have actually argued pretty much in an opposite way.

I assume that with the Galactic freeze, we'll remove the `rosbag2_transport` class altogether. Yet, I don't really anticipate a real API change for the `Player`/`Recorder` classes, that is both still get a node passed in from an external application context. With this in mind, I'd very much only change and adjust the logger configuration for that node and not for two separate loggers. I am also not sure at that point if one can configure loggers which are not attached to a ROS node.

emersonknapp:
I think I just had to sit with it for a few more minutes - I am convinced by your reasoning about Player/Recorder. And you're right, I plan for us to remove `Rosbag2Transport` class altogether.

emersonknapp (approved):

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14263/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9054/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11946/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14362/)

Karsten1987:
I can reproduce the test failure locally, yet I really don't have a clue how this PR introduces a segfault in the rosbag2_cpp converter impl:
```
 ➭ ros2 bag record -a -f some_rmw
[INFO] [1617724218.388506984] [rosbag2_cpp]: No plugin found providing serialization format. Falling back to checking RMW implementations.
[INFO] [1617724218.388555046] [rosbag2_cpp]: No plugin found providing serialization format. Falling back to checking RMW implementations.
[ERROR] [1617724218.401780979] [rosbag2_cpp]: Could not initialize RMWImplementedConverter: No RMW implementation found supporting serialization format some_rmw
Segmentation fault: 11
```

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14266/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9057/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14365/)

PR #716, opened 2021-04-06
Remove rosbag2_transport class

Karsten1987:
Promote Player and Recorder to public API. rosbag2_py will then use these player/recorder classes directly.


emersonknapp:
Looking forward to this change - once it is in I plan to expose time control Services via the `Player` node.

Karsten1987:
@emersonknapp don't get your hopes up too early ;-) 

This is indeed quite a mess and involves a ton of file changes. That is, the test infrastructure is pretty badly impacted by this change and I don't really see a clean way of not going down this rabbit hole.

Karsten1987:
Also, I'd recommend to run the benchmark suite for this before merging.

emersonknapp:
> This is indeed quite a mess and involves a ton of file changes. That is, the test infrastructure is pretty badly impacted by this change and I don't really see a clean way of not going down this rabbit hole.

Maybe there is a slightly less broad change that could be made. Instead of moving the existing Player and Recorder up to public scope, perhaps we can split uses of `Rosbag2Transport` into 2 new public-scope interfaces that wrap the existing `Player` and `Recorder`? I am happy to whip up a draft of this approach today, to see if it's any more manageable

emersonknapp:
You know my style - I very much prefer incremental PRs :)

emersonknapp:
Let me know what you think of https://github.com/ros2/rosbag2/pull/721 as a first step. It's fairly small in terms of diff lines-of-code, and passes all the tests locally. It might be nice to start there and then build up.

emersonknapp:
I believe this is now complete - closing. Reopen if I was wrong

PR #718, opened 2021-04-06
(WIP) Jump in time implementation for PlayerClock class

MichaelOrlov:
Part of #696
Depends on #704

This is initial implementation for jump functionality in `PlayerClock`.

- Added jump(ROSTimePoint time_point) with calling registered callbacks before and after `jump`
-  Added `PlayerClock::JumpHandler::SharedPtr create_jump_handler(
    const JumpHandler::pre_callback_t & pre_callback,
    const JumpHandler::post_callback_t & post_callback,
    const rcl_jump_threshold_t & threshold)`
- Added `void add_jump_calbacks(PlayerClock::JumpHandler::SharedPtr handler)`
- Added `void remove_jump_callbacks(PlayerClock::JumpHandler::SharedPtr handler)`

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
@@ -20,6 +20,7 @@
 #include <memory>
 
 #include "rcutils/time.h"
+#include "rclcpp/clock.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
 namespace rosbag2_cpp
@@ -40,6 +41,8 @@ public:
    */
   typedef std::function<std::chrono::steady_clock::time_point()> NowFunction;
 
+  using JumpHandler = rclcpp::JumpHandler;
+
   ROSBAG2_CPP_PUBLIC
   virtual ~PlayerClock() = default;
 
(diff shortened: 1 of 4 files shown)


MichaelOrlov:
@emersonknapp I've made initial implementation for the `jump` functionality with registering callbacks. \
However I haven't added any unit tests for it yet. \
Could you please take a look on my implementation and comment if you have any concerns or suggestions?

emersonknapp:
Overall the approach to jump handlers looks fine - is it mostly copied from `rclcpp::Clock`?

One thing I have realized, is that:
`jump` should trigger the callbacks for arbitrary time jumps, and bust the queue in the Player. This probably needs new functionality in the Reader API to seek as well. However, "play_next while paused" shouldn't actually need all of that.

I know that "play_next" feature was a priority of yours, let me know if you have a less heavy approach you want to implement first - maybe `jump(callbacks=false)`, so that we don't have to incur all that overhead for just fast-forwarding to the next message timestamp.

MichaelOrlov:
@emersonknapp I wouldn't say that it's mostly copied from rclcpp, but rather have similar approach and design. \
I've looked on the implementation of such callbacks mechanism in `rclcpp` and it turns out it's over complicated. It's diving deep on `rcl` layer and most of the code written on `"C"`. I adopted some data types from `rclcpp` and made my own lightweight implementation which should better satisfy our needs and to be more clear for understanding.

WRT to the new reader API for arbitrary seek. Yes it would be nice to have such extension. Although I thought about that it would make sense to implement this feature for sequential reader first. Not sure yet how we will advance in time. May be just reading out messages and skipping them until we will found required timestamp. It could be just straight forward temporary workaround without adding new API to the reader.

And I agree that it make sense to implement this feature with multiple iterations and PRs by adding some essential functionality to unblock development for another features as "Play next in pause".

I think that current API and implementation should cover dependencies for "Play next in pause" feature.

emersonknapp:
> Not sure yet how we will advance in time. May be just reading out messages and skipping them until we will found required timestamp. It could be just straight forward temporary workaround without adding new API to the reader.

Yes, I think this would be a good first implementation. For "go backward in time" - maybe "reopen the bag at the beginning and read messages until the required timestamp". I have a feeling that  this will run into performance problems quickly - but at least it would get the feature in front of people to complain about the performance :)

> I think that current API and implementation should cover dependencies for "Play next in pause" feature.

OK! I just wanted to make sure you didn't feel like this feature was blocking you from getting to the one you really wanted to do.

MichaelOrlov:
Closing this PR as wouldn't merge since we've split implementation on https://github.com/ros2/rosbag2/pull/754 and https://github.com/ros2/rosbag2/pull/775

PR #722, opened 2021-04-08
Remove -Werror from CMakeLists.txt, enable it in Action CI

### LINKED ISSUE #720: Remove `-Werror` from CMakeLists.txt
While it is generally appreciated to compile without warnings, there are situations where we don't want to fail the compilation on the first warning, such as a buildfarm run or trying to (cross-)compile rosbag2 on different platforms and OS's.

One can still compile locally with `-Werror` enabled by passing in the respective cmake args to `colcon` and their-like.

emersonknapp:
Closes #720


clalancette (approved):
Looks good to me assuming CI is happy.

Karsten1987 (approved):
thanks @emersonknapp

PR #728, opened 2021-04-09
Add rosbag2_interfaces package with playback service definitions

emersonknapp:
These .srv definitions will be used to control playback on the Player. Recording controls not yet defined, so none added.

Part of #696
Design at https://github.com/ros2/rosbag2/blob/master/docs/design/rosbag2_playback_time.md


jikawa-az (approved):

rosbag2_interfaces/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+cmake_minimum_required(VERSION 3.8)
+project(rosbag2_interfaces)
```
Karsten1987, line 2: I'd actually prefer `rosbag2_srvs` as a project name.

rosbag2_interfaces/srv/IsPaused.srv
```diff
@@ -0,0 +1,2 @@
+---
```
Karsten1987, line 1: why not using http://docs.ros.org/en/api/std_srvs/html/srv/Trigger.html

rosbag2_interfaces/srv/Pause.srv
```diff
@@ -0,0 +1 @@
+---
```
Karsten1987, line 1: http://docs.ros.org/en/api/std_srvs/html/srv/Empty.html

Karsten1987 (commented):
I'd vote to rename the project to `rosbag2_srvs`. 
I don't mind too much about the actual services, but some of them can be replaced by `std_srvs`.

rosbag2_interfaces/srv/IsPaused.srv
emersonknapp: My thinking on ROS best-practices was that it's generally better to define a semantically-specific interface, even if a `std_msgs`/`std_srvs` has the same fields. I can't remember where I first got this idea, though

rosbag2_interfaces/CMakeLists.txt
emersonknapp: I was thinking about leaving us open to potentially defining messages, or actions. Would you prefer stylistically that (if they come to be needed) those would be separate packages?

rosbag2_interfaces/srv/IsPaused.srv
emersonknapp: Reasoning something like "`rosbag2_interfaces/SetRate` is self-documenting, but `generic_srvs/SetFloat` doesn't tell you anything about what the service is for".

And in this case, `Trigger` semantically means "make something happen" - which could be confusing considering it's really a "GetBool" operation.

rosbag2_interfaces/srv/Pause.srv
emersonknapp: These are more accurately `Trigger`s, in terms of what they do - I just couldn't think of a failure case at the time. After doing some integration - I'm thinking that if the player is constructed, but not yet started, that could return `success=false`

Karsten1987:
it's all good. you can go ahead and merge as is. I just wanted to bring it up.

emersonknapp:
> it's all good. you can go ahead and merge as is. I just wanted to bring it up.

I'm just presenting where my thinking came from - I don't have strong opinions here, I am definitely open to making changes if there are good counterpoints.

For the package naming - I was thinking that potentially `Play`, `Record`, or `Convert/Rewrite` could be controlled by an Action with progress feedback - I don't have any Messages in mind right now though.

rosbag2_interfaces/srv/IsPaused.srv
emersonknapp: I can see a counterpoint being - "if you use `std_srvs`, then you don't introduce a dependency on `rosbag2` in order to have a remote-controller". I think here the problem is that `std_srvs` is so limited, that you still need a custom service for e.g. `Seek` and `GetRate` - so given that the dependency exists anyways, might as well have custom definitions.

Karsten1987 (approved):
I agree with your reasoning here. Thanks for the explanations.

david-prody (approved):
This looks good

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/20e6f6611bb120321b7294df2f6974bb/raw/9560335eef310dd4ebe284927edc903bb23f43a2/ros2.repos
BUILD args: --packages-up-to rosbag2 rosbag2_interfaces
TEST args: --packages-select rosbag2 rosbag2_interfaces
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8207
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14349/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9130/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12029/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14448/)

PR #729, opened 2021-04-09
Expose pause/resume related services on the Player

emersonknapp:
Depends on #728 
Depends on #704 

Exposes "Pause", "Resume", "TogglePaused", and "IsPaused" services for the Player.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -333,6 +333,7 @@ void Player::prepare_publishers()
 
 void Player::prepare_clock(rcutils_time_point_value_t starting_time)
 {
+  // TODO(emersonknapp) move clock setup into the constructor
   double rate = play_options_.rate > 0.0 ? play_options_.rate : 1.0;
   clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(starting_time, rate);
 
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -96,39 +96,38 @@ Player::Player(
   storage_options_(storage_options),
   play_options_(play_options)
 {
-
   srv_pause_ = create_service<rosbag2_interfaces::srv::Pause>(
     "~/pause",
     [this](
-      const std::shared_ptr<rmw_request_id_t> /* request_header */,
-      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Request> /* request */,
-      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Response> /* response */)
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Response>/* response */)
     {
       pause();
     });
   srv_resume_ = create_service<rosbag2_interfaces::srv::Resume>(
     "~/resume",
     [this](
-      const std::shared_ptr<rmw_request_id_t> /* request_header */,
-      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Request> /* request */,
-      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Response> /* response */)
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Response>/* response */)
     {
       resume();
     });
   srv_toggle_paused_ = create_service<rosbag2_interfaces::srv::TogglePaused>(
     "~/toggle_paused",
     [this](
-      const std::shared_ptr<rmw_request_id_t> /* request_header */,
-      const std::shared_ptr<rosbag2_interfaces::srv::TogglePaused::Request> /* request */,
-      const std::shared_ptr<rosbag2_interfaces::srv::TogglePaused::Response> /* response */)
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::TogglePaused::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::TogglePaused::Response>/* response */)
     {
       toggle_paused();
     });
   srv_is_paused_ = create_service<rosbag2_interfaces::srv::IsPaused>(
     "~/is_paused",
     [this](
-      const std::shared_ptr<rmw_request_id_t> /* request_header */,
-      const std::shared_ptr<rosbag2_interfaces::srv::IsPaused::Request> /* request */,
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::IsPaused::Request>/* request */,
       const std::shared_ptr<rosbag2_interfaces::srv::IsPaused::Response> response)
     {
       response->paused = is_paused();
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -95,7 +95,45 @@ Player::Player(
   reader_(std::move(reader)),
   storage_options_(storage_options),
   play_options_(play_options)
-{}
+{
+
+  srv_pause_ = create_service<rosbag2_interfaces::srv::Pause>(
+    "~/pause",
+    [this](
+      const std::shared_ptr<rmw_request_id_t> /* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Request> /* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Pause::Response> /* response */)
+    {
+      pause();
+    });
+  srv_resume_ = create_service<rosbag2_interfaces::srv::Resume>(
+    "~/resume",
+    [this](
+      const std::shared_ptr<rmw_request_id_t> /* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Request> /* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Resume::Response> /* response */)
+    {
+      resume();
+    });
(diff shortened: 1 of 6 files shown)


emersonknapp:
I'm opening this up for review because this is the implementation - but we will definitely not merge without tests. 

I am manually testing with this simple Python app for keyboard control and it's working great! (need to `pip3 install getkey`). 
I'm wondering, Should we maybe make a `rosbag2_examples` package for example usage, like this key controller? We could also put usage examples of the `rosbag2_py` and `rosbag2_cpp` APIs there, to make it easier to understand by example how to use rosbag2 programmatically. This would probably be better than putting code examples in tutorial docs, because at least we can ensure these compile with updates.

```
#!/usr/bin/env python3
import threading

from getkey import getkey, keys

import rclpy
from rclpy.node import Node

from rosbag2_interfaces.srv import (
    Pause,
    Resume,
    TogglePaused,
    IsPaused,
)

class PlayController(Node):
    def __init__(self):
        super().__init__('playcontrol')
        self.pause_client = self.create_client(Pause, '/rosbag2_player/pause')
        self.resume_client = self.create_client(Resume, '/rosbag2_player/resume')
        self.toggle_client = self.create_client(TogglePaused, '/rosbag2_player/toggle_paused')
        self.is_paused_client = self.create_client(IsPaused, '/rosbag2_player/is_paused')
        self.timer = self.create_timer(1, self.state_timer)

    def print_state(self, future):
        res = future.result()
        if res is not None:
            self.get_logger().info(f'IsPaused: {res.paused}')
        else:
            self.get_logger().error(f'Exception while calling service: {future.exception()}')

    def trigger_result_cb(self, name):
        def impl(future):
            res = future.result()
            if res is not None:
                self.get_logger().info(f'Response received for: {name}')
            else:
                self.get_logger().error(f'Exception for service: {future.exception()}')

        return impl

    def state_timer(self):
        req = IsPaused.Request()
        future = self.is_paused_client.call_async(req)
        future.add_done_callback(self.print_state)

    def pause(self):
        req = Pause.Request()
        future = self.pause_client.call_async(req)
        future.add_done_callback(self.trigger_result_cb('Pause'))

    def resume(self):
        req = Resume.Request()
        future = self.resume_client.call_async(req)
        future.add_done_callback(self.trigger_result_cb('Resume'))

    def toggle_paused(self):
        req = TogglePaused.Request()
        future = self.toggle_client.call_async(req)
        future.add_done_callback(self.trigger_result_cb('Toggle'))


def spin_fn(node):
    node.get_logger().info('Starting spin thread')
    rclpy.spin(node)
    node.get_logger().info('Spin thread done')


def main():
    rclpy.init()
    node = PlayController()
    spin_thread = threading.Thread(target=spin_fn, args=[node])
    spin_thread.start()

    try:
        while True:
            c = getkey()
            if c == keys.SPACE:
                node.toggle_paused()
            elif c == keys.P:
                node.pause()
            elif c == keys.R:
                node.resume()
            else:
                print(f'Ignored: {c}')
    except KeyboardInterrupt:
        pass

    node.destroy_node()
    rclpy.shutdown()
    spin_thread.join()


if __name__ == "__main__":
    main()
```

emersonknapp:
@Karsten1987 this is ready to go, could you give it a look when you have a chance?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -157,6 +194,39 @@ void Player::play()
   }
 }
 
+void Player::pause()
+{
+  if (clock_) {
```
Karsten1987, line 199: How can `clock_` ever be `nullptr`? On that note, where is `clock_` currently initialized?

same below.

rosbag2_transport/test/rosbag2_transport/rosbag2_play_test_fixture.hpp
```diff
@@ -30,7 +30,7 @@ class RosBag2PlayTestFixture : public Rosbag2TransportTestFixture
     sub_ = std::make_shared<SubscriptionManager>();
   }
 
-  ~RosBag2PlayTestFixture() override
+  virtual ~RosBag2PlayTestFixture()
```
Karsten1987, line 33: why this change?

rosbag2_transport/test/rosbag2_transport/test_play_services.cpp
```diff
@@ -0,0 +1,236 @@
...
+      std::bind(&PlaySrvsTest::topic_callback, this, std::placeholders::_1));
+
+    exec_.add_node(player_);
+    exec_.add_node(client_node_);
+    spin_thread_ = std::thread(
+      [this]() {
+        exec_.spin();
+      });
+
+    // Make sure all expected services are present before starting any test
```
Karsten1987, line 78: Does it make sense to offload this into a `service_manager` similar to `publication` and `subscription` manager?
Not a show-stopper, just food for thought.

Karsten1987 (approved):
lgtm with green CI.

rosbag2_transport/test/rosbag2_transport/rosbag2_play_test_fixture.hpp
emersonknapp: Oh - I'll change it back, I wasn't thinking very hard about it. I see that declaring `override` _forces_ the parent class to also have a virtual destructor. I just wasn't used to seeing that pattern.

rosbag2_transport/test/rosbag2_transport/test_play_services.cpp
emersonknapp: I think that would make sense to me as a followup refactor if we want to use this pattern in other test fixtures. I like to avoid premature architecture.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: Clock is currently intitialized in `prepare_clock`, which call from `play`. I would like to initialize it in the constructor (and therefore avoid having it be on the heap) - but we don't have access to the "initial time" until the reader is opened, which only happens when we call `play`.

It may make sense, to me, to open the reader in the constructor, now that we have all arguments present at that point. I wouldn't want to make that change here, though. What do you think?

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/bd7f6cd9db15a825e6522b7e5f9efad4/raw/2cdbc90e9cf69cd9c850cfca808846fce74cc5bf/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2 rosbag2_py rosbag2_tests
TEST args: --packages-select rosbag2_transport rosbag2 rosbag2_py rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8288
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14443/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9214/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12114/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14546/)

rosbag2_transport/src/rosbag2_transport/player.cpp
Karsten1987: I am okay if you don't address it here, but I'd like to have it noted in an GH issue or an TODO to keep track of it. I think we should - now with its public header - review the overall API and offload some of the initialization work from `play` to the constructor.

emersonknapp: I agree with all of that

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14447/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9218/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14551/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14450/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9221/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12121/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14554/)

emersonknapp:
Just trying OSX for the mutex concern [Build Status](https://ci.ros2.org/job/ci_osx/12123/)

emersonknapp:
Happy OSX, logic fixed up. last (hopefully) run

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14455/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9225/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14557/)

rosbag2_transport/test/rosbag2_transport/test_play_services.cpp
```diff
@@ -0,0 +1,236 @@
...
+  : RosBag2PlayTestFixture(),
+    client_node_(std::make_shared<rclcpp::Node>("test_play_client"))
+  {}
+
+  ~PlaySrvsTest() override
+  {
+    exec_.cancel();
+    rclcpp::shutdown();
+    spin_thread_.join();
+    play_thread_.join();
```
MichaelOrlov, line 54: For the sake of defensive programming I would recommend to check if threads still running before calling join otherwise it will cause program termination.
i.e.
```cpp
if (spin_thread_.joinable()) {
  spin_thread_.join();
}
if (play_thread_.joinable()) {
  play_thread_.join();
}
```

emersonknapp: > before calling join otherwise it will cause program termination

That is not true according to  https://en.cppreference.com/w/cpp/thread/thread/join - which states that 

> Blocks the current thread until the thread identified by *this finishes its execution.

This is the intention - the threads will exit after `rclcpp::shutdown` - we want to wait until they do.

MichaelOrlov: I meant the case when thread have already finished for some unexpected reasons or haven't even had a chance to start. e.g. caught some exception before calling
```cpp
spin_thread_ = std::thread(
      [this]() {
        exec_.spin();
      });
```
But never mind, this is just in tests. It will not affect quality of the core rosbag2 code. \
Sorry for attracting attention about it, I missed that it's just in tests.

PR #741, opened 2021-04-20
use public recorder api in tests

Karsten1987:
sits on top of #740

PR #742, opened 2021-04-20
remove rosbag2_transport header

Karsten1987:
as per title, finally triggering the switch.

sits on top of #741


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14430/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9202/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12102/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14530/)

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -156,20 +171,28 @@ class Recorder
...
     }
 
-    rosbag2_transport::Recorder impl(writer);
-    impl.record(storage_options, record_options);
+    auto recorder = std::make_shared<rosbag2_transport::Recorder>(
+      std::move(writer), storage_options, record_options);
+    recorder->record();
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.add_node(recorder);
+    auto spin_thread = std::thread(
```
emersonknapp, line 189: Don't you want to start spinning before calling record? IIRC `record` doesn't exit until it's cancelled.

Karsten1987: hmm. I honestly don't know. I had the same question at #741 within the tests - this code is actually really just copied from these. 
I thought it would make sense to give `record` the change to setup all subscriptions and potential error handling and really only spin once that's completed. 

Why do you advocate for spinning first?

rosbag2_transport/include/rosbag2_transport/recorder.hpp
```diff
@@ -39,25 +40,27 @@ class Writer;
...
     const rosbag2_storage::StorageOptions & storage_options,
     const rosbag2_transport::RecordOptions & record_options,
     const std::string & node_name = "rosbag2_recorder",
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
 
+  ROSBAG2_TRANSPORT_PUBLIC
   virtual ~Recorder();
 
+  ROSBAG2_TRANSPORT_PUBLIC
   void record();
```
emersonknapp, line 64: Now that we are making these public I think we need a serious docstring pass. Not blocking this, but we should make an effort to document every public header in the "cleanup sprint"

rosbag2_py/src/rosbag2_py/_transport.cpp
emersonknapp: Oh - `Recorder::record` returns right away after setup and runs in the background, until the `Recorder` object is destroyed. API confusion from the previous blocking API. 

Not something we have to handle here, but maybe we ought to consider a slightly more obvious API - if recording is controlled by the object lifecycle, then should recording just start in the constructor? Or, alternatively should there be a way to stop recording without destroying the object?

emersonknapp (approved):

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14435/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9207/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12107/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14537/)

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -121,8 +123,19 @@ class Player
...
+    auto player = std::make_shared<rosbag2_transport::Player>(
+      std::move(reader), storage_options, play_options);
+
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.add_node(player);
+    auto spin_thread = std::thread(
+      [&exec]() {
+        exec.spin();
+      });
+    player->play();
```
emersonknapp, line 135: it seems that this does not handle `play_options.loop` anymore - that loop was in `rosbag2_transport.cpp` - or am I missing its new location?

emersonknapp (changes requested):

rosbag2_py/src/rosbag2_py/_transport.cpp
emersonknapp: Oh - my mistake, yes I remember now that you added it to `Player` in the previous PR

emersonknapp (approved):

Karsten1987:
@emerson, could it be that the GH actions don't observe the `xfail` label?
There seems to be a discrepancy between the CI buildfarm results and the actions.

As a side note, I believe there's simply a bug in the `record_end_to_end` test. We expect a positive exit code, yet the command is clearly invoked incorrectly. So the system actually behaves correctly, the test is wrong in my opinion.

emersonknapp:
> could it be that the GH actions don't observe the xfail label?

That's correct - it is not configured out of the box to do anything specific about `xfail` - that's something we could configure for this specific repo, or we could make it a default behavior. I'm more inclined to allow it to be configured on a per-repo basis - given that it's not "built-in" to the Jenkins jobs, it's just build args `--ctest-args -LE xfail --pytest-args -m "not xfail"`

emersonknapp:
https://github.com/ros2/rosbag2/pull/748

Issue #737, opened 2021-04-19
Allow control over which executor is used for recording and/or consider changing the default


adamdbrw:
## Description

rosbag2 uses SingleThreadedExecutor and runs spin() for recording. This is not great for some cases, notably when recording streams with high number of executables per second, since spin() consumes a lot of CPU in such case: https://github.com/ros2/rclcpp/issues/1637. This is especially the case with rosbag2 record, which:

- should not care that much about minimizing latency for callback execution (a few ms added is probably ok).
- certainly does not want to minimize latency at the cost of high CPU use, since it disturbs the recorded system running on the same machine (e. g. an automotive stack) 

Changing the spin() in recorder to spin_some() or spin_all() + a very small sleep strictly improves performance in all important metrics (for myself at least) (a drop to half the CPU use).

## Related Issues

https://github.com/ros2/rclcpp/issues/1637
https://github.com/ros2/rosbag2/issues/679

## Completion Criteria

API is extended to allow different executors (or different use of executors, some lambdas?) and/or default spinning is changed.

## Testing Notes / Suggestions

rosbag2 performance benchmarks can be used to run a desired setup of publishers easily.

emersonknapp:
I think it would be acceptable to just change the default behavior to the better-performing one. 

Adding a new configuration option adds complexity that I don't think we need yet.

adamdbrw:
@emersonknapp I proposed a minimal in LOC but impactful PR

MichaelOrlov:
- Closing as wouldn't be done since it was proved here https://github.com/ros2/rosbag2/issues/679#issuecomment-809032204 that Multithreaded executed with the current cache mechanism will not give a tangible gain in performance. 

- The root cause rather in the suboptimal implementation of the `SingleThreadedExecutor` which is going to be improved soon by not recreating waitset entity after each callback.
- Also performance issue in the `should_split_bagfile()` was addressed for the MCAP file format.

PR #743, opened 2021-04-20
Recording CPU load decrease through spin_some()

adamdbrw:
The rationale behind this PR is explained here: https://github.com/ros2/rosbag2/issues/737 and here:  https://github.com/ros2/rclcpp/issues/1637.

In short, it is cutting CPU load from `ros2 bag record` from 110% to 55% in a certain use-case on a certain platform without apparent drawbacks. Notably, the use-case is quite a representative automotive one and the platform is also unexceptional. I believe this extrapolates well into reducing CPU use overall when the number of executables (e. g. subscriptions) to process each second is high. There might be some additional testing needed to build even more confidence in the broader conclusion.

Update: I have no confidence in the generalized claim anymore, see comment below.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/rosbag2_transport.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/rosbag2_transport.cpp
@@ -98,7 +98,11 @@ void Recorder::record(
     exec.add_node(recorder);
     auto spin_thread = std::thread(
       [&exec]() {
-        exec.spin();
+        const auto time_step = std::chrono::milliseconds(1);
+        while (rclcpp::ok()) {
+          exec.spin_some();
+          std::this_thread::sleep_for(time_step);
+        }
       });
     auto exit = rcpputils::scope_exit(
       [&exec, &spin_thread]() {


emersonknapp:
I think I would like to understand why this different is so big, before merging it. Is the standard Executor used in `spin` underperforming in some way that we can easily fix instead? (That being said, since core is frozen and we are not, we could make this change in rosbag for Galactic with a TODO note to remove in favor of core change).

Would this difference show in a development machine using the rosbag2_performance benchmarks? The output of that suite would also help make this review easier.

adamdbrw:
Of course, let me explain why the difference is so big. I tried to do so in the rclcpp issue linked in the description, notably this part:

>With the case of ~4k executables per second, when calling wait_for_work, 3.3k calls to rcl_wait per each second are made, so it is mostly only one executable that is returned each time, which seems quite inefficient. I am not sure if this is by design (since it perhaps minimizes latency), certainly collections used to gather a bunch of executables for each rcl_wait call are underused.

>When a 1 ms sleep is introduced after we miss the cache (before/after rcl_wait), only ~600 calls to rcl_wait per second are made while successfully executing the same number of callbacks per second.

Basically, this change causes the executables "cache" to be actually used `(get_next_ready_executable_from_map` is used more often in place of rcl_wait), a significantly cheaper operation.

CPU measurements before and after, exactly the same setup:

before

after

The bag file played to get this comparison (play into record) contains some non-public messages and I agree that it would be great to get something that is easy to replicate. I will find some time to do this.

Bag file rough profile: 3.7GB in 452k messages over 100 seconds with about 200 topics.

This should be fully reproducable with benchmarking publishers.

adamdbrw:
Running benchmarks didn't confirm that this gain is more generic, and in some cases (lower executables/sec count) CPU load is apparently even increased by this patch. So I am not a fan of this change anymore. I am investigating further. What is so crucially different in these scenarios? These results don't make much sense to me yet.

emersonknapp:
Thanks for taking a deep dive into this, it is reassuring to see this level of rigor applied to an important issue.

I would be interested in a profiling result, to see exactly where in the code the 50% CPU usage difference comes from in your sample case - a function profile will probably tell us a lot.

MichaelOrlov:
@emersonknapp As far as I understand from @christophebedard [analysis](https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077/10?u=michaelorlov) main overhead in executor in `get_next_ready_executable` that it checks its lists of `timers/subscriptions/services/clients/waitables` and returns once it has found one that is ready to execute. 

Introducing sleep between `spin_some` just yielding CPU resources for another threads instead of dummy crawling  by `get_next_ready_executable` in it's huge lists. However this is not "deterministic"  approach and as @adamdbrw mentioned it could work beneficial for some scenarios and platforms and could make performance worse for another configurations.

I don't see an easy fix for this problem now.

While executors have a good abstraction and easy to use API they have a lot of overhead inside. \
At the moment I anticipate that only using `wait_sets` directly in recorder instead of executors could help with substantial performance improvements.

emersonknapp:
Perhaps using https://github.com/ros2/design/pull/305 the "Events Executor" instead of a `waitset`-based one would help - supposedly the performance is way better.

adamdbrw:
I will dive deeper into this

emersonknapp:
@adamdbrw is this PR still relevant, or should we close it as stale?

adamdbrw:
Closing as stale

PR #744, opened 2021-04-20
Include utility to quiet cpplint.

clalancette:

This should fix the nightlies: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1578/testReport/junit/rosbag2_py/cpplint/build_include_what_you_use__4____home_jenkins_agent_workspace_nightly_linux_aarch64_debug_ws_src_ros2_rosbag2_rosbag2_py_src_rosbag2_py__transport_cpp_123_/

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -16,6 +16,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 #include "rosbag2_compression/compression_options.hpp"


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14413/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9186/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12086/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14510/)

Karsten1987 (approved):
strange that we didn't catch this earlier. Thanks for the fix.

emersonknapp:
> strange that we didn't catch this earlier

It's not that strange - the linter versions aren't pinned and have broken out from underneath us numerous times before. Or can even have different behavior between development installs and buildfarm installs - since according to the official setup instructions, one environment might install from `apt` and one might install from `pip` - giving even different major  versions.

emersonknapp:
But also - maybe we weren't running it every time... https://github.com/ros2/rosbag2/pull/745

Rant stands, but that may not have been what it actually was this time ;P

clalancette:
> It's not that strange - the linter versions aren't pinned and have broken out from underneath us numerous times before. Or can even have different behavior between development installs and buildfarm installs - since according to the official setup instructions, one environment might install from `apt` and one might install from `pip` - giving even different major versions.

I understand your frustration with that, and share it to some degree.  But even if we were to make this consistent between the buildfarm and CI jobs on Linux, for instance, there would still inevitably be drift between it and macOS and Windows.  You could talk about pinning *all* of them to the lowest-common-denominator, but that has problems of its own.  Anyway, that is a bit off-topic here.

> Rant stands, but that wasn't what the problem actually was this time ;P. We must have accidentally let through a yellow CI run PR at some point in there

Cool, thanks for fixing that up!

PR #747, opened 2021-04-21
correct exit code assertion

### LINKED ISSUE #746: record end to end test has wrong expectancy for wrong rmw format.
I believe there's simply a bug in the `record_end_to_end` test. We expect a positive exit code, yet the command is clearly invoked incorrectly. So the system actually behaves correctly, the test is wrong in my opinion.

```
TEST_F(RecordFixture, record_fails_gracefully_if_plugin_for_given_encoding_does_not_exist) {
  internal::CaptureStderr();
  auto exit_code =
    execute_and_wait_until_completion("ros2 bag record -a -f some_rmw", temporary_dir_path_);
  auto error_output = internal::GetCapturedStderr();

  EXPECT_THAT(exit_code, Eq(EXIT_SUCCESS));
  EXPECT_THAT(
    error_output, HasSubstr("Could not find converter for format some_rmw"));
}
```

I don't think we should expect a `EXIT_SUCCESS` here, rather a `EXIT_FAILURE`.

Karsten1987:
fixes #746 

as described in the linked issue, the exit code should rather be `EXIT_FAILURE` than `EXIT_SUCCESS`.

### DIFF
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
@@ -584,7 +584,7 @@ TEST_F(RecordFixture, record_fails_gracefully_if_plugin_for_given_encoding_does_
     execute_and_wait_until_completion("ros2 bag record -a -f some_rmw", temporary_dir_path_);
   auto error_output = internal::GetCapturedStderr();
 
-  EXPECT_THAT(exit_code, Eq(EXIT_SUCCESS));
+  EXPECT_THAT(exit_code, Eq(EXIT_FAILURE));
   EXPECT_THAT(
     error_output, HasSubstr("Could not find converter for format some_rmw"));
 }


Karsten1987:
once ci is available again, I want to run the rosbag2_tests package without the `xfail` label to see if that fix recovers already the `record_end_to_end` test.

emersonknapp (approved):

MichaelOrlov:
@Karsten1987 Sorry that I am writing in closed PR. \
Could you please clarify how or why you come to the conclusion that it should be `EXIT_SUCCESS` rather than `EXIT_FAILURE`.

The reason why I am asking is because I am getting failure on my development branch in this test exactly at the same place where you changed expectation. \
I have tried to make a brief analysis and come to conclusion that it should be `EXIT_FAILURE`. \
I might be overlook something, but here is my thoughts:
1. In this test we are expecting to observe message "Could not find converter for format some_rmw" in the output.
2. This message only appears as text in throwing exception in https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/src/rosbag2_cpp/converter.cpp#L49-L56
3. I've tried to trace this exception up to the stack and seems we are not handling it anywhere. Please correct me if I am wrong. But if we are not handling this exception we should expect `EXIT_FAILURE` result in this case.

It seems first time we are creating format converter in [void SequentialReader::check_converter_serialization_format](https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L236)
which we are calling from [void SequentialReader::open(..)](https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L132) \
First time we are calling reader_->open(storage_options_, {"", rmw_get_serialization_format()}); in [constructor](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/player.cpp#L110) of the player class which we invoke in [rosbag2_py::Player::play()](https://github.com/ros2/rosbag2/blob/master/rosbag2_py/src/rosbag2_py/_transport.cpp#L126) which is finally invokes from [`ros2bag verb play.py`](https://github.com/ros2/rosbag2/blob/master/ros2bag/ros2bag/verb/play.py#L120-L121) interface

None of them have try catch clause on the back trace.

MichaelOrlov:
@Karsten1987 I apologies, I messed up. You made it right  - you changed `EXIT_SUCCESS` to the `EXIT_FAILURE`
 and this is exactly what I am expecting to observe. Sorry my bad.

PR #749, opened 2021-04-22
remodel publication manager

Karsten1987:
A few things here:

* I renamed the `publisher_manager` to `publication_manager`.
* The publication manager sets up a node in its constructor 
  * It can't thus be default initialized as a class member var as `rclcpp::init` wasn't triggered just yet.
  * A separate call to setup publishers before running it, eases up discovery of the node and the publishers before actually calling publish
  * We only have one node for all publishers within the test. Before that, each publisher had its own node, which certainly didn't help discovery.
* General cleanup of the `record_end_to_end_tests`


Karsten1987:
CI without `xfail`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9212/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14542/)

I don't necessarily expect a clean CI here, but the record end to end tests should pass.

Karsten1987:
master is currently broken due to https://github.com/ros2/rmw_cyclonedds/pull/306

Karsten1987:
building without CycloneDDS:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14442/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9213/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12113/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14545/)

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -318,27 +302,24 @@ TEST_F(RecordIntegrationTestFixture, duration_and_noncompatibility_policies_mixe
 TEST_F(RecordIntegrationTestFixture, topic_qos_overrides)
 {
   const auto num_msgs = 3;
-  test_msgs::msg::Strings strict_msg;
-  strict_msg.string_value = "strict";
-  const auto strict_topic = "/strict_topic";
+  test_msgs::msg::Strings string_msg;
```
emersonknapp, line 305: This whole test might not do anything anymore, since we added the subscriber-qos-adaptation feature. I think even without an override the recording gets everything here now.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -0,0 +1,103 @@
...
+      [publisher, message, repetition, interval](bool verbose = false) {
+        for (auto i = 0u; i < repetition; ++i) {
+          if (rclcpp::ok()) {
+            publisher->publish(message);
+            if (verbose) {
+              RCLCPP_INFO(
+                rclcpp::get_logger("publication_manager"),
+                "publish on topic %s", publisher->get_topic_name());
+            }
+            std::this_thread::sleep_for(interval);
```
emersonknapp, line 71: Would there be a slightly simpler way to build this with timers instead? Not sure, just thinking out loud

emersonknapp (approved):
LGTM - i like that the code got smaller

rosbag2_transport/test/rosbag2_transport/test_record.cpp
Karsten1987: are you saying we could just remove this test altogether?

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Karsten1987: I am actually in favor of this solution. Timers involve node spinning and executors. Here I can simply call publish and sleep.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
Karsten1987: done in e78ddbe

Karsten1987:
with cyclone
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14446/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9217/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12117/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14550/)

PR #751, opened 2021-04-22
Move Player::play setup logic mostly into constructor

emersonknapp:
Not sure if this is the exact pattern we want, but it moves publisher and clock setup into the constructor. And, makes the clock live on the stack so it can't be NULL.

Put here for consideration.


emersonknapp:
Closing in favor of #754

PR #753, opened 2021-04-23
Expose GetRate/SetRate services for playback

emersonknapp:
Depends on #729

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -250,7 +250,7 @@ double Player::get_rate() const
   if (clock_) {
     return clock_->get_rate();
   }
-  return 0.;
+  return 1.0;
 }
 
 bool Player::set_rate(double rate)
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -132,6 +132,24 @@ Player::Player(
     {
       response->paused = is_paused();
     });
+  srv_get_rate_ = create_service<rosbag2_interfaces::srv::GetRate>(
+    "~/get_rate",
+    [this](
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::GetRate::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::GetRate::Response> response)
+    {
+      response->rate = get_rate();
+    });
+  srv_set_rate_ = create_service<rosbag2_interfaces::srv::SetRate>(
+    "~/set_rate",
+    [this](
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::SetRate::Request> request,
+      const std::shared_ptr<rosbag2_interfaces::srv::SetRate::Response>/* response */)
+    {
+      set_rate(request->rate);
+    });
 }
 
 Player::~Player()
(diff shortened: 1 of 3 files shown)


emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/b384e0eaab9d55c633a49977d2997716/raw/8bbbd327f7627c3b5b4134f6358a383249be526f/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_transport rosbag2_tests rosbag2
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8299
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14456/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9226/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14558/)

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -227,6 +245,25 @@ bool Player::is_paused() const
   return true;
 }
 
+double Player::get_rate() const
+{
+  if (clock_) {
+    return clock_->get_rate();
+  }
+  return 0.;
```
Karsten1987, line 253: would it make sense to return a `1.0` here as a default?

Karsten1987 (approved):

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -227,6 +245,25 @@ bool Player::is_paused() const
...
+  if (clock_) {
+    return clock_->get_rate();
+  }
+  return 1.0;
+}
+
+bool Player::set_rate(double rate)
+{
+  if (rate <= 0.0) {
+    return false;
```
MichaelOrlov, line 259: @emersonknapp I thought about that it would be better to put this check
```cpp
if (rate <= 0) {
```
inside `clock_->set_rate(rate);` and return boolean from `set_rate(rate)`. i.e. API changes require.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -335,7 +372,16 @@ void Player::prepare_clock(rcutils_time_point_value_t starting_time)
...
+  bool paused = false;
+  // Copy over important settings
+  if (clock_) {
+    rate = clock_->get_rate();
+    paused = clock_->is_paused();
+  }
   clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(starting_time, rate);
+  if (paused) {
+    clock_->pause();
+  }
```
MichaelOrlov, line 384: @emersonknapp It seems that we need to add `paused` inside constructor of `rosbag2_cpp::TimeControllerClock` as a parameter or initialize it by default to true. Otherwise if we had been in pause and doing clock re-init here, the clock might advance since creation when we will hit pause here.

emersonknapp: Good comment - this is handled by https://github.com/ros2/rosbag2/pull/754 since we no longer re-create the clock.

rosbag2_transport/src/rosbag2_transport/player.cpp
emersonknapp: I would love to be able to run time in revers - do you think that makes sense as a possible future feature? I was thinking that the "rate > 0" requirement is a current limitation of the `Player` - rather than being a limitation of the `PlayerClock`

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Yeahh.. I just reviewed it. I agree that it will not cause problem when it will be in constructor.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: So well, we are doing abstraction in player around clock. And if clock would be able to run in revers why not? \
It least I don't see any problem in Player class if it would be negative. Please correct me if I am wrong.

Issue #663, opened 2021-02-18
Recording - limited buffer with triggerable snapshot


d-walsh:
## Description

The snapshot capability was added to the ROS1 version of rosbag and it would be great to support this use case with ROS2. https://github.com/ros/rosbag_snapshot

Feature high level:
* Don't record to disk, instead keep a full buffer of a configurable size
* On a trigger (probably a Service call), dump the contents of the buffer to disk

This allows for capturing data around events of interest without having to store the data for an entire recording duration, saving on disk usage for resource limited use case.

## Testing Suggestions

emersonknapp:
Really cool feature :+1:

emersonknapp:
Would like to have more details of the feature on this ticket itself. Putting to bottom of backlog for now.

kielczykowski:
Hey, it feels like i can give a hand with implementing this feature.

Would it be more suitable to open PR and new discussion there or continue designing solution in this thread?

emersonknapp:
As with most features of any significant weight, I'd prefer a design proposal first before looking at an implementation (feels slower - but slow is smooth and smooth is fast). You could open up a PR with a new document in the folder https://github.com/ros2/rosbag2/tree/master/docs/design - then we could tie this feature request ticket to that design.

There's room for flexibility - if you think it's faster to throw up a sample implementation than to write a doc - but generally I find it easier to review the doc, that way any disagreement over definitions or behaviors can be hashed out before a lot of time is lost on making it work well enough for a proof-of-concept.

Luckily, we're getting to a place where it should be a lot easier to implement something like this. The `Recorder` has been converted to be a `rclcpp::Node` which will make it easier to expose new APIs via services (I am currently implementing the same idea for playback control on `Player`)

In a design, I would be looking for specifics on expected user inputs / disk outputs. It doesn't have to focus too heavily on the exact implementation details, but pseudocode API or diagrams with an idea of how they tie to existing structures would help make the review go smoothly. Feel free to join the [Tooling WG](https://docs.google.com/document/d/1UXWaEkT6ji0Z0uTtJJSni5Y6fiopM9qOOzv4ldeNaAM) meetings if you want to have open discussion.

JWhitleyWork:
@emersonknapp I am apparently blind but thanks for pointing this out! Is there documentation for usage?

emersonknapp:
You know what - that made me realize there isn't! I'll look into adding some now.

[...]

PR #755, opened 2021-04-23
Recording triggerable snapshot design proposal

kielczykowski:
# Description
As suggested in https://github.com/ros2/rosbag2/issues/663, this PR contains a design proposal for rosbag2 triggerable snapshot feature implementation.

For now I found multiple inconveniences with design decision making and I'm looking for a feedback and tips on what are the expectations for this feature design.


docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,63 @@
...
+  * Resuming continues the process of buffering after pause. Resuming would clear all buffers.
+
+
+## Proposal
+
+As thinking of this design in the past `rosbag_snapshot` was written as different package than `rosbag`. It was defined this way not to break the existent API of `rosbag`. Taken this into consideration it could be done so this time too. However it would require duplicating a large part of code which is already written and being optimized as a part of `rosbag2` repository. So as not being sure which way would be the most appropriate to implement I propose multiple approaches.
+
+### Outside rosbag2
+
+First approach implements `ros2 bag snapshot` as a different package. Mostly it requires rewriting `Recorder` node completly. Completly means in this case diving into implementation of `Writer`, `SequentialWriter` and `BaseWriterInterface` to copy and adjust writing to disk algorithms.
```
emersonknapp, line 30: I don't think that this makes sense, and I would honestly remove it from the proposal - you might add a section at the end called "alternate approaches considered" - but top-level I like to see a design being a single proposal of what should be implemented, not a selection of choices.

I think everyone would agree that this is a useful feature for rosbag2 to do - somebody could implement something in a separate repository/package without doing any design here - I think our goal is to determine a design that _does_ fit into `rosbag2` directly.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,63 @@
...
+* Many new functions would be redefined or even redundant
+* Breaking the initial desing of writers/readers in favor of putting more effort to custom mechanism
+* Not being part of `rosbag2` and `ros2bag` (cli incompatibility)
+
+### Writer Interfaces changes + `Recorder` modifications
+
+Second approach would base on making changes to `Recorder`, writers and their interfaces. This approach bases on adding service to `Recorder` which handles triggering buffer save. Although current design of writers bases on `BaseWriterInterface` which would need to be changed to enable some kind of `save()/flush()` functionality. I believe this is not an elegant approach but I haven't find other way to trigger buffer save from `Recorder`, through `Writer` to `SequentialWriter`. For sure it requires implementing new `SnapshotWriter` since it needs to take care of buffering the messages from subscribers instead of passing them straight to `MessageCache` and `CacheConsumer`.
+
+Pros:
+* Maintaining snapshot inside `ros2 bag snapshot ...` (or even `ros2 bag record snapshot ...`) and `Recorder` functionality
```
emersonknapp, line 45: I'm not sure that this makes sense as a separate verb. I would think that the circular-buffer + snapshot approach should be considered a behavioral modification of recording, just like `rosbag2 play --loop`

I'd like to see a specific section for "User Interface" - how do I interact with this feature (without knowing how it is built)?

I think maybe something like the following:

```
# Recording is invoked with this feature via an extra option to "record"
ros2 bag record --buffer-snapshot

# Snapshot is exposed as a Service on the Recorder node (so it can be triggered remotely)
ros2 service call /rosbag2_recorder rosbag2_interfaces/Snapshot

# C++ API
* RecordingOptions has a new value "buffer_snapshot"
* Recorder has a new API `Recorder::snapshot`
```

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,63 @@
...
+* Reuse most of already written code (producer-consumer in save, abstract interfacing between writers)
+
+Cons:
+* `BaseWriterInterface` redefinition (adding at least one pure abstract function)
+* May require many changes in current implementation of the above mentioned classes
+
+### Custom `Recorder` implementation
+
+Last but not least it is possible to write new alternative to `Recorder` node inside `rosbag2_transport` package. `Snapshoter` would implement functionalities of subscribing the messages, buffering them at high level node and most probably passing data straight to `MessageCache` and `CacheConsumer`.
```
emersonknapp, line 54: I think that the `Recorder` actually won't know much about this behavior. The `SequentialWriter` has the `message_cache_` and is the entity that will probably implement this behavior. That's the sort of proposal I'm looking for here, like "what owns what functionality, logically". The Recorder itself I think should just pass through configuration, its responsibility is the transport layer, e.g. Subscriptions, and the Writer should handle behavior around writing.

emersonknapp:
Hopefully the above comments help clarify what we're looking for here - overall I'd structure it like:

1. What's the feature, what does it do?
2. How does a user interact with it? CLI, ros communications, language APIs
3. What are the major components involved and what portion of the functionality do they own

Clarifying the above things via this document makes it feel to me like there won't be any big surprises or rabbit-holes in the implementation/review stage. Let me know if I can help clarify anything else.

kielczykowski:
After long time I found some time to design the feature in more details. Still haven't figured out some of the concerns, but I believe code-like documentation will help dispel them.

 I am waiting for more suggestions and opinions.

kielczykowski:
It feels like at this point of development a validation of my idea would be helpful. 
@emersonknapp could you take a look at current design and share your thoughts about it?

kielczykowski:
Managed to get some community review, but would love maintainers feedback (@emersonknapp @Karsten1987)

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+### Cons
+
+* Prevents bag splitting
+* Problematic configuration, pausing and resuming feature (no direct way to change behavior during runtime)
+* Problematic messages counting and rosbag start timestamping
+* (see `TODO` section)
+
+## In-depth structure changes
+
+```diff
```
emersonknapp, line 93: I don't know if I accidentally implied that you should do this diff thing - but please remove the entire diff section. The design review is not the place to put actual code - this will be out of date as soon as we actually implement it. The design should be the place for high level concepts that don't get out of date.

If you would like to open a separate PR with all these changes actually implementing this feature, that would be great! We can link them together in the descriptions as companions of each other.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+* "Recorder" - rosbag2 node which records data right after program start until killed
+
+## Feature description
+
+Rosbag in ROS1 had functionality of creating a [snapshot](https://github.com/ros/rosbag_snapshot). Snapshotter was a program which subscribed to topics and buffered messages until a `trigger` for snapshot was sent. As in Rosbag2 we want to support the following features:
+
+* Save data to rosbag on trigger
+  * Enable triggering buffer save to disk with service call
+
+* Pause and resume buffering process
```
emersonknapp, line 19: I don't understand this feature very well - my understanding for the intention of this feature is:

* rosbag2 is always keeping a circular buffer of the latest information - but not writing anything to disk
* when we trigger a snapshot - dump the entire buffer to disk

Why does a user want to pause buffering? This sounds to me like it makes rosbag2 just drop data, such that you wouldn't be able to take a snapshot of it.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+  * Pausing refers to stopping buffering new messages until resumed or write is triggered.
+  * Resuming continues the process of buffering after pause. Resuming would clear all buffers.
+
+* Node and buffer configuration
+
+  Enable user to specify `cli` arguments which could be shared with `ros2 bag record` command. User should be able to configure the following arguments:
+
+    | parameter name | description   | behavior  |
+    | -------------- | ------------- | --------- |
+    | buffer size    | maximum memory per topic | <ul><li> defined by new `--max-buffer-size` argument, default value is 0 [bytes]  </li><li>if not set/set to non-positive value only duration parameter will be relevant (*note: either buffer size or duration must be specified*) </li></ul> |
```
emersonknapp, line 29: Why would "buffer size" be different than the "cache size" that is currently used for double-buffered writing? It seems to me that we already have a configuration for an in-memory buffer of data. When snapshot is enabled, we would use this buffer circularly rather than triggering a write-to-disk when it reaches the configured size.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+    | duration       | maximum time difference between newest and oldest message | <ul><li> defined by new `--max-buffer-duration` argument, default value is 10.0 [s] </li><li> if not set/set to non-positive value only buffer size parameter will be relevant (*note: either buffer size or duration must be specified*) </li></ul> |
```
emersonknapp, line 30: Same as above comment - I think we have a feature request for a maximum duration of the write cache, this should probably be the same parameter, to minimize the number of parameters.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
emersonknapp: Additionally, why would this be per-topic. It seems better to use this as a total value so that the user has a guarantee of maximum memory usage. Regardless, we don't currently have per-topic caches, so that would require a rewrite of the caching system.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+ros2 service call /rosbag2_recorder/snapshot rosbag2_interfaces/srv/Snapshot "{ }"
+
+# Reconfigure snapshot max memory configuration
+ros2 service call /rosbag2_recorder/snapshot/memory rosbag2_interfaces/srv/SetMemory "{memory: 1000000}"
+
+# Reconfigure snapshot max duration configuration
+ros2 service call /rosbag2_recorder/snapshot/duration rosbag2_interfaces/srv/SetDuration "{duration: 50.0}"
+
+# Clear snapshot buffer
+ros2 service call /rosbag2_recorder/snapshot/clear_buffer rosbag2_interfaces/srv/ClearBuffer "{ }"
```
emersonknapp, line 56: I'm a little confused by this feature (and it doesn't seem to be mentioned in the introduction) - a snapshot write should clear the buffers, but why would we want to do it without a write?

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+
+* `CacheBufferFactory` check if returns good `CacheBufferInterface` type instances, dependent on passed arguments
+* `MessageSnapshotBuffer` check pushing behavior
+* `MessageCache` check synchronization and start timestamp calculation
+* `SequentialWriter` test if snapshot forces using cache, starting timestamp change in snapshot mode and bag split when calling snapshot
+* `Recorder` service call check
+
+#### TODO
+
+* Deleting empty rosbag after node closure
```
emersonknapp, line 587: This is a separate feature that already has an issue open, you don't need to worry about it for this design.

docs/design/rosbag2_snapshot_recorder/rosbag2_snapshot_recorder.md
```diff
@@ -0,0 +1,738 @@
...
+* Resolve how to implement pause/resume functionality
+* Think about configuration change during runtime
```
emersonknapp, line 589: I think that this feature can be intentionally excluded from this design. Maybe add an "Out of scope" section near the top, something like


```
## Out of scope

* Reconfiguration of recording settings during runtime
```

emersonknapp (commented):
I can see that this exercise is maybe not something you're used to. I would also be happy to review a PR that starts with just the API changes, so that we can review the interfaces first.

My main comment across this review is:

We should try to keep the interface as simple as possible. We already have `--max-cache-size` argument, which could be used to implement the first version of this feature. I would rather review a simplified feature that does one thing well than everything you might want it to do. It seems likely that we could maintain bag splitting and everything, just changing the `MessageCache` to have a different implementation, so that it doesn't provide a queue to the Consumer/Writer until triggered, rather than the current implementation where it provides the buffer once it is full.

danthony06:
I'll throw in my two cents here, I think my comments are high level enough that I'm not going to insert them into the review process.

- I really like the overall idea. I have not seen the rosbag_snapshot package before, but I wish I had, because the functionality looks very useful, and ROS 2 will benefit from the same type of system.
- The ROS 1 implementation used a separate node that cached all of the requested topics, and then wrote them out on command to a bag file using the ROS 1 bag interface. Looking through the comments [here](https://github.com/ros/ros_comm/pull/1414), I can see why. Fully implementing the functionality adds a lot of configurable options to the command line interface, and added a good amount of code to the rosbag codebase. I see why moving this feature into a separate node is attractive, because the maintenance burden could be high.
- With that being said, does the rosbag2 code offer the functionality to reimplement this functionality? Very crude pseudocode would be something like:
```
bag = rosbag2.open(file_path)
for message in message_cache:
    bag.write(message)
bag.close
```

emersonknapp:
> The main advantage that I can possibly see in embedding this functionality in the main rosbag2 codebase is re-using the message caches to reduce the memory footprint of the system. If done cleverly enough, we could avoid duplicating the cache of snapshot messages and the messages to be written out to a bag file, but this could get annoyingly complicated when things like splitting bag files, limiting bag file sizes, and running a snapshot while bag recording is paused, are considered.

The main point I am trying to make is that this _shouldn't_ be annoyingly complicated. All of the functionality to do splitting, bag sizes, etc is already handled by the `Writer` functionality - and there should be a strong separation of concerns between the `Writer` and the in-memory `MessageCache`. Instead of trying to mix it all up, you should be able to get the advantage of all the `Writer` functionality without any changes whatsoever, by providing a different `CircularTriggeredMessageCache` that provides a different way to talk to the writer.

I think that strong separation of concerns makes for much more simple implementations - I think we can manage that here.

emersonknapp:
Closing in favor of https://github.com/ros2/rosbag2/pull/837

PR #761, opened 2021-04-28
[test] Run Action CI on Windows/OSX as well

emersonknapp:
I'm not sure of a way to avoid copying the logic, because we can't use containers on Win/Mac, but we definitely want to on Linux - it's a lot faster.


emersonknapp:
closing for now until the underlying actions actually are expected to work

PR #765, opened 2021-05-10
Recording with --all and --exclude fix

pijaro:
During regex/exclude filtering `take` flag works incorrectly. This results in dropping all messages after first `take` is set to `false`.

This PR fixes this by resetting `take` flag before each topic check.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -157,8 +157,8 @@ Recorder::get_requested_or_available_topics()
 
   std::regex topic_regex(record_options_.regex);
   std::regex exclude_regex(record_options_.exclude);
-  bool take = record_options_.all;
   for (const auto & kv : filtered_topics_and_types) {
+    bool take = record_options_.all;
     // regex_match returns false for 'empty' regex
     if (!record_options_.regex.empty()) {
       take = std::regex_match(kv.first, topic_regex);
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -153,24 +153,12 @@ Recorder::get_requested_or_available_topics()
     return filtered_topics_and_types;
   }
 
-  std::unordered_map<std::string, std::string> filtered_by_regex;
-
-  std::regex topic_regex(record_options_.regex);
-  std::regex exclude_regex(record_options_.exclude);
-  for (const auto & kv : filtered_topics_and_types) {
-    bool take = record_options_.all;
-    // regex_match returns false for 'empty' regex
-    if (!record_options_.regex.empty()) {
-      take = std::regex_match(kv.first, topic_regex);
-    }
-    if (take) {
-      take = !std::regex_match(kv.first, exclude_regex);
-    }
-    if (take) {
-      filtered_by_regex.insert(kv);
-    }
-  }
-  return filtered_by_regex;
+  return topic_filter::filter_topics_using_regex(
+    filtered_topics_and_types,
+    record_options_.regex,
+    record_options_.exclude,
+    record_options_.all
+  );
 }
 
 std::unordered_map<std::string, std::string>
--- a/rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
+++ b/rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
@@ -39,6 +39,15 @@ filter_topics_with_more_than_one_type(
   const std::map<std::string, std::vector<std::string>> & topics_and_types,
   bool include_hidden_topics = false);
 
+ROSBAG2_TRANSPORT_PUBLIC
+std::unordered_map<std::string, std::string>
+filter_topics_using_regex(
+  const std::unordered_map<std::string, std::string> & topics_and_types,
+  const std::string & filter_regex_string,
+  const std::string & exclude_regex_string,
+  const bool & all_flag
+);
+
 }  // namespace topic_filter
 }  // namespace rosbag2_transport
 
--- a/rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
+++ b/rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
@@ -45,7 +45,7 @@ filter_topics_using_regex(
   const std::unordered_map<std::string, std::string> & topics_and_types,
   const std::string & filter_regex_string,
   const std::string & exclude_regex_string,
-  const bool & all_flag
+  bool all_flag
 );
 
 }  // namespace topic_filter
--- a/rosbag2_transport/test/rosbag2_transport/test_topic_filter.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_topic_filter.cpp
@@ -25,6 +25,19 @@
 
 using namespace ::testing;  // NOLINT
 
+class RegexFixture : public Test
+{
+protected:
+  std::unordered_map<std::string, std::string> topics_and_types_ = {
+    {"/planning", "planning_topic_type"},
+    {"/invalid_topic", "invalid_topic_type"},
+    {"/invalidated_topic", "invalidated_topic_type"},
+    {"/localization", "localization_topic_type"},
+    {"/invisible", "invisible_topic_type"},
+    {"/status", "status_topic_type"}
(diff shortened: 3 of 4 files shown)


emersonknapp (changes requested):
This change looks good - can you add a regression test that would fail without this fix?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -157,8 +157,8 @@ Recorder::get_requested_or_available_topics()
 
   std::regex topic_regex(record_options_.regex);
   std::regex exclude_regex(record_options_.exclude);
-  bool take = record_options_.all;
   for (const auto & kv : filtered_topics_and_types) {
+    bool take = record_options_.all;
     // regex_match returns false for 'empty' regex
     if (!record_options_.regex.empty()) {
```
MichaelOrlov, line 163: It would be nice to override `regex` by `record_options.all` in the logic as well.
```cpp
 if (!record_options.all && !record_options.regex.empty())
```
Please see my comment with explanation here https://github.com/ros2/rosbag2/pull/765#issuecomment-837263558

MichaelOrlov (changes requested):

pijaro:
@emersonknapp @pijaro I also struggle with a valid regression test for this issue. However I will give it another shot next week.

As for the priority for `-a` and `-x` - I really like the idea of prioritizing `-a`. If we all agree that this is the way to go, I will include it in this PR. What do you think @emersonknapp ?

carlossvg:
> @emersonknapp @MichaelOrlov I also struggle with a valid regression test for this issue. However I will give it another shot next week.

Hi @pijaro,

We (Apex.AI) created a regression test for this bug very similar to [the existing one](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/test/rosbag2_transport/test_record_regex.cpp#L97-L168). I found that the test was flaky with Fast-RTPS due to some issue in the topic discovery. I was not able to reproduce this issue in rolling though. I wonder if the struggle you mentioned above is somehow related or it's something else.

MichaelOrlov:
@carlossvg To solve flakiness with discovery we could add `wait_for_matched` in `publication_manager` similar as I am adding it in `subscription_manager` in this [PR](https://github.com/MichaelOrlov/rosbag2/blob/michaelorlov/play_next/rosbag2_test_common/include/rosbag2_test_common/subscription_manager.hpp#L76-L118) but instead of doing spin with executor need just sleep for some interval.

emersonknapp:
> As for the priority for -a and -x - I really like the idea of prioritizing -a. If we all agree that this is the way to go, I will include it in this PR. What do you think @emersonknapp ?

I agree with this - to me `--all` should always mean "all topics are recorded", it's confusing to me if that's not the case. We can note this in the `argparse` documentation to make it more clear.

emersonknapp:
>  I also struggle with a valid regression test for this issue

I think that this could be done with a _unit_ test, rather than an _integration_ test - which should make it faster and less prone to flakiness. 

The way I would do this - is factor out `get_requested_or_available_topics` into a separate utility function with a signature like

```
get_requested_or_available_topics(
  const std::map<std::string, std::vector<std::string>> & topics_and_types, 
  const rosbag2_transport::RecordOptions & record_options
);
```

From within `Recorder` - you could call this like

```
auto topics_to_subscribe = filter_topics(this->get_topic_names_and_types(), record_options_);
```

Then - you can simply call this one function in a unit test to assert that it works the way we think it should.

To me - that would be a better solution than keeping it embedded in a stateful class, and trying to figure out if it works by running a full recording.

mitsudome-r:
> I agree with this - to me --all should always mean "all topics are recorded", it's confusing to me if that's not the case. We can note this in the argparse documentation to make it more clear.

I personally disagree with this. Default rosbag takes whitelist approach to specify with topic to listen to. However, `-x` is excluding option, which is blacklist approach of specifying topic names. I think that many users would like to specify all topics before setting a blacklist. Also, it does say in the document for `-x` that it works with `-a` option.

We still have a workaround to select all topics first with `-e ".*"` instead of `-a` so I'm not **strongly** against it, though.

 https://github.com/ros2/rosbag2/blob/c488315bc9a19b67961a176e10b1127019fd2f72/ros2bag/ros2bag/verb/record.py#L49.

carlossvg:
I agree with @mitsudome-r,  it makes perfect sense to combine the `all` and `exclude` options. Note, the change proposed by @MichaelOrlov in https://github.com/ros2/rosbag2/pull/765#discussion_r629655198 would only prioritize `all` over `regex`.  Combining `all` with `exclude` would still work as before.

emersonknapp:
You're both right - I hadn't thought all the way through it. `--all --exclude` should definitely work as you're describing - everything except the explicitly excluded topics.

pijaro:
Thank you for your suggestions, really appreciate it!

I refactored code to extract regex filtering in order to make unit test like @emersonknapp suggested. I didn't pull out whole `get_requested_or_available_topics` method but only the regex part and moved it to the `topic_filter.cpp`. I think this is more aligned with existing code. Change suggested by @MichaelOrlov is also included.

I added requested unit test with bunch of checks as well.

rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
```diff
@@ -39,6 +39,15 @@ filter_topics_with_more_than_one_type(
   const std::map<std::string, std::vector<std::string>> & topics_and_types,
   bool include_hidden_topics = false);
 
+ROSBAG2_TRANSPORT_PUBLIC
+std::unordered_map<std::string, std::string>
+filter_topics_using_regex(
+  const std::unordered_map<std::string, std::string> & topics_and_types,
+  const std::string & filter_regex_string,
+  const std::string & exclude_regex_string,
+  const bool & all_flag
```
emersonknapp, line 48: It doesn't make sense to take a boolean by const-reference - this argument should be by value.

rosbag2_transport/test/rosbag2_transport/test_topic_filter.cpp
```diff
@@ -122,3 +122,82 @@ TEST(TestTopicFilter, filter_topics) {
...
+    bool all_flag = true;
+
+    auto filtered_topics = rosbag2_transport::topic_filter::filter_topics_using_regex(
+      topics_and_types,
+      filter_regex_string,
+      exclude_regex_string,
+      all_flag
+    );
+
+    EXPECT_THAT(filtered_topics, SizeIs(6));
```
emersonknapp, line 201: Nitpick: I think that each of these blocks would be more appropriate as a separate test case. You could define `topics_and_types` in a fixture or just as a static constant.

emersonknapp (changes requested):
This looks really good - thanks for the revision. I just have the 2 small comments then we're ready to go.

rosbag2_transport/src/rosbag2_transport/topic_filter.hpp
pijaro: Good point, fixed.

rosbag2_transport/test/rosbag2_transport/test_topic_filter.cpp
pijaro: I moved `topics_and_types` to a fixture and split whole test into multiple smaller ones.

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/6ab205d61a54c1bae965aa9b00e28e94/raw/b1aad07ae943c37eba6b8141271efb087df6fa97/ros2.repos
BUILD args: --packages-up-to rosbag2 rosbag2_transport rosbag2_tests
TEST args: --packages-select rosbag2 rosbag2_transport rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8478
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14656/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9409/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12320/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14800/)

wep21:
@emersonknapp Will this be backported to `galactic`?

emersonknapp:
I'm always happy to review a backport PR!

PR #768, opened 2021-05-13
Design document for keyboard handler

MichaelOrlov:
Design document for keyboard handler package.

Relates to the #735 

Part of the #696

### DIFF
--- a/docs/design/keyboard_handler.md
+++ b/docs/design/keyboard_handler.md
@@ -60,7 +60,8 @@ not better than C or C++ for this case. And here is the rational for that:
    enough and have low quality.
 2. Most of the ROS2 code written in C++ and it will be much easy to support and use solution
    written on the same language.
-3. Some of the POSIX compatible OSes doesn't have Python interpreter. For instance [QNX for safety](https://blackberry.qnx.com/en/software-solutions/embedded-software/qnx-os-for-safety) 
+3. Some of the POSIX compatible OSes doesn't have Python interpreter. For instance 
+   [QNX for safety](https://blackberry.qnx.com/en/software-solutions/embedded-software/qnx-os-for-safety) 
 
 ## Specific of handling input from keyboard on different platforms
 Processing of the keyboard handling is differ for different operating systems.
@@ -68,7 +69,7 @@ There are two major different approaches:
 1. Unix/MacOS aka POSIX "compatible" OSes
 2. Windows family OSes
 
-Posix compatible OSes approach based on readout from standard input linked to the terminal.
+POSIX compatible OSes approach based on readout from standard input linked to the terminal.
 Pressed key combination represented as sequence of 8 bit values. Number of the 8 bit values in sequence
 could vary from pressed key combination and usually in range of 1-8 chars. Note: There are no
 '\0' value representing null terminator in providing sequence of bytes. i.e. those byte
@@ -82,6 +83,27 @@ integer values. Simple alphanumerical keys represented as single integer. Contro
 function keys represented as two integer values. Note: for some key combinations first integer 
 value could contain '\0' value representing null terminator in regular strings. 
 i.e it is not possible to treat such sequence of bytes as regular strings.
+
+###Specific of handling input from keyboard on POSIX compatible platforms
+POSIX systems support two basic modes of input: canonical and noncanonical. In canonical input 
+processing mode, terminal input is processed in lines terminated by newline '\n', EOF, or 
+EOL characters. No input can be read until an entire line has been typed by the user. Most 
+programs use canonical input mode, because this gives the user a way to edit input line by line 
+and this is what many of users accustom to see when typing commands in terminal window. \
+However to be able to handle single key press, terminal input need to be switched to the 
+noncanonical mode. By design keyboard handler will switch current terminal session to the 
+noncanonical mode during construction and return it to the canonical mode in destructor.
+
+##Handling abnormal program termination via Ctrl+C
+By design keyboard handler not providing ability to transfer `Ctrl+C` key press event to its 
+clients via callbacks. It could be considered as current design limitation.  
+On POSIX compatible platforms when user pressing `Ctrl+C` OS generates specific `SIGINT` signal 
+and if this signal became unhandled current process will be terminated. \
+On POSIX compatible platforms keyboard handler will register it's own signal handler for the 
+`SIGINT` to be able to return terminal input in canonical mode after abnormal program termination 
+via pressing `Ctrl+C`. This signal handling could overlap with registered specific `SIGINT` 
+handler on upper level or inside keyboard handler's clients. It could cause side effects and 
+considering as current design and implementation limitations. 
 
 ## Handling cases when client's code subscribed to the key press event got destructed before keyboard handler.
 There are two options to properly handle this case:
@@ -132,6 +154,7 @@ There are two options to properly handle this case:
         void callback_func(KeyboardHandler::KeyCode key_code);
       }
       ```
+      
    2. Using [`shared_from_this()`](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/enable_shared_from_this). 
       Note in this case client class should be inherited from 
       [`std::enable_shared_from_this`](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/enable_shared_from_this) 
@@ -155,10 +178,10 @@ There are two options to properly handle this case:
           keyboard_handler.add_key_press_callback(callback, KeyboardHandler::KeyCode::CURSOR_UP);
         }
       ```
+      
       `Player` class could be instantiated as 
       `std::shared_ptr<Player> player_shared_ptr(new Player());`
-   
-   
+      
 ## Handling cases when standard input from terminal or console redirected to the file or stream.
 By design keyboard handler rely on the assumption that it will poll on keypress event and then 
 readout pressed keys from standard input. When standard input redirected to be read from the 
@@ -186,6 +209,7 @@ skipped during unit test. For instance:
     // Handle exceptions
   }
 ```
+
 As you can see it will be impossible to separate exception handling for the case if keyboard 
 handler will throw exception when it's running under the `gtest`.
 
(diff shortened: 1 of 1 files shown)


docs/design/keyboard_handler.md
```diff
@@ -0,0 +1,87 @@
+# Keyboard handler
+Package providing ability to handle keyboard input via simple interface with callbacks. 
+
+## Goal
+We need to be able to handle keyboard input in unified way with cross-platform implementation.
+
+## Design Proposal
+It would be great to have some simple interface to be able to subscribe to the specific key
```
emersonknapp, line 8: This is good - for readability I would put this as a list of **design requirements**:

```
Design requirements:
* Subscribe to keyboard events via callbacks
* Subscriptions work with key modifiers (e.g. Ctrl+C)
* Multiple clients can subscribe to the same event
```

docs/design/keyboard_handler.md
```diff
@@ -0,0 +1,87 @@
...
+    KeyboardHandler keyboard_handler;
+    client1.register_key_press_callbacks(keyboard_handler);
+    client2.register_callbacks_for_keyboard_handler(keyboard_handler);
+```
+Inside client class registering for the callbacks could be organized as simple as:
+```cpp
+  ClientClass1::register_key_press_callbacks(KeyboardHandler & keyboard_handler) 
+  {
+    keyboard_handler.add_key_press_callback(callback_fn, KeyboardHandler::KeyCode::CURSOR_UP);
+    keyboard_handler.add_key_press_callback(callback_fn, KeyboardHandler::KeyCode::SHIFT_F1);
```
emersonknapp, line 26: Maybe this is "key + optional modifier", instead of defining predetermined combinations of keys:

```
# Subscribe to Shift+Ctrl+F1
keyboard_handler.subscribe(callback=callback_fn, key=KeyCode::F1, modifiers=KeyMod::SHIFT | KeyMod::CTRL);
```

docs/design/keyboard_handler.md
```diff
@@ -0,0 +1,87 @@
...
+polling "`int kbhit()`" system function call until it will return non zero value indicating that
+some key was pressed. After that need to read integer value(s) from the standard input to
+determine which key was pressed. Pressed key combinations could be represented as one or two 
+integer values. Simple alphanumerical keys represented as single integer. Control keys and specific
+function keys represented as two integer values. Note: for some key combinations first integer 
+value could contain '\0' value representing null terminator in regular strings. 
+i.e it is not possible to treat such sequence of bytes as regular strings.
+
+## Considerations of how to handle cases when client code subscribed to the key press event got destructed before keyboard handler.
+TBD.
```
emersonknapp, line 87: By returning some handle to the subscription that the client has to own, you can have it unregister itself in the destructor.

emersonknapp (commented):
The way this design sounds - I still am questioning whether this keyboard utility should live in its own repository. 

It also makes me wonder whether we should just be pulling in a 3rd party dependency rather than having to maintain code - like SDL or https://github.com/jkuhlmann/gainput

docs/design/keyboard_handler.md
MichaelOrlov: It would great to have such possibility to use such key modifiers which could be applied with "or" mask. But it is almost impossible or very difficult to implement in a cross-platform approach. \
Also in this case will be an open question how to detect to which key and modifier corresponds event when callback will arrive?
```cpp
 void callback_function(KeyboardHandler::KeyCode key_code)
  {
    using KeyCode = KeyboardHandler::KeyCode;
    switch (key_code) {
    // ...
```

MichaelOrlov: Ok I've tried to consider adding key modifiers for representing `SHIFT`, `ALT` and `CTRL`. \
It is possible to add this feature, although in this case it will add constraint on the values in `enum class KeyboardHandlerBase::KeyCode` that key codes with modifiers should be strictly ordered and follow in sequence in the same order as original key codes without modifiers. It will be much more error prone if someone will try to add new key codes to the enum and it's impossible to check if key codes will be out of order or if one them will be missed.

Giving that I have a strong opinion against such initiative. And I think that explicit prefix e.g. `SHIFT_F1` looks not so bad and pretty clear for usage.

MichaelOrlov:
@emersonknapp 
> The way this design sounds - I still am questioning whether this keyboard utility should live in its own repository.
> 
> It also makes me wonder whether we should just be pulling in a 3rd party dependency rather than having to maintain code - like SDL or https://github.com/jkuhlmann/gainput

Though it became pretty big, self contained and independent component with it's own unit tests and design document. It looks like it would be better to have it in a separate package.

As regards to the consideration if we should use some third party libraries. I've tried to consider some, but most of them have a big "baggage" in addition and extra other dependencies like  https://github.com/jkuhlmann/gainput or not well written/designed and documented. 

I would prefer to support our own well designed, well written and well documented cross-platform implementation rather than have some dependencies from others and struggle with inability to fix it if some issue will arise.

docs/design/keyboard_handler.md
emersonknapp: > n this case it will add constraint on the values in enum class KeyboardHandlerBase::KeyCode that key codes with modifiers should be strictly ordered and follow in sequence in the same order as original key codes without modifier

I was confused by this - I'm not quite sure what you mean.

The way I am thinking about it, you have something like:

```
enum KeyCode:
  a,
  b,
  c,
  F1,
  LEFT_BRACKET,
  etc...

enum KeyMod:
  CTRL,
  SHIFT,
  ALT
```

This is very easy to define - there is a linear number of each type, and events are given by combination. This is how I understand that most input libraries handle this interface.

The method you're proposing, e.g. `SHIFT_F1` means that you have to define M*M*M*N enum entries, where M is number of keymods and N is number of keycodes. This is not scalable to an entire keyboard.

emersonknapp: I'm looking at https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h for example of keycodes.

For pseudocode:

```
func on_keydown_event(KeyCode key) {
  if (key in MODIFIER_KEYS) {
    current_mods_ |= key;
  } else {
    auto callback = registered_callbacks_.find(current_mods_, key);
    if (callback) {
      callback();    
    }
  }
}

func on_keyup_event(KeyCode key) {
  if (key in MODIFIER_KEYS) {
    current_mods_ &= ~key;
  }
}


```

MichaelOrlov: It easy to define in design but not so simple to implement.

The problem is that I don't have such a low level API to be able to detect if some key modifier got pressed.
```cpp
if (key in MODIFIER_KEYS) {
    current_mods_ |= key;
```
Look on this
https://github.com/MichaelOrlov/rosbag2/blob/michaelorlov/keyboard_handler/keyboard_handler/src/keyboard_handler_unix_impl.cpp#L104-L120

It's just returning the sequence of codes like array and for instance for 'a' key corresponds one byte = '97' for `CTRL`+'a' it also returns one byte with value = '1' . \
And there are no way to detect if just only key modifiers pressed. Read function doesn't respond on them. \
In case of functional keys it's even more sophisticated. For instance:
```
static constexpr char             F11[] = {27, 91, 50, 51, 126, '\0'};
static constexpr char SHIFT_F11[] = {27, 91, 50, 51, 59, 50, 126, '\0'};
```
`Shift` modifier adding `59, 50` somewhere in the middle. The situation is even worse keyboard handler running in terminal under some application. For instance CLION IDE modifying returning sequence for F11 (refer to the [note](https://github.com/MichaelOrlov/rosbag2/blob/michaelorlov/keyboard_handler/keyboard_handler/src/default_unix_key_map.cpp#L16-L22))
```
constexpr char F11[] = {27, 91, 50, 51, 126, 27, '\0'};  // Extra 27 at the end
```
In current implementation I have simple luck up table with hash table for detecting key codes. This is relatively simple and straight forward approach.  And it would be relatively easy to extend API to provide ability to alter LUT in runtime with custom key codes without recompilation source code in case if someone will need a quick workaround for redefining existing key binding or need to add new key code.

If go to the design approach when modifiers will be separate from physical keys it will require to write some parser for returning key sequences with some heuristic inside.  In this case if someone will need to change corresponding key mapping it will require to change logic in those parser.

I can try to implement some parser for key modifiers but I am not sure how it will align with cross platform specific. Give me a couple days to try it.

MichaelOrlov: @emersonknapp I am in the process of redesigning keyboard handler to be able support key modifiers separate from base key press.

I made implementation for Linux with basic alphanumeric keys, although supporting functional and control keys with key modifiers is non trivial and there are no descent solution for it. The difficulties that such modifiers as `SHIFT`, `CTRL` and `ALT` among inserting some extra symbols in the middle of sequence also changing base values for some keys.
https://github.com/MichaelOrlov/rosbag2/blob/michaelorlov/keyboard_handler/keyboard_handler/src/default_unix_key_map.cpp#L41-L67
For instance:
```cpp
static constexpr char F1[]  = {27, 79, 80, '\0'};
static constexpr char SHIFT_F1[]  = {27, 91, 49, 59, 50, 80, '\0'};
```
The second byte for `F1` changing from 79 to 91.

I tend to agree and accept idea for using new API with modifiers i.g.
```cpp
/// \brief Type for callback functions
  using callback_t = std::function<void (KeyCode, KeyModifiers)>;
```
but **without supporting key modifiers for functional and control keys**. \
I would propose to put this restriction as design/implementation limitation and known issue.

Are you ok with it?

emersonknapp: Yes, I am OK with that limitation - let's make sure to document it.

MichaelOrlov: @emersonknapp I've prepared new version of the keyboard handler with changed API. Please review it at https://github.com/ros-tooling/keyboard_handler/pull/1.

I've documented design/implementation limitations in Doxygen comments for relevant classes. Also going to move and update design document in a separate PR.

MichaelOrlov: @emersonknapp I've updated design document and created new PR https://github.com/ros-tooling/keyboard_handler/pull/2 in ros-tooling please review.

emersonknapp: Approved the keyboard_handler PR, let's close this one

MichaelOrlov:
Closing this PR as would not merge in favor of the newly opened https://github.com/ros-tooling/keyboard_handler/pull/2

PR #771, opened 2021-05-17
Prepare release 0.9.0

emersonknapp:
Will bloom into Rolling and Galactic - and create `galactic` branch from this commit


Karsten1987:
will 0.9 be a galactic or rolling release?

emersonknapp:
I was planning to do both, as stated in the PR description. My thinking was to create the `galactic` branch at the squashed commit created by this PR.

Karsten1987:
hmm, given that they are different distros we can use the same version in both, I guess.

emersonknapp:
Is that not generally OK? My understanding was that Rolling contains releases that are also in other distros, since the idea is that the new distro is directly snapshotted from Rolling.

emersonknapp:
However, I'm not entirely sure how the "Rolling->Galactic" snapshot works, when it happens or if it already happened. So I thought I would do it manually.

Karsten1987:
I guess I'd done it slightly differently. Branch off, release rosbag2 as 0.9.0 into galactic, release 0.10.0 on master into rolling. But your approach would work as well I believe.

emersonknapp:
Yeah - right now `rolling` contains `0.8.0-1` and `galactic` contains `0.8.0-2`. I'd like to do less stuff and keep the branching as minimal as possible - so I'd rather share the changelog and do the same situation for 0.9.0

jhdcs:
Yeah, I thought Rolling was just sort-of the always-in-development branch. Where all the new stuff gets added. And then as each new release comes along (`Galactic`, h-turtle, etc...) we'd just bud off from Rolling.

emersonknapp:
> And then as each new release comes along (Galactic, h-turtle, etc...) we'd just bud off from Rolling.

That's correct but the confusion is coming from the fact that this already happened on April 19. However everything that has been done on `master` since then should still go into `Galactic` - so I'm suggesting we set common ancestor of Galactic and Rolling at today (the result of this PR).

Karsten1987:
> Yeah - right now rolling contains 0.8.0-1 and galactic contains 0.8.0-2. I'd like to do less stuff and keep the branching as minimal as possible - so I'd rather share the changelog and do the same situation for 0.9.0

That makes sense. Still, the moment we release a new version into rolling, we have to make sure to bump the minor version on master twice (or bump the major). Otherwise, we'd have two version 0.9 in rolling as well as galactic where they differ in terms of content though.

> That's correct but the confusion is coming from the fact that this already (technically) happened on April 19. However everything that has been done on master since then should still go into Galactic

I guess I didn't communicate this correctly back then. While I mentioned to @cottsay to skip the automatic/scripted branching to galactic for rosbag2, I believe I should have stated it more clearly that I expected the galactic release to be skipped as well. Setting a new common ancestor makes sense to me.

jhdcs:
If the work done since then can be reasonably assumed to be correct (passing all tests), I'm okay with it. Ideally, I'd like it to be more battle-tested with users in the field, but I think we're still somewhat working on getting things to the state that they CAN be battle-tested. No offense, of course. We've come a long way, but there's always work to be done, it seems...

emersonknapp:
> Still, the moment we release a new version into rolling, we have to make sure to bump the minor version on master twice (or bump the major)

I don't understand that this would be true. Rolling will have 0.9.0 after this release. The next time we release it, the release tools will bump from 0.9.0 to 0.10.0 automatically, we won't have to do it twice. The below is the plan for the branches

```
*        (rolling) some latest rolling stuff
|   *    (galactic) some galactic stuff
|   |
|  /
* (tag: 0.9.0) 0.9.0 release for Rolling && Galactic
```


> If the work done since then can be reasonably assumed to be correct (passing all tests), I'm okay with it. Ideally, I'd like it to be more battle-tested with users in the field, but I think we're still somewhat working on getting things to the state that they CAN be battle-tested. No offense, of course. We've come a long way, but there's always work to be done, it seems...

Sorry, this comment seems off-topic to me. Galactic has to be released today - it's the date for the release candidate (RC) for Galactic. It's not frozen in stone for eternity - we can backport bugfixes for the entire life of the distro.

jhdcs:
> > If the work done since then can be reasonably assumed to be correct (passing all tests), I'm okay with it. Ideally, I'd like it to be more battle-tested with users in the field, but I think we're still somewhat working on getting things to the state that they CAN be battle-tested. No offense, of course. We've come a long way, but there's always work to be done, it seems...
> 
> Sorry, this comment seems off-topic to me. Galactic has to be released today - it's the date for the release candidate (RC) for Galactic. It's not frozen in stone for eternity - we can backport bugfixes for the entire life of the distro.

Ah, whoops. Seems I'm misunderstanding what's going on here. Ignore me!

clalancette:
> That makes sense. Still, the moment we release a new version into rolling, we have to make sure to bump the minor version on master twice (or bump the major). Otherwise, we'd have two version 0.9 in rolling as well as galactic where they differ in terms of content though.

Well, you just need to make sure to bump the minor once in Rolling (the major is still zero).  That leaves you with room in the patch numbers to do bugfixes for Galactic.

My suggestion is to go forward with this PR as-is, releasing 0.9.0 into both rolling and galactic.  From there, you can go one of two ways:

1. Just leave this situation as-is until the next time you do a release into Rolling.  At that point, you *must* remember to bump the minor for Rolling, otherwise you take away the ability for Galactic to do point releases in non-confusing ways.
2. Immediately after this merge+release, do another release into Rolling as 0.10.0.  This ensures that you have the versions correct going forward, at the expense of burning another number (minor), and having a true branch (so bugfixes will have to be ported between the branches).

We do both 1 and 2 all over in the ROS 2 core, so the choice is yours.

emersonknapp:
Yes - I see that you mean we don't want different `0.9.1` in Rolling and Galactic. We will make a point to reserve `0.9.x` for Galactic, and the next Rolling release will be either a minor or major bump.

emersonknapp:
That being clarified - anybody want to approve the review so I can get the releasing done?

Karsten1987 (approved):

clalancette:
One other thing (which is probably obvious, but I'll state it anyway) is that you should make a 'galactic' branch right off of this commit.  Once you have done that, you'll need to update:

* https://github.com/ros2/ros2/blob/galactic/ros2.repos
* https://github.com/ros/rosdistro/blob/a8912fcdd49c57cb1107721d7b29ff91a58c0c4e/galactic/distribution.yaml#L2219-L2251
* https://github.com/ros2-gbp/rosbag2-release/blob/230f1f2665c4422e065d1e062672b565bc3a6976/tracks.yaml#L116

to point to that branch.

(we're aware that it is not ideal that you have to update this in 3 places; this is something that we are going to look into for next time, but right now we unfortunately need to do this)

emersonknapp:
Noted! I will make those changes immediately.

Issue #776, opened 2021-06-04
Add in `clang-tidy` as a static analyzer


jhdcs:
## Description

Clang Tidy is a static analysis tool that provides a wide variety of checks on C++ and Python codebases (The manual page is [here](https://clang.llvm.org/extra/clang-tidy/ ))

## Completion Criteria
Either we include Clang Tidy as a static analyzer, or reject this issue

## Implementation Notes / Suggestions

It would probably be worthwhile to run Clang Tidy on the codebase outside of CI until we are able to find a set of configurations that we feel fits the project. There are a lot of options, so this may take some time.

PR #784, opened 2021-06-18
[WIP] Testing adding clang-tidy to github actions

jhdcs:
This relates to issue #776. Just doing some testing as to what all would be involved, and would show up.

Distro A, OPSEC #4584

Signed-off-by: Jacob Hassold <jhassold@dcscorp.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [cppcheck, cpplint, uncrustify]
+          linter: [clang-tidy, cppcheck, cpplint, uncrustify]
     steps:
     - uses: actions/checkout@v2
     - uses: ros-tooling/action-ros-lint@v0.1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-          linter: [cppcheck, cpplint, uncrustify]
+          linter: [clang_tidy, cppcheck, cpplint, uncrustify]
     steps:
     - uses: actions/checkout@v2
     - uses: ros-tooling/action-ros-lint@v0.1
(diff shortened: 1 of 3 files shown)


jhdcs:
Strange. All the checks disappeared.

jhdcs:
Or not. Just the UI being strange.

emersonknapp:
Hm - maybe clang-tidy has to be run post-compile? Maybe running via action-ros-lint won't work then. Worth trying to enable it via a package CMake and see the result in the `Test rosbag2` job

jhdcs:
Ah, yes, it needs a `compile_commands.json` file to have been generated (which we can get via `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`). I'll try and add in those modifications and see where this leaves us.

jhdcs:
[clang_tidy_output.txt](https://github.com/ros2/rosbag2/files/6696818/clang_tidy_output.txt)

For our information, this is what Clang-Tidy is returning when I run it locally.

MichaelOrlov:
FYI. `Clang-Tidy` built-in in `CLion` IDE and highlights all these findings by the red color by default and even sometimes suggesting quick fixes or giving extra info about issue. Here is an example:
Screenshot from 2021-06-22 14-43-55
However very often warnings from Clang-Tidy not critical and code written as it was designed. 

You can try to run `Clang-Tidy` with very simple example when passing shared pointer in function as argument by value - it will complain that it should be passed by const reference. Which is odd and not intent to be by design.

jhdcs:
> FYI. `Clang-Tidy` built-in in `CLion` IDE and highlights all these findings by the red color by default and even sometimes suggesting quick fixes or giving extra info about issue.

Ah! That's good to know! Yes, people should be able to check and see if their preferred IDE has options for using Clang-Tidy to lint their code as they go along.


> However very often warnings from Clang-Tidy not critical and code written as it was designed.

This is common with static analysis tools - Their priority is to make sure that they don't miss issues, at the expense of sometimes flagging non-issues. Usually the tools have ways that you can add a comment to your code to suppress the warning, or the tool can be configured to turn off certain checks (if you are _absolutely_ certain that the check in question is not applicable to your codebase).

The results I generated above were using the default configuration of [`ament_clang_tidy`](https://github.com/ament/ament_lint/tree/master/ament_cmake_clang_tidy) for ROS 2.

emersonknapp:
I notice that a significant amount of the output is from generated code in `build/` - is there a way to only run on the "written" code in `src/`?

I think one safe one to disable would be `google-runtime-references` - I can't even find this rule in https://google.github.io/styleguide/cppguide.html#Reference_Arguments and it rubbed me the wrong way. There are lots of perfectly good reasons to have a non-const reference parameter - such as for output parameters.

jhdcs:
> I notice that a significant amount of the output is from generated code in `build/` - is there a way to only run on the "written" code in `src/`?

`Clang-Tidy` is a bit of an odd duck, it seems - it requires the generation of a `compile_commands.json` file (which gets placed in the build folder) and does all of its checks based on that. However, I _thought_ that it should be only showing output for the `/src` directory... Then again, I also thought it would only be showing warnings, rather than all these little notes...

In any case, I will look into this...

> I think one safe one to disable would be `google-runtime-references` - I can't even find this rule in https://google.github.io/styleguide/cppguide.html#Reference_Arguments any more (was it removed?) and it rubbed me the wrong way. There are lots of perfectly good reasons to have a non-const reference parameter - such as for output parameters.
> 
> Edit: haha yup Google came around and removed the ban on mutable references about a year ago https://github.com/google/styleguide/pull/553/files

This has been disabled (hopefully), doing some testing now.

emersonknapp:
I'm going to close this as stale, please feel free to reopen if new work is done.

Issue #787, opened 2021-06-22
ros2 bag record exit code seems to be 2 when there's no error


russkel:
## Description
I noticed at the exit end of my launch logs any rosbag2 recording sessions would report an ERROR. It looks like `ros2 bag record` returns 2 even if it was successful. 

```
^C[WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
[rosbag-3] [INFO] [1624258420.697845659] [rclcpp]: signal_handler(signal_value=2)
[ERROR] [rosbag-3]: process has died [pid 26949, exit code 2, cmd 'ros2 bag record -a -o /ros2_ws/rosbag'].
```

### Expected Behavior
Exit code of ros2 bag record command should be 0.

### Actual Behavior
```
ros@russ:/ros2_ws$ ros2 bag record -a -o rosbag2
[INFO] [1624329432.042607717] [rosbag2_storage]: Opened database '/ros2_ws/rosbag2/rosbag2_0.db3' for READ_WRITE.
[INFO] [1624329432.042690322] [rosbag2_recorder]: Listening for topics...
[INFO] [1624329432.043308341] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1624329432.043664406] [rosbag2_recorder]: Subscribed to topic '/parameter_events'
^C[INFO] [1624329435.136884441] [rclcpp]: signal_handler(signal_value=2)
[INFO] [1624329435.155487408] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
ros@russ:/ros2_ws$ echo $?
2
```

## To Reproduce
As per Actual Behaviour

## System (please complete the following information)
 - OS: focal
 - ROS 2 Distro: galactic
 - Version: ros-galactic-ros2bag/focal,now 0.9.0-1focal.20210521.032102 amd64 [installed,automatic]

emersonknapp:
Fixed in #788

russkel:
@emersonknapp, scholar and a gentleman. Thanks.

PR #788, opened 2021-06-22
Avoid passing exception KeyboardInterrupt to the upper layer

Barry-Xu-2018:
Address #787

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -204,7 +204,11 @@ class RecordVerb(VerbExtension):
         record_options.include_hidden_topics = args.include_hidden_topics
 
         recorder = Recorder()
-        recorder.record(storage_options, record_options)
+
+        try:
+            recorder.record(storage_options, record_options)
+        except KeyboardInterrupt:
+            pass
 
         if os.path.isdir(uri) and not os.listdir(uri):
             os.rmdir(uri)
--- a/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_unix.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_unix.hpp
@@ -66,7 +66,7 @@ void stop_execution(const ProcessHandle & handle)
   waitpid(handle, &child_return_code, 0);
   // this call will make sure that the process does execute without issues before it is killed by
   // the user in the test or, in case it runs until completion, that it has correctly executed.
-  EXPECT_THAT(WEXITSTATUS(child_return_code), Not(Eq(EXIT_FAILURE)));
+  EXPECT_THAT(WEXITSTATUS(child_return_code), EXIT_SUCCESS);
 }
 
 #endif  // ROSBAG2_TEST_COMMON__PROCESS_EXECUTION_HELPERS_UNIX_HPP_
--- a/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
+++ b/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
@@ -122,14 +122,17 @@ public:
     const auto database_path = get_bag_file_path(0);
 
     while (true) {
-      std::this_thread::sleep_for(50ms);
-      if (database_path.exists()) {
-        break;
+      try {
+        std::this_thread::sleep_for(50ms);  // wait a bit to not query constantly
+        if (database_path.exists()) {
+          rosbag2_storage_plugins::SqliteWrapper db{
+            database_path.string(), rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY};
+          return;
+        }
+      } catch (const rosbag2_storage_plugins::SqliteException & ex) {
+        (void) ex;  // still waiting
       }
     }
-
-    // Wait a while for topic discovered to avoid missing published messages.
-    std::this_thread::sleep_for(200ms);
   }
 
   size_t count_stored_messages(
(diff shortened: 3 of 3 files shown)


emersonknapp (approved):
looks good, thanks. It might be nice to have a regression test for this, if you could add that

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/31a0f4acbf7a9178dfbedb80dccd2d01/raw/7573fb48ec7c6d3994fa585d98eb137bf3928e3b/ros2.repos
BUILD args: --packages-up-to ros2bag
TEST args: --packages-select ros2bag
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8611
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14796/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9545/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12462/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14943/)

Barry-Xu-2018:
@emersonknapp  

Thanks for the review.  

> looks good, thanks. It might be nice to have a regression test for this, if you could add that

I find there is a check.  
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_unix.hpp#L67-L69

But it checks if return is EXIT_FAILURE(1).   
As you known, original codes return 2 (reference below code)
https://github.com/ros2/ros2cli/blob/aaeda774f6e8cc4fe43c97888c5fd633e010a56a/ros2cli/ros2cli/cli.py#L66-L70

So the check always pass. This is incorrect behavior.   
So I modify this check as https://github.com/ros2/rosbag2/pull/788/commits/38fbbac8898728dbdf9394678d1c80cbe1d42a1b.

Barry-Xu-2018:
> Rpr__rosbag2__ubuntu_focal_amd64  --- failure
According to https://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/881/testReport/, 3 test case failed.  

The failure reason includes 2 points. 
1. While publishing the message, sometimes rosbag doesn't discover the topic to be recorded. 
e.g. 
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L121-L129

This leads to the error as the below (The number of publishing message doesn't match the number of recorded message)
https://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/881/testReport/rosbag2_tests/RecordFixture/record_end_to_end_test_with_cache/

Now we have no interface to make sure rosbag finish discovering topic, so I want to add sleep before publishing message to make sure discover is finished.

2. It is possible that  2 processes access the created sqlite3 database at the same time. 
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L125-L127

start_execution() will start rosbag command in new process.  
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp#L120-L131
As the above codes, current process also access sqlite3 database.   
This leads that the lock occurs. 
```
[ RUN      ] RecordFixture.rosbag2_record_and_play_multiple_topics_with_filter
[INFO] [1624441972.716871010] [rosbag2_storage]: Opened database '/tmp/tmp_test_dir_OVIieg/rosbag2_record_and_play_multiple_topics_with_filter/rosbag2_record_and_play_multiple_topics_with_filter_0.db3' for READ_WRITE.
[INFO] [1624441972.716994585] [rosbag2_recorder]: Listening for topics...
[INFO] [1624441972.718430617] [rosbag2_recorder]: Subscribed to topic '/test_topic1'
[INFO] [1624441972.719189448] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
Error when processing SQL statement. SQLite error (5): database is locked
```

Barry-Xu-2018:
@emersonknapp 

I made a fix and want to discuss with you on how to fix.  
So please take a look https://github.com/ros2/rosbag2/pull/788/commits/faa17439b8891609d2a71a156556f75e34f4ffeb.

emersonknapp:
Thanks for analyzing the flaky test!

I don't like there being an arbitrary sleep for the OK case. I think we can have timeouts, but much of the flaky behavior comes from these sleep statements - I don't know if the fix you have proposed is appropriate.

However, your fix for the exit code is good, so I would prefer to break out the flaky-test fix into a separate PR, and move this fix forward independently since it is definitely OK.

Barry-Xu-2018:
@emersonknapp 

> I don't like there being an arbitrary sleep for the OK case. I think we can have timeouts, but much of the flaky behavior comes from these sleep statements - I don't know if the fix you have proposed is appropriate.

Yeah. I also don't want to use sleep.   
I will create another issue to discuss how to fix and we can continue to discuss it.  

> However, your fix for the exit code is good, so I would prefer to break out the flaky-test fix into a separate PR, and move this fix forward independently since it is definitely OK.

Okay.  
I will new commit to revert the latest commit https://github.com/ros2/rosbag2/commit/faa17439b8891609d2a71a156556f75e34f4ffeb

emersonknapp (approved):

Issue #786, opened 2021-06-22
Delay option for ros2 bag play


hayato-m126:
## Feature request

Add a commandline option `--delay`, which sleeps for the specified amount of time before starting to publish messages.

This delay should also happen in between loops, when looping playback.

## Original Ticket

I want to use the delay option of bag play that was implemented in ROS1, but it seems that it is not implemented in ROS2.
Is there any reason why it is not implemented in ROS2?

TierIV implemented the delay option in the forked repository below, if we submit a pull request to this official repository, will it be accepted?

https://github.com/tier4/rosbag2/pull/3/files

emersonknapp:
> Is there any reason why it is not implemented in ROS2?

No reason, just hasn't been built

> if we submit a pull request to this official repository, will it be accepted?

Assuming we iron out anything that comes up during code review, yes!

Note: i've edited your original ticket (leaving all the original context) to lay out the feature request. Please check that it matches what you have in mind.

hayato-m126:
Thank you for reply.
The edited issue is as I intended.

MichaelOrlov:
@hayato-m126 Could you please clarify what a use cases for usage this `delay` API?

hayato-m126:
@MichaelOrlov 
The delay option is assumed to be used when calling ros bag play from ros launch.
When ros bag play is launched with multiple nodes such as sensing, percepiton, localization, etc., sometimes data be output before each node is fully initialized, so this option is used to delay data playback.

MichaelOrlov:
@hayato-m126 I would ecomend to implement `wait_for_matched` API for Player class instead for your cases.

hayato-m126:
@MichaelOrlov 
Do you mean to implement an option to wait for playback based on generic conditions, not just time?
If so, what are the possible use cases for using conditions other than time?

Since delay is an option that existed in ROS1, Tier4 simply implemented the same functionality in the pull request associated with this issue.

MichaelOrlov:
@hayato-m126 I understand that you are trying to repeat functionality from ROS1.
However there are exists possibility to wait for matched number of subscribers for publisher rather than just blind delay.
i.e. Player will not start playback until publisher will will discover number of specified subscribers. It means until subscribers nodes will be initialized and visible for publisher via discovery.
[...]

PR #789, opened 2021-06-23
Add delay option

kosuke55:
Hi, thank you for your work.
In this PR I added `--delay` option related to https://github.com/ros2/rosbag2/issues/786.
If we run for the following, the rosbag will play after a 5 second sleep.
```
ros2 bag play <rosbag2 file> -d 5
```
It also works with `--loop` like
```
ros2 bag play <rosbag2 file> -d 5 -l
```

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -119,7 +119,7 @@ class PlayVerb(VerbExtension):
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
-        play_options.delay=args.delay
+        play_options.delay = args.delay
 
         player = Player()
         player.play(storage_options, play_options)
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -84,7 +84,7 @@ class PlayVerb(VerbExtension):
                  'Value must be positive. Defaults to not publishing.')
         parser.add_argument(
             '-d', '--delay', type=float, default=0.0,
-            help='Sleep SEC seconds after play. Valid range > 0.0')
+            help='Sleep SEC seconds before play. Valid range > 0.0')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -82,6 +82,9 @@ class PlayVerb(VerbExtension):
             '--clock', type=positive_float, nargs='?', const=40, default=0,
             help='Publish to /clock at a specific frequency in Hz, to act as a ROS Time Source. '
                  'Value must be positive. Defaults to not publishing.')
+        parser.add_argument(
+            '-d', '--delay', type=float, default=0.0,
+            help='Sleep SEC seconds after play. Valid range > 0.0')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -116,6 +119,7 @@ class PlayVerb(VerbExtension):
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
+        play_options.delay=args.delay
 
         player = Player()
         player.play(storage_options, play_options)
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -44,6 +44,8 @@ public:
   // Rate in Hz at which to publish to /clock.
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
+
+  float delay = 0.0;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -45,7 +45,7 @@ public:
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
 
-  // Sleep SEC seconds after play. Valid range > 0.0.
+  // Sleep SEC seconds before play. Valid range > 0.0.
   float delay = 0.0;
 };
 
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -45,6 +45,7 @@ public:
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
 
+  // Sleep SEC seconds after play. Valid range > 0.0.
   float delay = 0.0;
 };
 
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -215,7 +215,7 @@ void Player::play()
   is_in_play_ = true;
 
   float delay;
-  if (play_options_.delay > 0.0) {
+  if (play_options_.delay >= 0.0) {
     delay = play_options_.delay;
   } else {
     RCLCPP_WARN(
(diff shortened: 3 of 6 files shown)


rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -44,6 +44,8 @@ struct PlayOptions
   // Rate in Hz at which to publish to /clock.
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
+
```
zmichaels11, line 47: Can you add a comment here stating what `delay` is and the unit?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -215,6 +215,13 @@ void Player::play()
   is_in_play_ = true;
   try {
     do {
+      float delay =
```
zmichaels11, line 218: Can you move `delay` out of the loop? I don't think it can change between iterations.

I'd also suggest raising a warning or exception when delay is an invalid value. That way the user knows when they enter an invalid value.

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
```diff
@@ -43,6 +43,8 @@ TEST_F(RosBag2PlayTestFixture, play_bag_file_twice) {
   const size_t read_ahead_queue_size = 1000;
   const float rate = 1.0;
   const bool loop_playback = false;
+  double clock_publish_frequency = 0.0;
+  const float delay = 1.0;
```
zmichaels11, line 47: Can you also test for an invalid delay value?

zmichaels11 (commented):
Thanks for your PR!

I left a couple comments/suggestions.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
kosuke55: I added in https://github.com/ros2/rosbag2/pull/789/commits/d10ee937e26570f29298545e446e51ebd6bfd24d.

rosbag2_transport/src/rosbag2_transport/player.cpp
kosuke55: I moved it out of the loop. And raising warning and set to 0 if dealy value is invalid. https://github.com/ros2/rosbag2/pull/789/commits/0b4687f53c8c3dde2fa69e5235137a66ff390873

danthony06:
I know at one point there was a plan to implement node lifecycles for rosbag. Is this still the case, because this seems like a case where it would fit in really well.

rosbag2_transport/test/rosbag2_transport/test_play_loop.cpp
kosuke55: I added tests for delay in https://github.com/ros2/rosbag2/pull/789/commits/dfee76ad21b4e396eef0e8d2f12bd1f8b084ea4e.

kosuke55:
@zmichaels11 
Thaks for your review. I modified so please check them.

zmichaels11 (approved):

ros2bag/ros2bag/verb/play.py
```diff
@@ -116,6 +119,7 @@ def main(self, *, args):  # noqa: D102
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
+        play_options.delay=args.delay
```
emersonknapp, line 122: linter is failing on this line - it will help you to run the tests locally before putting up for review!

emersonknapp:
> I know at one point there was a plan to implement node lifecycles for rosbag. Is this still the case, because this seems like a case where it would fit in really well.

There isn't an active effort going on - I don't personally have a concrete idea what this would look like. We could take that discussion to another issue if we want to pick it up. With some quick diagrams or something showing what we would want.

ros2bag/ros2bag/verb/play.py
kosuke55: Thaks for your review and advice. I modified it and checked that `test_flake8.py` paased. https://github.com/ros2/rosbag2/pull/789/commits/1c0cbffc9815858a106474661008c53df884312b

emersonknapp:
Looking very close - you will need to sign your commits for the DCO check to pass. You can do that with `git rebase master --signoff`

kosuke55:
@emersonknapp 
Oh, sorry and thaks. I did that.

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/df22ce71e23066faa49f1ad4b777faee/raw/e66eb02dae79a8dbe4184fb2c6bf28be493f1741/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_py rosbag2_transport rosbag2 rosbag2_tests
TEST args: --packages-select ros2bag rosbag2_py rosbag2_transport rosbag2 rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8646
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14831/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9581/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12497/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14988/)

emersonknapp:
The Windows runners are very slow - and so timing-based tests don't always work the way you would expect there. Is there any deterministic way to check on this test, without depending on actual timing? (worst case, we could just expand the error margin to accommodate these slow runners)

kosuke55:
@emersonknapp 
hmm, I can't think of any other way than using the actual time.
This test was made with reference to other parts. Other tests have a fairly large margin .
https://github.com/ros2/rosbag2/blob/c8b10325f1a43b25a47f207cd969c430240ba424/rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp#L122-L123
I don't think there are many cases where we actually want to delay by 0.x seconds, so how about simply increasing the margin (1.0, etc.)?

kosuke55:
And this test is included in the rate test, but it's better to separate them.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -44,6 +44,9 @@ struct PlayOptions
...
+  // Sleep SEC seconds after play. Valid range > 0.0.
```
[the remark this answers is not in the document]
kosuke55: Oh, thanks that't better and fixed it https://github.com/ros2/rosbag2/pull/789/commits/f7d5d609e128e1be89a978c4fc7c4d2170f44777. (I meant "after running play command", but it's confusing)

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -191,3 +191,64 @@ TEST_F(PlayerTestFixture, playing_respects_rate)
...
+  {
+    play_options_.delay = -5.0;
+    std::chrono::duration<float> delay_uppper(delay_margin);
+
+    auto prepared_mock_reader = std::make_unique<MockSequentialReader>();
+    prepared_mock_reader->prepare(messages, topics_and_types);
+    auto reader = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
+    auto player = std::make_shared<rosbag2_transport::Player>(
+      std::move(
+        reader), storage_options_, play_options_);
```
MichaelOrlov, line 246: Nitpick. It would be better if 
```cpp
    std::move(
        reader), storage_options_, play_options_);
```
would be on one line
```cpp
    std::move(reader), storage_options_, play_options_);
```

kosuke55: OK, I modified all same part in https://github.com/ros2/rosbag2/pull/789/commits/75eca9f527e9a5fcba9d88af42420715d8f13a07.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14841/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9591/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12508/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14998/)

kosuke55:
@emersonknapp 
Is there anything else I need to fix in this code? If not, I would like to send a same PR for the galactic branch as well.

emersonknapp:
No - nothing left. I just didn't look at it over the holiday weekend. Please feel free to open a Galactic backport PR

Issue #790, opened 2021-06-24
Flaky test in test_rosbag2_record_end_to_end


Barry-Xu-2018:
## Description
test cases for test_rosbag2_record_end_to_end sometimes fails

### Expected Behavior
Get stabilization result

### Actual Behavior
Sometimes failure

## To Reproduce
Repeat run test_rosbag2_record_end_to_end in rosbags_test  
colcon test --event-handlers console_direct+ --packages-select rosbag2_tests --ctest-args -R test_rosbag2_record_end_to_end --repeat-until-fail 1000

## System (please complete the following information)
 - OS: Ubuntu 20.04
 - ROS 2 Distro: rolling
 - Version: master

## Additional context
This problem also occurs in CI system

Barry-Xu-2018:
This problem was found at https://github.com/ros2/rosbag2/pull/788#issuecomment-866715732.  
And then I can reproduce this in my local environment if execute enough times.

Barry-Xu-2018:
I have analyzed problems at https://github.com/ros2/rosbag2/pull/788#issuecomment-866715732, but the solution need to be discussed here.  
So I copy my analysis to here.

---
I find CI report failure.  
> Rpr__rosbag2__ubuntu_focal_amd64  --- failure
According to https://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/881/testReport/, 3 test case failed.  

The failure reason includes 2 points. 
1. While publishing the message, sometimes rosbag doesn't discover the topic to be recorded. 
e.g. 
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L121-L129

This leads to the error as the below (The number of publishing message doesn't match the number of recorded message)
https://build.ros2.org/job/Rpr__rosbag2__ubuntu_focal_amd64/881/testReport/rosbag2_tests/RecordFixture/record_end_to_end_test_with_cache/

Now we have no interface to make sure rosbag finish discovering topic, so I want to add sleep before publishing message to make sure discover is finished.

2. It is possible that  2 processes access the created sqlite3 database at the same time. 
https://github.com/ros2/rosbag2/blob/6f7503eafede497bda4f04c8e4b2de798b9cca99/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L125-L127

[...]

PR #797, opened 2021-06-28
Add spin_and_wait_for_matched to PublicationManager and update test c…

Barry-Xu-2018:
Address #790

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
@@ -15,6 +15,7 @@
 #ifndef ROSBAG2_TEST_COMMON__PUBLICATION_MANAGER_HPP_
 #define ROSBAG2_TEST_COMMON__PUBLICATION_MANAGER_HPP_
 
+#include <cstring>
 #include <functional>
 #include <future>
 #include <memory>
@@ -58,7 +59,9 @@ public:
   {
     auto publisher = pub_node_->create_publisher<MsgT>(topic_name, qos);
 
-    publishers_.push_back(
+    publishers_.push_back(publisher);
+
+    publishers_fcn_.push_back(
       [publisher, message, repetition, interval](bool verbose = false) {
         for (auto i = 0u; i < repetition; ++i) {
           if (rclcpp::ok()) {
@@ -76,15 +79,18 @@ public:
 
   void run_publishers(bool verbose = false) const
   {
-    for (const auto & pub_fcn : publishers_) {
+    for (const auto & pub_fcn : publishers_fcn_) {
       pub_fcn(verbose);
     }
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.spin_node_some(pub_node_);
+    std::this_thread::sleep_for(100ms);
   }
 
   void run_publishers_async(bool verbose = false) const
   {
     std::vector<std::future<void>> futures;
-    for (const auto & pub_fcn : publishers_) {
+    for (const auto & pub_fcn : publishers_fcn_) {
       futures.push_back(std::async(std::launch::async, pub_fcn, verbose));
     }
     for (auto & publisher_future : futures) {
@@ -92,10 +98,59 @@ public:
     }
   }
 
+  /// \brief Wait until publisher with specified topic will be connected to the subscribers or
+  ///    timeout occur.
+  /// \tparam Timeout Data type for timeout duration from std::chrono:: namespace
+  /// \param topic_name topic name to find corresponding publisher
+  /// \param timeout Maximum time duration during which discovery should happen.
+  /// \param n_subscribers_to_match Number of subscribers publisher should have for match.
+  /// \return true if not find publisher by topic name or publisher has specified number of
+  ///    subscribers, otherwise false.
+  template<typename Timeout>
+  bool spin_and_wait_for_matched(
+    const char * topic_name,
+    Timeout timeout, size_t n_subscribers_to_match = 1)
+  {
+    rclcpp::PublisherBase::SharedPtr publisher = nullptr;
+    for (const auto pub : publishers_) {
+      if (!std::strcmp(pub->get_topic_name(), topic_name)) {
+        publisher = pub;
+        break;
+      }
+    }
+
+    if (publisher == nullptr) {
+      return false;
+    }
+
+    if (publisher->get_subscription_count() +
+      publisher->get_intra_process_subscription_count() >= n_subscribers_to_match)
+    {
+      return true;
+    }
+
+    using clock = std::chrono::steady_clock;
+    auto start = clock::now();
+
+    rclcpp::executors::SingleThreadedExecutor exec;
+    do {
+      exec.spin_node_some(pub_node_);
+
+      if (publisher->get_subscription_count() +
+        publisher->get_intra_process_subscription_count() >= n_subscribers_to_match)
+      {
+        return true;
+      }
+    } while ((clock::now() - start) < timeout);
+
+    return false;
+  }
+
 private:
   std::shared_ptr<rclcpp::Node> pub_node_;
   using PublicationF = std::function<void (bool)>;
-  std::vector<PublicationF> publishers_;
+  std::vector<PublicationF> publishers_fcn_;
+  std::vector<rclcpp::PublisherBase::SharedPtr> publishers_;
 };
 
 }  // namespace rosbag2_test_common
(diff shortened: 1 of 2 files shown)


MichaelOrlov:
@Barry-Xu-2018 I am kind of confused about your implementation and this is not what I expected to see.
Why do you need to spin node for PublicationManager?
It does make sense to spin node only for subscription.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -76,26 +79,78 @@ class PublicationManager
...
+    } while ((clock::now() - start) < timeout);
+
+    return false;
+  }
+
 private:
   std::shared_ptr<rclcpp::Node> pub_node_;
   using PublicationF = std::function<void (bool)>;
-  std::vector<PublicationF> publishers_;
+  std::vector<PublicationF> publishers_fcn_;
```
emersonknapp, line 152: Is there any reason why we should maintain two separate vectors? I would prefer 

```
std::vector<std::pair<rclcpp::PublisherBase::SharedPtr, PublicationF>> publishers_;
```

This way it's impossible for the vectors to get out of sync.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -76,26 +79,78 @@ class PublicationManager
 
   void run_publishers(bool verbose = false) const
   {
-    for (const auto & pub_fcn : publishers_) {
+    for (const auto & pub_fcn : publishers_fcn_) {
       pub_fcn(verbose);
     }
+    rclcpp::executors::SingleThreadedExecutor exec;
+    exec.spin_node_some(pub_node_);
```
emersonknapp, line 86: Is spinning necessary for publishing?

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -76,26 +79,78 @@ class PublicationManager
...
+    std::this_thread::sleep_for(100ms);
```
emersonknapp, line 87: Is 100ms meaningful? Where does this number come from? Why do we need to sleep? Does this make the publishing "bursty"?

Barry-Xu-2018:
@MichaelOrlov 

> I am kind of confused about your implementation and this is not what I expected to see.
Why do you need to spin node for PublicationManager?
It does make sense to spin node only for subscription.

According to your suggestion in https://github.com/ros2/rosbag2/issues/790#issuecomment-868945959
```
For the Problem 1 related to the discovery it would be more appropriate to add wait_for_matched(..) API for the PublicationManager. Similar as made spin_and_wait_for_matched in SubscriptionManager.
```
My understanding is to add similar API like spin_and_wait_for_matched in SubscriptionManager to PublicationManager.  
Spin node is to confirm the number of connected subscription (from rosbag) for specified publisher.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Barry-Xu-2018: Not necessary.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Barry-Xu-2018: Not necessary. I forgot to remove them.

MichaelOrlov:
@Barry-Xu-2018 It seems you misinterpreted my suggestion. I intentionally mentioned `wait_for_matched(..)` API without word spin. Also mentioned **`similar`** as `spin_and_wait_for_matched(..)`. I meant that `similar` doesn't means "the same as".

Anyway, thank you for your PR, I will take a look more closely on it since you have addressed some review comments.

Barry-Xu-2018:
@MichaelOrlov 

> It seems you misinterpreted my suggestion. I intentionally mentioned wait_for_matched(..) API without word spin. Also mentioned similar as spin_and_wait_for_matched(..). I meant that similar doesn't means "the same as".

Sorry for my misunderstanding.   
Spin is  unnecessary.

> Anyway, thank you for your PR, I will take a look more closely on it since you have addressed some review comments.

Please wait a moment for the review.   
I will update codes to remove spin.

Barry-Xu-2018:
@MichaelOrlov @emersonknapp 

Addressed your comments.   
Please review again.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -71,31 +72,90 @@ class PublicationManager
...
+    auto sleep_time = std::chrono::milliseconds(100);
+
+    if (timeout < std::chrono::seconds(1)) {
+      sleep_time = timeout;
+    }
+
+    using clock = std::chrono::steady_clock;
+    auto start = clock::now();
+
+    rclcpp::executors::SingleThreadedExecutor exec;
```
MichaelOrlov, line 141: It seems forgot to remove 
```
```

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -71,31 +72,90 @@ class PublicationManager
...
+      return false;
+    }
+
+    if (publisher->get_subscription_count() +
+      publisher->get_intra_process_subscription_count() >= n_subscribers_to_match)
+    {
+      return true;
+    }
+
+    auto sleep_time = std::chrono::milliseconds(100);
```
MichaelOrlov, line 132: I would recommend to reduce default sleep timeout. I would chose some number from 1 to 10 mSec.

MichaelOrlov (commented):
@Barry-Xu-2018 LGTM among a few nitpicks.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Barry-Xu-2018: Sorry. My mistake.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Barry-Xu-2018: Okay. I change to use 10ms as default.

MichaelOrlov (approved):
@Barry-Xu-2018 All looks good now.
Thank you for your contribution.

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
```diff
@@ -71,31 +72,89 @@ class PublicationManager
...
   }
 
+  /// \brief Wait until publisher with specified topic will be connected to the subscribers or
+  ///    timeout occur.
+  /// \tparam Timeout Data type for timeout duration from std::chrono:: namespace
+  /// \param topic_name topic name to find corresponding publisher
+  /// \param timeout Maximum time duration during which discovery should happen.
+  /// \param n_subscribers_to_match Number of subscribers publisher should have for match.
+  /// \return true if not find publisher by topic name or publisher has specified number of
+  ///    subscribers, otherwise false.
```
MichaelOrlov, line 105: Ups. I overlooked mistake in description for return value. It should be other way around. i.e. return false if fail and true if success. 
@Barry-Xu-2018 Could you please fix it before merge?

MichaelOrlov (changes requested):

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/8f92e1829d0071ceb0844bff7ad53c52/raw/e7160639b7b34096e84da97c39d90e499def5c70/ros2.repos

BUILD args: --packages-up-to rosbag2_test_common  rosbag2_tests
TEST args: --packages-select rosbag2_test_common  rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8641/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14826/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12492/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14981/)

rosbag2_test_common/include/rosbag2_test_common/publication_manager.hpp
Barry-Xu-2018: Thank you.  
The description about return value is incorrect.

MichaelOrlov (approved):

Issue #795, opened 2021-06-25
Clang Thread Safety Analysis warnings in time_controller_clock.cpp


sloretz:
## Description

The latest [`nightly_linux_clang_libcxx`](https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/) job shows new thread safety analysis warnings.

### Expected Behavior
No thread safety analysis warnings

### Actual Behavior
Clang reports thread safety analysis warnings

## To Reproduce
1. Go to https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/829/clang-tidy/new/
2. Scroll down to the 'Issues' at the bottom of the page
3. Expand all the `time_controller_clock.cpp` warnings

## System (please complete the following information)
 - OS: Ubuntu Focal
 - ROS 2 Distro: Latest development
 - Version: `master` / bb9c7ae783d740cf7fd47a3a10a0b4ff6dcdc1a5

## Additional context
All the offending code appears to come from this commit https://github.com/ros2/rosbag2/commit/6f7503eafede497bda4f04c8e4b2de798b9cca99 #779

```
[ 17%] Building CXX object CMakeFiles/rosbag2_cpp.dir/src/rosbag2_cpp/clocks/time_controller_clock.cpp.o
/home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp:143:5: warning: reading variable 'cv' requires holding mutex 'state_mutex' [-Wthread-safety-analysis]
    cv.notify_all();
    ^
/home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp:202:7: warning: calling function 'process_callback' requires holding mutex 'callback_list_mutex_' exclusively [-Wthread-safety-analysis]
      process_callback(handler, time_jump, true);
      ^
/home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros2/rosbag2/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp:210:7: warning: calling function 'process_callback' requires holding mutex 'callback_list_mutex_' exclusively [-Wthread-safety-analysis]
      process_callback(handler, time_jump, false);
```

MichaelOrlov:
[...]

PR #799, opened 2021-06-30
Workaround for false positive findings by clang thread safety analysi…

MichaelOrlov:
Workaround for false positive findings by clang thread safety analysis in time controller jump callbacks API.
Address #795 

Moved cv.notify_all() inside mutex lock scope, also made method process_callback() as non-static

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
@@ -138,9 +138,9 @@ public:
     {
       std::lock_guard<std::mutex> lock(state_mutex);
       snapshot(ros_time);
+      cv.notify_all();
     }
     process_callbacks_after_jump(time_jump);
-    cv.notify_all();
   }
 
   /**
@@ -211,9 +211,9 @@ private:
     }
   }
 
-  static void process_callback(
+  void process_callback(
     PlayerClock::JumpHandler::SharedPtr handler, const rcl_time_jump_t & time_jump,
-    bool before_jump) RCPPUTILS_TSA_REQUIRES(callback_list_mutex_)
+    bool before_jump) const RCPPUTILS_TSA_REQUIRES(callback_list_mutex_)
   {
     bool is_clock_change = time_jump.clock_change == RCL_ROS_TIME_ACTIVATED ||
       time_jump.clock_change == RCL_ROS_TIME_DEACTIVATED;


MichaelOrlov:
Gist:
https://gist.githubusercontent.com/MichaelOrlov/5fd70f1953b536179399721f25846a73/raw/fc9b346ecfc7d32e68197e58608448b6d2f3a1a9/ros2.repos

BUILD args: --packages-up-to rosbag2_cpp
TEST args: --packages-select rosbag2_cpp
Job: ci_launcher
ci_linux_clang_libcxx ran: https://ci.ros2.org/job/ci_linux_clang_libcxx/5/
[Build Status](https://ci.ros2.org/job/ci_linux_clang_libcxx/5/)

emersonknapp (approved):
Looks good - however I will say that these are _not_ false positives.

* `cv` was used without the mutex
* A static function does not have access to non-static member variables. The declaration of `static void REQUIRES` was ill-formed.

MichaelOrlov:
@emersonknapp As regards to the 
> - cv was used without the mutex

`cv.notify_all()` should be used when mutex already unlocked. According to the https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all
> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock.

Implemented workaround putting `cv.notify_all()` under the mutex only works due to the specific optimization implemented in GCC. Tomorrow new type or version of compiler will be used or with some flags turning off this optimization and we will get random errors with missed notifications. I am actually not happy with such type of workarounds.  

As regards to the:
> A static function does not have access to non-static member variables. The declaration of static void REQUIRES(member_) was ill-formed for requiring a non-static variable.

Not sure if I understand you correctly but `process_callback()` function doesn't have access to the member variables of the class and purely operates with input parameters.

MichaelOrlov:
@emersonknapp Oh. BTW fun fact. 
I didn't mind to declare `process_callback()` as static from the very beginning, but `clang-tidy` gave me a warning that this function could be declared as static and proposed to put static keyword for it. I reviewed this function and come to the conclusion so well why not. 

Now we have contradiction between two tools `clang-tidy` and `clang thread safety check` :)

emersonknapp:
> cv.notify_all() should be used when mutex already unlocked

according to the documented usage hints. But, we explicitly said in the code that `cv REQUIRES(mutex)` - so the TSA is doing exactly what we told it to. It's not a false positive - it's an accurate result of the code that has been written.

> Not sure if I understand you correctly but process_callback() function doesn't have access to the member variables of the class and purely operates with input parameters.

The `static` function was declared `REQUIRES(non_static_member_)` - there is no way for the static function to have access to that non-static-member. The compiler seems to allow this annotation without failing, but it's logically inconsistent so it would always raise a warning that the mutex is not held.

PR #801, opened 2021-06-30
Run xfail tests in a separate step that is not required to pass

emersonknapp:


zmichaels11 (approved):
Is it possible to have variables in the `run` field?
Can you do something like:
```yaml
run: |
  base_paths=$(colcon list -p --packages-select rosbag2)
  packages=$(colcon list -n --base_paths ${base_paths})
  source /opt/ros/rolling/setup.sh && colcon test --packages-select ${packages}
```

jikawa-az (approved):

emersonknapp:
> Is it possible to have variables in the run field?

Yes, it's arbitrary bash and we can do multiline scripts, so I've tweaked it to pull out some of that longer logic

zmichaels11 (approved):

PR #806, opened 2021-07-02
Update Galactic CI to run against APT binaries

emersonknapp:
Don't check out sources since distro should be stable to build against


lihui815 (approved):
lgtm

PR #767, opened 2021-05-11
Expose play_next() API via service

MichaelOrlov:
Expose play_next() API via service

Part of the #696

Update: Rebased on top of master

PR #810, opened 2021-07-05
Started working on jump service feature

Kaju-Bubanja:
I would like to implement Expose jump services on the Player from #696. I checked how it was done for the set_rate service #767, but the whole package is quite complex and I'm not sure how rosbag2_cpp interacts with rosbag2_transport. I'm not sure if one can be seen as the backend and the other as the frontend and if I should expose the jump service in rosbag2_transport, like the others or if it should reside in rosbag2_cpp like I did now. I currently can't get it to build, but before investing more time, I would be thankful for some advice, if this is the correct package to add the service.

### DIFF
--- a/rosbag2_cpp/CMakeLists.txt
+++ b/rosbag2_cpp/CMakeLists.txt
@@ -51,6 +51,7 @@ find_package(rosidl_runtime_c REQUIRED)
 find_package(rosidl_runtime_cpp REQUIRED)
 find_package(rosidl_typesupport_cpp REQUIRED)
 find_package(rosidl_typesupport_introspection_cpp REQUIRED)
+find_package(rosbag2_interfaces REQUIRED)
 
 add_library(${PROJECT_NAME} SHARED
   src/rosbag2_cpp/cache/cache_consumer.cpp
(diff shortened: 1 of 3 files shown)


emersonknapp:
>  I'm not sure if one can be seen as the backend and the other as the frontend

`rosbag2_cpp` is the generic rosbag2 C++ API, it doesn't interact with "communications", it only deals with the "bag" portion. `rosbag2_transport` is the package that has Publishers and Subscribers, and manages actual "transport" of messages.

>  if I should expose the jump service in rosbag2_transport, like the others

Yes, this service should be with the others, for a cohesive design. It is not good practice to split up related functionality into different places - if we migrate it, it would all be at once.

In this case, though, the service belongs to the `Player`, which owns a `PlayerClock` that controls time. So, the service to interact with this clock will also live on the `Player` in `rosbag2_transport`.

Additionally, ROS 2 Services are a "transport"/"communications" concept, which we want to keep isolated to the `rosbag2_transport` package

MichaelOrlov:
@Kaju-Bubanja @emersonknapp Unfortunately `Player::jump(..)` API have not yet fully implemented and I am afraid it will be a blocker for relevant service implementation.

emersonknapp:
I'll close this for now as stale until the relevant underlying features are ready to expose

PR #814, opened 2021-07-08
Add wait for player to be ready for playback in Player::play_next()

### LINKED ISSUE #813: Need to wait until player will start to be able to use play_next()
## Description
In current implementation to be able to use `play_next()` API need to accomplish following steps:
1. Create instance of the `Player` class.
2. Put player in pause mode i.e. `player->pause();`
3. Start playback in separate thread i.e.
```cpp
 auto player_future = std::async(std::launch::async, [&player]() -> void {player->play();});
```
4. Wait for some time while player will accomplish it's initialization and fill internal queue with messages
5. Start using `player->play_next()` API

For step `4` in our unit tests we are using internal [MockPlayer::wait_for_playback_to_start()](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/test/rosbag2_transport/test_play_next.cpp#L52-L57) API which is available only for `MockPlayer`. 

The problem is that `wait_for_playback_to_start()` API is not publicly available in Player class and potential user of the `play_next()` API will have to put some ambient delay before start calling `play_next()` method. 
If don't wait after spawning thread with playback `Player::play_next()` will be returning false until player will fill it's internal queue with messages. At the same time returning false from `play_next()` also means that we reached end of the bag and nothing to play.
Using some ambient delay before start calling `play_next()` method in user code is non-deterministic and doesn't fit to the expectations from `play_next()` API which was specifically designed for deterministic replay.

To avoid this ambiguous behavior and race conditions with internal queue initialization in player class it would be more reasonable if `Player::play_next()` will wait internally while playback will be started and queue will be filled with messages. In this case we will not need to worry about delay and call for `wait_for_playback_to_start()` method.
This solution also align with the current behavior since we already have constrained for `play_next()` method that [`If internal player queue is starving and storage has not been completely loaded, this method will wait until new element will be pushed to the queue`](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/include/rosbag2_transport/player.hpp#L119-L120)

### Expected Behavior
`Player::play_next()` method should wait until `Player::play(..)` method will be called in the separate thread and internal `Player` queue will be filled with the messages to be ready for playback.

### Actual Behavior
`Player::play_next()` method returning false if `Player::play(..)` hasn't been called yet or if method `play()` haven't reached the point when internal queue filled with messages and player ready to play messages. 

## To Reproduce
1. Create instance of the `Player` class.
2. Put player in pause mode i.e. `player->pause();`
3. Start playback in separate thread i.e.
```cpp
 auto player_future = std::async(std::launch::async, [&player]() -> void {player->play();});
```
5. Start using `player->play_next()` API

`player->play_next()` will return false and message will not be played.

 - OS: [Ubuntu 18.04]
 - ROS 2 Distro: [Rolling]
 - Version: [master branch]

## Additional context
I am going to handle this issue by myself.

MichaelOrlov:
Add wait for player to be ready for playback in `Player::play_next()` method.

Closes #813

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -212,16 +212,12 @@ bool Player::is_storage_completely_loaded() const
 
 void Player::play()
 {
-  is_in_play_ = true;
-
   float delay;
   if (play_options_.delay >= 0.0) {
     delay = play_options_.delay;
   } else {
     RCLCPP_WARN(
-      this->get_logger(),
-      "Invalid delay value: %f. Delay is disabled.",
-      play_options_.delay);
+      this->get_logger(), "Invalid delay value: %f. Delay is disabled.", play_options_.delay);
     delay = 0.0;
   }
 
@@ -237,18 +233,28 @@ void Player::play()
         reader_->get_metadata().starting_time.time_since_epoch()).count();
       clock_->jump(starting_time);
 
-      storage_loading_future_ = std::async(
-        std::launch::async,
-        [this]() {load_storage_content();});
+      storage_loading_future_ = std::async(std::launch::async, [this]() {load_storage_content();});
 
       wait_for_filled_queue();
+      {  // Notify play_next() that we are ready for playback
+        std::lock_guard<std::mutex> lk(ready_to_play_from_queue_mutex_);
+        is_ready_to_play_from_queue_ = true;
+        ready_to_play_from_queue_cv_.notify_all();
+      }
       play_messages_from_queue();
+      {
+        std::lock_guard<std::mutex> lk(ready_to_play_from_queue_mutex_);
+        is_ready_to_play_from_queue_ = false;
+        ready_to_play_from_queue_cv_.notify_all();
+      }
       reader_->close();
     } while (rclcpp::ok() && play_options_.loop);
   } catch (std::runtime_error & e) {
     RCLCPP_ERROR(this->get_logger(), "Failed to play: %s", e.what());
   }
-  is_in_play_ = false;
+  std::lock_guard<std::mutex> lk(ready_to_play_from_queue_mutex_);
+  is_ready_to_play_from_queue_ = false;
+  ready_to_play_from_queue_cv_.notify_all();
 }
 
 void Player::pause()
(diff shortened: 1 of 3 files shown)


rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -299,14 +305,19 @@ rosbag2_storage::SerializedBagMessageSharedPtr * Player::peek_next_message_from_
...
     return false;
   }
 
+  // Temporary take over playback from play_messages_from_queue()
+  std::lock_guard<std::mutex> main_play_loop_lk(skip_message_in_main_play_loop_mutex_);
   skip_message_in_main_play_loop_ = true;
+  // Wait for player to be ready for playback messages from queue i.e. wait for Player:play() to
+  // be called if not yet and queue to be filled with messages.
+  {
+    std::unique_lock<std::mutex> lk(ready_to_play_from_queue_mutex_);
```
emersonknapp, line 318: Will this block forever if `play` has not been called yet?

MichaelOrlov: Yes, it's going to wait forever while `play` will not be called. 
Otherwise it will be non deterministic behavior and race conditions will not be solved.
BTW. I will need the same synchronization for `Player::seek/jump` it's going to be the same problem with race conditions.

MichaelOrlov: @emersonknapp gentle ping to finish review.

MichaelOrlov:
Gist:
https://gist.githubusercontent.com/MichaelOrlov/52594ce9ab23765032d28daf0f23a29c/raw/84d8e9c0c3e4b9f833d45611e7fabccd512b2e06/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8742
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14936/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9678/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12612/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15100/)

PR #822, opened 2021-07-16
WIP Add Player::seek/jump API

### LINKED ISSUE #821: Add jump/seek API for the Player class
## Description

This feature is part of the #696 
According to the design jump/seek API suppose to allow jump back or forward to an arbitrary point in time during playback. 
During the jump, messages lying in the interval between current timestamp and target time point should not be published.

## Related Issues
- Relates #696 
- Blocked by #813 (will need the same wait)
- Relates #824

## Completion Criteria

1. - [x] jump/seek API able to jump forward to an arbitrary time point during active playback.
2. - [x] jump/seek API able to jump forward to an arbitrary time point when player in pause mode.
3. - [x] jump/seek API able to jump backward to an arbitrary time point during active playback.
4. - [x] jump/seek API able to jump backward to an arbitrary time point when player in pause mode.

We can use the same mechanism for taking over playback as we do in play_next() API, however updating internal message queue is going to be tricky part. 
Also it could be a few nasty race conditions related to the adjustment of the reader position during active playback. Will need to call close-open for reader from separate thread to be able adjust reader position when jumping back in time.

Perhaps it would be able to write non-flaky tests using MockReader and MockPlayerClock classes.

MichaelOrlov:
- Closes #821

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -124,6 +124,9 @@ public:
   bool play_next();
 
 protected:
+  bool is_ready_to_play_from_queue_{false};
+  std::mutex ready_to_play_from_queue_mutex_;
+  std::condition_variable ready_to_play_from_queue_cv_;
   std::atomic<bool> playing_messages_from_queue_{false};
   rclcpp::Publisher<rosgraph_msgs::msg::Clock>::SharedPtr clock_publisher_;
   std::unordered_map<std::string, std::shared_ptr<rclcpp::GenericPublisher>> publishers_;
@@ -140,7 +143,8 @@ private:
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
   static const std::chrono::milliseconds queue_read_wait_period_;
 
-  std::unique_ptr<rosbag2_cpp::Reader> reader_;
+  std::mutex reader_mutex_;
+  std::unique_ptr<rosbag2_cpp::Reader> reader_ RCPPUTILS_TSA_GUARDED_BY(reader_mutex_);
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::PlayOptions play_options_;
   moodycamel::ReaderWriterQueue<rosbag2_storage::SerializedBagMessageSharedPtr> message_queue_;


rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -140,7 +143,8 @@ class Player : public rclcpp::Node
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
   static const std::chrono::milliseconds queue_read_wait_period_;
 
-  std::unique_ptr<rosbag2_cpp::Reader> reader_;
+  std::mutex reader_mutex_;
+  std::unique_ptr<rosbag2_cpp::Reader> reader_ RCPPUTILS_TSA_GUARDED_BY(reader_mutex_);
```
MichaelOrlov, line 147: @emersonknapp This is what I was talking on today's meeting. Just adding a few member variables causing memory corruption in constructor.  

TEST_F(RosBag2PlayTestFixture, player_can_resume_after_play_next) fails with memory corruption in Player constructor.

clalancette: I'm also tracking down a memory corruption in https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1296 .  I'm wondering if they are related.  If you create the node with `NodeOptions.use_clock_thread(false)`, does that make any difference?

MichaelOrlov: @clalancette Nop. The corruption still there even with `NodeOptions.use_clock_thread(false)`

clalancette: > @clalancette Nop. The corruption still there even with `NodeOptions.use_clock_thread(false)`.
> To be precise I observe exact the same issue with exact the same report from Valgrind.

All right, thanks for giving it a shot.  Seems like a different problem then.

MichaelOrlov:
@emersonknapp @clalancette I see memory corruptions when running tests under the CLion.
After some analysis I found out that `colcon build` start using `16 byte` memory alignment for data structures and classes.  
But CLion uses `8 byte` alignment as GCC doing it by default.  
it causes memory corruption when linking library/executable with different default memory alignment 
i.e in CLion `sizeof(Player) = 1784` and when compiling via `colcon build` `sizeof(Player) = 1792` .

It seems `colcon build` start using 16 bytes alignment relatively recently because we are in ApexAI using ROS2 fork from Dashing and there `colcon build` using 8 bytes alignment. I also remember that the same tests was passing on my local machine for some time ago.

Do you know when and where and may be for what reason default alignment was changed in `colcon build`?

MichaelOrlov:
It also could be the case that some compile time define not visible or not defined under the CLion and size of the data structures encapsulated in Player class evalutes with smaller size. I see that memory corruption happening first after the services i.e after rclcpp::Service<rosbag2_interfaces::srv::Pause>::SharedPtr srv_pause_; in memory and their after
it might be some compilation flags not set for  those generated code. 
Anyone have any idea?

MichaelOrlov:
@emersonknapp @clalancette Sorry for the confusion in the previous posts. 
I dig deeper and figured out that this is not a global data type alignments.
The issue is that incorrectly evaluates `sizeof(moodycamel::ReaderWriterQueue<rosbag2_storage::SerializedBagMessage>)` which is defining type for our message queue in the `Player` class. 

When compiling with `colcon buiild`  the size is **88 bytes**, although when compiling in CLion the size is **80 bytes**. This is actually the root cause of memory corruptions.

MichaelOrlov:
Issue solved by adding default `colcon` configuration in `ros2-ade` via MR https://gitlab.apex.ai/apex.ai-public/ros2-ade/-/merge_requests/26

PR #826, opened 2021-07-28
Add jump/seek API for Player class

### LINKED ISSUE #821: Add jump/seek API for the Player class

MichaelOrlov:
- Closes #821


MichaelOrlov:
I am considering to add one more unit test and probably switch to the real rosbag2 file instead of mock reader.

MichaelOrlov:
Ready for review.

emersonknapp:
Thanks for the update. I was on vacation last week but will review it this week

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -39,6 +39,20 @@
 
 namespace
 {
+/**
+ * Trivial std::unique_lock wrapper providing constructor that allows Clang Thread Safety Analysis.
+ * The std::unique_lock does not have these annotations.
+ */
+class RCPPUTILS_TSA_SCOPED_CAPABILITY TSAUniqueLock : public std::unique_lock<std::mutex>
```
emersonknapp, line 46: Non-blocking: I wonder if we should add this to `rcpputils`. I don't anticipate clang-libcxx adding it any time soon, so we may want it in a shared place.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -328,6 +350,134 @@ bool Player::play_next()
...
+      while (!message_found && reader_->has_next()) {
+        current_message = reader_->read_next();
+        if (current_message->time_stamp >= time_point) {
+          message_found = true;
+        }
+      }
+      if (message_found) {
+        // Enqueue current_message and rest of the queue to the queue again to preserve order of
+        // messages
+        if (message_queue_.peek() != nullptr) {  // if message_queue_ not empty
```
emersonknapp, line 431: nit: it might be nice to pull this logic into a separate function so it's easier to read - like `empty_queue_into(from, into)`.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -328,6 +350,134 @@ bool Player::play_next()
...
+        } else {
+          message_queue_.enqueue(current_message);
+          enqueue_up_to_boundary(play_options_.read_ahead_queue_size);  // refill queue
+        }
+      } else {
+        restore_message_queue_from_storage(message_before_seek);
+      }
+    } else {  // jump_backward
+      // Purge queue
+      while (message_queue_.pop()) {}
```
emersonknapp, line 452: would it make more sense to reconstruct the queue rather than pop each element individually?

emersonknapp (approved):
Overall this looks fine - a few non-blocking comments that you can address if you like

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: But how to properly reconstruct the `message_queue_` if we are allocating it on the stack?
I can try to use a "hack" with move constructor or move assignment operator. But it will be implementation specific and might not properly work in future, since in general case object shouldn't be used after move.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: I would like to keep it as is. wrapped in function it will be unclear what it does why.  
It's just a few lines of code in place. I can't move out whole block of code since current message is not empty already and should be accounted.
```cpp
         decltype(message_queue_) temp_message_queue;
          temp_message_queue.enqueue(current_message);
          // Copy residual messages from message_queue_ to the temp_message_queue
          while (message_queue_.try_dequeue(current_message)) {
            temp_message_queue.enqueue(current_message);
          }
```
It would be possible to move out only 
```cpp
          while (message_queue_.try_dequeue(current_message)) {
            temp_message_queue.enqueue(current_message);
          }
```
But it's really doesn't make sense to do this IMO.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: While agree with you that we need to move this definition in to the rcpputils, I don't have time and resources to do this currently.

MichaelOrlov:
Gist:
https://gist.githubusercontent.com/MichaelOrlov/ad810b7bb85445a829ac57a03dcc14c5/raw/f87ec0131d7459819d794fa0f9ec69a0f161c505/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8969
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9910/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12870/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15369/)

MichaelOrlov:
@emersonknapp CI fails with errors in `rmw_cyclonedds_cpp` package which I didn't touch.
```
--- stderr: rmw_cyclonedds_cpp
19:52:53 /home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp: In function ‘CddsPublisher* create_cdds_publisher(dds_entity_t, dds_entity_t, const rosidl_message_type_support_t*, const char*, const rmw_qos_profile_t*)’:
19:52:53 /home/jenkins-agent/workspace/ci_linux-aarch64/ws/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:2039:22: error: ‘dds_is_loan_available’ was not declared in this scope; did you mean ‘is_loan_available’?
19:52:53  2039 |     is_fixed_type && dds_is_loan_available(pub->enth);
19:52:53       |                      ^~~~~~~~~~~~~~~~~~~~~
19:52:53       |                      is_loan_available
```

Should I modify `ros2.repos` to refer to a some specific version of the cyclonedds?

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15191/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9911/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12871/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15370/)

MichaelOrlov:
Rerunning Windows build. It seems some unrelevant issue that CI script can't delete temp file from TEMP folder.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15372/)

Issue #824, opened 2021-07-22
Seek/Jump API for Reader/Storage 


emersonknapp:
## Description

Provide a way to "seek" the "read head" of a bag to an arbitrary timestamp. This will allow for more flexible exploration and playback of bags.

## Related Issues

#696 - will enable full `jump` functionality 

## Completion Criteria

* New `rosbag2_storage::storage_interfaces::ReadOnlyInterface` API with signature (probably) `seek(timestamp)` - which clamps value to duration of the bag
* Implement this API for the default sqlite3 storage implementation
* Expose this to the `rosbag2_cpp::Reader` and `rosbag2_py.Reader` so that it can be used for API-driven use case
* Make sure "seek"ing across split bag boundaries works correctly both backwards and forwards

Out of scope:
* Exposing this functionality to `Player` - that will be a followup issue, which will require careful management of the play queue

## Implementation Notes / Suggestions

`rqt_bag` for ROS 2 already implements similar functionality with direct sqlite3 queries, it can provide implementation hints

## Testing Notes / Suggestions

* Unit testing for new C++/Python APIs
* Test to make sure "seek"ing across split bag boundaries works correctly both backwards and forwards

emersonknapp:
FYI @MichaelOrlov - my team will be trying to implement this within the next few weeks, which should make part of the `jump` function easier to do

PR #836, opened 2021-08-03
added seek interface

lihui815:
Per #824, created `seek(timestamp)` method in `rosbag2_storage::storage_interfaces::ReadOnlyInterface` API.

- added `seek(timestamp)`
- default sqlite implementation:
  - `seek` implementation keeps the storage filter, but will re-run the query with a new seek_time
  - updated `set_filter`/`reset_filter` to be able to be called in between reads in place. This means that if the message topics were in the order of:
     ```
     1 AAA
     2 BBB
     3 CCC
     4 AAA
     5 BBB
     6 AAA
     ```
     and we set filter on topic `AAA` after reading two messages (`AAA` & `BBB`), the next message read would be the `AAA` on the 4th row.
- added seek implementation to sequential reader (assumes multi-file is in time order)
  - needs to reset the current file to the first file in order to be able to skip backwards
  - `read_next()` calls `has_next()` to ensure that reader rolls over to next file at end of current file
  - `has_next()` calls itself recursively until either there is a next row or there are no more files (doesn't assume that the next file will definitely contain valid row)
- added tests for seeking/filtering both single and multi-file bags

#### NOTE

Because of continued confusion in the comments, I'm highlighting here the difference between
- sqlite implementation of `storage_interfaces::ReadOnlyInterface->seek(t)`
- `SequentialReader->seek(t)`

The differences are as follows:
- reading in time order:
  - `SqliteStorage` DOES guarantee time-ordered reads regardless of write order. This means that calling `seek(t)` will also guarantee subsequent `read_next()` calls to return messages in time order until the end of the file.
  - `SequentialReader` DOES NOT guarantee time-ordered reads if individual files within a bag are not in non-overlapping sequential time order (#842). This means that, with or without calling `seek(t)`, there may be messages that occur at an earlier timestamp than previously-read messages when one file rolls over to the next file. `SequentialReader->seek(t)` will only guarantee that all messages and only messages with timestamp >= `t` will be read (satisfying any topic filters).
- In-place setting/resetting of topic filters:
  - in `SqliteStorage`, messages are queried in time order. The implicit tiebreaker between messages of the same timestamp is on the row id. This PR makes this second-degree ordering explicit. When we modify topic filters "in place", we actually rerun the query of the db file. In order to prevent re-reading previously read messages, we filter the timestamp to prevent reading "older" messages, AND we filter row-id to prevent reading already-read messages at the current timestamp. Note that if we don't assume write order (consistent with existing SqliteStorage implementation), we need to do both. For example:
    ```
     # write order
      row0 t1 msgA
      row1 t0 msgB
      row2 t0 msgC
      row3 t2 msgD
      row4 t1 msgE

      # read order
      row1 t0 msgB
      row2 t0 msgC  <-- same timestamp as msgB but loses row_id tiebreak
      row0 t1 msgA  <-- later timestamp than msgB & msgC, but lower row-id
      row4 t1 msgE  <-- same timestamp as msgA but loses row_id tiebreak
      row3 t2 msgD
     ```
      From this example, if we set topic filter between reading `msgA` and `msgE`, then we need to check against the last read message time `t1` to prevent previously read messages `msgB` and `msgC` to be re-read, and to check against the row id `row0` to make sure that `msgA` doesn't get re-read. This is why in the `SqliteStorage` implementation we update `seek_time_` and `seek_row_id_` with every read, to make sure that we can make the correct in-place update query. Thus the filter query is necessarily `(timestamp > last_t) or (timestamp = last_t and row_id > last_rid)`.
  - in `SequentialReader`, this is not applicable. When setting/resetting topic filters the "in place pointer" is kept by reference to the current file. We can assume that all files previous to the current file have already been read, so the setting/resetting of topic filters just occurs on the current & subsequent files. Within the current file, the "in place" message pointer is already handled by the single-file storage implementation.


rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
```diff
@@ -45,6 +47,8 @@ class ROSBAG2_STORAGE_PUBLIC ReadOnlyInterface
   virtual void set_filter(const StorageFilter & storage_filter) = 0;
 
   virtual void reset_filter() = 0;
+
+  virtual void seek(const rcutils_time_point_value_t & timestamp) = 0;
```
emersonknapp, line 51: I know the rest of this interface isn't documented yet - but can you document this new signature? I'd like to see some language like "next message from `read_next` will be greater than or equal to `timestamp`". Also some note about the bounds behavior, such as does it automatically clamp the value to `[start, end]`?

emersonknapp: Second comment - I think this should probably `rclcpp::Time` instead of an `rcutils` type. Open to counterpoint, though

rosbag2_storage_default_plugins/include/rosbag2_storage_default_plugins/sqlite/sqlite_storage.hpp
```diff
@@ -110,8 +111,12 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
+
+  rcutils_time_point_value_t seek_time_ = 0;
+  int seek_row_id_ = 0;
   rosbag2_storage::StorageFilter storage_filter_ {};
 
+
```
emersonknapp, line 119: rm unrelated newline?

rosbag2_transport/test/rosbag2_transport/mock_sequential_reader.hpp
```diff
@@ -79,6 +79,11 @@ class MockSequentialReader : public rosbag2_cpp::reader_interfaces::BaseReaderIn
     filter_ = rosbag2_storage::StorageFilter();
   }
 
+  void seek(const rcutils_time_point_value_t & /*timestamp*/) override
+  {
+    filter_ = rosbag2_storage::StorageFilter();
```
emersonknapp, line 84: seek resets the filter?

rosbag2_py/test/test_sequential_reader_multiple_files.py
```diff
@@ -0,0 +1,123 @@
...
+    # TL;DR This makes class_loader work when using a python extension compiled with libc++.
+    #
+    # For the fun RTTI ABI details, see https://whatofhow.wordpress.com/2015/03/17/odr-rtti-dso/.
+    sys.setdlopenflags(os.RTLD_GLOBAL | os.RTLD_LAZY)
+
+from common import get_rosbag_options  # noqa
+import rosbag2_py  # noqa
+
+
+def test_sequential_reader():
```
emersonknapp, line 31: name this file/test more specifically as a `seek` test?

rosbag2_py/test/test_sequential_reader.py
```diff
@@ -79,6 +79,57 @@ def test_sequential_reader():
             assert msg.data == f'Hello, world! {msg_counter}'
             msg_counter += 1
 
+    # Seek No Filter
```
emersonknapp, line 82: should this maybe be a separate test case? this case looks really long

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -191,13 +194,25 @@ void SequentialReader::set_filter(
...
+{
+  seek_time_ = timestamp;
   if (storage_) {
-    storage_->reset_filter();
+    // reset to the first file
+    current_file_iterator_ = file_paths_.begin();
+    preprocess_current_file();
+    storage_options_.uri = get_current_file();
+    // open and set filters
+    storage_ = storage_factory_->open_read_only(storage_options_);
```
emersonknapp, line 209: Is there much overhead to this opening? We don't have to block first pass on this consideration, but I might put a TODO here to do something more performant - unless it's not a problem or there's a strong reason we _have_ to reload the storage.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -143,9 +143,10 @@ bool SequentialReader::has_next()
     if (!current_storage_has_next && has_next_file()) {
       load_next_file();
       storage_options_.uri = get_current_file();
-
+      // open and set filters
       storage_ = storage_factory_->open_read_only(storage_options_);
-      storage_->set_filter(topics_filter_);
+      storage_->seek(seek_time_);
```
emersonknapp, line 148: Would it make sense for this `if`-block logic to go into `load_next_file` instead of being here in `has_next`?

emersonknapp (changes requested):

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
MichaelOrlov: @emersonknapp We have defined `virtual void jump(rcutils_time_point_value_t time_point) = 0;` API with rcutils timestamp in `PlayerClock` class and I am going to operate with it on upper `rosbag2_transport` layer as well. 
It would be good to have consistency in types in all API levels to avoid extra conversions or checks.

MichaelOrlov (changes requested):

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
lihui815: I was just keeping it consistent with the type returned by `read_next()`, if `PlayerClock::jump` uses same type, then i vote not changing it for now.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
lihui815: @MichaelOrlov for the same reason that we had `->set_filter(topics_filter_)`. `seek_time_` is basically just a time filter in the query, and when you roll over to a new file, you need to setup the filters of the query to be run in the new storage reader.

The `sequential_reader` kind of assumes that the files in a split bag are arranged in non-overlapping time order, so setting the `seek_time_` is redundant given that assumption, but there isn't actually anything that reinforces this assumption, I can easily create a multi-file bag that breaks it, and the reader wouldn't know any better (see #842). While setting the `seek_time_` here wouldn't necessarily resolve this assumption break, it will ensure that reads across multiple files will consistently obey the seek time filter, and appear in weakly monotonic time order.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
lihui815: are you asking why we need to reopen files here? there's no guarantee that the seek time is in the current file. if we have per-file metadata around the start/end time of each file, i guess we can just set to the first file that satisfies. i'm not sure why we're particularly concerned about this reopen as it happens once per seek. we already do it on every roll over. It's true that if the bag is split into many files the seek + subsequent rollovers before hitting the first message might add up.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/read_only_interface.hpp
emersonknapp: Good counterpoint!

jhdcs: If I may be so bold as to comment on an issue I'm not actively reviewing, perhaps we should create a separate issue where we discuss the merits of both `rclcpp::Time` and `rcutils`, and use that to decide which way the package as a whole should go?

My vote is also for not changing the API yet. But that's mostly because I like having one PR per one issue, and that feels a bit beyond scope here.

emersonknapp (approved):
LGTM

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
emersonknapp: My concern was more around repeated seeks within a single file, such as in a "video scrubber" GUI interface - however we can leave this consideration for a real use case and apply performance testing rather than having me speculate on it. Let's move forward with this implementation.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15016/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9755/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12705/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15178/)

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
MichaelOrlov: @lihui815 Hold on. Seek is not a filter. 
At least it's supposed to be a one time command for adjusting pointer to current message with requested time.
Am I missed something?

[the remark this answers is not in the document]
MichaelOrlov: If user calling `has_next()` it should return if bag really has next message independently from `seek()` operation.
Per your explanation logic will  get broken and will have side effect not as we expected.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -158,9 +154,13 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::read_ne
 {
   if (storage_) {
     // performs rollover if necessary
-    has_next();
-    auto message = storage_->read_next();
-    return converter_ ? converter_->convert(message) : message;
+    if (has_next()) {
+      auto message = storage_->read_next();
+      // update seek time
+      seek_time_ = message->time_stamp;
```
MichaelOrlov, line 160: We should not use time_stamp from read_next message as seek_time. 
Seek time should be used only directly from `void SequentialReader::seek(const rcutils_time_point_value_t & timestamp)` API call.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
MichaelOrlov: @lihui815 @emersonknapp IMO #842 is a completely separate issue and is matter of discussion how this possibility of writing out of ordered messages could be workarounded or even fixed. But it certainly shouldn't be addressed in this PR using `seek()` API as filter.

Playback must replay in exact order as messages was recorded. **We need this condition for deterministic replay.** Otherwise it will be difficult to reproduce exact behavior of the entire system if player will do some internal reordering of the messages which he was not asked for.
If messages were recorder out of order this is problem of recorder or post processing tool.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
lihui815: this is to track time as we proceed through the read. If we set/reset filter, we want to do so "in place" without resetting the seek time back to a previous place in time.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -191,25 +191,42 @@ void SequentialReader::set_filter(
...
+}
+
+void SequentialReader::seek(const rcutils_time_point_value_t & timestamp)
+{
+  seek_time_ = timestamp;
   if (storage_) {
-    storage_->reset_filter();
+    // reset to the first file
+    current_file_iterator_ = file_paths_.begin();
+    load_current_file();
```
MichaelOrlov, line 203: We are resetting to the first file and trying to `seek` only in it. But what if requested timestamp in second or third file?
It seems we are not going to try to search out of the first file. 
Am I missed some logic?

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
emersonknapp: This conversation is off topic for this feature - if you have a specific issue with the current behavior (order by timestamp) - open a new issue to discuss.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
lihui815: When we `read_next()`, we call `has_next()`, which will call `load_next_file()` if we reach the end of the file and there is another file. It will do so until either there is a next message found, or until there are no more files to open. Resetting to the first file guarantees that you will be able to roll through to subsequent files while reading. Not resetting to the first file means that you will miss messages that occurred before the current file.

Example:

```
##### File 0 #####
t0 msg0
t1 msg1
t2 msg2

##### File 1 #####
t3 msg3
t4 msg4
t5 msg5
```

Assume we just finished reading `msg3`, and the next message is `msg4`. When we call `seek(t1)`, it will set the current file to `File 0`, and open `File 0` and set filter and seek time. Then subsequent calls to `read_next()` will return:

```
msg1
msg2
msg3  <-- rollover to File 1 happens on the has_next() call in this read_next() call
msg4
msg5
```

If we don't reset the current file to the first file, then instead, we will only be seeking within the current file before the `seek(t1)` call and our reads will return:

```
msg3  <-- First msg in File 1 to occur after t1
msg4
msg5
```

Now consider the case of calling `seek(t5)`. If we reset to the first file (File 0), then on the first `read_next()`, we will be calling `has_next()` which calls `storage_->has_next()`, which will return `false`. This will trigger the reader to check for and load the next file (File 1), and we proceed to seeking within File 1:

```
// Finds no messages in File 0, rolls over to File 1
msg5 <-- First message in File 1 that occurs on or after t5
```

lihui815: btw the new test cases in rosbag2_py is essentially testing that `seek(t)` behaves as expected backward, forward and across file splits. You're welcome print out the entirety of the `wbag` messages and follow along.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -477,12 +485,24 @@ rosbag2_storage::BagMetadata SqliteStorage::get_metadata()
...
+  set_filter(rosbag2_storage::StorageFilter());
+}
+
+void SqliteStorage::seek(const rcutils_time_point_value_t & timestamp)
+{
+  // reset row id to 0 and set start time to input
+  // keep topic filter and reset read statement for re-read
+  seek_row_id_ = 0;
+  seek_time_ = timestamp;
+  read_statement_ = nullptr;
```
MichaelOrlov, line 505: I would prefer to leverage heavy weighting operations by making new request to the DB here inside `seek` API rather than just resetting `read_statement_` pointer.

It will affect performance in some cases. 
Someone could reasonably assume that `seek` API really doing necessary heavy lifting preparations for retrieving next message with requesting timestamp.
As for the example if we would have bag file with 16 GB of data and will do `seek` and than `has_next` `read_next`. seek  operation will complete almost instantaneously although `has_next` will take substantial amount of time to reared this 16 GB in memory. For instance `play_next()` call will highly likely hiccup after the `seek` operation.

MichaelOrlov: To optimize performance please consider to iterate throw messages in current `message_result_` to try to find message with requested timestamp before doing new SQL request to the DB.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
MichaelOrlov: @lihui815 Thank you for detailed explanation. I missed the part that we start calling `has_next` inside `read_next` API recently.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
MichaelOrlov: @lihui815 Thanks for the explanation, I've figured it out.
 But still have opinion that we shouldn't use seek time as a filter.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
MichaelOrlov: After digging deeper in to the SQLite implementation I figured out that iterating through messages in current `message_result_` will not be faster than making new request to the database since when iterating throw `message_result_` we are actually doing new SQL request to the DB for the next row. 
I was thinking that `read_statement_->execute_query<..>` doing actual SQL request to the DB and caching it in the memory, but this is not true.

MichaelOrlov (changes requested):
I am requesting for changes in `sqlite_storage.cpp` to cleanup logic since in current stage it's misleading further in sequential_reader

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
MichaelOrlov: @emersonknapp @lihui815 Of course I can create follow up PR. Just want to let you know that current implementation will introduce **bug**.
With current implementation in case when we would have second DB file with first message in it with timestamp early then the in last message in the first file (as described in #842) we are going to skip this first message in second DB file. 
Not playing message just because it was recorder out of order will be considered as a bug and will lead to the non-deterministic replay. 
Just imagine if it would be some key frame for some algorithm or triggering message and we will not play it.

IMO seek `seek_time_` should be updated only from `seek(timestamp)` API.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -301,6 +301,11 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SqliteStorage::read_next(
   bag_message->time_stamp = std::get<1>(*current_message_row_);
   bag_message->topic_name = std::get<2>(*current_message_row_);
 
+  // set start time to current time
+  // and set seek_row_id to the new row id up
+  seek_time_ = bag_message->time_stamp;
+  seek_row_id_ = std::get<3>(*current_message_row_) + 1;
```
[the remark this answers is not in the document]
lihui815: you're right if you assume row_id is in time order, but it might not be. row_id is used as a tie breaker to preserve message pointer at the same timestamp, otherwise, the timestamp is used for that purpose.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
MichaelOrlov: @lihui815 Sorry, but I am still not get it.
Why introduce bug in first place and link it to some issue for further consideration, rather then just fix it on code review?
It will be obviously a buggy behavior with skipping message. 
I would disagree with this
>skipping messages that appear out of order in next file - yes i considered this, however this is something i would wrap into issue of #842 because it's part of the generic issue of messages being able to be read out of time order

Before this PR behavior was consistent and we didn't skip messages which was recorded out of order. 

I agree that seek operation doesn't make sense if messages stored in bag in out of timestamp order.
And 
> so I'd rather not make implementation workarounds for this

This is exactly what I am asking for. Please don't try to workaround situation with messages stored out of order in this PR.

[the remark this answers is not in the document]
MichaelOrlov: Currently we don't have real cases with situation when messages stored out of timestamp order in different files. 
Issue #842 describe potential problem if someone will potentially try to incorrectly use direct API when saving messages to bag files.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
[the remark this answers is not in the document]
MichaelOrlov: @lihui815 But how `seek_row_id_` will help to preserve message pointer at the same timestamp when `row_id` is not in timestamp order? 
I don't see a logic for this in your implementation. Could you please explain, clarify?
 
And again the case with situation when `row_id` stored in out of time stamp order should't be considered in this PR at first place.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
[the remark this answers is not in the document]
lihui815: Agree changing behavior not best practice. removed `seek_time_` updating in `SequentialReader->read_next()`, and clarified that `SequentialReader->seek(t)` will just ensure that all subsequent reads will return all messages at or after time `t`, and will not guarantee time order.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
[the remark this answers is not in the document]
emersonknapp: OK - the behavior is changed to maintain previous behavior. Now it moves to the first instance of the time in question, but does not try to maintain that time when loading new files.

We can try to address out-of-order files separately from this feature.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15052/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9786/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12739/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15220/)

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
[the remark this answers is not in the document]
MichaelOrlov: @lihui815 Thanks for trying to introduce clarity in your logic with updates in the description.

Although I am pretty mach assure that your assumption is wrong. You are constructing extra level of ordering using  manual indexing via `seek_row_id_` to handle situation when written two or more messages with the same time stamp. i.e.
> write order
  row0 t1 msgA
  row1 t0 msgB
  row2 t0 msgC

But this situation is not possible in regular scenario. Since we have only one common timer and it's accuracy up to the  nanosecond. Even if two cores will ask timestamp at the same processors tact they will get different timestamp from the timer. 
The rule could be violated if someone will **hack** rosbag and insert record with the same timestamp without really taking it from timer. But in this case it will be **ill formed** rosbag and no guarantee that `row_id` will be unique as well (e.g. it could be hacked the similar way). 
IMO. any way we should not make workarounds inside our readers for ill formed bag files.

PR #837, opened 2021-08-04
Design proposal for snapshot feature

camm73:
See ros2/rosbag2#663 for details on the original feature request.

### DIFF
--- a/docs/design/rosbag2_snapshot_mode.md
+++ b/docs/design/rosbag2_snapshot_mode.md
@@ -1,0 +1,46 @@
+# Adding a Snapshot Service for rosbag2
+
+## Context
+In [ros2/rosbag2#663](https://github.com/ros2/rosbag2/issues/663), it was recommended that a "snapshot" feature be added to rosbag2.
+
+## Motivation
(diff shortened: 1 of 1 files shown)


docs/design/rosbag2_snapshot_mode.md
```diff
@@ -0,0 +1,46 @@
...
+For example, if a rare, anomalous behavior occurs, this feature will allow the user to record the most recent ROS topic messages when the behavior is observed, instead of having to record the entire execution.
+
+## Example Usage
+
+Starting `rosbag2` in snapshot mode with a specified cache size:
+```
+$ ros2 bag record --max-cache-size 100000 --snapshot-mode [topics [topics ...]]
+```
+
+Triggering a snapshot:
```
emersonknapp, line 19: This is good, nitpick: I would probably shorten the language to `~/snapshot` (note `~` notation for "root namespace of node").

I think you should also make an additional note that this should also be usable via keyboard handler on the recording process. We don't have any keyboarding hooked up _yet_, but it should be an intention for the feature. Not necessary to tie it to a key, but we could say something like "Also triggerable via key press on recording process, suggestion: 's'"

emersonknapp (approved):
This looks good! I think it's clear enough to start implementation based on this. Let's merge after the small tweak on the invocation section

emersonknapp (approved):

PR #839, opened 2021-08-05
fix sequential reader rollover-to-next-file

lihui815:
- `has_next()` - calls recursively until next message OR no next file found
- `read_next()` - calls `has_next()` to ensure necessary rollovers are made if necessary

Signed-off-by: Sonia Jin <diegothemuich@gmail.com>

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
@@ -105,6 +105,8 @@ void SequentialCompressionReader::open(
 std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialCompressionReader::read_next()
 {
   if (storage_ && decompressor_) {
+    // roll over if necessary
+    has_next();
     auto message = storage_->read_next();
     if (compression_mode_ == rosbag2_compression::CompressionMode::MESSAGE) {
       decompressor_->decompress_serialized_bag_message(message.get());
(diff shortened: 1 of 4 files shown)


emersonknapp (approved):
Nice fix

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14992/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9731/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12678/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15154/)

MichaelOrlov:
@emersonknapp @lihui815 I am sorry for my late response to this closed issue, but IMO it was not a real issue it was working as designed. We just didn't have Doxygen documentation for these API.
I believe it was designed this way on purpose for the sake of performance to avoid calling `has_next() ` twice. 
We are calling `has_next()` and `read_next()` always sequentially in 
[Player::enqueue_up_to_boundary(uint64_t boundary)](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/player.cpp#L356-L366) 
```cpp
void Player::enqueue_up_to_boundary(uint64_t boundary)
{
  rosbag2_storage::SerializedBagMessageSharedPtr message;
  for (size_t i = message_queue_.size_approx(); i < boundary; i++) {
    if (!reader_->has_next()) {
      break;
    }
    message = reader_->read_next();
    message_queue_.enqueue(message);
  }
}
```
and this is only one place where we are calling it. 

I would suggest to revert this PR and add relevant Doxygen comment to the `read_next()` API that if it would be called without checking that `has_next()` returning true it will be undefined behavior.

MichaelOrlov:
After some detailed consideration I will step back. Sorry for disturbing.
We actually will need this extra call for `has_next()` inside `read_next()` to make sure that seek timestamp will be applied correctly and calling one more time for the `reader->has_next()` will translates to the one more call to the `storage->has_next()` call which is going to check that iterator doesn't point to the end and it will not affect performance.

PR #840, opened 2021-08-06
add per-file information to metadata.yaml

WJaworskiRobotec:


emersonknapp:
Hope you don't mind - I marked this as draft while WIP. Feel free to move back to ready for review when you want a review on it

WJaworskiRobotec:
@emersonknapp sure no problem. It's now ready for review.

WJaworskiRobotec:
@emersonknapp could you please review this PR ?

emersonknapp:
Yes - I have been on vacation but will review this week. Thank you for your patience

WJaworskiRobotec:
@MichaelOrlov sure, I will add this and let you know once it's ready. Thanks for your comment.

WJaworskiRobotec:
@emersonknapp @MichaelOrlov Message count per-file added. Please review this PR

emersonknapp (changes requested):
High level, I feel like it would make more sense to replace the existing `relative_file_paths` list of strings, with a `files` list of dictionaries that have a path as well as other metadata

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -31,12 +31,21 @@ struct TopicInformation
   size_t message_count;
 };
 
+struct FileInformation
+{
+  std::string relative_file_path;
```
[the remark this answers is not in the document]
WJaworskiRobotec: Ok, i've changed the structure of metadata to remove the duplicated data

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -135,6 +135,57 @@ struct convert<std::vector<rosbag2_storage::TopicInformation>>
...
+    metadata.starting_time =
+      node["starting_time"].as<std::chrono::time_point<std::chrono::high_resolution_clock>>();
+    metadata.duration = node["duration"].as<std::chrono::nanoseconds>();
+    metadata.message_count = node["message_count"].as<uint64_t>();
+    return true;
+  }
+};
+
+template<>
+struct convert<std::vector<rosbag2_storage::FileInformation>>
```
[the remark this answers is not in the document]
WJaworskiRobotec: You are right, not needed code removed

WJaworskiRobotec:
I've applied the changes @emersonknapp. However i see we should take a look at all the tests (together with fixing bug mentioned in  #841). Looks to me like some of the tests are ignoring fields, also replaying the bag split by time does not work properly on current master as i observed. Please re-review the changes and as the next step i'd like to take a look at #841 and i can take care of the tests in general.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -91,13 +91,22 @@ void SequentialReader::open(
   // This is necessary for non ROS2 bags (aka ROS1 legacy bags).
   if (metadata_io_->metadata_file_exists(storage_options.uri)) {
     metadata_ = metadata_io_->read_metadata(storage_options.uri);
-    if (metadata_.relative_file_paths.empty()) {
-      ROSBAG2_CPP_LOG_WARN("No file paths were found in metadata.");
-      return;
+    if (metadata_.version < 5) {
```
emersonknapp, line 94: Let's please keep this versioning logic out of the `sequential_reader`.  It's already a bit dangerous, so we should contain it completely in `metadata_io` and `bag_metadata (there may be instances where this is not true right now, but that's not a good reason for us to make the problem worse haha)

see my comment on `BagMetadata` about the redundant fields - maybe this ends up looking like:

```
// resolve_relative_filepaths checks the version (and maybe in the future it's actually a method on BagMetadata to get that logic out of this file)
file_paths_ = details::resolve_relative_paths(storage_options.uri, metadata_);
```

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -31,12 +31,21 @@ struct TopicInformation
...
+};
+
 struct BagMetadata
 {
-  int version = 4;  // upgrade this number when changing the content of the struct
+  int version = 5;  // upgrade this number when changing the content of the struct
   uint64_t bag_size = 0;  // Will not be serialized
   std::string storage_identifier;
   std::vector<std::string> relative_file_paths;
+  std::vector<FileInformation> files;
```
emersonknapp, line 48: i think ideally we should just remove `relative_file_paths` here - so that there is not redundant information. Perhaps `relative_file_paths` is a getter method that parses from `files`

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -189,14 +212,19 @@ struct convert<rosbag2_storage::BagMetadata>
       node["compression_format"] = metadata.compression_format;
       node["compression_mode"] = metadata.compression_mode;
     }
+    if (metadata.version < 5) {
+      node["relative_file_paths"] = metadata.relative_file_paths;
+    }
+    if (metadata.version >= 5) {
```
emersonknapp, line 218:
```suggestion
    } else {
```

emersonknapp: However, it's interesting that we even have versioning information here - `encode` should always be encoding the latest version, in this case 5. I would not be opposed to removing all versioning from `encode`.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -209,6 +237,13 @@ struct convert<rosbag2_storage::BagMetadata>
       metadata.compression_format = node["compression_format"].as<std::string>();
       metadata.compression_mode = node["compression_mode"].as<std::string>();
     }
+    if (metadata.version < 5) {
+      metadata.relative_file_paths = node["relative_file_paths"].as<std::vector<std::string>>();
+    }
+    if (metadata.version >= 5) {
```
emersonknapp, line 243:
```suggestion
    } else {
```

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -209,6 +237,13 @@ struct convert<rosbag2_storage::BagMetadata>
       metadata.compression_format = node["compression_format"].as<std::string>();
       metadata.compression_mode = node["compression_mode"].as<std::string>();
     }
+    if (metadata.version < 5) {
+      metadata.relative_file_paths = node["relative_file_paths"].as<std::vector<std::string>>();
```
emersonknapp, line 241: so maybe based on prior comments, this looks like

```
auto relative_paths = node["relative_file_paths"].as<std::vector<std::string>>();
for (const auto & relative_path : relative_paths) {
  FileInformation fi;
  fi.starting_time = -1;  // maybe use a negative time for "unknown"
  fi.duration = -1;
  fi.path = relative_path;
  metadata.files.push_back(fi);
}
```

emersonknapp (changes requested):

WJaworskiRobotec:
@emersonknapp as removing relative_file_paths requires a lot of changes (mostly in tests, that should be fixed anyway), my suggestion is that in this PR i will move back to the "clean" state with added per-file information and I will create separate PR with metadata and encode function cleanup (removing all version conditions from encode, removing relative_file_paths from metadata). What is your opinion ?

emersonknapp:
It should be easy to remove version conditions from `encode`. However, I understand that removing `relative_file_paths` might be messy, and I'm OK with leaving that for later.

WJaworskiRobotec:
@emersonknapp I applied changes, but I feel i made a little mess here because of merging master branch here instead of rebasing. I think it might be more clear to create new "clean" branch with only relevant changes and create this PR once again.

WJaworskiRobotec:
If you agree to make it in more clean way in the second PR, i created this one : #870 and i will remove this messy PR and branch.

emersonknapp:
You can force-push to this source branch with any cleaned-up version that you create - you shouldn't need to recreate a PR. However, if that's what you want to do, then go ahead.

WJaworskiRobotec:
Yeah, it'll be more clear. Please review #870, and i'm closing this one

PR #844, opened 2021-08-11
Circular Message Cache implementation for snapshot feature

camm73:
Part of #663

Implements the first part of the [new snapshot feature](https://github.com/ros2/rosbag2/blob/master/docs/design/rosbag2_snapshot_mode.md) by creating a `MessageCacheCircularBuffer` class which implements a circular buffer used to store messages until the buffers are flipped. This class is used by `CircularMessageCache` to create two circular buffers where the primary buffer is continuously filled with messages, overwriting old messages when full, until a buffer swap is triggered. Once the swap occurs, the recorded messages for the snapshot can now be accessed.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef ROSBAG2_CPP__CACHE_INTERFACES__BASE_CACHE_BUFFER_INTERFACE_HPP_
-#define ROSBAG2_CPP__CACHE_INTERFACES__BASE_CACHE_BUFFER_INTERFACE_HPP_
+#ifndef ROSBAG2_CPP__CACHE__CACHE_BUFFER_INTERFACE_HPP_
+#define ROSBAG2_CPP__CACHE__CACHE_BUFFER_INTERFACE_HPP_
 
 #include <memory>
 #include <vector>
@@ -23,14 +23,13 @@
 
 namespace rosbag2_cpp
 {
-using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
-namespace cache_interfaces
+namespace cache
 {
-
-class ROSBAG2_CPP_PUBLIC BaseCacheBufferInterface
+using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
+class ROSBAG2_CPP_PUBLIC CacheBufferInterface
 {
 public:
-  virtual ~BaseCacheBufferInterface() {}
+  virtual ~CacheBufferInterface() {}
 
   virtual bool push(buffer_element_t msg) = 0;
 
@@ -41,7 +40,7 @@ public:
   virtual const std::vector<buffer_element_t> & data() = 0;
 };
 
-}  // namespace cache_interfaces
+}  // namespace cache
 }  // namespace rosbag2_cpp
 
-#endif  // ROSBAG2_CPP__CACHE_INTERFACES__BASE_CACHE_BUFFER_INTERFACE_HPP_
+#endif  // ROSBAG2_CPP__CACHE__CACHE_BUFFER_INTERFACE_HPP_
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
@@ -25,18 +25,46 @@ namespace rosbag2_cpp
 {
 namespace cache
 {
-using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
+/**
+* This class provides the interface for all CacheBuffer implementations. The
+* role of these buffers is store messages that are received via topic subscriptions
+* before they are written to the bagfile.
+*
+* Any class that implements CacheBufferInterface is reponsible for handling
+* synchronization that is needed for the buffer to operate properly.
+*/
 class ROSBAG2_CPP_PUBLIC CacheBufferInterface
 {
 public:
+  using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
   virtual ~CacheBufferInterface() {}
 
+  /**
+   *   Pushes a SerializedBagMessage into the cache buffer.
+   *
+   *   \param msg SerializedBagMessage to add to the buffer.
+   *   \return whether message was successfully added to the cache.
+   */
   virtual bool push(buffer_element_t msg) = 0;
 
+  /**
+   *   Clears the buffer by removing all remaining messages.
+   */
   virtual void clear() = 0;
 
+  /**
+   *   Check the buffer message count.
+   *
+   *   \return number of elements in the buffer.
+   */
   virtual size_t size() = 0;
 
+  /**
+   *   Get the data/messages stored in the buffer. This should only be
+   *   called once no more messages are being added to the buffer.
+   *
+   *   \return a vector containing messages in the buffer.
+   */
   virtual const std::vector<buffer_element_t> & data() = 0;
 };
 
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
@@ -20,6 +20,8 @@
 #include <string>
 
 #include "rosbag2_cpp/cache/message_cache_circular_buffer.hpp"
+#include "rosbag2_cpp/cache_interfaces/base_cache_interface.hpp"
+#include "rosbag2_cpp/cache_interfaces/base_cache_buffer_interface.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
 #include "rosbag2_storage/serialized_bag_message.hpp"
@@ -38,21 +40,23 @@ namespace cache
 {
 
 class ROSBAG2_CPP_PUBLIC CircularMessageCache
+  : public rosbag2_cpp::cache_interfaces::BaseCacheInterface
 {
 public:
   explicit CircularMessageCache(uint64_t max_buffer_size);
 
   /// Puts msg into circular buffer, replacing the oldest msg when buffer is full
-  void push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg);
+  void push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg) override;
 
   /// get current buffer to consume
-  std::shared_ptr<MessageCacheCircularBuffer> consumer_buffer();
+  std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseCacheBufferInterface>
+  consumer_buffer() override;
 
   /// Swap the primary and secondary buffer before consumption.
   /// NOTE: consumer_buffer() should be called sequentially after
   /// swap_buffer() to ensure expected behavior. Calling swap_buffer()
   /// while accessing consumer_buffer() will be undefined behavior.
-  void swap_buffers();
+  void swap_buffers() override;
 
 private:
   /// Double buffers
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
@@ -20,7 +20,7 @@
 #include <string>
 
 #include "rosbag2_cpp/cache/message_cache_circular_buffer.hpp"
-#include "rosbag2_cpp/cache_interfaces/base_cache_interface.hpp"
+#include "rosbag2_cpp/cache_interfaces/base_message_cache_interface.hpp"
 #include "rosbag2_cpp/cache_interfaces/base_cache_buffer_interface.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
@@ -40,7 +40,7 @@ namespace cache
 {
 
 class ROSBAG2_CPP_PUBLIC CircularMessageCache
-  : public rosbag2_cpp::cache_interfaces::BaseCacheInterface
+  : public rosbag2_cpp::cache_interfaces::BaseMessageCacheInterface
 {
 public:
   explicit CircularMessageCache(uint64_t max_buffer_size);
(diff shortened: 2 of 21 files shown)


rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -0,0 +1,66 @@
...
+namespace cache
+{
+/**
+* This class implements a circular buffer message cache. Since the buffer
+* size is limited by total byte size of the storage messages rather than
+* a fix number of messages, a linked list is used instead of a vector since
+* older messages can always be dropped from the front and new messages added
+* to the end. This buffer may consume more byte than max_cache_size, however
+* it will only exceed the limit by the excess data in the most recent message.
+*/
```
MichaelOrlov, line 31: It does make sense to move description to the class in relevant header `hpp` file, where all other interface descriptions and doxygen comments.

MichaelOrlov (changes requested):
@camm73 Thank you for your contribution.
Implementation of `MessageCacheCircularBufer` with linked list looks very reasonable, however I would recommend  to keep the same interface as for `MessageCacheBuffer` i.e. will need to change return type for `MessageCacheCircularBuffer::data()` from list to the vector.
Inside `MessageCacheCircularBuffer::data()` you can create new `std::vector<buffer_element_t>` for returning value and push elements in it from your internal `buffer_` which is linked list. 
We are operating with `shared_pointers` to the messages in linked list and vector and copying elements will not cause a significant overhead since it will be copied only a few bytes for each shared pointer. And whole purpose of the circular buffer is to get data occasionally by request. There are no expectations that request for data from circular buffer will occur in tight loop.

Adhere to the common interface with `MessageCacheBuffer` will allow to better design structure of the classes. It will be more reasonable to create common pure virtual interface class or base class for message cache buffers.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /// Get number of elements in the buffer
+  size_t size();
+
+  /// Get buffer data
+  const std::list<buffer_element_t> & data();
+
+private:
+  std::list<buffer_element_t> buffer_;
+  uint64_t buffer_bytes_size_ {0u};
+  const uint64_t max_bytes_size_;
```
MichaelOrlov, line 62: In case when will be used default constructor `max_bytes_size_` will be uninitialized. It will lead to the undefined behavior for further usage of default constructed instance of the class.
Since it's declared as const will need to explicitly define default constructor without parameters and initialize it to zero.

emersonknapp:
Sorry to put this out there so late - but did you see https://github.com/cameron314/readerwriterqueue/blob/master/readerwritercircularbuffer.h - that the same lock-free queue library we were using has a circular implementation? It would be worth trying this out, as it's probably much more tuned and performant than anything we could build ourselves.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
emersonknapp: You could explicitly delete the default constructor to avoid this concern.

```
MessageCacheCircularBuffer() = delete;
```

rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace rosbag2_cpp
+{
+namespace cache
+{
+
+class ROSBAG2_CPP_PUBLIC CircularMessageCache
+{
+public:
+  explicit CircularMessageCache(uint64_t max_buffer_size);
```
emersonknapp, line 43: This seems like it should inherit from a common interface as `MessageCache` - so that it can be passed to the same interfaces downstream without modification. I can see that right now `MessageCache` has the `swap` function built-in to logic we don't want for the snapshot version - but this could also be a configuration on the `MessagCache`. In that case maybe we could have a generic `MessageCache<BufferType>`, with one or two configuration variables, instead of a totally separate type that the consumers will have to understand?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -0,0 +1,66 @@
...
+  buffer_.clear();
+  buffer_bytes_size_ = 0u;
+}
+
+size_t MessageCacheCircularBuffer::size()
+{
+  return buffer_.size();
+}
+
+const std::list<MessageCacheCircularBuffer::buffer_element_t> & MessageCacheCircularBuffer::data()
```
emersonknapp, line 60: I think I disagree that a list is a good choice for this data structure. In my view, unless you need to arbitrarily remove and insert items _from the middle of your data_, there is no good justification for a linked-list. Its memory layout is non-contiguous, so accessing items in order requires indirects for every single item. I think this would still be best be implemented as a vector - preallocated to the full size, with iterators/indices tracking the current "beginning" and "end" point within the vector, allowing for wraparound.

emersonknapp (changes requested):
See my comment about the linkedlist implementation, and about the Cache type hierarchy. We might be better off using somebody else's implementation rather than creating our own (though it is a fun exercise to do) - but if we do build it ourselves, a vector should provide much better performance than a linked list.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
camm73: Discussed with @emersonknapp offline and decided to move to a `std::deque` implementation

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
camm73: Going to go with deleting the default constructor since we'll require a non-zero `max-cache-size` to be provided by the user whenever snapshot mode is enabled.

MichaelOrlov: Ok guys, if you want to delete default constructor, how are you going to use this class with `shared_ptr`?
As far as I am understand the code is not going to compile in the way how it's uses.

camm73: Currently we're constructing each `std::shared_ptr` for MessageCacheCircularBuffer via `std::make_shared` and explicitly passing the `max-cache-size` as a parameter, so I don't think deleting the default constructor is an issue here. From testing, it also compiles without issue and passes all test cases.

MichaelOrlov: Ok, then.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
```diff
@@ -45,32 +46,31 @@ namespace cache
 * size.
 */
 class ROSBAG2_CPP_PUBLIC MessageCacheCircularBuffer
+  : public rosbag2_cpp::cache_interfaces::BaseCacheBufferInterface
```
MichaelOrlov, line 49: Perhaps it could be shortened as:
```suggestion
  : public CacheBufferInterface
```

rosbag2_cpp/include/rosbag2_cpp/cache_interfaces/base_message_cache_interface.hpp
```diff
@@ -0,0 +1,51 @@
...
+{
+
+class ROSBAG2_CPP_PUBLIC BaseMessageCacheInterface
+{
+public:
+  virtual ~BaseMessageCacheInterface() {}
+
+  virtual void push(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg) = 0;
+
+  virtual std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseCacheBufferInterface>
```
MichaelOrlov, line 36: Perhaps it could be shortened as:
```suggestion
  virtual std::shared_ptr<CacheBufferInterface>
```

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -115,7 +116,7 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   std::unique_ptr<Converter> converter_;
 
   bool use_cache_;
-  std::shared_ptr<rosbag2_cpp::cache::MessageCache> message_cache_;
+  std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseMessageCacheInterface> message_cache_;
```
MichaelOrlov, line 119: Perhaps it could be shortened as:
```suggestion
  std::shared_ptr<MessageCacheInterface> message_cache_;
```

rosbag2_cpp/src/rosbag2_cpp/cache/cache_consumer.cpp
```diff
@@ -23,7 +23,7 @@ namespace cache
 {
 
 CacheConsumer::CacheConsumer(
-  std::shared_ptr<MessageCache> message_cache,
+  std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseMessageCacheInterface> message_cache,
```
MichaelOrlov, line 26: Perhaps it could be shortened as:
```suggestion
  std::shared_ptr<rMessageCacheInterface> message_cache,
```

rosbag2_cpp/src/rosbag2_cpp/cache/circular_message_cache.cpp
```diff
@@ -38,7 +39,8 @@ void CircularMessageCache::push(std::shared_ptr<const rosbag2_storage::Serialize
   primary_buffer_->push(msg);
 }
 
-std::shared_ptr<MessageCacheCircularBuffer> CircularMessageCache::consumer_buffer()
+std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseCacheBufferInterface> CircularMessageCache::
```
MichaelOrlov, line 42: Perhaps it could be shortened as:
```suggestion
std::shared_ptr<rCacheBufferInterface> CircularMessageCache::
```

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache.cpp
```diff
@@ -87,7 +88,8 @@ void MessageCache::wait_for_buffer()
   std::swap(primary_buffer_, secondary_buffer_);
 }
 
-std::shared_ptr<MessageCacheBuffer> MessageCache::consumer_buffer()
+std::shared_ptr<rosbag2_cpp::cache_interfaces::BaseCacheBufferInterface> MessageCache::
+consumer_buffer()
```
MichaelOrlov, line 92: Perhaps it could be shortened as:
```suggestion
std::shared_ptr<CacheBufferInterface> MessageCache::consumer_buffer()
```

MichaelOrlov (changes requested):
It's certainly getting better, however I would like to request several changes, mostly a nitpicks but still very important:

- Please remove `Base` prefix from newly defined interfaces `BaseMessageCacheInterface` and `BaseCacheBufferInterface` it's really interfaces without any implementations in it.
Word `base` is extra in this case and introduce confusion. 

- Please move interfaces out of  the `cache_interfaces` and put them inside `rosbag2_cpp/cache/` folder and namespace,  for the sake of the simplification and avoiding long names in the code.

-  See my other comments in code.

camm73:
Thanks for the comments @MichaelOrlov. Just pushed changes for those

rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <vector>
+
+#include "rosbag2_cpp/visibility_control.hpp"
+#include "rosbag2_storage/serialized_bag_message.hpp"
+
+namespace rosbag2_cpp
+{
+namespace cache
+{
+using buffer_element_t = std::shared_ptr<const rosbag2_storage::SerializedBagMessage>;
```
emersonknapp, line 28: nit: I think this `using` should be inside the interface Class definition.

rosbag2_cpp/include/rosbag2_cpp/cache/cache_buffer_interface.hpp
```diff
@@ -0,0 +1,46 @@
...
+class ROSBAG2_CPP_PUBLIC CacheBufferInterface
+{
+public:
+  virtual ~CacheBufferInterface() {}
```
emersonknapp, line 32: Please add docstrings for class-level and methods of both `CacheBufferInterface` and `MessageCacheInterface`. These should clearly explain the expected behavior of these interfaces on the opaque API level. Implementing classes can add more documentation if it differs in some way from the virtual interface.

emersonknapp (changes requested):
Looking great - just asking for a quick namespace and docstring update then I think we're good to go

rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
MichaelOrlov: Nitpick. could be not addressed.
It's better to use `size_t` type for memory related variables to automatically accommodate for the underlying HW architecture.

MichaelOrlov (approved):
LGTM

emersonknapp (approved):

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15194/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9914/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12874/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15375/)

emersonknapp:
Oof just a couple of compiler warnings - so close! I'll try and push a commit fixing those, since I think @camm73 may not be reachable for a few days

emersonknapp:
Rebased and definitely fixed Clang warnings - took a best guess at Windows, it seems like something changed underneath us for `create_subscription`, the warning came from a package untouched in this PR. In the interest of going green, touched that file in `rosbag2_transport`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9915/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12875/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15376/)

emersonknapp:
🤞 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9928/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12888/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15390/)

PR #847, opened 2021-08-13
keyboard controls for pause/resume toggle and play-next:

lihui815:
Closes https://github.com/ros2/rosbag2/issues/55
Depends on https://github.com/ros2/ros2/pull/1184

adds keyboard controls for `ros2 bag play`

action-ros-ci-repos-supplemental: https://gist.githubusercontent.com/lihui815/5edccb5d36176ccc6a779fd8a26122db/raw/5b9ce9db3628fddce2329118bbf2b6ab6a9ce2a3/kh.repos

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -85,6 +85,9 @@ class PlayVerb(VerbExtension):
         parser.add_argument(
             '-d', '--delay', type=positive_float, default=0.0,
             help='Sleep duration before play (each loop), in seconds. Negative durations invalid.')
+        parser.add_argument(
+            '--disable-keyboard-controls', action='store_true',
+            help='disables keyboard controls for playback')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -120,6 +123,7 @@ class PlayVerb(VerbExtension):
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
         play_options.delay = args.delay
+        play_options.disable_keyboard_controls = args.disable_keyboard_controls
 
         player = Player()
         player.play(storage_options, play_options)
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -22,6 +22,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
+find_package(keyboard_handler REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
@@ -43,6 +44,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 ament_target_dependencies(${PROJECT_NAME}
+  keyboard_handler
   rcl
   rclcpp
   rcutils
@@ -75,6 +77,7 @@ ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(export_${PROJECT_NAME})
 ament_export_dependencies(
+  keyboard_handler
   rosbag2_cpp
   rosbag2_compression
   rosbag2_interfaces
@@ -158,6 +161,11 @@ function(create_tests_for_rmw_implementation)
 
   rosbag2_transport_add_gmock(test_record_all_no_discovery
     test/rosbag2_transport/test_record_all_no_discovery.cpp
+    LINK_LIBS rosbag2_transport
+    AMENT_DEPS test_msgs rosbag2_test_common)
+
+  rosbag2_transport_add_gmock(test_keyboard_controls
+    test/rosbag2_transport/test_keyboard_controls.cpp
     LINK_LIBS rosbag2_transport
     AMENT_DEPS test_msgs rosbag2_test_common)
 
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -20,6 +20,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "keyboard_handler/keyboard_handler.hpp"
 #include "rclcpp/duration.hpp"
 #include "rclcpp/qos.hpp"
 
@@ -48,6 +49,11 @@ public:
 
   // Sleep before play. Negative durations invalid. Will delay at the beginning of each loop.
   rclcpp::Duration delay = rclcpp::Duration(0, 0);
+
+  bool disable_keyboard_controls = false;
+  // keybindings
+  KeyboardHandler::KeyCode pause_resume_toggle_key = KeyboardHandler::KeyCode::SPACE;
+  KeyboardHandler::KeyCode play_next_key = KeyboardHandler::KeyCode::CURSOR_RIGHT;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -21,6 +21,9 @@
 #include <queue>
 #include <string>
 #include <unordered_map>
+#include <vector>
+
+#include "keyboard_handler/keyboard_handler.hpp"
 
 #include "moodycamel/readerwriterqueue.h"
 
@@ -76,6 +79,15 @@ public:
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
 
   ROSBAG2_TRANSPORT_PUBLIC
+  Player(
+    std::unique_ptr<rosbag2_cpp::Reader> reader,
+    std::shared_ptr<KeyboardHandler> keyboard_handler,
+    const rosbag2_storage::StorageOptions & storage_options,
+    const rosbag2_transport::PlayOptions & play_options,
+    const std::string & node_name = "rosbag2_player",
+    const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
+
+  ROSBAG2_TRANSPORT_PUBLIC
   virtual ~Player();
 
   ROSBAG2_TRANSPORT_PUBLIC
@@ -87,11 +99,11 @@ public:
   // Playback control interface
   /// Pause the flow of time for playback.
   ROSBAG2_TRANSPORT_PUBLIC
-  void pause();
+  virtual void pause();
 
   /// Start the flow of time for playback.
   ROSBAG2_TRANSPORT_PUBLIC
-  void resume();
+  virtual void resume();
 
   /// Pause if time running, resume if paused.
   ROSBAG2_TRANSPORT_PUBLIC
@@ -117,7 +129,7 @@ public:
   /// this method will wait until new element will be pushed to the queue.
   /// \return true if player in pause mode and successfully played next message, otherwise false.
   ROSBAG2_TRANSPORT_PUBLIC
-  bool play_next();
+  virtual bool play_next();
 
   /// \brief Advance player to the message with closest timestamp >= time_point.
   /// \details This is blocking call and it will wait until current message will be published
@@ -154,6 +166,13 @@ private:
 
   std::mutex reader_mutex_;
   std::unique_ptr<rosbag2_cpp::Reader> reader_ RCPPUTILS_TSA_GUARDED_BY(reader_mutex_);
+
+  void add_key_callback(
+    KeyboardHandler::KeyCode key,
+    const std::function<void()> & cb,
+    const std::string & op_name);
+  void add_keyboard_callbacks();
+
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::PlayOptions play_options_;
   moodycamel::ReaderWriterQueue<rosbag2_storage::SerializedBagMessageSharedPtr> message_queue_;
@@ -172,6 +191,10 @@ private:
   rclcpp::Service<rosbag2_interfaces::srv::GetRate>::SharedPtr srv_get_rate_;
   rclcpp::Service<rosbag2_interfaces::srv::SetRate>::SharedPtr srv_set_rate_;
   rclcpp::Service<rosbag2_interfaces::srv::PlayNext>::SharedPtr srv_play_next_;
+
+  // defaults
+  std::shared_ptr<KeyboardHandler> keyboard_handler_;
+  std::vector<KeyboardHandler::callback_handle_t> keyboard_callbacks_;
 };
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -113,11 +113,26 @@ Player::Player(
   const rosbag2_transport::PlayOptions & play_options,
   const std::string & node_name,
   const rclcpp::NodeOptions & node_options)
+: Player(std::move(reader),
+    // only call KeyboardHandler when using default keyboard handler implementation
+    std::shared_ptr<KeyboardHandler>(new KeyboardHandler()),
+    storage_options, play_options,
+    node_name, node_options)
+{}
+
+Player::Player(
+  std::unique_ptr<rosbag2_cpp::Reader> reader,
+  std::shared_ptr<KeyboardHandler> keyboard_handler,
+  const rosbag2_storage::StorageOptions & storage_options,
+  const rosbag2_transport::PlayOptions & play_options,
+  const std::string & node_name,
+  const rclcpp::NodeOptions & node_options)
 : rclcpp::Node(
     node_name,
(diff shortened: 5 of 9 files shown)


emersonknapp:
High level - I'm inclined to not have the keys be configurable via CLI options. Simplifies this code a bit - if we want to have some remapping later, maybe we do it via Node parameters for the Player Node. if anything is configurable, maybe it's enable/disable keyboard handler. Thoughts?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -466,4 +480,48 @@ bool Player::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr mess
...
+        play_options_.play_next_key <<
+        " for play-next!");
+    throw std::invalid_argument("Invalid key binding.");
+  }
+  // create callback
+  KeyboardHandler::callback_t callback = [this, pause_resume_toggle_key, play_next_key](
+    KeyboardHandler::KeyCode key_code,
+    KeyboardHandler::KeyModifiers /*key_modifiers*/)
+    {
+      if (key_code == pause_resume_toggle_key) {
```
emersonknapp, line 512: stylistically - i would lean towards separate callbacks

```
keyboard_handler_.add_key_press_callback(KeyCode::SPACE, std::bind(Player::toggle_paused, this));
keyboard_handler_.add_key_press_callback(KeyCode::n, std::bind(Player::play_next, this));
```

And if not doing that, would lean towards a `switch` rather than `elif`-chain

lihui815: k done

lihui815:
> High level - I'm inclined to not have the keys be configurable via CLI options. Simplifies this code a bit - if we want to have some remapping later, maybe we do it via Node parameters for the Player Node. if anything is configurable, maybe it's enable/disable keyboard handler. Thoughts?

Ok. I'll take that out and add disable to the cli

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -48,6 +49,11 @@ struct PlayOptions
 
   // Sleep before play. Negative durations invalid. Will delay at the beginning of each loop.
   rclcpp::Duration delay = rclcpp::Duration(0, 0);
+
+  bool disable_keyboard_controls = false;
+  // keybindings
+  KeyboardHandlerBase::KeyCode pause_resume_toggle_key = KeyboardHandlerBase::KeyCode::SPACE;
+  KeyboardHandlerBase::KeyCode play_next_key = KeyboardHandlerBase::KeyCode::S;
```
MichaelOrlov, line 56: IMO it would be more naive to use `KeyboardHandlerBase::KeyCode::CURSOR_RIGHT`  for play next command.

MichaelOrlov (commented):
I would prefer to hard code key codes for `play_next` and `pause` operations rather than providing them via player options. We have to many options and this key bindings pretty much trivial just need to agree upon on them.

Also have some concern about using time measurements in unit tests for validation. 
It's used to lead to the flakiness on CI.

ros2bag/ros2bag/verb/play.py
```diff
@@ -85,6 +85,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '-d', '--delay', type=positive_float, default=0.0,
             help='Sleep duration before play (each loop), in seconds. Negative durations invalid.')
+        parser.add_argument(
+            '--disable-keyboard-controls', action='store_true',
+            help='disables keyboard controls for pause/resume/play-next')
```
emersonknapp, line 90: might remove the `pause/resume/play-next` wording because i think we will add even more controls and i'd rather not enumerate them all here (or have to remember to update it)

rosbag2_transport/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ find_package(rosbag2_storage REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(shared_queues_vendor REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
+find_package(keyboard_handler REQUIRED)
```
emersonknapp, line 36: alphasort `find_package`

rosbag2_transport/include/rosbag2_transport/play_options.hpp
emersonknapp: That seems good to me too. I was thinking about arrow keys for faster/slower rate, but that could use UP/DOWN

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -42,6 +43,8 @@
 #include "rosbag2_transport/play_options.hpp"
 #include "rosbag2_transport/visibility_control.hpp"
 
+#include "keyboard_handler/keyboard_handler.hpp"
```
emersonknapp, line 46: alphasort include blocks

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -122,6 +134,28 @@ class Player : public rclcpp::Node
   ROSBAG2_TRANSPORT_PUBLIC
   bool play_next();
 
+  // \brief add keyboard callbacks
```
emersonknapp, line 137: I think the auto-generated brief of the function name would have added this much context already ;)

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -160,6 +194,14 @@ class Player : public rclcpp::Node
   rclcpp::Service<rosbag2_interfaces::srv::GetRate>::SharedPtr srv_get_rate_;
   rclcpp::Service<rosbag2_interfaces::srv::SetRate>::SharedPtr srv_set_rate_;
   rclcpp::Service<rosbag2_interfaces::srv::PlayNext>::SharedPtr srv_play_next_;
+
+  // defaults
+  std::shared_ptr<KeyboardHandlerBase> keyboard_handler_;
+  std::vector<KeyboardHandlerBase::callback_handle_t> keyboard_callbacks_;
+
+  int num_paused_ = 0;
```
emersonknapp, line 202: these counts seem purely for testing purposes (unless i'm misreading). You should be able to expect call counts via `EXPECT_CALL().Times()` for the Mock KeboardHandler methods without adding extra instrumentation to the production code. See https://google.github.io/googletest/gmock_cheat_sheet.html#UsingMocks

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -309,6 +336,8 @@ bool Player::play_next()
     return false;
   }
 
+  RCLCPP_INFO_STREAM(this->get_logger(), "Playing next message.");
```
emersonknapp, line 339: While we're adding these infos - maybe a `WARN` for the "not paused" case above? `Play next requested but not paused` something like that

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -471,4 +503,56 @@ bool Player::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr mess
...
 
+void Player::add_keyboard_callbacks()
+{
+  // skip if disabled
+  if (play_options_.disable_keyboard_controls) {
+    return;
+  }
+  RCLCPP_INFO_STREAM(this->get_logger(), "Adding keyboard callbacks.");
+  // check keybindings
+  if (play_options_.pause_resume_toggle_key == KeyboardHandlerBase::KeyCode::UNKNOWN) {
```
emersonknapp, line 514: Dedupe with generic utility?

```
void add_callback(key, cb, op_name) {
 if (key == UNKNOWN) {
    ERROR("INvalid key " << key << " for operation " << op_name);
    throw;
  }
  callbacks_.push_back(hangler->add_callback(cb, key));
  INFO("Press " << key << " to " << op_name);
}

void add_callbacks() {
  add_key_callback(play_options_.pause_resume_toggle_key, []() { dostuff; }, "pause/resume toggle");
  add_key_callback(play_options_.play_next_key, []() { dostuff; }, "play next message");
}
```

emersonknapp (changes requested):

rosbag2_transport/include/rosbag2_transport/player.hpp
lihui815: so i'm not trying to mock the player, but that the callbacks perform correctly for the real player. i just pulled it into a wrapper class.

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -120,7 +132,16 @@ class Player : public rclcpp::Node
...
+  virtual bool play_next();
+
+  ROSBAG2_TRANSPORT_PUBLIC
+  void add_key_callback(
+    KeyboardHandlerBase::KeyCode key,
+    const std::function<void()> & cb,
+    const std::string & op_name);
+
+  ROSBAG2_TRANSPORT_PUBLIC
+  void add_keyboard_callbacks();
```
emersonknapp, line 144: should these methods be private/protected? they feel like implementation details

emersonknapp (approved):
LGTM

MichaelOrlov (changes requested):
@lihui815 thank you for taking actions and fixing previous comments/requests.

However I would like to request more changes.

Generic comment:
 - Don't need to use base class `KeyboardHandlerBase` to refer to the key codes and basic functionality defined in it.
Please use `KeyboardHandler` alias directly instead.

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -87,11 +99,11 @@ class Player : public rclcpp::Node
   // Playback control interface
   /// Pause the flow of time for playback.
   ROSBAG2_TRANSPORT_PUBLIC
-  void pause();
+  virtual void pause();
```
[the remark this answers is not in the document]
MichaelOrlov: After some consideration I have to step back in my request to avoid virtual methods. It seems there are no good way to mock those methods without defining them as virtual. 
Let's keep them virtual.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -277,16 +300,18 @@ void Player::play()
 void Player::pause()
 {
   clock_->pause();
+  RCLCPP_INFO_STREAM(this->get_logger(), "Pausing play.");
```
emersonknapp, line 303: Minor stylistic nitpick (across diffs in this file), I prefer to omit `this` in C++. This codebase may not be fully consistent on it, but I think on average it's `this`-less

```suggestion
  RCLCPP_INFO_STREAM(get_logger(), "Pausing play.");
```

rosbag2_transport/test/rosbag2_transport/mock_keyboard_handler.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "keyboard_handler/keyboard_handler.hpp"
+
+// press one specific key code one second at a time
+class MockKeyboardHandlerImpl : public KeyboardHandler
```
emersonknapp, line 26: Nit: `Impl` is usually reserved for "hidden" implementations that are only used by a public interface - such as in the pimpl pattern. In this case, I would just call this `MockKeyboardHandler`

```suggestion
class MockKeyboardHandler : public KeyboardHandler
```

emersonknapp (approved):
Only tiny nitpicks

rosbag2_transport/test/rosbag2_transport/test_keyboard_controls.cpp
```diff
@@ -0,0 +1,165 @@
...
+    play_options_.play_next_key);
+
+  // play bag in thread
+  auto player = std::make_shared<CountPlayer>(
+    std::move(reader), std::shared_ptr<KeyboardHandler>(keyboard_handler),
+    storage_options_, play_options_);
+  std::thread player_thread = std::thread([player]() {player->play();});
+
+  EXPECT_THAT(player->is_paused(), false);
+  keyboard_handler->play_next();
```
MichaelOrlov, line 149: It doesn't make sense to call `play_next()` when player is not paused.
1. The test could be flaky. Since playback starting in the separate thread it could turns out that playback will already finish before we will try to test keyboard handling commands.
2. It's actually do nothing rather then returns with false if player is not in pause mode.

I think it would be better to call 
```cpp
keyboard_handler->pause_resume();
  EXPECT_THAT(player->is_paused(), true);
```
before
```cpp
std::thread player_thread = std::thread([player]() {player->play();});
```
and then one time after checking that `play_next()` key bynding working.

rosbag2_transport/test/rosbag2_transport/test_keyboard_controls.cpp
```diff
@@ -0,0 +1,165 @@
...
+  EXPECT_THAT(player->is_paused(), false);
+  keyboard_handler->pause_resume();
+  EXPECT_THAT(player->is_paused(), true);
+  keyboard_handler->play_next();
```
MichaelOrlov, line 153: @lihui815 Could you please redesign `MockKeyboardHandler` such a way that it will have only one extra method 
i.e. 
```cpp
void simulate_key_press(KeyboardHandler::KeyCode key_code, 
                        KeyboardHandler::KeyModifiers key_modifiers = KeyboardHandler::KeyModifiers::NONE);
```
instead of two others `keyboard_handler->play_next();` and `keyboard_handler->pause_resume();`

In tests you can use `play_options_.pause_resume_toggle_key` and `play_options_.play_next_key` as arguments for `keyboard_handler->simulate_key_press(key_code)`

It will be much cleaner to what is going on behind the scene when reading dedicated unit test.

MichaelOrlov (changes requested):
@lihui815 Implementation looks good to me. 
However I would encourage you to make some "nitpick" changes in unit tests.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15289/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10003/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12970/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15481/)

emersonknapp:
I think the `KEYBOARD_HANDLER_PUBLIC` delcarations caused it to be a `dllimport` on the Mock class. They're not needed within the test library, so removal should fix the windows build. Also addressed last review comments

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15295/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10010/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12976/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15487/)

rosbag2_transport/test/rosbag2_transport/test_keyboard_controls.cpp
```diff
@@ -135,24 +135,20 @@ TEST_F(RosBag2PlayTestFixture, test_keyboard_controls)
...
-    std::move(reader), std::shared_ptr<KeyboardHandler>(keyboard_handler),
+    std::move(reader), keyboard_handler,
     storage_options_, play_options_);
   std::thread player_thread = std::thread([player]() {player->play();});
 
   EXPECT_THAT(player->is_paused(), false);
-  keyboard_handler->play_next();
-  EXPECT_THAT(player->is_paused(), false);
-  keyboard_handler->pause_resume();
+  keyboard_handler->simulate_key_press(play_options_.pause_resume_toggle_key);
```
MichaelOrlov, line 147: @emersonknapp Thanks for fixing tests.

However I think it would better to call 
```cpp
  EXPECT_THAT(player->is_paused(), false);
  keyboard_handler->simulate_key_press(play_options_.pause_resume_toggle_key);
  EXPECT_THAT(player->is_paused(), true);
```
before 
```cpp
std::thread player_thread = std::thread([player]() {player->play();});
```
to avoid race conditions and flakiness in tests.

MichaelOrlov (approved):

MichaelOrlov:
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/15491/)

emersonknapp:
Cancelled above windows build - it was probably going to have the same warning as before. Pushed a fix commit

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15299/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10014/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12980/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15493/)

emersonknapp:
Tests were failing because test environment is not a TTY. Mocked the constructor to have fake implementations for the OS stuff.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15303/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10016/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15496/)

emersonknapp:
NOTE: still can't merge this until https://github.com/ros2/ros2/pull/1184 is merged

PR #848, opened 2021-08-16
Fix discovery silently stops after unknown msg type is found.

llapx:
Signed-off-by: Lei Liu <Lei.Liu.AP@sony.com>

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/typesupport_helpers.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/typesupport_helpers.hpp
@@ -34,6 +34,10 @@ std::shared_ptr<rcpputils::SharedLibrary>
 get_typesupport_library(const std::string & type, const std::string & typesupport_identifier);
 
 ROSBAG2_CPP_PUBLIC
+std::string get_typesupport_library_path(
+  const std::string & package_name, const std::string & typesupport_identifier);
+
+ROSBAG2_CPP_PUBLIC
 const rosidl_message_type_support_t *
 get_typesupport_handle(
   const std::string & type,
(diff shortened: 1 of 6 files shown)


fujitatomoya:
@Barry-Xu-2018 would you mind taking time for review since this comes from us. I'll do the same.

Barry-Xu-2018:
@fujitatomoya 

> would you mind taking time for review since this comes from us. I'll do the same.

Yeah. I will check this PR.

Barry-Xu-2018:
I have no comments on the latest updating.

jhdcs (approved):
Looks good to me, as long as it passes CI

emersonknapp:
Before I approve this PR - I am wondering if there is a simpler way to implement this. Does the existing behavior throw an exception when it tries to create a subscription to an unknown type, thus cancelling the discovery behavior? Can we simply catch this exception and print a warning message - instead of implementing complex filtering logic?

llapx:
> Does the existing behavior throw an exception when it tries to create a subscription to an unknown type, thus cancelling the discovery behavior?
Yes, when create a subscription with an unknown type, will cancelling the discovery process.

> Can we simply catch this exception and print a warning message
The simple way is try to skip create_subscription with unknown type and print warning like this:

```C++
...
std::shared_ptr<rclcpp::GenericSubscription>
Recorder::create_subscription(
{
  // catch the exception below 
  auto subscription = this->create_generic_subscription(
    topic_name,
    topic_type,
    qos,
    [this, topic_name](std::shared_ptr<rclcpp::SerializedMessage> message) {
      auto bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
...
```
but this will print warning in a loop if someone continue sending the unknown type messages, this may  flush the log.

> - instead of implementing complex filtering logic?
so I tried to filter out the unknown type messages, and keep the unknown types to make it only print warning once, I think this may good for performance.

PR #849, opened 2021-08-17
Race condition in Player::peek_next_message_from_queue()

MichaelOrlov:
There are possible race condition i.e. context switches between calling to
```cpp
rosbag2_storage::SerializedBagMessageSharedPtr * message_ptr = message_queue_.peek()
```
and
```cpp
if (message_ptr == nullptr && !is_storage_completely_loaded() && rclcpp::ok())
```
in [`Player::peek_next_message_from_queue()`](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/player.cpp#L287-L288)
It could lead to the situation that last message from bag will not be played when message queue is starving.

While this is a very rare case in real life it could cause flakiness on CI.

To fix this issue it will be enough to recheck 
```cpp
if (message_ptr == nullptr) {
    message_ptr = message_queue_.peek();
  }
```
before returning from method.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -295,6 +295,11 @@ rosbag2_storage::SerializedBagMessageSharedPtr * Player::peek_next_message_from_
       message_ptr = message_queue_.peek();
     }
   }
+  // Workaround for race condition between peek and is_storage_completely_loaded()
+  // Don't sync with mutex for the sake of the performance
+  if (message_ptr == nullptr) {
+    message_ptr = message_queue_.peek();
+  }
   return message_ptr;
 }
 


MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/f0e3524b0aba568b4c699df9c8b61013/raw/d014063bba4c8647528eb6a92e8085373304c84e/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/8860
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15068/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9800/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12753/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15235/)

emersonknapp (approved):

PR #850, opened 2021-08-18
Implement snapshot mechanism and corresponding ROS Service

camm73:
Part of #663
Depends on #844

This PR implements the mechanism for taking a snapshot into the `SequentialWriter`, `Writer`, and `BaseWriterInterface` classes. It implements the `take_snapshot` function as a pure virtual function in `BaseWriterInterface` to ensure that all future Writers are built with snapshot mode compatibility.

Furthermore, a ROS service `~/snapshot` is created within the `Recorder` node to provide an endpoint for triggering snapshots.

Finally, `StorageOptions` was updated to add a `snapshot_mode` field which is required by `SequentialWriter` to determine when the circular snapshot buffers should be created.


rosbag2_transport/test/rosbag2_transport/test_record_services.cpp
```diff
@@ -0,0 +1,147 @@
...
+  rclcpp::Subscription<test_msgs::msg::Strings>::SharedPtr topic_sub_;
+};
+
+TEST_F(RecordSrvsTest, trigger_snapshot)
+{
+  auto & writer = recorder_->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+  EXPECT_THAT(mock_writer.get_messages().size(), Eq(0u));
+  successful_call<Snapshot>(cli_snapshot_);
```
MichaelOrlov, line 145: I would recommend to add extra check with wait for some timeout that `mock_writer.snapshot_buffer_.size() > 0` before calling `successful_call<Snapshot>(cli_snapshot_);`
I am afraid that it could be a race condition when we returned from sending service request for `snapshot` while no one test messages has been recorded yet. It could potentially cause flakiness in test.

MichaelOrlov (changes requested):
@camm73 Thank you for you contribution.
I assume that circular message cache and inner circular buffer will be implemented and reviewed in #844.
I didn't review them in the scope of this PR.
While entire implementation looks good for me, I would suggest a few changes in code.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -118,6 +125,8 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   std::shared_ptr<rosbag2_cpp::cache::MessageCache> message_cache_;
   std::unique_ptr<rosbag2_cpp::cache::CacheConsumer> cache_consumer_;
 
+  std::shared_ptr<rosbag2_cpp::cache::CircularMessageCache> circular_message_cache_;
```
emersonknapp, line 128: I don't think it makes sense to have two separate caches here - it makes a lot more sense to have a single `MessageCache` interface that is either polymorphic or templated to do what we need.

rosbag2_cpp/include/rosbag2_cpp/writer_interfaces/base_writer_interface.hpp
```diff
@@ -45,6 +45,8 @@ class ROSBAG2_CPP_PUBLIC BaseWriterInterface
   virtual void remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type) = 0;
 
   virtual void write(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message) = 0;
+
+  virtual void take_snapshot() = 0;
```
emersonknapp, line 49: This needs a docstring please.

Maybe it should return a `bool`? If a Writer is not configured to do snapshot-writing, then it would make sense to provide some indication that "no i won't do this" or "nothing happens"

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -163,6 +174,26 @@ void SequentialWriter::close()
   storage_factory_.reset();
 }
 
+void SequentialWriter::take_snapshot()
+{
+  circular_message_cache_->swap_buffers();
```
emersonknapp, line 179: Should we check first that the output buffer is done being consumed? The behavior might be pretty strange if it gives back a partially-consumed buffer, where we will still have those older messages, that could then be written out alongside newer messages in a future snapshot.

I am thinking that if we have to prioritize a behavior, I think it would be "fully consume previous snapshot" before trying to do the new one.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -163,6 +174,26 @@ void SequentialWriter::close()
...
+  auto msg_list = circular_message_cache_->consumer_buffer()->data();
+
+  std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>>
+  msg_vector(msg_list.size());
+
+  // Copy list into a vector needed by the storage writer
+  std::copy(msg_list.begin(), msg_list.end(), msg_vector.begin());
```
emersonknapp, line 186: I commented on this in the CircularBuffer PR, I think we should just be using a vector, the linkedlist is not giving us value, and this copy isn't necessarily cheap.

rosbag2_interfaces/srv/Snapshot.srv
```diff
@@ -0,0 +1 @@
+---
```
emersonknapp, line 1: This should probably have a `bool success` return

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -46,6 +46,10 @@ struct StorageOptions
   // Storage specific configuration file.
   // Defaults to empty string.
   std::string storage_config_uri = "";
+
+  // Enable snapshot mode.
+  // Defaults to disabled.
+  bool snapshot_mode = false;
```
emersonknapp, line 52: This should be a `RecordOptions` setting - it is not a property of the Storage , which knows nothing about this (`Play` also uses `StorageOptions`).

emersonknapp (changes requested):
A few structural comments - mostly looks good

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
camm73: Decided to leave this in `StorageOptions` for now since `RecordOptions` is not accessible within `SequentialWriter`. May move this to `RecordOptions` in a future refactor

emersonknapp (commented):
Adding some comments. I think I might be taking over this PR so they may be simply "notes to self"

emersonknapp (approved):

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15223/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9943/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12903/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15406/)

emersonknapp:
@MichaelOrlov I have reworked this PR a bit to make the double-buffering logic easier to follow (I think). Would you mind taking another look at this modified implementation and let me know what you think?

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache.hpp
```diff
@@ -72,14 +76,12 @@ class ROSBAG2_CPP_PUBLIC MessageCache
...
-  /// Set the cache to consume-only mode for final buffer flush before closing
-  void finalize() override;
+  /// Gets a consumer buffer.
+  /// In this greedy implementation, swap buffers before providing the buffer.
+  std::shared_ptr<CacheBufferInterface> consumer_buffer() override;
 
-  /// Notify that flushing is complete
-  void notify_flushing_done() override;
+  /// Notify that consumer_buffer has been fully used. Unlock.
+  void return_consumer_buffer() override;
```
lihui815, line 84: nit: don't like using "return" unless actually returning, confusing. release?

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15294/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10009/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12975/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15486/)

MichaelOrlov:
@emersonknapp I've just recently got to your ping in this issue. 
I am sorry I didn't catch it in time early.
I see that you made a significant overhaul in `sequential_writer` and `message_cache` classes.  
Actually I didn't expect it in the scope of this PR, but nevertheless thanks for the improvements and clean up. 
I will do post merge review to make sure that we didn't missed something or brake some logic.

MichaelOrlov:
@emersonknapp It seems I found a breakage in logic. Split doesn't work as before. i.e. after finalizing bag split, messages in next bag will not be written.
I have a quick fix in mind, however would like to spend some more time to find more optimal solution to address problem related to the high risk of  dropping messages during the split.

MichaelOrlov:
@emersonknapp Also found out that in case of using `CircularMessageCache` it will be busy loop in `CacheConsumer` class.
The consequences will be that in `Snapshot mode` rosbag2 recorder will consume 100% of CPU resources on one core.
Also thinking how to proper address this issue. Don't want to do a quick patchy fix.

PR #854, opened 2021-08-24
Add stopRecording into rosbag2_py

afonsofonbraga:
This PR implements a command to stop recording a rosbag without the need of using ctrl+c

In my development, I was trying to find a way to stop recording the rosbag using the python API without the need of closing the terminal. I've tried to kill the process using the PID I received from the python Recorder() the value differs from the PID used inside the rosdbag2_transport wrapper. In this case, I've created a method that when it is called it sends a SIGTERM to terminate the process without the need of closing my application.

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -153,13 +153,13 @@ public:
 class Recorder
 {
 private:
-  rclcpp::executors::SingleThreadedExecutor * exec_;
+  std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_;
 
 public:
   Recorder()
   {
     rclcpp::init(0, nullptr);
-    exec_ = new rclcpp::executors::SingleThreadedExecutor;
+    exec_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
     std::signal(
       SIGTERM, [](int /* signal */) {
         rclcpp::shutdown();
@@ -168,7 +168,6 @@ public:
 
   virtual ~Recorder()
   {
-    delete exec_;
     rclcpp::shutdown();
   }
 
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -153,12 +153,13 @@ public:
 class Recorder
 {
 private:
-  std::promise<void> cancel_;
+  rclcpp::executors::SingleThreadedExecutor * exec_;
 
 public:
   Recorder()
   {
     rclcpp::init(0, nullptr);
+    exec_ = new rclcpp::executors::SingleThreadedExecutor;
     std::signal(
       SIGTERM, [](int /* signal */) {
         rclcpp::shutdown();
@@ -167,6 +168,7 @@ public:
 
   virtual ~Recorder()
   {
+    delete exec_;
     rclcpp::shutdown();
   }
 
@@ -202,21 +204,18 @@ public:
     auto recorder = std::make_shared<rosbag2_transport::Recorder>(
       std::move(writer), storage_options, record_options);
     recorder->record();
-    rclcpp::executors::SingleThreadedExecutor exec;
-    exec.add_node(recorder);
 
-    cancel_ = std::promise<void>();
-    auto future = cancel_.get_future();
+    exec_->add_node(recorder);
     // Release the GIL for long-running record, so that calling Python code can use other threads
     {
       py::gil_scoped_release release;
-      exec.spin_until_future_complete(future);
+      exec_->spin();
     }
   }
 
   void cancel()
   {
-    cancel_.set_value();
+    exec_->cancel();
   }
 };
 
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -152,6 +152,9 @@ public:
 
 class Recorder
 {
+private:
+  std::promise<void> cancel_;
+
 public:
   Recorder()
   {
@@ -201,13 +204,19 @@ public:
     recorder->record();
     rclcpp::executors::SingleThreadedExecutor exec;
     exec.add_node(recorder);
-    auto spin_thread = std::thread(
-      [&exec]() {
-        exec.spin();
-      });
 
-    exec.cancel();
-    spin_thread.join();
+    cancel_ = std::promise<void>();
+    auto future = cancel_.get_future();
+    // Release the GIL for long-running record, so that calling Python code can use other threads
+    {
+      py::gil_scoped_release release;
+      exec.spin_until_future_complete(future);
+    }
+  }
+
+  void cancel()
+  {
+    cancel_.set_value();
   }
 };
 
@@ -270,5 +279,6 @@ PYBIND11_MODULE(_transport, m) {
   py::class_<rosbag2_py::Recorder>(m, "Recorder")
   .def(py::init())
   .def("record", &rosbag2_py::Recorder::record)
+  .def("cancel", &rosbag2_py::Recorder::cancel)
   ;
 }
--- a/rosbag2_py/test/test_transport.py
+++ b/rosbag2_py/test/test_transport.py
@@ -82,8 +82,6 @@ def test_record_cancel(tmp_path):
         pub.publish(msg)
 
     recorder.cancel()
-    msg.data = 'Spining one last time'
-    pub.publish(msg)
 
     time_counter = 0
     while (not os.path.exists(Path(bag_path) / 'metadata.yaml') or
--- a/rosbag2_py/test/test_transport.py
+++ b/rosbag2_py/test/test_transport.py
@@ -17,10 +17,9 @@ from pathlib import Path
 import os
 import sys
 import threading
-import time
 
 
-from common import get_rosbag_options
+from common import get_rosbag_options, wait_for
 import rclpy
 from rclpy.qos import QoSProfile
 from std_msgs.msg import String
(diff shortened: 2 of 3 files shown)


emersonknapp:
This is a good consideration, but I do not believe that it is the right approach.

It is perhaps a bit awkward that `rosbag2_py.Recorder` does an `rclcpp::shutdown()` in its destructor - maybe you want to run that node without it managing the global context. However - we definitely shouldn't have to shutdown an entire process to cancel recording - this is overkill and nonportable.

I think you can accomplish your goal of cancelling the `Recorder` by having the embedded C++ process call `rclcpp::shutdown()`  - instead of doing an OS signal.

emersonknapp:
Another thought - can you cause the recording to shut down just be forcing the Recorder object to go out of scope? I guess the problem right now is that `.record()` would be waiting on `spin_thread.join()` which won't exit until the executor stops spinning.

This could possibly be made better by using `spin_until_future_complete` with a Future that we can cancel manually via a `stop_recording` API

afonsofonbraga:
Hi, thanks for the answer. I'm trying to implement it by using the ``` spin_until_future_complete ```. I'm still new to the ROS2 and I've done some research but I couldn't think of a solution without creating a service for it and I think it will add too much complexity. Could you give me more details of how to implement this feature the way you have mentioned?

emersonknapp (changes requested):
Adding my review to clear from my queue until next revision

afonsofonbraga:
Just finished the test for the record and cancel command. After calling the cancel, I had to send another rosmessage to release the ```spin_until_future_complete```. I'm not so sure if this is the best approach but it is working fine.

rosbag2_py/test/test_transport.py
```diff
@@ -12,10 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import os
 import sys
+import threading
+from pathlib import Path
 
+from common import get_rosbag_options  # noqa
```
emersonknapp, line 21: Why is this noqa?

rosbag2_py/test/test_transport.py
```diff
@@ -40,3 +46,50 @@ def test_options_qos_conversion():
...
+        daemon=True)
+    record_thread.start()
+
+    node = rclpy.create_node('test_record_cancel')
+    executor = rclpy.executors.SingleThreadedExecutor()
+    executor.add_node(node)
+    pub = node.create_publisher(String, 'chatter', 10)
+
+    i = 0
+    import time
```
emersonknapp, line 80: move import to top of file

rosbag2_py/test/test_transport.py
```diff
@@ -40,3 +46,50 @@ def test_options_qos_conversion():
...
+
+    record_options = rosbag2_py.RecordOptions()
+    record_options.all = True
+    record_options.is_discovery_disabled = False
+    record_options.rmw_serialization_format = ""
+    record_options.topic_polling_interval = datetime.timedelta(milliseconds=100)
+
+    record_options.compression_mode = 'file'
+    record_options.compression_queue_size = 1
+    record_options.compression_threads = 0
```
emersonknapp, line 65: I'd remove the compression options, the defaults are reasonable, and `file` compression adds extra complexity we don't need in this test

```suggestion
```

rosbag2_py/test/test_transport.py
```diff
@@ -40,3 +46,50 @@ def test_options_qos_conversion():
...
+        msg = String()
+        msg.data = 'Hello World: {0}'.format(i)
+        i += 1
+        pub.publish(msg)
+        time.sleep(0.1)
+
+    recorder.cancel()
+    msg = String()
+    msg.data = 'Spining one last time'
+    pub.publish(msg)
```
emersonknapp, line 91: why is it necessary to "spin one more time"? And, instead of sleeping for an arbitrary time, could you instead wait for the expected files to exist, and fail on some timeout? That should be more reliable than hoping that "0.1 seconds is probably long enough"

emersonknapp (changes requested):
Implementation looks good :), just some small comments on the test

rosbag2_py/test/test_transport.py
afonsofonbraga: I noticed if you are not publishing anything when you call the ```recorder.cancel()``` the ```spin_until_future_complete``` stays blocked until there is another spin. That's why I added one last message to be sent so the thread will finish properly creating the metafile.yaml . Removing the sleep of line 86 might also prevent this behavior because when the cancel command is activated there might be some messages in the buffer. However, I think is wise to leave the spining one last time message. 
This behavior can be reproduced if you are calling the recorder and cancel without any publishers.

emersonknapp: This _should_ work without new spinning/events coming in. The `spin_until_future_complete` loop is supposed to wake "immediately" when the future is marked complete, in both rclpy and rclcpp.

I'm going to run a quick experiment locally. I wonder if it has something to do with the GIL or something.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -201,13 +204,19 @@ class Recorder
...
+    // Release the GIL for long-running record, so that calling Python code can use other threads
+    {
+      py::gil_scoped_release release;
+      exec.spin_until_future_complete(future);
+    }
+  }
+
+  void cancel()
+  {
+    cancel_.set_value();
```
emersonknapp, line 219: I was able to reproduce the problem locally as well - It's not clear why setting the value on the promise does not wake the executor. However, did we already rule out the following? 

Make `exec_` a SingleThreadedExecutor member variable, and do the following (could get rid of `cancel_`)

```suggestion
    // Release the GIL for long-running record, so that calling Python code can use other threads
    {
      py::gil_scoped_release release;
      exec_.spin();
    }
  }

  void cancel()
  {
    exec_.cancel();
```

afonsofonbraga:
Thanks for taking the time to test it and I just made some changes following your suggestion.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -153,12 +153,13 @@ class Player
 class Recorder
 {
 private:
-  std::promise<void> cancel_;
+  rclcpp::executors::SingleThreadedExecutor * exec_;
```
emersonknapp, line 156: please avoid using bare pointers in modern C++ - if we were to use a pointer, I would suggest `unique_ptr`. In this case, though, it should be perfectly safe to use a non-pointer member object.
```suggestion
  rclcpp::executors::SingleThreadedExecutor exec_;
```

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -167,6 +168,7 @@ class Recorder
 
   virtual ~Recorder()
   {
+    delete exec_;
```
emersonknapp, line 171:
```suggestion
```

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -153,12 +153,13 @@ class Player
...
 {
 private:
-  std::promise<void> cancel_;
+  rclcpp::executors::SingleThreadedExecutor * exec_;
 
 public:
   Recorder()
   {
     rclcpp::init(0, nullptr);
+    exec_ = new rclcpp::executors::SingleThreadedExecutor;
```
emersonknapp, line 162:
```suggestion
    exec_ =  std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
```

rosbag2_py/test/test_transport.py
```diff
@@ -40,3 +47,49 @@ def test_options_qos_conversion():
...
+    msg = String()
+
+    while rclpy.ok() and i < 10:
+        msg.data = 'Hello World: {0}'.format(i)
+        i += 1
+        pub.publish(msg)
+
+    recorder.cancel()
+
+    time_counter = 0
```
emersonknapp, line 86: This is a little bit of an odd way to do timeout waiting - generally you want to add a timeout in terms of time, rather than a "number of times to sleep". In `rosbag2_test_common` we have a utility function `spin_and_wait_for` - https://github.com/ros2/rosbag2/blob/master/rosbag2_test_common/include/rosbag2_test_common/wait_for.hpp#L25

I think it would make sense to give yourself a utility function in `common.py` called simply `wait_for`, that could look like this:

```
def wait_for(condition, timeout: rclpy.Duration):
  clock = rclpy.Clock(RCL_STEADY_TIME)
  start = clock.now()
  while !condition():
    if clock.now() - start > timeout:
      return false
    # could make this configurable as well if you like - it's just a rate throttle so this loop doesn't eat full CPU
    time.sleep(0.1)
  return true
```

Then you could write the test as nice and simple, like:

```
metadata_path = bag_path / 'metadata.yaml'
db3_path = bag_path /  'test_record_cancel_0.db3'
assert wait_for(lambda: metadata_path.is_file() and db3_path.is_file(), timeout=rclpy.Duration(seconds=3)) 
```

That keeps the implementation details out and makes the test very self-documenting

emersonknapp (changes requested):
I think we're about done with one last quick round to clean things up. Thanks for going through the revisions!

afonsofonbraga:
All suggestions were applied to the code. Thanks for taking your time to help in this PR.

emersonknapp (approved):
LGTM - I will run CI soon

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15212/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9932/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12892/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15394/)

Issue #831, opened 2021-07-30
Generic Bag(s)->Bag(s) Rewriter


emersonknapp:
## Description

Create an API pipeline (possibly hooked up to CLI) that takes in any number of rosbags, and outputs a possibly different number of rosbags, using any storage(etc)Options for all inputs and outputs.

## Related Issues

* Covers much of the functionality discussed in https://github.com/ros2/rosbag2/issues/717, but isn't an exact replica
* https://github.com/ros2/rosbag2/issues/668 join or stitch, handled by this overall feature set
* https://github.com/ros2/rosbag2/issues/599 "post-splitting"
 
## Completion Criteria

See description.

PR #864, opened 2021-09-04
Add filter verb

wep21:
I added filter verb to filter topics with python expression.

### DIFF
--- a/ros2bag/ros2bag/verb/filter.py
+++ b/ros2bag/ros2bag/verb/filter.py
@@ -1,0 +1,81 @@
+# Copyright 2021 Tier IV, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


wep21:
@emersonknapp Could you review this?

emersonknapp (commented):
While I think this is a useful feature - I think that it can be built a little more generically.

@gbiggs is currently working on https://github.com/ros2/rosbag2/issues/831. This `ros2 bag convert` feature could take the filter as an extra argument, rather than having a whole separate verb `ros2 bag filter`.

I would like to hold off on reviewing this PR until we have that base feature in place, so that we could build this as a simple extension to it. Because of this, I won't leave my full comments here (I would have commented on hardcoding `cdr` for the StorageOptions, for example. And I would say that we need tests, and control over the StorageOptions for the output bag. However, if the generic feature is built first, then this won't have to worry about most of that)

I'll try to see about the timeline on that feature. To accelerate development we may put in the singular bag->bag version first, then move on to the multi-bag implementation.

wep21:
Closed by #921.

emersonknapp:
A note that #921 doesn't support arbitrary Python expression. It only supports filtering in the sense of choosing which topic names to forward.

This PR is more related to https://github.com/ros2/rosbag2/issues/885, which is still open at this point.

Issue #856, opened 2021-08-25
Visiting invalid pointer addresss when using "MESSAGE" compression mode to write data


0xbabeface:
## Description
I wrote testing codes like: 
```cpp
rosbag2_compression::CompressionOptions compression_options{
    "zstd",
    rosbag2_compression::CompressionMode::MESSAGE,
    10,
    1,
};
writer_ = std::make_unique<rosbag2_cpp::Writer>(
    std::make_unique<rosbag2_compression::SequentialCompressionWriter>(compression_options));
// Do somthing to initialize the writer_, like getting the path and opening it with some options
for (;;) {
  writer_->write(msg_, topic_, rclcpp::Clock().now());
}
```
The program will throw out a "std::bad_alloc" exception when writing bag.

### Expected Behavior
Work normally.

### Actual Behavior
In the function in rosbag2_cpp/writer.cpp:
```cpp
void Writer::write(
  const rclcpp::SerializedMessage & message,
  const std::string & topic_name,
  const std::string & type_name,
  const rclcpp::Time & time)
{
  auto serialized_bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
  serialized_bag_message->topic_name = topic_name;
  serialized_bag_message->time_stamp = time.nanoseconds();
  
  // temporary store the payload in a shared_ptr.
  // add custom no-op deleter to avoid deep copying data.
  serialized_bag_message->serialized_data = std::shared_ptr<rcutils_uint8_array_t>(
    const_cast<rcutils_uint8_array_t *>(&message.get_rcl_serialized_message()),
    [](rcutils_uint8_array_t * /* data */) {});

  // ---
  // The write behaviour will result in `write` function in SequentialCompressionWriter:   writer_impl_->write(message);
  // And serialized_bag_message will be pushed in to a queue to zstd compressor in another thread
  // However, message in this thread may have been destructed when visiting serialized_bag_message.
  // So that the pointor `serialized_bag_message->serialized_data` is hanged 
  // and the compressor visits pointer with a invalid address
  // ---
  return write(
    serialized_bag_message, topic_name, type_name, rmw_get_serialization_format());
}

```

## To Reproduce
Maybe some codes for testing are needed.

## System (please complete the following information)
 - OS: focal
 - ROS 2 Distro: galactic
 - Version: release

## Additional context
I made a deepcopy from the input parameter `message` to a new piece of memory. Then the program works.
[...]

PR #866, opened 2021-09-06
Fix a bug on invalid pointer address when using "MESSAGE" compressio…

Barry-Xu-2018:
Address #856


Barry-Xu-2018:
This modification affects the performance for a case `No compression and cache size is set to 0`.  
For this case, message need not be duplicated.   
If not change `class Writer`, I have no simple way to get the information of cache size in Writer::write.  
So do duplication for all cases.

Barry-Xu-2018:
I made new PR #867  to add new interface to help judgement whether message need to be duplicated.

emersonknapp (approved):
This looks good - but is there any way to add a unit test for this case?

Barry-Xu-2018:
@emersonknapp   

I submit a new commit (45c9a78)
In this commit, I mark the old interface as deprecated and add a new write interface instead of that.  
I think we should not make a lot of change for fixing the problem from deprecated interface.

Barry-Xu-2018:
After checking, the cause is relevant to set `-Werror=deprecated-declarations`.  
In code, I marked the old interface as `deprecated`.

rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -144,14 +144,38 @@ class ROSBAG2_CPP_PUBLIC Writer final
...
+   * Write a serialized message to a bagfile.
+   * The topic will be created if it has not been created already.
+   *
+   * \warning after calling this function, the serialized data will no longer be managed by message.
+   *
+   * \param message rclcpp::SerializedMessage The serialized message to be written to the bagfile
+   * \param topic_name the string of the topic this messages belongs to
+   * \param type_name the string of the type associated with this message
+   * \param time The time stamp of the message
+   * \throws runtime_error if the Writer is not open or duplicating message is failed.
```
fujitatomoya, line 171: Is this needed?

```suggestion
   * \throws runtime_error if the Writer is not open.
```

Barry-Xu-2018: Thanks. It should be removed.

emersonknapp (approved):

rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -144,14 +144,38 @@ class ROSBAG2_CPP_PUBLIC Writer final
...
    * \param time The time stamp of the message
-   * \throws runtime_error if the Writer is not open.
+   * \throws runtime_error if the Writer is not open or duplicating message is failed.
    */
+  [[deprecated(
+    "Use write(std::shared_ptr<rclcpp::SerializedMessage> message," \
+    " const std::string & topic_name," \
+    " const std::string & type_name," \
+    " const rclcpp::Time & time) instead."
+  )]]
```
emersonknapp, line 154: Line 199 in this file is causing the Windows warning - since the "non-serialized ROS message" template function calls this "reference version".

I think that the solution is to create a shared_ptr to the SerializedBagMessage, instead of a local instance.

Something like:

```
  template<class MessageT>
  void write(
    const MessageT & message,
    const std::string & topic_name,
    const rclcpp::Time & time)
  {
    auto serialized_msg = std::make_shared<rclcpp::SerializedMessage>();
    rclcpp::Serialization<MessageT> serialization;
    serialization.serialize_message(&message, serialized_msg.get());
    return write(serialized_msg, topic_name, rosidl_generator_traits::name<MessageT>(), time);
```

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/01dc4976a14df75a802c9e347d0309f8/raw/e057b1e91a54bdfd8b8b165e3d5fad7a61652ea0/ros2.repos
BUILD args: --packages-up-to rosbag2_cpp rosbag2_transport rosbag2 rosbag2_tests
TEST args: --packages-select rosbag2_cpp rosbag2_transport rosbag2 rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9190
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15430/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10133/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13116/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15649/)

PR #867, opened 2021-09-07
Topic fix writer interface issue 2

Barry-Xu-2018:
Address #856  

Comparing with fix #866, add new interface `is_async_process_message` to BaseWriterInterface.  
This new interface can help to get information about if the processing message is asynchronous.    
So we can know whether message need to be duplicated.


Barry-Xu-2018:
Change interface name from `is_async_process_message` to `request_owned_serialized_data`.

jhdcs (commented):
This looks fairly good to me. Just a few mild suggestions. The one I'm most interested in seeing is the preservation of the comment in `writer.cpp`, the others are mostly just mild style stuff.

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -112,11 +114,44 @@ void Writer::write(
...
-  serialized_bag_message->serialized_data = std::shared_ptr<rcutils_uint8_array_t>(
-    const_cast<rcutils_uint8_array_t *>(&message.get_rcl_serialized_message()),
-    [](rcutils_uint8_array_t * /* data */) {});
+  if (writer_impl_->request_owned_serialized_data()) {
+    // For asynchronous processing message, duplicate message
+    serialized_bag_message->serialized_data = std::shared_ptr<rcutils_uint8_array_t>(
+      new rcutils_uint8_array_t,
+      [](rcutils_uint8_array_t * msg) {
+        auto fini_return = rcutils_uint8_array_fini(msg);
+        delete msg;
```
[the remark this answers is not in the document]
Barry-Xu-2018: Not work.  
`rcutils_uint8_array_fini()` is C function.   And In `rcutils_uint8_array_fini()`, not free msg.

jhdcs (approved):
Looks good to me!

Barry-Xu-2018:
@jhdcs 

I added a new commit since I found below case.  
If compression mode is file, `SequentialCompressionWriter` will call SequentialWriter::write.

https://github.com/ros2/rosbag2/blob/891e08128e2d6ff36452871abda2cd776b8a1566/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp#L306-L315

So for this case, need to call  SequentialWriter::request_owned_serialized_data() to check.

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -112,11 +114,46 @@ void Writer::write(
   serialized_bag_message->topic_name = topic_name;
   serialized_bag_message->time_stamp = time.nanoseconds();
 
-  // temporary store the payload in a shared_ptr.
-  // add custom no-op deleter to avoid deep copying data.
-  serialized_bag_message->serialized_data = std::shared_ptr<rcutils_uint8_array_t>(
-    const_cast<rcutils_uint8_array_t *>(&message.get_rcl_serialized_message()),
-    [](rcutils_uint8_array_t * /* data */) {});
+  if (writer_impl_->request_owned_serialized_data()) {
```
emersonknapp, line 117: While we are adding these workarounds, can you also mark this function signature as `[[deprecated()]]` in the header? Based on the need to duplicate every message like this, I think we want to discourage its use and remove it later

rosbag2_cpp/include/rosbag2_cpp/writer_interfaces/base_writer_interface.hpp
```diff
@@ -45,6 +45,8 @@ class ROSBAG2_CPP_PUBLIC BaseWriterInterface
   virtual void remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type) = 0;
 
   virtual void write(std::shared_ptr<rosbag2_storage::SerializedBagMessage> message) = 0;
+
+  virtual bool request_owned_serialized_data() = 0;
```
emersonknapp, line 49: Non-blocking: This feels very much like an under-the-hood implementation detail. It would be nice if we could avoid exposing it to the global interface. However - I don't know if we can do that right now. We may just need to wait until we remove the `write(SerializedBagMessage &)` interface.

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
Barry-Xu-2018: But this interface is called now by below template interface  
https://github.com/ros2/rosbag2/blob/622b0ada5bf78da81a24e33f00c0ff89097c80c0/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L149-L170

So I think we also need to modify this template interface.  

I want to use this new interface to replace this interface to be deprecated.  
https://github.com/Barry-Xu-2018/rosbag2/blob/5f18ae59f1358ae045e1e4b14afadbc333a8737d/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L149-L166

New interface also can be called instead of below codes. 
https://github.com/ros2/rosbag2/blob/622b0ada5bf78da81a24e33f00c0ff89097c80c0/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L222-L246

What do you think ?

rosbag2_cpp/include/rosbag2_cpp/writer_interfaces/base_writer_interface.hpp
Barry-Xu-2018: Currently, I not find better way instead of adding hook function to global interface.

Barry-Xu-2018:
Friendly ping @emersonknapp

Barry-Xu-2018:
Do rebase

Barry-Xu-2018:
There is new way to avoid adding a new interface.  
I will try to implement it and create another PR.

Barry-Xu-2018:
@emersonknapp @jhdcs 

After deep consider, I think we should not modify so much codes for a deprecated interface.  
Please look my PR in #866.  (https://github.com/ros2/rosbag2/pull/866/commits/45c9a78939593a1fd1fa77ca9f5b9c7c9b792fef)
I add new write interface instead of old one. 
For the old interface, we didn't have to make a lot of code changes to fix the problems found.

PR #873, opened 2021-09-23
Use Reader's seek() method for seeking/jumping in Player

### LINKED ISSUE #824: Seek/Jump API for Reader/Storage

lihui815:
Closes #824 

The current `Player->seek()` was implemented before storage and by extension the reader exposed `seek` interface (#824). This PR updates the implementation to use the Reader's `seek()` method, which reduces the implementation complexity.

Other details:
- Reader's `seek()` method does not require closing and reopening. Thus it's preferable to just keep the reader open from the constructor, and close it in the destructor. Other methods can just assume an open reader and set `seek()` as needed.
- Removed the `release_reader()` method as the player is basically useless after releasing the reader anyway. This removes one more way the reader can get closed before play ends.
- Player `seek()` method no longer special cases seeking to a timestamp after the end of the bag. Previously, if no messages are found after the seek time, the Player will reset to the original time stamp before the seek method was called. Now, seek will always be performed, and seeking to a later timestamp will either end bag play or loop according to player options. This makes seek behavior to be consistent with the behavior of playing until the end of the bag. There is also no reason to return `true`/`false` depending on the success of seek, as the seek will always be successful.

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -93,9 +93,6 @@ public:
   ROSBAG2_TRANSPORT_PUBLIC
   void play();
 
-  ROSBAG2_TRANSPORT_PUBLIC
-  rosbag2_cpp::Reader * release_reader();
-
   // Playback control interface
   /// Pause the flow of time for playback.
   ROSBAG2_TRANSPORT_PUBLIC
@@ -133,13 +130,15 @@ public:
 
   /// \brief Advance player to the message with closest timestamp >= time_point.
   /// \details This is blocking call and it will wait until current message will be published
-  /// and message queue will be refilled. Could take substantial amount of time to find
-  /// required message or return player to the original stage.
+  /// and message queue will be refilled.
+  /// If time_point is before the beginning of the bag, then playback time will be set to the
+  /// beginning of the bag.
+  /// If time_point is after the end of the bag, playback time will be set to the end of the bag,
+  /// which will then end playback, or if loop is enabled then will start playing at the beginning
+  /// of the next loop.
   /// \param time_point Time point in ROS playback timeline.
-  /// \throw runtime_error if fail to restore message queue.
-  /// \return true if valid time in bag duration and successful seek operation, otherwise false.
   ROSBAG2_TRANSPORT_PUBLIC
-  bool seek(rcutils_time_point_value_t time_point);
+  void seek(rcutils_time_point_value_t time_point);
 
 protected:
   bool is_ready_to_play_from_queue_{false};
@@ -157,9 +156,6 @@ private:
   void play_messages_from_queue();
   void prepare_publishers();
   bool publish_message(rosbag2_storage::SerializedBagMessageSharedPtr message);
-  void restore_message_queue_from_storage(
-    rosbag2_storage::SerializedBagMessageSharedPtr origin_message) RCPPUTILS_TSA_REQUIRES(
-    reader_mutex_);
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
   static const std::chrono::milliseconds queue_read_wait_period_;
   std::atomic_bool cancel_wait_for_next_message_{false};
@@ -183,6 +179,8 @@ private:
   std::mutex skip_message_in_main_play_loop_mutex_;
   bool skip_message_in_main_play_loop_ RCPPUTILS_TSA_GUARDED_BY
     (skip_message_in_main_play_loop_mutex_) = false;
+
+  rcutils_time_point_value_t starting_time_;
 
   rclcpp::Service<rosbag2_interfaces::srv::Pause>::SharedPtr srv_pause_;
   rclcpp::Service<rosbag2_interfaces::srv::Resume>::SharedPtr srv_resume_;
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -137,16 +137,15 @@ Player::Player(
   {
     std::lock_guard<std::mutex> lk(reader_mutex_);
     reader_ = std::move(reader);
+    // keep reader open until player is destroyed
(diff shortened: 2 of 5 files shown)


rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -245,26 +237,21 @@ void Player::play()
         std::chrono::nanoseconds duration(delay.nanoseconds());
         std::this_thread::sleep_for(duration);
       }
-      rcutils_time_point_value_t starting_time = 0;
+      // no need to open and close reader all the time
+      // just seek to the right spot
```
emersonknapp, line 241:
```suggestion
```

This comment seems to serve to justify the diff, and only make sense with the historical context that we _were_ opening and closing all the time. I think the code makes sense without the comment

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -121,13 +118,11 @@ class Player : public rclcpp::Node
 
   /// \brief Advance player to the message with closest timestamp >= time_point.
   /// \details This is blocking call and it will wait until current message will be published
-  /// and message queue will be refilled. Could take substantial amount of time to find
-  /// required message or return player to the original stage.
+  /// and message queue will be refilled. If the given time stamp is after the end of the bag,
+  /// then the player will either end or loop consistent with player option settings.
```
emersonknapp, line 122: Can we clarify this language even more? I think it's 
* if time_point is before the beginning of the bag, then playback time will be set to the beginning of the bag
* if time_point is after the end of the bag, playback time will be set to the end of the bag, which will then end playback, or if loop is enabled then will start playing at the beginning of the next loop"

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -383,103 +353,21 @@ bool Player::seek(rcutils_time_point_value_t time_point)
...
-      } else {
-        restore_message_queue_from_storage(message_before_seek);
-      }
-    }
-  }
-
-  if (message_found) {
+    // Purge current messages in queue.
+    while (message_queue_.pop()) {}
+    // no need to close & reopen, just seek to right spot in reader
```
emersonknapp, line 361:
```suggestion
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -383,103 +353,21 @@ bool Player::seek(rcutils_time_point_value_t time_point)
...
+    reader_->seek(time_point);
+    // clock jumps to new time
```
emersonknapp, line 363:
```suggestion
```

emersonknapp (commented):
This implementation is so nice and clean - I love it! 

I have one concern related to the edge-case behavior noted in the `Player::seek` docstring comment

If we seek to a time before the bag, e.g. bag_start minus 1 minute - this implementation sets the clock back there, which means that the full minute will have to elapse in real time before the beginning of the bag starts playing. I am thinking that we should "clamp" the `clock->jump` calls to the bag_start/end range. What do you think?

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/d58facdff381d74d9fc79f29e4165abc/raw/789a662a1682a36f3a38dcf78ef6dd7b34dc3604/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_transport rosbag2_tests rosbag2
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9083
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15313/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10025/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12993/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15506/)

rosbag2_transport/include/rosbag2_transport/player.hpp
lihui815: ok

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/d4c53f3799483d9438a478bdc793b532/raw/2d45f2d76b98e4754577b94965a7e9f31a09ad52/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_transport rosbag2_tests rosbag2
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9096
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15327/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10038/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13007/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15520/)

PR #800, opened 2021-06-30
Add code samples for writing bag files in Foxy

gbiggs:
This PR adds some sample code for writing to bag files programmatically. The samples use the Foxy API, so this PR needs to be merged into the Foxy branch.

A corresponding tutorial has been submitted as a [PR to the ROS 2 documentation](https://github.com/ros2/ros2_documentation/pull/1663).

PR #878, opened 2021-10-06
[foxy] Default to C++14

jacobperron:
C++14 is supported version for Foxy.
This fixes a bunch of deprecation warnings on Windows as some included headers are using APIs that are deprecated in C++17.

Example build with warnings: https://ci.ros2.org/job/ci_windows/15579/msbuild/new/

The package was introduced in https://github.com/ros2/rosbag2/pull/800

I didn't review the code, but it builds for me locally simply by updating the standard in the CMakeLists.txt. So I guess C++17 features are not required.


clalancette (approved):
Yes, this is clearly correct for Foxy.  Looks good to me with green CI.

gbiggs (approved):

PR #889, opened 2021-10-19
Don't preprocess a storage file more than once

emersonknapp:
Fixes exception on file-compression playback, which would try to decompress the already-decompressed file

Added regression test that fails without the fix.


lihui815 (approved):
lgtm

PR #892, opened 2021-10-26
[galactic] allow for implementation of composable recorder via inheritance

berndpfrommer:
This PR is meant to be a stop gap measure until a full implementation of a composable Node is complete.

It will allow developers to write their own recorder node under galactic by inheriting from rosbag2_transport::Recorder and const-casting the references returned by the accessors. To make this work the constructor was changed to no longer throw an exception, but instead log a warning.

The motivation for this PR is the need to record high bandwidth and frequency data that interprocess communication was not able to handle.

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -78,6 +78,12 @@ public:
   ROSBAG2_TRANSPORT_PUBLIC
   const rosbag2_cpp::Writer & get_writer_handle();
 
+protected:
+  std::shared_ptr<rosbag2_cpp::Writer> writer_;
+  rosbag2_storage::StorageOptions storage_options_;
+  rosbag2_transport::RecordOptions record_options_;
+  std::atomic<bool> stop_discovery_;
+
 private:
   void topics_discovery();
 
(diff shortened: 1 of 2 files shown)


berndpfrommer:
If I want this commit also to find its way into new releases, should I submit a separate PR against the master branch or is there a way by which you can cherry pick the commit into master?

emersonknapp:
Generally speaking we go the other direction: develop on master and cherry-pick backports to released distros. I am considering this one a kind of "end of the line" patch to make this usable in Galactic, where we should probably develop the more complete feature moving forward.

hidmic:
@emersonknapp shall we open a ticket for this? Intra-process comms are an important feature for many.

emersonknapp:
Yeah, an issue in the tracker is the best way to go

hidmic:
FYI https://github.com/ros2/rosbag2/issues/902.

berndpfrommer:
I just checked and as of today 2/22/2022 the modified header files have not made it into the official ubuntu galactic packages. Is there a schedule for when that will happen?

berndpfrommer:
@emersonknapp sorry to bother you again with this but it's 9 months now and the currently distributed Ubuntu galactic packages don't seem to have this commit. I don't understand how the release mechanism works or where to ask such questions. Could you please let me know how to get this moving?
Thanks!

roncapat:
Is there an example implementation of such composable recorder?
FYI @Fixit-Davide

berndpfrommer:
Here is an example implementation: https://github.com/berndpfrommer/rosbag2_composable_recorder

roncapat:
Thank you so much!

PR #893, opened 2021-10-27
Add simple keyboard control for playback rate

emersonknapp:


rosbag2_transport/test/rosbag2_transport/test_keyboard_controls.cpp
```diff
@@ -147,6 +155,9 @@ TEST_F(RosBag2PlayTestFixture, test_keyboard_controls)
   keyboard_handler->simulate_key_press(play_options_.pause_resume_toggle_key);
   EXPECT_THAT(player->is_paused(), true);
 
+  keyboard_handler->simulate_key_press(play_options_.increase_rate_key);
+  keyboard_handler->simulate_key_press(play_options_.decrease_rate_key);
```
MichaelOrlov, line 159: Sorry for posting in closed PR. 
Just in case for future. It would be better to check rate va `player->get_rate()` before changing it again.

PR #904, opened 2021-11-04
Add --start-paused option to `ros2 bag play`

ivanpauno:
I think this was proposed in https://github.com/ros2/rosbag2/pull/443.

That PR ended up being closed in favor of https://github.com/ros2/rosbag2/issues/696, but an option to start the player in a paused state hasn't been added AFAIS.

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -88,6 +88,9 @@ class PlayVerb(VerbExtension):
         parser.add_argument(
             '--disable-keyboard-controls', action='store_true',
             help='disables keyboard controls for playback')
+        parser.add_argument(
+            '-p', '--start-paused', action='store_true', default=False,
+            help='Start the playback player paused.')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -124,6 +127,7 @@ class PlayVerb(VerbExtension):
         play_options.clock_publish_frequency = args.clock
         play_options.delay = args.delay
         play_options.disable_keyboard_controls = args.disable_keyboard_controls
+        play_options.paused = args.start_paused
 
         player = Player()
         player.play(storage_options, play_options)
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -50,6 +50,9 @@ public:
   // Sleep before play. Negative durations invalid. Will delay at the beginning of each loop.
   rclcpp::Duration delay = rclcpp::Duration(0, 0);
 
+  // Start paused.
+  bool paused = false;
+
   bool disable_keyboard_controls = false;
   // keybindings
   KeyboardHandler::KeyCode pause_resume_toggle_key = KeyboardHandler::KeyCode::SPACE;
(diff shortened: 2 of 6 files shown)


emersonknapp (approved):
LGTM with green CI, no concerns. Thanks for the upgrade!

ivanpauno:
This cppcheck error looks like a false positive:

> Error: [rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp:62]: (error: selfInitialization) Member variable 'paused' is initialized by itself.

I've suppressed it in https://github.com/ros2/rosbag2/pull/904/commits/32caf26f22575df734d8be36e0331c557a66680a, if another approach is prefered (e.g. renaming the argument), let me know and I will address that.

emersonknapp:
Another option would be to name the ctor parameter `bool start_paused`, which should then not trigger the `selfinitialization` issue, right? Not going to block on that, though

emersonknapp (approved):

ros2bag/ros2bag/verb/play.py
```diff
@@ -88,6 +88,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--disable-keyboard-controls', action='store_true',
             help='disables keyboard controls for playback')
+        parser.add_argument(
+            '-p', '--start-paused', action='store_true', default=False,
+            help='Start the playback player paused.')
```
jhdcs, line 93: Would it be a bit more clearer if the help text was re-worded as "Start the playback player in a paused state." ? Minor quibble, of course, but I could see some confusion for people not familiar with what this option does.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -50,6 +50,9 @@ struct PlayOptions
   // Sleep before play. Negative durations invalid. Will delay at the beginning of each loop.
   rclcpp::Duration delay = rclcpp::Duration(0, 0);
 
+  // Start paused.
+  bool paused = false;
```
jhdcs, line 54: Is this boolean saying if we're _starting_ paused? Or if we're _currently_ paused? The comment says the first, but the variable name implies the other.

jhdcs (approved):
This looks good to me code-wise. Only two (minor) comments on help text and variable names, but I'm not sure if we want to hold the PR back because of it.

Basically, I'm just trying to think about potential confusion points for future users and maintainers.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
ivanpauno: It's saying we're starting paused.
I agree that the variable naming is confusing, fixed in https://github.com/ros2/rosbag2/pull/904/commits/4ff72610999249babfabf2b97ae1d1cb177182bf.

ros2bag/ros2bag/verb/play.py
ivanpauno: Both options sound the same to me, though I'm not a native speaker.
Fixed in https://github.com/ros2/rosbag2/pull/904/commits/4ff72610999249babfabf2b97ae1d1cb177182bf.

ivanpauno:
CI config:

- Build: --packages-up-to rosbag2_tests ros2bag rosbag2_py rosbag2_transport
- Test: --packages-select rosbag2_tests ros2bag rosbag2_py rosbag2_transport

Jobs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15570/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10254/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13238/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15788/)

ivanpauno:
CI is passing, @jhdcs @emersonknapp let me know if this can be merged or if something else is missing.

PR #917, opened 2021-11-23
Add logging macros for rosbag2_transport

emersonknapp:
Related to #831 

The others packages have these. I will be adding more features soon that will need it as well.

The original reason this didn't exist was because everything happened within a Node. However, the "convert" feature will happen mostly within rosbag2_transport, but will not happen inside a Node.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/logging.hpp
+++ b/rosbag2_transport/src/rosbag2_transport/logging.hpp
@@ -1,0 +1,61 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSBAG2_TRANSPORT__LOGGING_HPP_
+#define ROSBAG2_TRANSPORT__LOGGING_HPP_
+
+#include <sstream>
+#include <string>
+
+#include "rcutils/logging_macros.h"
+
+#define ROSBAG2_TRANSPORT_PACKAGE_NAME "ROSBAG2_TRANSPORT"
+
+#define ROSBAG2_TRANSPORT_LOG_INFO(...) \
+  RCUTILS_LOG_INFO_NAMED(ROSBAG2_TRANSPORT_PACKAGE_NAME, __VA_ARGS__)
+
+#define ROSBAG2_TRANSPORT_LOG_INFO_STREAM(args) do { \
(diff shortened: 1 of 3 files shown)


lihui815 (approved):
makes sense

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/186a052e7ecedd2b558f0f141f72a079/raw/72fb858a7485626561780bebcfe51d814de12a9b/ros2.repos
BUILD args: --packages-up-to rosbag2_transport rosbag2
TEST args: --packages-select rosbag2_transport rosbag2
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9384
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15653/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10331/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15874/)

rosbag2_transport/src/rosbag2_transport/logging.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "rcutils/logging_macros.h"
+
+#define ROSBAG2_TRANSPORT_PACKAGE_NAME "ROSBAG2_TRANSPORT"
+
+#define ROSBAG2_TRANSPORT_LOG_INFO(...) \
+  RCUTILS_LOG_INFO_NAMED(ROSBAG2_TRANSPORT_PACKAGE_NAME, __VA_ARGS__)
+
+#define ROSBAG2_TRANSPORT_LOG_INFO_STREAM(args) do { \
+    std::stringstream __ss; \
+    __ss << args; \
```
aprotyas, line 30: I think there might be a problem here if `args` contains an expression with an operator which has lower precedence than the `<<` operator, because the `<<` operator would be executed first. A problematic example is `args` == `1 ^ 2`.

Same comment for other instances of `__ss << args;`

emersonknapp: I don't believe that solution works for this type of macro - the intended usage is like:

```
ROSBAG2_TRANSPORT_LOG_INFO_STREAM("foo" << bar << "baz");
```

If you put `args` in parentheses - then the generated code looks like:

```
__ss << ("foo" << bar << "baz");
```

which isn't valid. 

This code as written is the same implementation as is generated in https://github.com/ros2/rclcpp/blob/master/rclcpp/resource/logging.hpp.em#L145 - I see the concern with low-precedence operators, but given the constraints, I perceive that the burden is on the user to be aware of this and act accordingly. It will be true in their chained streams, e.g. `LOG_INFO_STREAM("foo" << (1^2) << "bar");`

emersonknapp: Thanks for paying close attention! But I'm going to move forward with this implementation for now, given that it at least matches the existing pattern in rclcpp/rosbag2 packages

aprotyas: My bad, I interleaved two ideas in my train of thought. Yes, for this situation the user should shoulder the burden of operator precedence within their macro argument. Thanks for the explanation. 😄

Issue #915, opened 2021-11-23
handle recording of messages larger than the sqlite blob limit (default is 1GB)


wjwwood:
## Description

Currently if you try to record a message larger than the sqlite blob/string limit, you get an exception and rosbag2 crashes due to an unhandled exception:

```
terminate called after throwing an instance of 'rosbag2_storage_plugins::SqliteException'
  what():  Error when binding SQL parameter 3. SQLite error (18): string or blob too big
Aborted (core dumped)
```

The default for sqlite (compile time configuration) is 1GB: https://sqlite.org/limits.html (`SQLITE_MAX_LENGTH`)

At the very least it should fail more gracefully, with an error or warning that tells you which topic the message came from and other details like the size of the message that was unsuccessfully written. I'm open to suggestion on what the behavior should be, but I'm leaning towards a warning that the message was not recorded (maybe with a warning "once per topic" or "throttled to once a second" or something), but continuing to record otherwise. I could see the argument for just stopping when this happens, but with a better error message.

Ideally, we'd extend rosbag2 so that it could split arbitrarily large messages across multiple database rows (pagination) as needed to avoid the blob limit.

## Related Issues

I ran into this working with a customer, and there's not a good work around other than avoiding the offending topic.

I'm working on a pull request to add a regression test and improve the error message, but I will not have time to implement pagination or anything like that.

## Completion Criteria

I think this issue should remain open until arbitrarily large messages can be recorded or until the maintainers decide to not support that use case.

## Implementation Notes / Suggestions

As I mentioned, the only idea I had for solving this would be to introduce pagination of large messages over multiple rows. I'm not sure what the implications are for the storage API or the "schema" for sqlite based rosbags.

## Testing Notes / Suggestions

Like I mentioned, I'm working on a regression test for this, but you should be able to trigger it by publishing anything over 1GB and recording it to trigger it.

emersonknapp:
Good ticket. Not sure if I'll have any resources to work on this - but it makes sense to me that rosbag2 should be able to support this use case. 

The one workaround I can think of off the bat would be to paginate the payload across multiple messages, rather than publishing a single payload as a 1GB+ individual message.

wjwwood:
> The one workaround I can think of off the bat would be to paginate the payload across multiple messages, rather than publishing a single payload as a 1GB+ individual message.

[...]

PR #919, opened 2021-11-24
Emit a warning rather than crash when a message is too big for sqlite

wjwwood:
Related to issue https://github.com/ros2/rosbag2/issues/915

This pull request changes the behavior of the sqlite storage plugin to emit a warning rather than throw an exception when a message is written that is too large for sqlite.

### DIFF
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -258,11 +258,11 @@ void SqliteStorage::write_locked(
         SQLITE_LIMIT_LENGTH,
         -1);
       ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_WARN_STREAM(
-        "Message on topic '" << message->topic_name << "' of size '"
-        << message->serialized_data->buffer_length
-        << "' bytes failed to write because it exceeds the maximum size sqlite can store ('"
-        << sqlite_limit << "' bytes): "
-        << exc.what());
+        "Message on topic '" << message->topic_name << "' of size '" <<
+          message->serialized_data->buffer_length <<
+          "' bytes failed to write because it exceeds the maximum size sqlite can store ('" <<
+          sqlite_limit << "' bytes): " <<
+          exc.what());
       return;
     } else {
       // Rethrow.
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -252,10 +252,16 @@ void SqliteStorage::write_locked(
     write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
   } catch (const SqliteException & exc) {
     if (SQLITE_TOOBIG == exc.get_sqlite_return_code()) {
+      // Get the sqlite string/blob limit.
+      size_t sqlite_limit = sqlite3_limit(
+        this->get_sqlite_database_wrapper().get_database(),
+        SQLITE_LIMIT_LENGTH,
+        -1);
       ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_WARN_STREAM(
         "Message on topic '" << message->topic_name << "' of size '"
         << message->serialized_data->buffer_length
-        << "' bytes failed to write because it exceeds the maximum size sqlite can store: "
+        << "' bytes failed to write because it exceeds the maximum size sqlite can store ('"
+        << sqlite_limit << "' bytes): "
         << exc.what());
       return;
     } else {
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -498,6 +498,14 @@ std::string SqliteStorage::get_storage_setting(const std::string & key)
   return database_->query_pragma_value(key);
 }
 
+SqliteWrapper & SqliteStorage::get_sqlite_database_wrapper()
+{
+  if (nullptr == database_) {
+    throw std::runtime_error("database not open");
+  }
+  return *database_;
+}
+
 }  // namespace rosbag2_storage_plugins
 
 #include "pluginlib/class_list_macros.hpp"  // NOLINT
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -248,7 +248,21 @@ void SqliteStorage::write_locked(
             "' has not been created yet! Call 'create_topic' first.");
   }
 
-  write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
+  try {
+    write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
+  } catch (const SqliteException & exc) {
+    if (SQLITE_TOOBIG == exc.get_sqlite_return_code()) {
+      ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_WARN_STREAM(
+        "Message on topic '" << message->topic_name << "' of size '"
+        << message->serialized_data->buffer_length
+        << "' bytes failed to write because it exceeds the maximum size sqlite can store: "
+        << exc.what());
+      return;
+    } else {
+      // Rethrow.
+      throw;
+    }
+  }
   write_statement_->execute_and_reset();
 }
 
--- a/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
+++ b/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
@@ -253,7 +253,7 @@ void SqliteStorage::write_locked(
   } catch (const SqliteException & exc) {
     if (SQLITE_TOOBIG == exc.get_sqlite_return_code()) {
       // Get the sqlite string/blob limit.
-      size_t sqlite_limit = sqlite3_limit(
+      const size_t sqlite_limit = sqlite3_limit(
         this->get_sqlite_database_wrapper().get_database(),
         SQLITE_LIMIT_LENGTH,
         -1);
--- a/rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
+++ b/rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
@@ -18,6 +18,7 @@
 #include <gtest/gtest.h>
 
 #include <cstdio>
+#include <cstring>
 #include <iostream>
 #include <fstream>
 #include <memory>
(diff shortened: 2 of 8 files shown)


rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -248,7 +248,21 @@ void SqliteStorage::write_locked(
             "' has not been created yet! Call 'create_topic' first.");
   }
 
-  write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
+  try {
+    write_statement_->bind(message->time_stamp, topic_entry->second, message->serialized_data);
+  } catch (const SqliteException & exc) {
+    if (SQLITE_TOOBIG == exc.get_sqlite_return_code()) {
+      ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_WARN_STREAM(
```
wjwwood, line 255: I thought about using one of the throttle or once type macros, but I hesitated for this since it's simpler. But I'm open to suggestions on what to do and, if throttle, how often to throttle it.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
```diff
@@ -156,22 +165,58 @@ class StorageTestFixture : public TemporaryDirectoryFixture
...
+
+      amount_to_write = remaining_buffer;
+      if (message.size() < amount_to_write) {
+        amount_to_write = message.size();
+      }
+      std::memcpy(buffer + amount_written, message.data(), amount_to_write);
+      amount_written += amount_to_write;
+
+      return amount_written;
+    }
```
wjwwood, line 219: So, I could have just removed the snprintf section, since I _think_ my `memcpy` one is equivalent, but I just wanted to limit my side effects with this change. Let me know what you guys think.

wjwwood:
This is what the error message looks like in practice:

```
[WARN] [1637792715.870418893] [rosbag2_storage]: Message on topic '/too_big_message' of size '1000000010' bytes failed to write because it exceeds the maximum size sqlite can store: Error when binding SQL parameter 3. SQLite error (18): string or blob too big
```

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
gbiggs: It looks equivalent to me. I'd be fine with removing the `snprintf` version to keep the code simpler.

gbiggs (approved):

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
danthony06: I agree with @gbiggs. And practically, it seems pretty unlikely that you'll have messages exceeding the maximum SQLite blob size that are also coming at a high rate, because you would need an unusually fast network connection for that to happen.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
```diff
@@ -248,7 +248,21 @@ void SqliteStorage::write_locked(
...
+        "Message on topic '" << message->topic_name << "' of size '"
+        << message->serialized_data->buffer_length
+        << "' bytes failed to write because it exceeds the maximum size sqlite can store: "
```
danthony06, line 258: Could we also include the SQLITE_MAX_LENGTH (https://www.sqlite.org/limits.html) value in this error message, so that the users can know what it's set at?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/test_sqlite_storage.cpp
```diff
@@ -373,3 +373,23 @@ TEST_F(StorageTestFixture, throws_on_invalid_pragma_in_config_file) {
...
+
+  // Get the sqlite string/blob limit.
+  size_t sqlite_limit = sqlite3_limit(
+    writable_storage->get_sqlite_database_wrapper().get_database(),
+    SQLITE_LIMIT_LENGTH,
+    -1);
+
+  // This should produce a warning, but not an exception.
+  std::string msg(sqlite_limit + 1, '\0');
+  this->write_messages_to_sqlite(
```
emersonknapp, line 391: It doesn't actually affect the functionality of the test, but I find it helps readability to use `EXPECT_NO_THROW({code_block...})` - then you don't have to use a comment.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
emersonknapp: Also fine with changing to purely `memcpy` to keep things consistent.

emersonknapp (approved):
This LGTM

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
wjwwood: So, this value doesn't appear to be in the headers from sqlite (it's a compiler definition, but it isn't stored in a header AFAIK). I can get it dynamically by "changing" the limit too `-1` and it will return the current value without modifying it, but this requires an sqlite database and I'm not sure I have one in scope here. I'll try.

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
wjwwood: Ok, will do.

rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp
danthony06: Ah, that's too bad. I don't think it's really critical, just a nice thing to know when it happens without needing to go to the documentation.

wjwwood: I got it, I forgot we were in a method of the class that holds the db.

wjwwood:
I had to refactor more of the test fixture's methods to avoid null terminating characters now that we're using `memcpy` and not `snprintf`.

gbiggs (approved):

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/701c5a90a6259e55494a5ecc055e44e3/raw/fe5656f5fc29f36e5f5c17b3152efece41fee6dd/ros2.repos
BUILD args: --packages-up-to rosbag2_storage_default_plugins rosbag2_tests rosbag2
TEST args: --packages-select rosbag2_storage_default_plugins rosbag2_tests rosbag2
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9437
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15711/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10393/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13376/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15953/)

wjwwood:
I'm looking into the test failures on macOS and Windows.

wjwwood:
Re-running CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15717/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10399/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13382/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15961/)

wjwwood:
@ros-pull-request-builder test this please

wjwwood:
So, either because of how Homebrew builds sqlite or because of a newer version of sqlite, the limit for a blob on my macOS machine and our buildfarm is just over 17GB. Which is causing a new failure in how we use the API. I'm gonna try one thing, but if that doesn't work, I might have to skip this test on machines where the limit is more than say 2GB.

wjwwood:
Ok, I found a solution in 51cb11c (and improved in 3f433b6), where I artificially lower the limit. This should work on all the machines.

New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15719/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10401/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13384/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15963/)

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
```diff
@@ -156,22 +161,45 @@ class StorageTestFixture : public TemporaryDirectoryFixture
...
-      message.c_str());
+    size_t remaining_buffer = buffer_capacity;
+    size_t amount_written = 0;
+    if (remaining_buffer == 0) {
+      return amount_written;
+    }
+
+    size_t amount_to_write = remaining_buffer;
+    std::string preamble = get_preamble();
+    if (preamble.size() < amount_to_write) {
```
hidmic, line 185: @wjwwood nit: would it make sense to `assert(buffer_capacity >= preamble.size() + message.size())`?

rosbag2_storage_default_plugins/test/rosbag2_storage_default_plugins/sqlite/storage_test_fixture.hpp
```diff
@@ -73,48 +73,53 @@ class StorageTestFixture : public TemporaryDirectoryFixture
       });
 
     serialized_data->buffer_length = message_size;
-    int written_size = write_data_to_serialized_string_message(
+    size_t written_size = write_data_to_serialized_string_message(
```
hidmic, line 76: @wjwwood would 
```suggestion
    (void)write_data_to_serialized_string_message(
```
have the same effect?

hidmic (approved):
Minor comments, all of them.

wjwwood:
@hidmic I addressed your feedback and fixed one more warning on windows, so if you could quickly re-review those I'd appreciate it.

CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15894/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10569/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13558/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16201/)

wjwwood:
CI again after fixing a new bug:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15905/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10580/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13569/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16225/)

wjwwood:
The `ament_uncrustify` lint will fail until rolling is synced, which is a known issue. Locally it passes.

sloretz (approved):
Approving just the two new commits (https://github.com/ros2/rosbag2/pull/919/commits/52f96e2ec18421be0d842151dc76f8515172ca28 , and https://github.com/ros2/rosbag2/pull/919/commits/ebdce6700324d82d9f65e4a3c7a1d719780985ad ) , I haven't reviewed the rest

Issue #929, opened 2021-12-03
Play: start and end time


dongfang-steven-yang:
## Description

Hello, 

I just wonder if `ros2 bag play` has functionality like `-s` as in `rosbag play` in ROS 1?
I would like to debug my algorithm at a specific time in a big bag file. For example, the duration of my bag is 10 mins, when I start to replay the topics, I would like to directly jump into 6 mins, so that every time I open a bag, I don't have to wait. 

If there is no such functionality, is there any plan to add it?
Or is there any workaround to solve this issue?

Thanks, 
DF

aprotyas:
I don't think the start/stop timestamp feature is exposed in the CLI tool yet, but the infrastructure exists following the work in and https://github.com/ros2/rosbag2/issues/696 and  https://github.com/ros2/rosbag2/issues/821.

aprotyas:
Implementation question for maintainers:
For the play start timestamp, can this feature boil down to adding a start timestamp to `rosbag2_py.PlayOptions`, which can then be used to invoke `rosbag2_transport::Player::seek(timepoint)` after construction of a `Player` but before calling `play()` on it? So, right after L115 here.

https://github.com/ros2/rosbag2/blob/2abc50904803a0eae58ee2094542977ec9262db7/rosbag2_py/src/rosbag2_py/_transport.cpp#L109-L128

emersonknapp:
These both make sense as additional `PlayOption` fields.

This feature makes sense to me, for starting time, it should be easy to implement by overriding `Player::starting_time_` in `Player::play`. 

The end time could be implemented in the dequeueing mechanism by returning `nullptr` from `Player::peek_next_message_from_queue`, which would stop playback at that time. Alternatively, it could be implemented in the enqueueing mechanism `Player::enqueue_up_to_boundary`, stopping enqueueing when a message past the given timestamp is reached. The result would be approximately the same - maybe slightly better performance to stop enqueueing messages, instead of adding messages to the queue that we don't intend to read.

gbiggs:
We have [a PR in our fork[(https://github.com/gbiggs/rosbag2/pull/5) that adds "play until timestamp" functionality. We're planning on upstreaming it later this week or early next week. It might need some revision though.

aprotyas:
[...]

PR #931, opened 2021-12-04
Add start time play option

aprotyas:
This PR adds a `--start-time` (optional) argument to the `ros2 bag play` verb. The desired use is as follows:

```python
ros2 bag play --start-time [positive-float-denoting-start-time-in-seconds]
```

In doing so, the `start_offset` field is added to `rosbag2_transport::PlayOptions` class and its corresponding PyBind11 wrapper.

`rosbag2_transport::Player` respects this option by adding `start_offset` - during construction - to the `starting_time` obtained from its readers metadata, but only does so if the start time is a valid non-negative value. Else, `Player` outputs to the warning stream about the invalid offset.

Partially addresses https://github.com/ros2/rosbag2/issues/929 and gets closer to parity with `rosbag play` from ROS 1.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -130,7 +130,7 @@ class PlayVerb(VerbExtension):
         play_options.delay = args.delay
         play_options.disable_keyboard_controls = args.disable_keyboard_controls
         play_options.start_paused = args.start_paused
-        play_options.starting_time = args.start_time
+        play_options.start_offset = args.start_time
 
         player = Player()
         player.play(storage_options, play_options)
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -91,7 +91,7 @@ class PlayVerb(VerbExtension):
             '-p', '--start-paused', action='store_true', default=False,
             help='Start the playback player in a paused state.')
         parser.add_argument(
-            '--start-time', type=check_positive_float, default=0.0,
+            '--start-offset', type=check_positive_float, default=0.0,
             help='Start the playback player this many seconds into the bag file.')
 
     def main(self, *, args):  # noqa: D102
@@ -130,7 +130,7 @@ class PlayVerb(VerbExtension):
         play_options.delay = args.delay
         play_options.disable_keyboard_controls = args.disable_keyboard_controls
         play_options.start_paused = args.start_paused
-        play_options.start_offset = args.start_time
+        play_options.start_offset = args.start_offset
 
         player = Player()
         player.play(storage_options, play_options)
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -90,6 +90,9 @@ class PlayVerb(VerbExtension):
         parser.add_argument(
             '-p', '--start-paused', action='store_true', default=False,
             help='Start the playback player in a paused state.')
+        parser.add_argument(
+            '--start-time', type=check_positive_float, default=0.0,
+            help='Start the playback player this many seconds into the bag file.')
 
     def main(self, *, args):  # noqa: D102
         qos_profile_overrides = {}  # Specify a valid default
@@ -127,6 +130,7 @@ class PlayVerb(VerbExtension):
         play_options.delay = args.delay
         play_options.disable_keyboard_controls = args.disable_keyboard_controls
         play_options.start_paused = args.start_paused
+        play_options.starting_time = args.start_time
 
         player = Player()
         player.play(storage_options, play_options)
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -16,7 +16,6 @@
 #define ROSBAG2_TRANSPORT__PLAY_OPTIONS_HPP_
 
 #include <cstddef>
-#include <cstdint>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -55,7 +54,7 @@ public:
   bool start_paused = false;
 
   // Time to start playback as an offset from the beginning of the bag.
-  rcutils_time_point_value_t start_offset = INT64_C(0);
+  rcutils_time_point_value_t start_offset = 0;
 
   bool disable_keyboard_controls = false;
   // keybindings
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -54,8 +54,8 @@ public:
   // Start paused.
   bool start_paused = false;
 
-  // Start time from which the player should play the bag.
-  rcutils_time_point_value_t starting_time = INT64_C(0);
+  // Time to start playback as an offset from the beginning of the bag.
+  rcutils_time_point_value_t start_offset = INT64_C(0);
 
   bool disable_keyboard_controls = false;
   // keybindings
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -16,6 +16,7 @@
 #define ROSBAG2_TRANSPORT__PLAY_OPTIONS_HPP_
 
 #include <cstddef>
+#include <cstdint>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -52,6 +53,9 @@ public:
 
   // Start paused.
   bool start_paused = false;
+
+  // Start time from which the player should play the bag.
+  rcutils_time_point_value_t starting_time = INT64_C(0);
 
   bool disable_keyboard_controls = false;
   // keybindings
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -144,7 +144,7 @@ Player::Player(
       metadata.starting_time.time_since_epoch()).count();
     // If a non-default (positive) starting time offset is provided in PlayOptions,
     // then add the offset to the starting time obtained from reader metadata
-    if (play_options_.start_offset < INT64_C(0)) {
+    if (play_options_.start_offset < 0) {
       RCLCPP_WARN_STREAM(
         get_logger(),
         "Invalid start offset value: " <<
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -191,6 +191,12 @@ void Player::play()
       "Invalid delay value: " << play_options_.delay.nanoseconds() << ". Delay is disabled.");
   }
 
+  // If a non-default (positive) starting time was
+  // provided, override the starting time from metadata
+  if (play_options_.starting_time > INT64_C(0)) {
+    starting_time_ = play_options_.starting_time;
+  }
+
   try {
     do {
       if (delay > rclcpp::Duration(0, 0)) {
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -191,8 +191,8 @@ void Player::play()
       "Invalid delay value: " << play_options_.delay.nanoseconds() << ". Delay is disabled.");
   }
 
-  // If a non-default (positive) starting time was
-  // provided, override the starting time from metadata
+  // If a non-default (positive) starting time is provided in PlayOptions,
+  // override the starting time obtained from reader metadata
   if (play_options_.starting_time > INT64_C(0)) {
     starting_time_ = play_options_.starting_time;
   }
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -142,6 +142,17 @@ Player::Player(
     auto metadata = reader_->get_metadata();
     starting_time_ = std::chrono::duration_cast<std::chrono::nanoseconds>(
       metadata.starting_time.time_since_epoch()).count();
+    // If a non-default (positive) starting time offset is provided in PlayOptions,
+    // then add the offset to the starting time obtained from reader metadata
+    if (play_options_.start_offset < INT64_C(0)) {
+      RCLCPP_WARN_STREAM(
+        get_logger(),
+        "Invalid start offset value: " <<
+          RCUTILS_NS_TO_S(static_cast<double>(play_options_.start_offset)) <<
+          ". Negative start offset ignored.");
+    } else {
+      starting_time_ += play_options_.start_offset;
+    }
     clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(
       starting_time_, std::chrono::steady_clock::now,
       std::chrono::milliseconds{100}, play_options_.start_paused);
@@ -189,12 +200,6 @@ void Player::play()
     RCLCPP_WARN_STREAM(
       get_logger(),
       "Invalid delay value: " << play_options_.delay.nanoseconds() << ". Delay is disabled.");
-  }
-
-  // If a non-default (positive) starting time is provided in PlayOptions,
-  // override the starting time obtained from reader metadata
-  if (play_options_.starting_time > INT64_C(0)) {
-    starting_time_ = play_options_.starting_time;
   }
 
   try {
--- a/rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
@@ -15,7 +15,6 @@
 #include <gmock/gmock.h>
 
 #include <chrono>
-#include <cstdint>
 #include <memory>
 #include <string>
 #include <vector>
@@ -194,7 +193,7 @@ TEST_F(PlayerTestFixture, play_ignores_invalid_delay)
 
 TEST_F(PlayerTestFixture, play_respects_start_offset)
 {
-  rclcpp::Duration start_delay(0, static_cast<uint32_t>(5e8));  // 5e8 ns == 0.5 s
+  rclcpp::Duration start_delay(0, static_cast<uint32_t>(RCUTILS_S_TO_NS(0.5)));
   rclcpp::Duration delay_margin(1, 0);
 
   // Start 0.5 seconds into the bag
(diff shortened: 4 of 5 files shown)


aprotyas:
Note for maintainers: this is still in draft because I'd like to get some preliminary feedback on the implementation. The main concern revolves around supplanting `starting_time_` in `Player::play()`. My rationale behind doing that is - if the `starting_time` provided through `PlayOptions` is to be respected, it must also serve as the base point in the `Player::seek()` method. As the `starting_time_` field sees no other use within the `Player` class, this should (?) be safe.

aprotyas:
CI:
Repos file: https://gist.githubusercontent.com/aprotyas/caee0f1688ca30d6f8c9abb787b7ec3d/raw/9c3e6b6b69ec8372c791ab9bff0573fd64ff063f/ros2.repos
Build args: `--packages-up-to ros2bag`
Test args: `--packages-select ros2bag rosbag2_py rosbag2_transport`
Job: https://ci.ros2.org/job/ci_launcher/9446/

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15720/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10402/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15964/)

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -190,3 +191,44 @@ TEST_F(PlayerTestFixture, play_ignores_invalid_delay)
...
+
+  auto start = clock.now();
+  player->play();
+  auto replay_time = clock.now() - start;
+
+  EXPECT_THAT(replay_time, Gt(lower_expected_duration));
+  EXPECT_THAT(replay_time, Lt(upper_expected_duration));
+}
+
+TEST_F(PlayerTestFixture, play_ignores_invalid_starting_time)
```
gbiggs, line 216: There should be some kind of warning to the user when the starting time is invalid.

aprotyas: That makes sense - I guess I overlooked it because the CLI invocation doesn't allow an invalid starting time, but that doesn't stop someone from populating a `rosbag2_transport::PlayOptions` object with the wrong value. Where should this invariant be exercised? It's convenient (not so for the user) to enforce the invariant lazily - i.e. during a `seek`/`play` function - but that may not be the greatest design.

aprotyas: I guess if we didn't want to push the negativity check down to the player, then `rosbag2_transport::PlayOptions::starting_time` could be wrapped as a `non_negative<rcutils_time_point_value_t>` with an `operator rcutils_time_point_value_t()` implementation for implicit conversions between the wrapper and the underlying integer type - wrapper in this gist:

[Non-negative signed integral type wrapper](https://gist.github.com/aprotyas/02803a4ade50059285ec8f7badbf2edd)

gbiggs: I like the concept that the player options must be correct as constructed, and making the `starting_time` a `non_negative<T>` value allows that - it would cause an error rapidly for the user. However if it's difficult to integrate that cleanly then I'd be happy with checking the options struct's values at the point of use and throwing an error there.

aprotyas: In hindsight, it may be hard to integrate this `non_negative<T>`, if for nothing else but for its "sole occurrence" nature in the `PlayOptions` class.
The warning/error for a negative `starting_time` definitely has to go somewhere, but its better to first sort out how the two `starting_time` sources - one from `PlayOptions` and one from `BagMetadata` - live together.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -53,6 +54,9 @@ struct PlayOptions
   // Start paused.
   bool start_paused = false;
 
+  // Start time from which the player should play the bag.
+  rcutils_time_point_value_t starting_time = INT64_C(0);
```
emersonknapp, line 58: nitpick: I would prefer to call this `start_time`. Or maybe even `start_offset`

The docstring should definitely be specific that this is an offset. "Time to start  playback as an offset from the beginning of the bag"

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -191,6 +191,12 @@ void Player::play()
       "Invalid delay value: " << play_options_.delay.nanoseconds() << ". Delay is disabled.");
   }
 
+  // If a non-default (positive) starting time is provided in PlayOptions,
+  // override the starting time obtained from reader metadata
+  if (play_options_.starting_time > INT64_C(0)) {
+    starting_time_ = play_options_.starting_time;
```
emersonknapp, line 197: > how - in general - PlayOptions::starting_time and BagMetadata::starting_time reside together - do they represent the same information?

I think that this option makes most sense as an "offset from the starting_time of the bag". `BagMetadata::starting_time` means "the timestamp of the first message in the bag" (though there is some discussion of changing it to mean "the timestamp at the time recording was started" - that is not the meaning presently). This value can be anything, as it depends on the recording system's steady clock.

However, the user likely does not care what the "absolute time" for starting is. E.g. if a bag has `starting_time=100`, and the user wants to start 20 seconds into the bag, they probably want to say `--start-time 20`, not `--start-time 120`, since the latter requires extra knowledge about the specific bag.

Given that, I would be more in favor this line looking like
```suggestion
    starting_time_ = metadata.starting_time + play_options_.start_offset;
```

emersonknapp (changes requested):
Left a longer comment inline - lmk what you think about that

rosbag2_transport/src/rosbag2_transport/player.cpp
aprotyas: > I think that this option makes most sense as an "offset from the starting_time of the bag". 

Yes, that's exactly how I expect this to be used. I concur that the user would mean play from 20 seconds into the bag when they will use `--start-time 20`, so I will incorporate this suggested change.

On a related note, where should error checking for the `play_options_.starting_time` go? Does it make sense to throw a runtime error if `play_options_.starting_time < 0` here? Or should an error/warning be raised at construction/assignment time? Check https://github.com/ros2/rosbag2/pull/931#discussion_r763526466 for more context.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
aprotyas: To be honest, `start_time` does not do justice to the difference in meaning between this field and `starting_time` from `BagMetadata::starting_time`. I'm inclined to call it something like `start_offset` or `start_from` instead - probably the former.

I'll take extra care to note its nature in documentation.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
emersonknapp: I think it makes OK sense to validate `start_offset` in `Player` usage. As it is semantically a Duration, in my view there's nothing wrong theoretically with the data having a negative duration. However, we would consider it a constraint of this specific `Player` implementation that it does not want to start playback before the beginning of the bag (though it would have no technical problem doing so, this use case feels like it's probably not something we would actually want to do in practice).

We're already doing validation of `delay` in `Player::play`, I wouldn't argue against validating `start_offset` there as well.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
aprotyas: Renamed in https://github.com/ros2/rosbag2/pull/931/commits/ec56a7a01c7ec726f9f3744ceaff410bd14171a4.

rosbag2_transport/src/rosbag2_transport/player.cpp
aprotyas: I added the `start_offset` to `starting_time` in the `Player` constructor rather than the `play` function - primarily because that means we only add to the `starting_time` only once. From that point onwards, it doesn't matter if an offset was specified in `play_options_`, I think.

Check https://github.com/ros2/rosbag2/pull/931/commits/c79d65cfacb565d6c8b7829ba2bdefce5d61cac0 for reference.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
aprotyas: So, as of https://github.com/ros2/rosbag2/pull/931/commits/c79d65cfacb565d6c8b7829ba2bdefce5d61cac0, if the `start_offset` is a negative value, `Player` outputs a message to the warning stream informing the user that it's invalid, but otherwise just ignores the negative value. Thoughts?

aprotyas:
Running CI after addressing feedback:
Repos file: https://gist.githubusercontent.com/aprotyas/caee0f1688ca30d6f8c9abb787b7ec3d/raw/9c3e6b6b69ec8372c791ab9bff0573fd64ff063f/ros2.repos
Build args: `--packages-up-to ros2bag`
Test args: `--packages-select ros2bag rosbag2_py rosbag2_transport`
Job: https://ci.ros2.org/job/ci_launcher/9469/

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15743/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10425/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15990/)

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -53,6 +54,9 @@ struct PlayOptions
   // Start paused.
   bool start_paused = false;
 
+  // Time to start playback as an offset from the beginning of the bag.
+  rcutils_time_point_value_t start_offset = INT64_C(0);
```
MichaelOrlov, line 58:
```suggestion
  rcutils_time_point_value_t start_offset = 0;
```
There are no reason for this ridiculous type cast.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -142,6 +142,17 @@ Player::Player(
     auto metadata = reader_->get_metadata();
     starting_time_ = std::chrono::duration_cast<std::chrono::nanoseconds>(
       metadata.starting_time.time_since_epoch()).count();
+    // If a non-default (positive) starting time offset is provided in PlayOptions,
+    // then add the offset to the starting time obtained from reader metadata
+    if (play_options_.start_offset < INT64_C(0)) {
```
MichaelOrlov, line 147:
```suggestion
    if (play_options_.start_offset < 0) {
```

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -190,3 +191,43 @@ TEST_F(PlayerTestFixture, play_ignores_invalid_delay)
   EXPECT_THAT(replay_time, Gt(lower_expected_duration));
   EXPECT_THAT(replay_time, Lt(upper_expected_duration));
 }
+
+TEST_F(PlayerTestFixture, play_respects_start_offset)
+{
+  rclcpp::Duration start_delay(0, static_cast<uint32_t>(5e8));  // 5e8 ns == 0.5 s
```
MichaelOrlov, line 197:
```suggestion
  rclcpp::Duration start_delay(0, static_cast<uint32_t>(RCUTILS_S_TO_NS(0.5)));
```

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -190,3 +191,43 @@ TEST_F(PlayerTestFixture, play_ignores_invalid_delay)
...
+  EXPECT_THAT(replay_time, Gt(lower_expected_duration));
+  EXPECT_THAT(replay_time, Lt(upper_expected_duration));
+}
+
+TEST_F(PlayerTestFixture, play_ignores_invalid_start_offset)
+{
+  rclcpp::Duration delay_margin(1, 0);
+
+  // Player should ignore an invalid (negative) start offset
+  play_options_.start_offset = INT64_C(-5);
```
MichaelOrlov, line 221:
```suggestion
  play_options_.start_offset = -5;
```

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -16,6 +16,7 @@
 #define ROSBAG2_TRANSPORT__PLAY_OPTIONS_HPP_
 
 #include <cstddef>
+#include <cstdint>
```
MichaelOrlov, line 19:
```suggestion
```

ros2bag/ros2bag/verb/play.py
```diff
@@ -90,6 +90,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '-p', '--start-paused', action='store_true', default=False,
             help='Start the playback player in a paused state.')
+        parser.add_argument(
+            '--start-time', type=check_positive_float, default=0.0,
```
MichaelOrlov, line 94: Naming parameter as `--start-time` and mapping it internally to the `start_offset` causing a lot of confusion what it really is.
It could be not clear that `start-time` doesn't really means start time in rosbag timestamp but rather time offset from the begging of the bag.
IMO it would be better to rename parameter `--start-time` to the `--start-offset`.

MichaelOrlov (changes requested):

rosbag2_transport/include/rosbag2_transport/play_options.hpp
aprotyas: Noted - my bad! I definitely over-engineered things here. :sweat_smile: 

Addressed in https://github.com/ros2/rosbag2/pull/931/commits/07543e0c770db06fbada46268105fd4eea1a9d3a.

rosbag2_transport/src/rosbag2_transport/player.cpp
aprotyas: Addressed in https://github.com/ros2/rosbag2/pull/931/commits/07543e0c770db06fbada46268105fd4eea1a9d3a.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
aprotyas: Addressed in https://github.com/ros2/rosbag2/pull/931/commits/07543e0c770db06fbada46268105fd4eea1a9d3a.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
aprotyas: Addressed in https://github.com/ros2/rosbag2/pull/931/commits/07543e0c770db06fbada46268105fd4eea1a9d3a.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
aprotyas: Addressed in https://github.com/ros2/rosbag2/pull/931/commits/07543e0c770db06fbada46268105fd4eea1a9d3a.

ros2bag/ros2bag/verb/play.py
aprotyas: Yes, I agree that `--start-offset` would be less confusing for developers. The reason I initially picked `--start-time` for the CLI argument was because I thought the wording would align with what (most) users expected - time offset from the beginning of the bag.

Having said that, it's better to not leave room for ambiguity. Addressed in https://github.com/ros2/rosbag2/pull/931/commits/975c65c9411d028cfaf7d140423c00e9679306a8.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -190,3 +190,43 @@ TEST_F(PlayerTestFixture, play_ignores_invalid_delay)
...
+  play_options_.start_offset = static_cast<rcutils_time_point_value_t>(RCUTILS_S_TO_NS(0.5));
+  // Subtract the delay since we've sought further into the bag
+  auto lower_expected_duration = message_time_difference - start_delay;
+  auto upper_expected_duration = message_time_difference - start_delay + delay_margin;
+  auto player = std::make_shared<rosbag2_transport::Player>(
+    std::move(reader), storage_options_, play_options_);
+
+  auto start = clock.now();
+  player->play();
+  auto replay_time = clock.now() - start;
```
jhdcs, line 209: Nitpick: Would this be slightly clearer if named `replay_duration`?

jhdcs (commented):
Just a minor naming thing to add, probably a much lower priority that the others' comments/suggestions.

emersonknapp (approved):
This LGTM with the updates. Thanks for the revision

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
aprotyas: Agreed! Renamed in https://github.com/ros2/rosbag2/pull/931/commits/7cd158f6163cc7f2538140a1528645669c55f641

aprotyas:
Thank you for the reviews everyone. Running CI again after the latest round of revisions:

Repos file: https://gist.githubusercontent.com/aprotyas/caee0f1688ca30d6f8c9abb787b7ec3d/raw/9c3e6b6b69ec8372c791ab9bff0573fd64ff063f/ros2.repos
Build args: `--packages-up-to ros2bag`
Test args: `--packages-select ros2bag rosbag2_py rosbag2_transport`
Job: https://ci.ros2.org/job/ci_launcher/9475/

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15750/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10431/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13415/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16001/)

jhdcs (approved):

aprotyas:
@emersonknapp I was wondering whether PRs which modify existing or add new features like this should be documented in the Humble release notes? Is that effort being tracked separately? I ask because I don't see any rosbag2 related entries in the release notes: https://docs.ros.org/en/rolling/Releases/Release-Humble-Hawksbill.html#changes-since-the-galactic-release

adiser:
Is there a backport of this feature planned for Foxy?

MichaelOrlov:
@adiser This PR will not be backported to the Foxy at least because it has breaking ABI changes.

PR #936, opened 2021-12-12
Bugfix for broken bag split when using cache

### LINKED ISSUE #939: Bag split doesn't work during recording when cache enabled
## Description
Bag split doesn't work during recording when cache enabled.

### Expected Behavior
Rosbag2 record multiple files when split enabled alongside with double buffering cache mode.

### Actual Behavior
Rosbag2 records only first file and when split happen dropping all other messages.

## To Reproduce
1. ros2 bag record -a -b 100000 --max-cache-size 1024

 - OS: Ubuntu 20.04
 - ROS 2 Distro: Rolling
 - Version: master 200c4c3590c87c38c02a4b0be0cb9104db80133e

## Additional context
Issue was discovered during post merge code review of the PR https://github.com/ros2/rosbag2/pull/850

MichaelOrlov:
- Closes #939

Add `start()` method to the `CacheConsumer` class and call it from
`sequential_writer` after split. Also added unit test to cover cases when bag splitting with enabled cache.

### DIFF
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -241,14 +241,14 @@ TEST_F(SequentialCompressionWriterTest, writer_creates_correct_metadata_relative
   message->topic_name = test_topic_name;
 
   writer_->write(message);
-  // bag size == max_bafile_size, no split yet
+  // bag size == max_bagfile_size, split
   writer_->write(message);
-  // bag size > max_bagfile_size, split
+  // bag size == max_bagfile_size, split
   writer_->write(message);
   writer_.reset();
 
   EXPECT_EQ(
-    intercepted_metadata_.relative_file_paths.size(), 2u);
+    intercepted_metadata_.relative_file_paths.size(), 3u);
 
   const auto base_path = tmp_dir_storage_options_.uri;
   int counter = 0;
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -257,6 +257,11 @@ void SequentialWriter::switch_to_next_storage()
   for (const auto & topic : topics_names_to_info_) {
     storage_->create_topic(topic.second.topic_metadata);
   }
+
+  if (use_cache_) {
+    // restart consumer thread for cache
+    cache_consumer_->start();
+  }
 }
 
 void SequentialWriter::split_bagfile()
@@ -347,8 +352,7 @@ bool SequentialWriter::should_split_bagfile() const
   if (storage_options_.max_bagfile_size !=
     rosbag2_storage::storage_interfaces::MAX_BAGFILE_SIZE_NO_SPLIT)
   {
-    should_split = should_split ||
-      (storage_->get_bagfile_size() > storage_options_.max_bagfile_size);
+    should_split = (storage_->get_bagfile_size() >= storage_options_.max_bagfile_size);
   }
 
   // Splitting by time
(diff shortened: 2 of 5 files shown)


adamdbrw:
This looks like a regression to me. Cache consumer worked without the need to explicitly start it previously. I will look into it now.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -257,6 +257,11 @@ void SequentialWriter::switch_to_next_storage()
   for (const auto & topic : topics_names_to_info_) {
     storage_->create_topic(topic.second.topic_metadata);
   }
+
+  if (use_cache_) {
```
adamdbrw, line 261: At first glance this looks incorrect since the callback needs to be changed. I see that the snapshot changes broke the cache mechanism and were apparently not tested with splitting.

Edit: After looking into previous PR I believe this is ok since the callback is now a member function and it is ensured that the callback is always good.

adamdbrw (approved):

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -241,14 +241,14 @@ TEST_F(SequentialCompressionWriterTest, writer_creates_correct_metadata_relative
   message->topic_name = test_topic_name;
 
   writer_->write(message);
-  // bag size == max_bafile_size, no split yet
+  // bag size == max_bafile_size, split
```
hidmic, line 244: @MichaelOrlov nit^2: I know it's preexisting, but `max_bafile_size` -> `max_bagfile_size`

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
hidmic: In which case, I think it'd be best not to bounce the consumer thread on every split. I won't block on it though.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
MichaelOrlov: Thanks for catching it up. Fixed via amend commit.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: @hidmic When we are doing bag split we are recreating `storage_` object and we need to prevent calling `staorge_->write(..)` from any other threads. 
There are two options stop/start `cache->consumer` thread or add extra lock inside consumer thread triggering outside i.e. from `SequentialWriter`. 
The last option is less preferable IMO, since it will over-complicate design and will be more prone for the deadlocks and race conditions.

MichaelOrlov:
Running CI
CI_BRANCH_TO_TEST: morlov/bugfix_for_broken_split
BUILD args: --packages-up-to rosbag2_cpp rosbag2_transport rosbag2_tests
TEST args: --packages-select rosbag2_cpp rosbag2_transport rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/9513
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15792/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10470/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16049/)

MichaelOrlov:
Rerun CI for Windows after trying to fix typecast warning in test.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16050/)

PR #944, opened 2021-12-18
WIP: Fix logic for regular expressions in topic_filter

MichaelOrlov:
Fixing logic for filtering messages in topic_filter to respect `record_options.all`, `regex` and `topic list`.

Current logic for filtering messages in `TopicFilter` doesn't respect `record_options.all` in case if  defined topic list at the same time. 
If topic name not in `topic list` but match with `regex` it's still will be filtered out. i.e. `regex` doesn't work if topic list not empty and topic name not in `topic list`.
Also if `record_options.all` equal `false` and `topic_name` doesn't match with other filters `TopicFilter` will not filter out such topics.

Expected behavior :
1. `record_options.alll` takes precedence over `regex` and `topic list`
1. `topic list` shouldn't have precedence over `regex`. They should be addition to each other.
1. `TopicFilter::take_topic(..)` should return false, if `record_options.alll == false`  and topic name doesn't match with any other filter conditions.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
@@ -75,7 +75,7 @@ is_leaf_topic(
   const std::string & topic_name, rclcpp::node_interfaces::NodeGraphInterface & node_graph)
 {
   auto subscriptions_info = node_graph.get_subscriptions_info_by_topic(topic_name);
-  return subscriptions_info.size() == 0;
+  return subscriptions_info.empty();
 }
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


emersonknapp:
> record_options.alll takes precedence over regex

I don't think this rule makes sense to me. If I specify `--all --regex foo` - should I really expect that the `regex` argument is invalidated? Should `all` only mix with `exclude`?

If that's the case, then I think we should have a large warning right at the beginning, if the user specifies `all` with either `topic_list` or `regex`, something like:

```
WARNING:  You provided a topic list or regex, but --all option was specified, which overrides both of those. Topic list and `--regex` will not be used.
```

MichaelOrlov:
> I don't think this rule makes sense to me. If I specify --all --regex foo - should I really expect that the regex argument is invalidated? Should all only mix with exclude?

I think the rule makes a lot of sense to be consistent in logic. I can envision a real case when command line could be patched by some script adding `--all` parameter to override all other filters.
I agree about idea adding a warning if `--all` overrides other filters specified in settings or command line. I will add proposed warning.

MichaelOrlov:
- Closing as wouldn't be merged. 
- The PR will be superseded by the #1480

PR #945, opened 2022-01-06
Update to pybind11 2.7.1

clalancette:
We can remove most of the hacks we were carrying around, but we also need to add in one more overload of the Python executable so pybind11 properly detects things on Windows Debug.

Depends on ros2/pybind11_vendor#10 (CI is over in that PR)

### DIFF
--- a/rosbag2_py/CMakeLists.txt
+++ b/rosbag2_py/CMakeLists.txt
@@ -29,18 +29,13 @@ find_package(PythonExtra REQUIRED)
 find_package(Python3 REQUIRED Interpreter Development)
 if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   # Set the python debug interpreter.
-  # pybind11 will setup the build for debug now.
+  # We need to set *both* PYTHON_EXECUTABLE (for use in this CMake file),
+  # as well as Python3_EXECUTABLE (which is what pybind11 uses).
   set(PYTHON_EXECUTABLE "${PYTHON_EXECUTABLE_DEBUG}")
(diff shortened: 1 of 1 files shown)


wjwwood:
@clalancette any idea about the uncrustify CI failure? I have the same thing in my pull request.

clalancette:
> @clalancette any idea about the uncrustify CI failure? I have the same thing in my pull request.

Based on circumstantial evidence, my best guess is that the GitHub action is using uncrustify 0.69.0 (installed from apt), while we've updated our vendored ament version of uncrustify to 0.72.  They have slightly different ideas of what certain pieces of code should look like.

rosbag2_py/CMakeLists.txt
```diff
@@ -26,46 +26,17 @@ find_package(rosbag2_transport REQUIRED)
 # Find python before pybind11
 find_package(python_cmake_module REQUIRED)
 find_package(PythonExtra REQUIRED)
+find_package(Python3 REQUIRED Interpreter Development)
 if(WIN32 AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   # Set the python debug interpreter.
-  # pybind11 will setup the build for debug now.
+  # We need to set *both* PYTHON_EXECUTABLE (for use in this CMake file),
```
sloretz, line 32: Same comment as on the rclpy PR about doing something like ros2/geometry2#494

clalancette: Cleaned up in 6919813

sloretz (approved):

PR #960, opened 2022-02-15
Add play-for functionality

gbiggs:
This PR adds the ability to play for a specified number of seconds.

This work was co-authored by @agalbachicar.

PR #963, opened 2022-02-17
Add play-next-n functionality

gbiggs:
This PR extends the play-next-message service/function with the ability to play the next N messages instead of just one message.

This work was co-authored by @agalbachicar.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -315,56 +315,59 @@ rosbag2_storage::SerializedBagMessageSharedPtr * Player::peek_next_message_from_
   return message_ptr;
 }
 
-bool Player::play_next(const std::optional<uint64_t> num_messages)
+uint64_t Player::play_next(const uint64_t num_messages)
 {
-  // Don't do anything if no messages were requested
-  if (num_messages.has_value() && *num_messages == 0) {
-    return true;
-  }
-
   if (!clock_->is_paused()) {
     RCLCPP_WARN_STREAM(get_logger(), "Called play next, but not in paused state.");
-    return false;
+    return 0u;
   }
-
-  RCLCPP_INFO_STREAM(get_logger(), "Playing next message.");
 
   // Temporary take over playback from play_messages_from_queue()
   std::lock_guard<std::mutex> main_play_loop_lk(skip_message_in_main_play_loop_mutex_);
+  skip_message_in_main_play_loop_ = true;
 
-  // Wait for player to be ready for playback messages from queue i.e. wait for Player:play() to
-  // be called if not yet and queue to be filled with messages.
-  {
-    std::unique_lock<std::mutex> lk(ready_to_play_from_queue_mutex_);
-    ready_to_play_from_queue_cv_.wait(lk, [this] {return is_ready_to_play_from_queue_;});
-  }
+  uint64_t message_counter = 0;
 
-  bool next_message_published = false;
-  uint64_t messages_remaining = num_messages.has_value() ? *num_messages : 1u;
-
-  while (messages_remaining > 0) {
-    // Reset state for the next iteration.
-    next_message_published = false;
-    skip_message_in_main_play_loop_ = true;
+  while (message_counter < num_messages) {
+    RCLCPP_INFO_STREAM(get_logger(), "Playing next message.");
+    // Wait for player to be ready for playback messages from queue i.e. wait for Player:play() to
+    // be called if not yet and queue to be filled with messages.
+    {
+      std::unique_lock<std::mutex> lk(ready_to_play_from_queue_mutex_);
+      ready_to_play_from_queue_cv_.wait(lk, [this] {return is_ready_to_play_from_queue_;});
+    }
 
     rosbag2_storage::SerializedBagMessageSharedPtr * message_ptr = peek_next_message_from_queue();
 
+    bool next_message_published = false;
     while (message_ptr != nullptr && !next_message_published) {
       {
         rosbag2_storage::SerializedBagMessageSharedPtr message = *message_ptr;
+
+        // Do not move on until sleep_until returns true
+        // It will always sleep, so this is not a tight busy loop on pause
+        clock_->resume();
+        while (rclcpp::ok() && !clock_->sleep_until(message->time_stamp)) {
+          if (std::atomic_exchange(&cancel_wait_for_next_message_, false)) {
+            break;
+          }
+        }
+        clock_->pause();
+
         next_message_published = publish_message(message);
-        clock_->jump(message->time_stamp);
       }
       message_queue_.pop();
       message_ptr = peek_next_message_from_queue();
     }
-    // There are no more messages of the built publishers to play.
-    if (message_ptr == nullptr) {
+    if (!next_message_published) {
       break;
     }
-    messages_remaining--;
+
+    message_counter++;
   }
-  return next_message_published;
+
+
+  return message_counter;
 }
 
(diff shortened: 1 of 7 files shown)


jhdcs:
I'm not seeing any obvious issues, apart from the CI failing. Once that's sorted out I'll give it another once-over just to triple-check.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -326,25 +331,38 @@ bool Player::play_next()
 
   // Temporary take over playback from play_messages_from_queue()
   std::lock_guard<std::mutex> main_play_loop_lk(skip_message_in_main_play_loop_mutex_);
-  skip_message_in_main_play_loop_ = true;
+
```
MichaelOrlov, line 334: It doesn't make sense to move `skip_message_in_main_play_loop_ = true;` inside loop while still holding relevant `skip_message_in_main_play_loop_mutex_` mutex.  
Please return it back.

rosbag2_transport/test/rosbag2_transport/test_play_next.cpp
```diff
@@ -158,6 +158,69 @@ TEST_F(RosBag2PlayTestFixture, play_next_playing_one_by_one_messages_with_the_sa
...
+  ASSERT_TRUE(
+    sub_->spin_and_wait_for_matched(player->get_list_of_publishers(), std::chrono::seconds(30)));
+
+  auto await_received_messages = sub_->spin_subscriptions();
+
+  player->pause();
+  ASSERT_TRUE(player->is_paused());
+
+  auto player_future = std::async(std::launch::async, [&player]() -> void {player->play();});
+  player->wait_for_playback_to_start();
```
MichaelOrlov, line 194: `player->wait_for_playback_to_start();` Is redundant here.  
I used to use it when I was need to measure exact time for play next operation to make sure that it will not wait inside for message timestamp.  
For this particular test we shouldn't use `player->wait_for_playback_to_start();` It's kind of opposite logic to check. We have expectation that `play_next` will take care about this wait synchronization internally. 
If for some circumstance it will not wait internally, test will fail occasionally and we will know that `play_next` performing not as expected.

rosbag2_transport/test/rosbag2_transport/test_play_next.cpp
```diff
@@ -158,6 +158,69 @@ TEST_F(RosBag2PlayTestFixture, play_next_playing_one_by_one_messages_with_the_sa
...
+
```
MichaelOrlov, line 195: If you will add here
```cpp
ASSERT_TRUE(player->is_paused());
  ASSERT_TRUE(player->play_next({0u}));
  // Yield CPU resources so messages are actually sent through.
  std::this_thread::sleep_for(std::chrono::milliseconds(50));
  EXPECT_THAT(sub_->get_received_messages<test_msgs::msg::BasicTypes>("/topic1"), SizeIs(0u));
```
You will save time and resource on creating and running one more `play_next_no_message_must_succeed` test. i.e. `play_next_no_message_must_succeed` will be redundant in this case.

MichaelOrlov (changes requested):
@gbiggs Thank you for your contribution.

I have a few requests for changes and some concerns/questions overall for the design of this feature.
I am curios about in what cases this feature aka playing multiple messages in "burst" mode without respecting their timings will be useful?   
I would (and I think many others) expect that if `play_next(N)` would exist it should respect timings between following messages.
Also it looks like if we will add such feature it will be more appropriate to change API for `play next N` operation to return number of played messages instead of  just`boolean` value.  
In current implementation for instance if someone requested to play **5** messages and for some circumstance `play_next(N)` will really play **3** of them `play_next(N)` will still return **true** and it will  not be possible to recognize whether those messages was lost on the transport layer or they are was just filtered out or missing inside bag.

gbiggs:
@MichaelOrlov: @agalbachicar has addressed your comments. Can you please re-review?

MichaelOrlov:
@gbiggs Following questions are still unclear for me.
> I am curios about in what cases this feature aka playing multiple messages in "burst" mode without respecting their timings will be useful?
> I would (and I think many others) expect that if `play_next(N)` would exist it should respect timings between following messages.

Could you please clarify?

MichaelOrlov (changes requested):
@gbiggs @agalbachicar 
I have a significant concerns about this feature and it's implementation. Therefore I can't approve this PR.
1. You didn't provide any explanations what would be a real world scenario for usage of this new feature and API changes. i.e. why following case doesn't work for you?
```cpp
  for (size_t i = 0; i < num_msgs_in_bag_; i++) {
    player->play_next();
  }
```
We use to creating new issue for feature development and discussion or at least need to provide justification and detailed explanations for proposed API changes and reasoning for them.
2. Current implementation broke generic behavior for `play_next()` which was before your changes. e.g `play_next()` shall play 1 message without delays.
3. **Tests start failing and you just quietly adjust timings between messages to let tests pass with new broken functionality. This is unfair and unacceptable!!!**
In particular in `play_next_playing_all_messages_without_delays` test.

Taking in to the account all above.
Please: 
- Provide justification and explanation what would be a reasoning for these changes and usage of this new API
- Revert timings and irrelevant changes in tests. I've made suggestions in line.
- Fix logic in core `player.cpp` to preserve existing functionality and to be able to pass existing tests. We need to keep backward compatibility.

gbiggs:
@MichaelOrlov Our use case is playing a subset of messages from a bag, but not in real-time, and doing it from the command line. A for-loop over `play_next` would work programmatically, but we want to be able to do it from the command line. This is useful for rapidly processing the data in a bag.

I guess what we are looking for is a "burst mode" playback. Would this PR be acceptable if we changed it to provide a `burst` function in the API rather than modifying the `play_next` function?

agalbachicar:
@MichaelOrlov , on top of @gbiggs comment , I just would like to add where all this started. Please take a look at the [meeting notes of the working group on 2021/05/21](https://docs.google.com/document/d/1UXWaEkT6ji0Z0uTtJJSni5Y6fiopM9qOOzv4ldeNaAM/edit?pli=1#). The scope of our use case extended from "play N seconds and pause" (see #960) to that and "play N messages".

Anyway, @gbiggs proposal stands. The `burst` verb might be more appropriate. We are eager to hear your thoughts.

MichaelOrlov:
@gbiggs @agalbachicar Will `play_for N seconds` API along side with another parameter **rate 10x** will work for you as burst mode?

gbiggs:
We would prefer to burst the messages as fast as possible and in controllable batch sizes.

MichaelOrlov:
@gbiggs You can set up even more aggressive rate **100**x for example.  
This is looks more logically correct and useful, since it's difficult to predict how many messages need to play for debug purpose.
It's more natural to operate with amount of time for replay rather than number of messages.  
Thoughts?
IMO I would refrain from creating separate `play_in_burst(N)`  API or changing existing `play_next()`.

gbiggs:
Our use case is not for debugging, it's for processing of data from a bag. For example, we may want to filter the topics in a bag down to a single image topic, then pull out 10 messages at a time from that topic and process them through a machine learning algorithm. This is not time-based, it is number-of-messages-based, and if we are doing massively parallel computation (as an example) then we want to get all the computation pipelines full as fast as possible.

MichaelOrlov:
@gbiggs Ok. Thanks for the explanation. 
In this case let's go with separate API aka `play_in_burst(num_messages)`

gbiggs:
@MichaelOrlov I have created [a new PR that implements burst mode](https://github.com/ros2/rosbag2/pull/977). I will close this PR as unmergeable.

Issue #967, opened 2022-03-03
Race condition in discovery of QoS prevents recording topics published with non-default QoS


theseankelly:
## Description
There is a race in the DDS participant discovery process such that `ros2 bag record` may discover the presence of a topic based on the existence of a DDS `DataReader` *before* the matching `DataWriter` is discovered.

When this race is lost, `rosbag2_transport` will discover no offered QoS profiles and will instead [subscribe with a default QoS](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/src/rosbag2_transport/qos.cpp#L110).

In cases where the publisher is providing an incompatible QoS with the default the recording will fail loudly (i.e. if the publisher offers `SensorDataQoS` we will see a warning about incompatible reliability) or silently (i.e. if the publisher offers `TRANSIENT_LOCAL` durability, such as previously published `/tf_static` messages will be dropped due to the subscriber's compatible request for `VOLATILE` durability).

Additionally, the `offered_qos_profiles` field in the resulting bag's `metadata.yaml` will be an empty string for the affected topics.

### Expected Behavior
Given a ROS graph with publishers and subscribers using non-default QoS, the expected behavior is to reliably record all the data with `ros2 bag record -a`.

### Actual Behavior
Occasionally (randomly) we miss data from non-default QoS publishers for the duration of a recording. The race in discovery appears to be exacerbated by size of the data and/or network load coming from the publisher.

For example, when using the [ros2_ouster](https://github.com/ros-drivers/ros2_ouster_drivers) driver with 3-4 subscribers, we see the following error 3 or 4 times out of 10 recording sessions:

```
[INFO] [1646323966.741931090] [rosbag2_recorder]: Listening for topics...
[INFO] [1646323966.743906916] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1646323966.744695252] [rosbag2_recorder]: Subscribed to topic '/points'
[INFO] [1646323966.746003963] [rosbag2_recorder]: Subscribed to topic '/parameter_events'
[WARN] [1646323966.749143844] [rosbag2_recorder]: New publisher discovered on topic '/points', offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
[...]

PR #968, opened 2022-03-03
Make unpublished topics unrecorded by default

theseankelly:
Topics with no publishers do not offer any QoS profiles. This causes the
recorder to subscribe with a default QoS, which is often disadvantageous.

This change makes subscribing to unpublished topics optional, off by
default. Once a publisher is discovered for a topic, discovery picks it
up and adds it to the bag.

This addresses #967

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -61,6 +61,11 @@ class RecordVerb(VerbExtension):
             help='Exclude topics containing provided regular expression. '
             'Works on top of --all, --regex, or topics list.')
         parser.add_argument(
+            '--include-unpublished-topics', action='store_true',
+            help='Discover and record topics which have no publisher. '
+            'Subscriptions on such topics will be made with default QoS unless otherwise '
+            'specified in a QoS overrides file.')
+        parser.add_argument(
             '--include-hidden-topics', action='store_true',
             help='Discover and record hidden topics as well. '
             'These are topics used internally by ROS 2 implementation.')
@@ -228,6 +233,7 @@ class RecordVerb(VerbExtension):
         record_options.compression_threads = args.compression_threads
         record_options.topic_qos_profile_overrides = qos_profile_overrides
         record_options.include_hidden_topics = args.include_hidden_topics
+        record_options.include_unpublished_topics = args.include_unpublished_topics
         record_options.start_paused = args.start_paused
         record_options.ignore_leaf_topics = args.ignore_leaf_topics
 
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -110,8 +110,8 @@ public:
   inline constexpr static const auto kPauseResumeToggleKey = KeyboardHandler::KeyCode::SPACE;
 
 protected:
-  std::unordered_map<std::string, std::string>
-  get_requested_or_available_topics();
+  ROSBAG2_TRANSPORT_EXPORT
+  std::unordered_map<std::string, std::string> get_requested_or_available_topics();
 
 private:
   void topics_discovery();
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -109,11 +109,12 @@ public:
 
   inline constexpr static const auto kPauseResumeToggleKey = KeyboardHandler::KeyCode::SPACE;
 
-private:
-  void topics_discovery();
-
+protected:
   std::unordered_map<std::string, std::string>
   get_requested_or_available_topics();
+
+private:
+  void topics_discovery();
 
   std::unordered_map<std::string, std::string>
   get_missing_topics(const std::unordered_map<std::string, std::string> & all_topics);
--- a/rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
@@ -1,0 +1,201 @@
+// Copyright 2022 Seegrid Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include <chrono>
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "test_msgs/msg/basic_types.hpp"
+#include "test_msgs/message_fixtures.hpp"
+
+#include "rosbag2_test_common/publication_manager.hpp"
+#include "rosbag2_test_common/wait_for.hpp"
+
+#include "rosbag2_transport/recorder.hpp"
+
+#include "record_integration_fixture.hpp"
+
+using namespace std::chrono_literals;  // NOLINT
+
+TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_false_ignores_unpublished)
+{
+  std::string string_topic = "/string_topic";
+  auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
+  auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
+    string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
+
+  rosbag2_transport::RecordOptions record_options = {true, false, {}, "rmw_format", 100ms};
+  record_options.include_unpublished_topics = false;
+  auto recorder = std::make_shared<rosbag2_transport::Recorder>(
+    std::move(writer_), storage_options_, record_options);
+  recorder->record();
+
+  start_async_spin(recorder);
+
+  auto & writer = recorder->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
+  rosbag2_test_common::wait_until_shutdown(
+    std::chrono::seconds(2),
+    [&mock_writer, &expected_messages]() {
+      return mock_writer.get_messages().size() > expected_messages;
+    });
+  (void) expected_messages;  // we can't say anything here, there might be some rosout
+
+  auto recorded_topics = mock_writer.get_topics();
+  EXPECT_EQ(0u, recorded_topics.count(string_topic));
+}
+
+TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_true_includes_unpublished)
+{
+  std::string string_topic = "/string_topic";
+  auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
+  auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
+    string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
+
+  rosbag2_transport::RecordOptions record_options = {true, false, {}, "rmw_format", 100ms};
+  record_options.include_unpublished_topics = true;
+  auto recorder = std::make_shared<rosbag2_transport::Recorder>(
+    std::move(writer_), storage_options_, record_options);
+  recorder->record();
+
+  start_async_spin(recorder);
+
+  auto & writer = recorder->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
+  rosbag2_test_common::wait_until_shutdown(
+    std::chrono::seconds(2),
+    [&mock_writer, &expected_messages]() {
+      return mock_writer.get_messages().size() > expected_messages;
+    });
+  (void) expected_messages;  // we can't say anything here, there might be some rosout
+
+  auto recorded_topics = mock_writer.get_topics();
+  EXPECT_EQ(1u, recorded_topics.count(string_topic));
+}
+
+TEST_F(
+  RecordIntegrationTestFixture,
+  record_all_include_unpublished_false_includes_later_published_default_qos)
+{
+  std::string string_topic = "/string_topic";
+  auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
+  auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
+    string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
+
+  rosbag2_transport::RecordOptions record_options = {true, false, {}, "rmw_format", 100ms};
+  record_options.include_unpublished_topics = false;
+  auto recorder = std::make_shared<rosbag2_transport::Recorder>(
+    std::move(writer_), storage_options_, record_options);
+  recorder->record();
+
+  start_async_spin(recorder);
+
+  // Start up a publisher on our topic *after* the recording has started
+  auto string_message = get_messages_strings()[0];
+  string_message->string_value = "Hello World";
+  rosbag2_test_common::PublicationManager pub_manager;
+
+  // Publish 40 messages at a 50ms interval for a steady 2 seconds worth of data
+  pub_manager.setup_publisher(
+    string_topic, string_message, 40, rclcpp::QoS{rclcpp::KeepAll()}, 50ms);
+
+  ASSERT_TRUE(pub_manager.wait_for_matched(string_topic.c_str()));
+
+  pub_manager.run_publishers();
+
+  auto & writer = recorder->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
+  auto ret = rosbag2_test_common::wait_until_shutdown(
+    std::chrono::seconds(2),
+    [&mock_writer, &expected_messages]() {
+      return mock_writer.get_messages().size() > expected_messages;
+    });
+  auto recorded_messages = mock_writer.get_messages();
+  EXPECT_TRUE(ret) << "failed to capture expected messages in time";
+
+  auto recorded_topics = mock_writer.get_topics();
+  EXPECT_EQ(1u, recorded_topics.count(string_topic));
+
+  // We expect to drop some messages due to the discovery process, but there
+  // should be at least one message received in the duration of this test.
+  auto string_messages = filter_messages<test_msgs::msg::Strings>(
+    recorded_messages, string_topic);
+  ASSERT_THAT(string_messages, SizeIs(Ge(1u)));
+  EXPECT_THAT(string_messages[0]->string_value, Eq("Hello World"));
+}
+
+TEST_F(
+  RecordIntegrationTestFixture,
+  record_all_include_unpublished_false_includes_later_published_sensor_data_qos)
+{
+  std::string string_topic = "/string_topic";
+  auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
+  auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
+    string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
+
+  rosbag2_transport::RecordOptions record_options = {true, false, {}, "rmw_format", 100ms};
+  record_options.include_unpublished_topics = false;
+  auto recorder = std::make_shared<rosbag2_transport::Recorder>(
+    std::move(writer_), storage_options_, record_options);
+  recorder->record();
+
+  start_async_spin(recorder);
+
+  // Start up a publisher on our topic *after* the recording has started
+  auto string_message = get_messages_strings()[0];
+  string_message->string_value = "Hello World";
+  rosbag2_test_common::PublicationManager pub_manager;
+
+  // Publish 40 messages at a 50ms interval for a steady 2 seconds worth of data
+  pub_manager.setup_publisher(string_topic, string_message, 40, rclcpp::SensorDataQoS(), 50ms);
+
+  ASSERT_TRUE(pub_manager.wait_for_matched(string_topic.c_str()));
+
+  pub_manager.run_publishers();
+
+  auto & writer = recorder->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
+  auto ret = rosbag2_test_common::wait_until_shutdown(
+    std::chrono::seconds(2),
+    [&mock_writer, &expected_messages]() {
+      return mock_writer.get_messages().size() > expected_messages;
+    });
+  auto recorded_messages = mock_writer.get_messages();
+  EXPECT_TRUE(ret) << "failed to capture expected messages in time";
+
+  auto recorded_topics = mock_writer.get_topics();
+  EXPECT_EQ(1u, recorded_topics.count(string_topic));
+
+  // We expect to drop some messages due to the discovery process, but there
+  // should be at least one message received in the duration of this test.
+  auto string_messages = filter_messages<test_msgs::msg::Strings>(
+    recorded_messages, string_topic);
+  ASSERT_THAT(string_messages, SizeIs(Ge(1u)));
+  EXPECT_THAT(string_messages[0]->string_value, Eq("Hello World"));
+}
--- a/rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
@@ -33,7 +33,7 @@ using namespace std::chrono_literals;  // NOLINT
 
 TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_false_ignores_unpublished)
 {
-  std::string string_topic = "/string_topic";
+  const std::string string_topic = "/string_topic";
   auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
   auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
     string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
@@ -50,13 +50,12 @@ TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_false_ignore
   MockSequentialWriter & mock_writer =
     static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
 
-  size_t expected_messages = 0;
   rosbag2_test_common::wait_until_shutdown(
     std::chrono::seconds(2),
-    [&mock_writer, &expected_messages]() {
-      return mock_writer.get_messages().size() > expected_messages;
+    [&mock_writer]() {
+      // Return false so we wait the full two seconds
+      return false;
     });
-  (void) expected_messages;  // we can't say anything here, there might be some rosout
 
   auto recorded_topics = mock_writer.get_topics();
   EXPECT_EQ(0u, recorded_topics.count(string_topic));
@@ -64,7 +63,7 @@ TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_false_ignore
 
 TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_true_includes_unpublished)
 {
-  std::string string_topic = "/string_topic";
+  const std::string string_topic = "/string_topic";
   auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
   auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
     string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});
@@ -81,13 +80,12 @@ TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_true_include
   MockSequentialWriter & mock_writer =
     static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
 
-  size_t expected_messages = 0;
   rosbag2_test_common::wait_until_shutdown(
     std::chrono::seconds(2),
-    [&mock_writer, &expected_messages]() {
-      return mock_writer.get_messages().size() > expected_messages;
+    [&mock_writer]() {
+      // Return false so we wait the full two seconds
+      return false;
     });
-  (void) expected_messages;  // we can't say anything here, there might be some rosout
 
   auto recorded_topics = mock_writer.get_topics();
   EXPECT_EQ(1u, recorded_topics.count(string_topic));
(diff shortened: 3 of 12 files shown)


rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+#include "rosbag2_transport/recorder.hpp"
+
+#include "record_integration_fixture.hpp"
+
+using namespace std::chrono_literals;  // NOLINT
+
+TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_false_ignores_unpublished)
+{
+  std::string string_topic = "/string_topic";
```
hidmic, line 36: @theseankelly nit: can be made `const`, same elsewhere below.

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
```diff
@@ -0,0 +1,201 @@
...
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
+  rosbag2_test_common::wait_until_shutdown(
+    std::chrono::seconds(2),
+    [&mock_writer, &expected_messages]() {
+      return mock_writer.get_messages().size() > expected_messages;
+    });
+  (void) expected_messages;  // we can't say anything here, there might be some rosout
```
hidmic, line 59: @theseankelly hmm, if the expectation below is to succeed, no messages should have been received

ros2bag/ros2bag/verb/record.py
```diff
@@ -60,6 +60,11 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             '-x', '--exclude', default='',
             help='Exclude topics containing provided regular expression. '
             'Works on top of --all, --regex, or topics list.')
+        parser.add_argument(
+            '--include-unpublished-topics', action='store_true',
```
hidmic, line 64: @theseankelly nit^2: would `--wait-for-publishers` be a more accurate name for this flag?

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
hidmic: Also, why not constrain recording to relevant topics to mitigate cross-talk?

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
```diff
@@ -0,0 +1,201 @@
...
+    std::move(writer_), storage_options_, record_options);
+  recorder->record();
+
+  start_async_spin(recorder);
+
+  auto & writer = recorder->get_writer_handle();
+  MockSequentialWriter & mock_writer =
+    static_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  size_t expected_messages = 0;
```
hidmic, line 53: @theseankelly nit: can be made `const`, same elsewhere below.

hidmic (commented):
First pass

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
theseankelly: thanks -- will do!

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
theseankelly: thanks -- will do!

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
theseankelly: That particular line (59) was copied from other tests in the same suite  (see `test_record_all_no_discovery.cpp` for example). I'm not crazy about this pattern but I thought consistency would be better in this case.  As an alternative I could just remove the `expected_messages` variable altogether. In fact, I should do so.

The reason not to constrain to relevant topics is because this test suite is a test of the `-a` flag

ros2bag/ros2bag/verb/record.py
theseankelly: The new behavior (don't add unpublished topics to the bag) is *off* by default in the current implementation. Naming it `--wait-for-publishers` would imply that the new behavior should be *on* by default.

Given the description in the underlying issue report, I feel like the most commonly desired behavior is to *not* record bags until they are published so that we avoid any QoS incompatibilities.

I could be persuaded if there's strong desire -- I do realize this change alters the default behavior of `ros2 bag record -a` w.r.t unpublished topics.

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
theseankelly: Actually I'm glad you brought this up because I misunderstood the usage of `rosbag2_test_common::wait_until_shutdown` -- I think some of my other tests may be ending prematurely. Stay tuned...

theseankelly:
Thanks very much for the review -- took care of what I could. The rest are pending the conversations above..

ros2bag/ros2bag/verb/record.py
hidmic: > The new behavior (don't add unpublished topics to the bag) is off by default in the current implementation. Naming it --wait-for-publishers would imply that the new behavior should be on by default.

Yeah, fair point.

> I could be persuaded if there's strong desire -- I do realize this change alters the default behavior of `ros2 bag record -a` w.r.t unpublished topics.

That's true, but I agree that the current default behavior is less than intuitive at best and harmful at worst (e.g. if `/tf_static` messages are silently dropped). 

It does beg the question though, what would be the use case for early subscription with default QoS policies? Grabbing the first few messages sent by a publisher? If there's no other subscriber from which to derive the topic type, DDS discovery latency and graph polling will make it impossible. CC @emersonknapp for feedback.

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
```diff
@@ -81,13 +80,12 @@ TEST_F(RecordIntegrationTestFixture, record_all_include_unpublished_true_include
...
 
-  size_t expected_messages = 0;
   rosbag2_test_common::wait_until_shutdown(
     std::chrono::seconds(2),
-    [&mock_writer, &expected_messages]() {
-      return mock_writer.get_messages().size() > expected_messages;
+    [&mock_writer]() {
+      // Return false so we wait the full two seconds
+      return false;
     });
```
hidmic, line 88: @theseankelly nit: looking at this code, I suspect that what you actually want is simply `rclcpp::sleep_for(std::chrono::seconds(2))`. Same elsewhere.

hidmic (commented):
Overall LGTM but for a few comments.

rosbag2_transport/test/rosbag2_transport/test_record_all_include_unpublished_topics.cpp
theseankelly: indeed. thanks for the simplication.

ros2bag/ros2bag/verb/record.py
theseankelly: I'm also interested in hearing what @emersonknapp has to say.

I think there are a couple (maybe contrived..) scenarios where the feature has value. For example, spinning up a background rosbag snapshotter as part of a robot topology - it might be useful to not have to worry about boot order of the various nodes in the system.

Definitely niche, and might only be useful with a couple other flags (`--qos-profiles-override-path` and `--no-discovery` or something).

MichaelOrlov (changes requested):
@theseankelly Thank you for your contribution.  
While implementation for fix looks good to me, I have some significant concerns in regards to the unit tests.

First of all waiting for 2-4 seconds in each test doesn't look good for me. We have hundreds of tests on CI and if wait for 4 seconds in each test our CI will be running for ever.
The second my concern is non-deterministic behavior in tests. Relying on sleep for 2 seconds in tests tend to make them flaky or false negative.

Although I have suggestion how to get rid from delay in tests and make them deterministic.
Need create `MockReader` derived from `rosbag2_transport::Reader`. Similar way how I made  [MockPlayer](https://github.com/ros2/rosbag2/blob/master/rosbag2_transport/test/rosbag2_transport/mock_player.hpp)
I envision `MockReader ` to be as
```cpp
class MockRecorder : public rosbag2_transport::Recorder
{
public:
  MockRecorder(
    std::shared_ptr<rosbag2_cpp::Writer> writer,
    const rosbag2_transport::RecordOptions & record_options)
  : Recorder(writer, storage_options, record_options)
  {}

  template<typename DurationRepT = int64_t, typename DurationT = std::milli>
  bool wait_for_topic_to_be_discovered(
    const std::string & topic_name_to_wait_for,
    std::chrono::duration<DurationRepT, DurationT> timeout = std::chrono::seconds(10))
  {
    bool discovered = false;
    using clock = std::chrono::steady_clock;
    auto start = clock::now();
    do {
      auto topic_names_and_types = this->get_topic_names_and_types();
      for (const auto &[topic_name, topic_types] : topic_names_and_types) {
        if (topic_name_to_wait_for == topic_name) {
          discovered = true;
          break;
        }
      }
      std::this_thread::sleep_for(std::chrono::milliseconds(20));
    } while (!discovered && (clock::now() - start) < timeout);
    return discovered;
  }

  bool topic_available_for_recording(const std::string & topic_name)
  {
    bool available_for_recording = false;
    auto topics_to_subscribe = get_requested_or_available_topics();
    for (const auto & topic_and_type : topics_to_subscribe) {
      if (topic_and_type.first == topic_name) {
        available_for_recording = true;
        break;
      }
    }
    return available_for_recording;
  }
};
```
As example one of the test rewritten with  `MockRecorder`
```cpp
TEST_F(
  RecordIntegrationTestFixture,
  record_all_include_unpublished_false_includes_later_published_default_qos)
{
  const std::string string_topic = "/string_topic";
  auto node = std::make_shared<rclcpp::Node>("test_string_msg_listener_node");
  auto string_msgs_sub = node->create_subscription<test_msgs::msg::Strings>(
    string_topic, 10, [](test_msgs::msg::Strings::ConstSharedPtr) {});

  rosbag2_transport::RecordOptions record_options = {true, false, {}, "rmw_format", 100ms};
  record_options.include_unpublished_topics = false;
  auto recorder = std::make_shared<MockRecorder>(writer_, storage_options_, record_options);
  recorder->record();
  start_async_spin(recorder);

  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(string_topic));
  ASSERT_FALSE(recorder->topic_available_for_recording(string_topic));

  // Start up a publisher on our topic *after* the recording has started
  auto string_message = get_messages_strings()[0];
  string_message->string_value = "Hello World";
  rosbag2_test_common::PublicationManager pub_manager;

  // Publish 10 messages at a 30ms interval for a steady 300 milliseconds worth of data
  pub_manager.setup_publisher(
    string_topic, string_message, 10, rclcpp::QoS{rclcpp::KeepAll()}, 30ms);

  ASSERT_TRUE(pub_manager.wait_for_matched(string_topic.c_str()));
  pub_manager.run_publishers();
  ASSERT_TRUE(recorder->topic_available_for_recording(string_topic));
}
```

theseankelly:
Thanks very much for the detailed feedback @MichaelOrlov -- My test implementation follows the patterns established in other unit tests in this package, however I share your concerns and I ought to have not taken the easy road out.

I'll dig in to the suggested `MockRecorder` and re-surface with reimplemented tests shortly!

theseankelly:
Thanks again, @MichaelOrlov -- I have added your `MockRecorder` and used it throughout. Neat way to limit the scope of the test surface; thanks for that idea. I did have to make `get_requested_or_available_topics` a `protected` member of `rosbag2_transport::Recorder` JFYI.

MichaelOrlov (approved):
@theseankelly Thank you for collaboration and support. Now it looks good enough to go. I approve this PR.
We will need to make sure that CI passing green and I think after that will be able to merge it.

ros2bag/ros2bag/verb/record.py
MichaelOrlov: @emersonknapp Gentle ping in this issue if you have any concerns or suggestions to the proposed solution it's good time to express it. 
Otherwise if CI will be green I think we will merge it Monday or Tuesday.

IMO alternative scenario described in #967 is a valid case which should be handled correctly
```
1. In terminal 1 create a subscriber by running ros2 topic echo --qos-reliability best_effort /foo std_msgs/msg/Int32
2. In terminal 2 run ros2 record -a
    a. NOTE in the output that the topic /foo was added to the bag -- found by the DataReader
3. In terminal 3 create a publisher via ros2 topic pub --qos-reliability best_effort /foo std_msgs/msg/Int32 '{value: 1}'

The DataWriter (and its offered QoS) will be discovered, and terminal 2 (rosbag record) will display the warning about incompatible publishers.
```

MichaelOrlov:
@theseankelly Unfortunately CI fails with error maessage:
```
FAIL: test_record.TestRecord.test_output
---------------------------------------------------------------------------------------------------------------------
  File "/home/michael/ros2_rolling_new/src/ros2/rosbag2/ros2bag/test/test_record.py", line 60, in test_output
    proc_output.assertWaitFor(
  File "/home/michael/ros2_rolling_new/build/launch_testing/launch_testing/io_handler.py", line 146, in assertWaitFor
    assert success, 'Waiting for output timed out'
AssertionError: Waiting for output timed out
----------------------------- Captured stdout call -----------------------------
[INFO] [launch]: All log files can be found below /home/michael/.ros/log/2022-03-20-13-46-28-333904-rosbag2_new-159318
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [ros2bag-cli-3]: process started with pid [159366]
[ros2bag-cli-3] stdin is not a terminal device. Keyboard handling disabled.[INFO] [1647809193.066783965] [rosbag2_recorder]: Press SPACE for pausing/resuming
[ros2bag-cli-3] [INFO] [1647809193.068556236] [rosbag2_storage]: Opened database '/tmp/tmppx8ettz6/ros2bag_test_record/ros2bag_test_record_0.db3' for READ_WRITE.
[ros2bag-cli-3] [INFO] [1647809193.068602576] [rosbag2_recorder]: Listening for topics...
[ros2bag-cli-3] [INFO] [1647809193.072296612] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [ros2bag-cli-3]: sending signal 'SIGINT' to process[ros2bag-cli-3]
[ros2bag-cli-3] [INFO] [1647809203.090175676] [rclcpp]: signal_handler(signum=2)
[ros2bag-cli-3] [INFO] [1647809203.099934439] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [ros2bag-cli-3]: process has finished cleanly [pid 159366]
----------------------------- Captured stderr call -----------------------------
test_output (test_record.TestRecord) ... FAIL
```
I was able to reproduce this error locally with `colcon test --packages-select ros2bag`. 
It seems regression in test. 
I tried to add `'--include-unpublished-topics'` option to the  [record_all_process](https://github.com/ros2/rosbag2/blob/master/ros2bag/test/test_record.py#L38)
And test start passing. 

Could you please dig deeper to figure out why this specific test not able to pass with default parameters?

MichaelOrlov:
@jacobperron It seems you was the author of the [ros2bag/test/test_record.py](https://github.com/ros2/rosbag2/blob/master/ros2bag/test/test_record.py).

Could you please explain how the test works?
I found where `ros2 bag record -a` command forming, but still can't figure out where topics publishing from?

theseankelly:
@MichaelOrlov yeah I'll dig in

jacobperron:
`test_record.py` expects that ros2bag will subscribe to the `/rosout` and  `/parameters_events` topics and by checking that the process prints the following lines:

```
Subscribed to topic '/rosout'
Subscribed to topic '/parameter_events'
```

It seems with this PR, ros2bag no longer automatically subscribes to the `/parameter_events` topic and so the test fails. Maybe this is correct behavior though and the test should be updated? It seems that `ros2 bag record` does not create a publisher to `/parameter_events`, so it may make sense that we don't subscribe to it initially.

MichaelOrlov:
@jacobperron Thanks for the explanation. 
What is still unclear for me is where from `/parameter_events` topic  supposed to be published in unit test?

jacobperron:
`/parameter_events` is a default topic advertised by ROS 2 nodes. A publisher is [created whenever a node is created](https://github.com/ros2/rclpy/blob/bed714f5c95c06885098989b4107421dc0b60b2a/rclpy/rclpy/node.py#L199-L200). It is expected to publish a message whenever there is a change to one of its parameters. So, in the case of the test, it is not necessarily expecting messages to be published, but it expecting the topic to exist and for the rosbag recorder to be subscribed.

theseankelly:
Hey, thanks for the details and sorry I've been slow on this one.

I'm in agreement with @jacobperron that I don't understand what about this change would have caused a test regression. For example, the output of just running `ros2 bag record -a` produces the following output:
```
$ ros2 bag record -a
1648028103.984134 [0]       ros2: selected interface "lo" is not multicast-capable: disabling multicast
[INFO] [1648028103.987354149] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1648028103.988991693] [rosbag2_storage]: Opened database 'rosbag2_2022_03_23-09_35_03/rosbag2_2022_03_23-09_35_03_0.db3' for READ_WRITE.
[INFO] [1648028103.989131022] [rosbag2_recorder]: Listening for topics...
[INFO] [1648028103.990820373] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1648028103.991566000] [rosbag2_recorder]: Subscribed to topic '/parameter_events'
```
Which ought to satisfy the unit test's checks.

Furthermore while the recording is active, there is indeed a publisher on the topic:
```
ros2 topic info -v /parameter_events
Type: rcl_interfaces/msg/ParameterEvent

Publisher count: 1

Node name: _ros2cli_daemon_0_654e377106084d5cbca689eaf0c92130
Node namespace: /
Topic type: rcl_interfaces/msg/ParameterEvent
Endpoint type: PUBLISHER
GID: 40.64.10.01.0f.e8.92.04.e2.1c.00.1a.00.00.07.03.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (1000)
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Subscription count: 2

Node name: rosbag2_recorder
Node namespace: /
Topic type: rcl_interfaces/msg/ParameterEvent
Endpoint type: SUBSCRIPTION
GID: 09.2f.10.01.04.40.f9.cf.04.2b.0c.0a.00.00.13.04.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (1000)
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Node name: rosbag2_recorder
Node namespace: /
Topic type: rcl_interfaces/msg/ParameterEvent
Endpoint type: SUBSCRIPTION
GID: 09.2f.10.01.04.40.f9.cf.04.2b.0c.0a.00.00.15.04.00.00.00.00.00.00.00.00
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (10)
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite
```

I will do some further investigation into the root cause before adding this parameter.

theseankelly:
Test consistently passes for me.
```
$ launch_test src/ros2/rosbag2/ros2bag/test/test_record.py 
[INFO] [launch]: All log files can be found below /home/skelly/.ros/log/2022-03-23-09-56-36-177449-skelly-l5550-22397
[INFO] [launch]: Default logging verbosity is set to INFO
test_output (test_record.TestRecord) ... [INFO] [ros2bag-cli-1]: process started with pid [22408]
[ros2bag-cli-1] stdin is not a terminal device. Keyboard handling disabled.1648029396.757729 [0]       ros2: selected interface "lo" is not multicast-capable: disabling multicast
[ros2bag-cli-1] [INFO] [1648029396.761074008] [rosbag2_recorder]: Press SPACE for pausing/resuming
[ros2bag-cli-1] [INFO] [1648029396.762860023] [rosbag2_storage]: Opened database '/tmp/tmp_i373pwf/ros2bag_test_record/ros2bag_test_record_0.db3' for READ_WRITE.
[ros2bag-cli-1] [INFO] [1648029396.762974865] [rosbag2_recorder]: Listening for topics...
[ros2bag-cli-1] [INFO] [1648029396.764584742] [rosbag2_recorder]: Subscribed to topic '/rosout'
[ros2bag-cli-1] [INFO] [1648029396.765429422] [rosbag2_recorder]: Subscribed to topic '/parameter_events'
ok

----------------------------------------------------------------------
Ran 1 test in 0.584s

OK
[INFO] [ros2bag-cli-1]: sending signal 'SIGINT' to process[ros2bag-cli-1]
[ros2bag-cli-1] [INFO] [1648029396.767560896] [rclcpp]: signal_handler(signum=2)
[ros2bag-cli-1] [INFO] [1648029396.867401609] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [ros2bag-cli-1]: process has finished cleanly [pid 22408]
test_exit_code (test_record.TestRecordAfterShutdown) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
```

theseankelly:
Oh interesting, I can reproduce this if I use FastDDS instead of CycloneDDS.

I'll keep digging now that I have a repro.

MichaelOrlov:
@theseankelly Thanks for catching it up.  
I am guessing that it could be a race condition in `node_graph` in the way that we are reading out information about topics names than taking information about each specific topic.
It could be that information about topic name became available first and than information about topic's info updates in `node_graph` non-atomically without protection for data integrity.  
However I don't know for sure. Need to verify my hypothesis.

jacobperron:
When I tried this PR locally, I do **not** see the log about subscribing to /parameter_events, using `ros2 bag record -a`. I think it might be picking up a publisher from the ros2 daemon in your case. First try killing the daemon with `ros2 daemon stop` and then try `ros2 bag record -a`.

theseankelly:
> When I tried this PR locally, I do **not** see the log about subscribing to /parameter_events, using `ros2 bag record -a`. I think it might be picking up a publisher from the ros2 daemon in your case. First try killing the daemon with `ros2 daemon stop` and then try `ros2 bag record -a`.

Aha -- yes indeed. I'll keep poking; I'm expecting to have more time in my schedule to see this through to completion this coming week.

theseankelly:
I rolled up my sleeves for a bit of a struggle root causing this but it turns out this is pretty easily explained...

From `rosbag2_transport/src/rosbag2_transport/recorder.cpp`:
```cpp
Recorder::Recorder(
  std::shared_ptr<rosbag2_cpp::Writer> writer,
  std::shared_ptr<KeyboardHandler> keyboard_handler,
  const rosbag2_transport::RecordOptions & record_options,
  const std::string & node_name,
  const rclcpp::NodeOptions & node_options)
: rclcpp::Node(node_name, rclcpp::NodeOptions(node_options).start_parameter_event_publisher(false)),

... etc ...
```

So the recorder node explicitly does not start a publisher on `/parameter_events`. With my change, this behavior is expected and correct.

I'm not sure what the best resolution is. I could conceive of:
1. Modify the test to record with `--include-unpublished-topics`
2. Modify the test assertions to drop the check for `Subscribed to topic /parmeter_events`
3. Modify the implementation of the recorder to explicitly subscribe on `/parameter_events` even if there are no publishers

I'd like the opinion of one of the upstream devs -- @MichaelOrlov @jacobperron ?

jacobperron:
Thanks for looking into it :+1: 

I think option 2 makes the most sense; remove the assertion that is no longer valid.

ros2bag/ros2bag/verb/record.py
emersonknapp: Sorry for late reply - my perspective is that "will not create a subscription until at least one publisher is discovered" should be the default behavior of discovery  That seems more intelligent behavior overall. In fact, I'm not sure I see any reason to turn it off, unless we're worried that the discovery loop is a performance overhead that a user may want to explicitly disable?

Even in that case, the default QoS is kind of an arbitrary baseline subscription to fall back to. It would probably make more sense that "if you provide a QoS override for an explicitly specified subscription topic, then the subscription will be created immediately using that override and will not impact discovery".

Given that, I'm not sure we need a new commandline option for this feature! I'll note that this is a good thing probably, we already have an unwieldy number of options.

MichaelOrlov:
@theseankelly I agree with option 2.

theseankelly:
Change made, rebased, force pushed.

MichaelOrlov (approved):

ros2bag/ros2bag/verb/record.py
MichaelOrlov: @emersonknapp  I also don't envision the case when we might need to use `--include-unpublished-topics` parameter.   

@theseankelly  Do you have any scenario in mind or in practice when we might need to use `--include-unpublished-topics`?  
If not I would ask you to remove this extra parameter from command line option.

IMO Possible concern that the discovery thread affecting performance doesn't count. Since it's very minor and used to be only for the begging of the recording.

I agree with @emersonknapp that 
> "if you provide a QoS override for an explicitly specified subscription topic, then the subscription will be created immediately using that override and will not use discovery".

However I would moved it as follow up feature request for the sake of saving time to merge current PR which is considering as a "bugfix".

theseankelly:
Hey @MichaelOrlov assuming the signoff is still good, could this be merged prior to humble being spun up?

clalancette:
> Hey @MichaelOrlov assuming the signoff is still good, could this be merged prior to humble being spun up?

Sorry, we are in a feature and API freeze for Humble at this point.  We can merge this in once Humble has been branched off.

theseankelly:
Thanks for the info @clalancette -- is that to say this fix will miss Humble entirely, or it will just need to come in after the Humble branches are created?

clalancette:
> Thanks for the info @clalancette -- is that to say this fix will miss Humble entirely, or it will just need to come in after the Humble branches are created?

If it doesn't break (public) API and isn't a new feature, the maintainers of rosbag2 can consider it for Humble.

theseankelly:
@emersonknapp @MichaelOrlov hoping to get the port-to-humble conversation started early. It seems to me that this is a pretty fundamental problem (we currently have to specify custom qos override files for all bag record/playback) and it'd be great to get it in to humble.

This doesn't break public API, but I wonder if the new command line flag would be a 'new feature'.

Could you comment? I'd be open to spinning up a version of this change which hard-codes the new default behavior without creating a new command line flag if that's going to be a problem.

MichaelOrlov:
Running CI
Gist: https://gist.githubusercontent.com/MichaelOrlov/2d562f3993ab6ef0d8af5e8b805502c9/raw/bbc96612ae3d03fa3d2a257da6fa1150e7d7eeb7/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_py rosbag2_transport rosbag2_tests
TEST args: --packages-select ros2bag rosbag2_py rosbag2_transport rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10262/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16663/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11249/)
* Windows [Build Status](http://ci.ros2.org/job/[ci_windows](https://ci.ros2.org/job/ci_windows/)/17058/)

MichaelOrlov:
@clalancette @emersonknapp Could you please advise how to proceed with this PR?  
We need to merge it to the `rolling/master` and to the `Humble` release.  
I don't have enough experience with process of how to correctly merge it to the release branch in current situation.
I
 wouldn't consider this PR as a new feature. This is more like a bug fix. 
It's kind of fixing unexpected behavior and providing option via command line argument to be able to return to the old behavior.
Technically we are not changing API, however we are adding one field to the `record_options` data structure which is passing as parameter for the `rosbag2::Recorder`. I think it could be considered as API changes, since new version of recorder will expect this new field to exist.

BTW. I have tried to run CI for the current branch. But it seems CI launcher is broken or get stuck somewhere for unknown reason.

clalancette:
> BTW. I have tried to run CI for the current branch. But it seems CI launcher is broken or get stuck somewhere for unknown reason.

Can you be more specific?  The [nightly builds](https://ci.ros2.org/view/nightly/) seem to be working fine, what problem are you having with the builds?

In general, the way forward with this PR is to get CI passing, get it approved, and merge it onto the master branch.  Once that is done, we can consider backporting it to Humble.  We are still before the release, so we can still break ABI at this point, but the time in which to do that is short; we'll stop taking ABI breaks on May 19th.

rosbag2_transport/include/rosbag2_transport/recorder.hpp
```diff
@@ -109,12 +109,13 @@ class Recorder : public rclcpp::Node
 
   inline constexpr static const auto kPauseResumeToggleKey = KeyboardHandler::KeyCode::SPACE;
 
-private:
-  void topics_discovery();
-
+protected:
   std::unordered_map<std::string, std::string>
   get_requested_or_available_topics();
```
MichaelOrlov, line 114: @theseankelly Windows build fails with linker error:
```
12:49:07 test_record_all_include_unpublished_topics.obj : error LNK2019: unresolved external symbol "protected: class std::unordered_map<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >,struct std::hash<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >,struct std::equal_to<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > >,class std::allocator<struct std::pair<class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const ,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > > > > __cdecl rosbag2_transport::Recorder::get_requested_or_available_topics(void)"
```
It seems to fix this linker error need to add 
ROSBAG2_TRANSPORT_EXPORT before declaration of the `get_requested_or_available_topics(void)`

MichaelOrlov:
@clalancette Thank you for information. Perhaps we will be able to merge it soon.   
As regards to the confusing comment about 
> BTW. I have tried to run CI for the current branch. But it seems CI launcher is broken or get stuck somewhere for unknown reason.

> Can you be more specific? The [nightly builds](https://ci.ros2.org/view/nightly/) seem to be working fine, what problem are you having with the builds?

Sorry for the confusion. At the time when I was writing those comment the CI launcher was reporting status **"not running"** for all three jobs and was hanging out in such state for a 30 minutes.

I want to clarify procedure for backporting PR to the Humble.  
Should we crate separate PR for backporting targeting https://github.com/ros2/rosbag2/tree/humble branch, get approval, pass CI to be green again and finally merge it?

clalancette:
> Sorry for the confusion. At the time when I was writing those comment the CI launcher was reporting status **"not running"** for all three jobs and was hanging out in such state for a 30 minutes.

Oh, right.  We don't always have enough executors (particularly on Windows), so it can just take some time for the jobs to run.

> Should we crate separate PR for backporting targeting https://github.com/ros2/rosbag2/tree/humble branch, get approval, pass CI to be green again and finally merge it?

Yes, that is generally correct.  At this point of the release, the one additional thing to do is to make sure to get @audrow 's approval before merging it into the `humble` branch.

rosbag2_transport/include/rosbag2_transport/recorder.hpp
MichaelOrlov, line 114:
```suggestion
  ROSBAG2_TRANSPORT_EXPORT
  std::unordered_map<std::string, std::string> get_requested_or_available_topics();
```

MichaelOrlov:
Re-running Windows build again after addressing linker errors:
* Windows [Build Status](http://ci.ros2.org/job/[ci_windows](https://ci.ros2.org/job/ci_windows/)/17070/)

PR #969, opened 2022-03-07
Use git hash for zstd vendor

sloretz:
CMake ExternalProject_add recommends using a specific git hash with
GIT_TAG because branches and tags can be updated to point to different
references.

https://cmake.org/cmake/help/latest/module/ExternalProject.html

This switches zstd_vendor to a git hash.
sqlite3_vendor and shared_queues_vendor don't need to be changed because
they use URL and URL_MD5 instead.

Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>

### DIFF
--- a/zstd_vendor/CMakeLists.txt
+++ b/zstd_vendor/CMakeLists.txt
@@ -32,7 +32,7 @@ macro(build_zstd)
   # We need to configure the CMake command to build from there instead.
   ExternalProject_Add(zstd-${zstd_version}
     GIT_REPOSITORY https://github.com/facebook/zstd.git
-    GIT_TAG v${zstd_version}
+    GIT_TAG 10f0e6993f9d2f682da6d04aa2385b7d53cbb4ee  # v${zstd_version}
     GIT_CONFIG advice.detachedHead=false
     # Suppress git update due to https://gitlab.kitware.com/cmake/cmake/-/issues/16419
     # See https://github.com/ament/uncrustify_vendor/pull/22 for details


jhdcs:
There appears to be a build error on Ubuntu Jammy - It's not able to find a build target for Python 3.9?

clalancette:
> There appears to be a build error on Ubuntu Jammy - It's not able to find a build target for Python 3.9?

Ubuntu Jammy is currently going through a (painful) Python 3.9 -> 3.10 transition.  Some of our underlying dependencies haven't been rebuilt, and so things are not working.  At the moment, we are relying strictly on builds from https://ci.ros2.org, which doesn't have this problem (due to the way it's Python is setup, which is different than for the PR jobs).  Once Jammy sorts itself out with the Python version, PR builds should come back.

clalancette (approved):

sloretz:
CI (build: `--packages-above-and-dependencies zstd_vendor` test: `--packages-select zstd_vendor`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16301/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16668/)

sloretz:
Looks like CI is green :tada: . @jhdcs if it looks good to you, may I merge this?

jhdcs (approved):
Looks good to me!

Issue #547, opened 2020-10-26
Design/Doc: sqlite directory storage format


chaoflow:
We're looking into storing additional files alongside a rosbag2, e.g. pcaps, pdfs, ... and are thinking to place these into the rosbag2 directory. Given that metadata.yaml lists the files relevent for reading rosbag2 this should not pose a problem, correct?

The rosbag2 [design article](https://github.com/ros2/design/blob/f69fbbd11848e3dd6866b71a158a1902e31e92f1/articles/rosbags.md) linked from [README](https://github.com/ros2/rosbag2/blob/master/README.md) does not specify the file system layout. Is there a more recent version?

---

Conclusions from discussions:

- [x] Adding additional files including subdirectories to a rosbag2 directory is allowed and encouraged if the files belong to the recording
- [x] metadata.yaml is primary location for metadata, db3-files may duplicate that data
- [ ] Additional metadata options:
  1. top-level ``metadata.yaml
  2. under top-level ``custom`` key; alternative name: ``userdata``
  3. under ``rosbag2_bagfile_information`` subkey ``custom/userdata`` -- this would enable the ``BagMetadata`` class to pass this information to a single-file storage format
- [ ] Message definitions will be stored in ``BagMetadata`` version ``>4``, probably ``TopicMetadata`` see #782

emersonknapp:
The directory contains metadata.yaml and all of the sqlite databases (if splitting is enabled) - putting extra files into the directory should not cause any problems as rosbag2 is currently implemented. It certainly does make sense to store this extra information with the bag if it is related to the recording session.

chaoflow:
@emersonknapp Great, thank you for confirming that. Would it make sense to extend the design article to include a section about file system layout or is there some other place better suited?

emersonknapp:
I have a hard time deciding on that - I feel like design/documentation should live here with the rosbag2 project, instead of in ros2/design, for better synchronization, including updating across releases without losing the context for any given release. @Karsten1987 do you have an opinion on that?

emersonknapp:
Considering starting a readthedocs page for this - or a dedicated documentation page like nav2

chaoflow:
[...]

PR #976, opened 2022-03-24
Add the ability to record any key/value pair in the 'custom' field in metadata.yaml

t0ny-peng:
Usage: `ros2 bag record --all --custom-data key1=value1 key2=value2`. The custom data will show up in the `metadata.yaml` file:
```yaml
rosbag2_bagfile_information:
  ...
  custom:
    key1: value1
    key2: value2
```

Closes [#547](https://github.com/ros2/rosbag2/issues/547)


rosbag2_storage/include/rosbag2_storage/yaml.hpp
```diff
@@ -41,6 +42,27 @@ void optional_assign(const Node & node, std::string field, T & assign_to)
...
+
+  static bool decode(const Node & node, std::unordered_map<std::string, std::string> & custom_data)
+  {
+    for (YAML::const_iterator it = node.begin(); it != node.end(); ++it) {
+      custom_data.emplace(it->first.as<std::string>(), it->second.as<std::string>());
+    }
+    return true;
+  }
+};
+
```
MichaelOrlov, line 65: I think it would make sense to put this `convert` specialization to the `metadata_io.cpp` as others `convert` template specializations related to the metadata.

rosbag2_storage/include/rosbag2_storage/yaml.hpp
```diff
@@ -41,6 +42,27 @@ void optional_assign(const Node & node, std::string field, T & assign_to)
...
+    for (const auto & it : custom_data) {
+      node[it.first] = it.second;
+    }
+    return node;
+  }
+
+  static bool decode(const Node & node, std::unordered_map<std::string, std::string> & custom_data)
+  {
+    for (YAML::const_iterator it = node.begin(); it != node.end(); ++it) {
+      custom_data.emplace(it->first.as<std::string>(), it->second.as<std::string>());
```
MichaelOrlov, line 60: `emplace` can fail to create new element. Need to add a sanity check for `emplace` result 
e.g.
```cpp
      if (!custom_data.emplace(it->first.as<std::string>(), it->second.as<std::string>()).second)
      {
        return false;
      }
```

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -225,6 +226,11 @@ struct convert<rosbag2_storage::BagMetadata>
       metadata.files =
         node["files"].as<std::vector<rosbag2_storage::FileInformation>>();
     }
+
+    if (node["custom"]) {
```
MichaelOrlov, line 230: Out of curiosity, is it really a valid check `if (node["custom"])` ?
Will not it always return true as in case with std::unordered_map?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -199,6 +199,7 @@ struct convert<rosbag2_storage::BagMetadata>
     node["compression_mode"] = metadata.compression_mode;
     node["relative_file_paths"] = metadata.relative_file_paths;
     node["files"] = metadata.files;
+    node["custom"] = metadata.custom_data;
```
MichaelOrlov, line 202: Suggestion to rename `custom` field to the `custom_data`.

MichaelOrlov (changes requested):
@t0ny-peng Thank you for contribution.  Overall looks good.
I have some minor findings. Please see them in inline comments.

I also have suggestion to rename `node["custom"]` to the `node["custom_data"]`.
name `custom`  is to generic and could confuse to what is relates for.

ros2bag/ros2bag/verb/record.py
```diff
@@ -198,6 +205,13 @@ def main(self, *, args):  # noqa: D102
             except (InvalidQoSProfileException, ValueError) as e:
                 return print_error(str(e))
 
+        # Prepare custom_data dictionary
+        custom_data = {}
+        if args.custom_data:
+            key_value_pais = [pair.split('=') for pair in args.custom_data]
```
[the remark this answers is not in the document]
t0ny-peng: Fixed

ros2bag/ros2bag/verb/record.py
```diff
@@ -198,6 +205,13 @@ def main(self, *, args):  # noqa: D102
...
+            custom_data = {pair[0]: pair[1] for pair in key_value_pais}
+            print(custom_data)
```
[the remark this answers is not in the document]
t0ny-peng: Added by mistake.  Removed

rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp
```diff
@@ -52,6 +53,7 @@ struct BagMetadata
   std::vector<TopicInformation> topics_with_message_count;
   std::string compression_format;
   std::string compression_mode;
+  std::unordered_map<std::string, std::string> custom_data;  // {key: value, ...}
```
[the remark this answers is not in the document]
t0ny-peng: Updated to 6.

ros2bag/ros2bag/verb/record.py
```diff
@@ -148,6 +148,13 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
                  'for storage (the defaut). This flag settings can still be overriden by '
                  'corresponding settings in the config passed with --storage-config-file.'
         )
+        parser.add_argument(
+            '--custom-data', type=str, metavar='KEY=VALUE', nargs='*',
+            help='Store the custom data in metadata.yaml'
+                 'under "rosbag2_bagfile_information/custom". The key=value pair can '
+                 'appear more than once.'
```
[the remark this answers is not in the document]
t0ny-peng: By appear more than once I mean the latter one will override the former ones. The description is updated.

rosbag2_storage/include/rosbag2_storage/yaml.hpp
t0ny-peng: Both `metadata_io.cpp` and `storage_options.cpp` needs this conversion so I placed it here

rosbag2_storage/include/rosbag2_storage/yaml.hpp
t0ny-peng: Since this method simply decode a yamlcpp Node to a `std::map`, I'm wondering in what case the `emplace` will fail? Note that the internal data structure in yamlcpp Node can already guarantee the uniqueness.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
t0ny-peng: `operator[]` of YAML-cpp Node will not create the element if the key doesn't exist.

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
t0ny-peng: ok changed

rosbag2_storage/include/rosbag2_storage/yaml.hpp
MichaelOrlov: From https://en.cppreference.com/w/cpp/container/unordered_map/emplace
```cpp
template< class... Args >
std::pair<iterator,bool> emplace( Args&&... args );
```
> Return value
> Returns a pair consisting of an iterator to the inserted element, or the already-existing element if no insertion happened, and a bool denoting whether the insertion took place (true if insertion happened, false if it did not).

rosbag2_storage/include/rosbag2_storage/yaml.hpp
MichaelOrlov: @t0ny-peng I don't really like that similar definitions spread out across the different files.  
It looks like that other `yaml convert` operators was incorrectly put in the `metadata_io.cpp` file from the very beginning.  
Since we have a dedicated file for `YAML` specific operations and definitions aka `yaml.hpp` it would more appropriate to all specific yaml conversions to this file.

Could you please move other `yaml` converter operators from  `metadata_io.cpp` to the `yaml.hpp` file?
i.e. everything under the `namespace YAML`.

I have tried it locally on my machine and it's compiles and passes unit tests.

MichaelOrlov (changes requested):
@t0ny-peng Thank you for the iteration in code review.
I have a few more requests for changes.
Also could you please next time push new changes as a separate commit?
One commit per one review iteration should be fine.
It's very difficult to review code outside of the Github e.g. in CLion, when all changes squashed in one commit and pushed with force command. i.e. there are no history for the changes outside of the PR.

Multiple commits will not clutter the final git history since we have mandatory squash for commits when we do merge on master. Although it's really helpful for code review.

rosbag2_storage/include/rosbag2_storage/yaml.hpp
t0ny-peng: Yes that is indeed the return value of `emplace` method. However, like I said, since we are decoding a YAML-cpp node to a STL map, the case of duplicate key, i.e., insertion failure, won't happen because YAML-cpp already did the deduplication.

rosbag2_storage/include/rosbag2_storage/yaml.hpp
t0ny-peng: @MichaelOrlov OK moved here

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -199,6 +199,9 @@ struct convert<rosbag2_storage::BagMetadata>
     node["compression_mode"] = metadata.compression_mode;
     node["relative_file_paths"] = metadata.relative_file_paths;
     node["files"] = metadata.files;
+    if (!metadata.custom_data.empty()) {
+      node["custom_data"] = metadata.custom_data;
+    }
```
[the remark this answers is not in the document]
t0ny-peng: Ok removed.

t0ny-peng:
@MichaelOrlov I've addressed your comments. I didn't split the first commit because it seems pretty atomic. The changes are all related and cannot be broken down to subcomponents. However moving the convert structs has its own commit. Please review. Thx

MichaelOrlov (approved):
@t0ny-peng Thanks, approving now.

As regards to the 
>  I didn't split the first commit because it seems pretty atomic. The changes are all related and cannot be broken down to subcomponents. However moving the convert structs has its own commit.

It's ok and this is exactly how I expected it to be.

MichaelOrlov:
Running CI
Gist: https://gist.githubusercontent.com/MichaelOrlov/f295a4dda3d8ce90076def0a0646c519/raw/620153ccb8176c963f45162d6b39e1903ee19bf2/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_cpp rosbag2_storage rosbag2_tests
TEST args: --packages-select ros2bag rosbag2_cpp rosbag2_storage rosbag2_tests
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10087/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16471/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11061/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16825/)

t0ny-peng:
@MichaelOrlov I'm not sure if increasing the version is the correct move here. `rosbag2_tests` has some database files that store version 5 at them, and the test failed because of version 6(from the code) and version 5(from the database file).


image

In this case, what's the correct way of updating the database? Using sqlite to modify it manually?

MichaelOrlov:
@t0ny-peng I think we need to fix tests by updating databases.  
We need to change the metadata version number each time as we are changing something in the `BagMetadata` staruct.  
https://github.com/ros2/rosbag2/blob/master/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L44

Need to look to the failing tests, may be there are some routines for generating test database.

BTW I see that tests fails not only in `rosbag2_tests` also in [rosbag2_cpp](https://ci.ros2.org/job/ci_linux/16471/testReport/rosbag2_cpp/TemporaryDirectoryFixture/)

MichaelOrlov:
@t0ny-peng **Update:** Don't need to change database files.
I was able to fix failing tests from `rosbag2_tests` altering version number in https://github.com/t0ny-peng/rosbag2/blob/master/rosbag2_tests/resources/reindex_test_bags/target_metadata/multiple_files/metadata.yaml#L2 and adding new section for the `custom_data` at the end.

However I was not able to reproduce failures in [`rosbag2_cpp`](https://ci.ros2.org/job/ci_linux/16471/testReport/rosbag2_cpp/TemporaryDirectoryFixture/) 
with message:
> C++ exception with description "Exception on parsing info file: invalid node; first invalid key: "custom_data"" thrown in the test body.

t0ny-peng:
@MichaelOrlov I think the error in `rosbag2_cpp` has already been fixed. Could you please try the latest commit(aae2fc7b7069fe30205f09b92285ffc4ddb44a38)?

image

WRT changing the version in the test, I don't think comparing the current version of `metadata` CPP struct with the version in the test yaml file is the best way. It means the YAML file has to be updated every time the CPP metadata version increases. However for now I'll just do that as a one time patch.

MichaelOrlov:
Running CI again after fix in rosbag2_tests
Gist: https://gist.githubusercontent.com/MichaelOrlov/f295a4dda3d8ce90076def0a0646c519/raw/620153ccb8176c963f45162d6b39e1903ee19bf2/ros2.repos
BUILD args: --packages-up-to ros2bag rosbag2_cpp rosbag2_storage rosbag2_tests rosbag2_transport
TEST args: --packages-select ros2bag rosbag2_cpp rosbag2_storage rosbag2_tests rosbag2_transport
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10089/
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16473/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11063/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16844/)

PR #982, opened 2022-03-29
0.14.0

clalancette:
@MichaelOrlov @emersonknapp I'd like to do a release of rosbag2 today so we can get all of the latest code into Rolling.  I know that there are some PRs in flight for this repository, but numbers are cheap and we can always do another release next week after the feature freeze.  Please let me know what you think.  Thanks!


theseankelly:
At the risk of being overly selfish I think pr #968 is on the verge of merging and the underlying bug is causing us a bunch of pain in our day to day operations -- would love to get this in the next rolling sync if at all possible

clalancette:
All right, going ahead with this.  Thanks for the feedback!

PR #988, opened 2022-04-13
Change the topic names in test_record.cpp

clalancette:
We've observed the 'test_record' series of tests sometimes
succeeding and sometimes failing on https://ci.ros2.org, particularly
in Windows builds.  When it does fail, it shows error messages
like this:

[ERROR] [1649769476.588754000] [ROSBAG2_TRANSPORT]: Topic '/chatter' has more than one type associated. Only topics with one type are supported

Given that we run tests sequentially (i.e. not in parallel), that
really shouldn't be the case.

However, we have noticed in the past that Windows can sometimes
be extremely slow to cleanup processes from previous tests.
Since many of our tests use a topic name of 'chatter', that is
likely what is happening in this case.  This change works around
this problem by using an alternate name for the topics.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_record.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record.cpp
@@ -91,7 +91,7 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
 TEST_F(RecordIntegrationTestFixture, qos_is_stored_in_metadata)
 {
   auto string_message = get_messages_strings()[1];
-  std::string topic = "/chatter";
+  std::string topic = "/qos_chatter";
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16562/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11148/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16947/)

gbiggs (approved):

clalancette:
The Rpr jobs are failing because of another flaky test (`test_play_services`).  That also needs to be investigated.  However, CI was happy here, so I'm going to go ahead and merge this one.  We'll investigate `test_play_services` separately.

PR #993, opened 2022-04-20
Make peek_next_message_from_queue return a SharedPtr.

clalancette:
It is unsafe to return a pointer to a SharedPtr here,
and then store it; if another thread comes along and
pops the queue, then we are now holding a pointer to an
object that has been freed.  Instead, dereference the
pointer so we get a SharedPtr (and a reference) immediately,
which ensures that the object won't be destroyed if we
race with another thread.

While I was in here, did a small refactoring of the method
to remove some duplicated code.

Before this PR, if I ran the `test_play_seek` test in a loop with stress on my machine, it would crash every once in a while.  After this PR, I no longer see that crash.  This should fix at least one of the flaky tests on the buildfarm.

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -193,7 +193,7 @@ private:
   std::unordered_map<std::string, std::shared_ptr<PlayerPublisher>> publishers_;
 
 private:
-  rosbag2_storage::SerializedBagMessageSharedPtr * peek_next_message_from_queue();
+  rosbag2_storage::SerializedBagMessageSharedPtr peek_next_message_from_queue();
   void load_storage_content();
   bool is_storage_completely_loaded() const;
   void enqueue_up_to_boundary(size_t boundary) RCPPUTILS_TSA_REQUIRES(reader_mutex_);
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -296,8 +296,8 @@ bool Player::set_rate(double rate)
 
 rosbag2_storage::SerializedBagMessageSharedPtr Player::peek_next_message_from_queue()
 {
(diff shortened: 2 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16576/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11160/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16964/)

gbiggs (approved):

MichaelOrlov (changes requested):
@clalancette Thank you for the code clean up.
I agree that using shared pointer instead of pointer to the shared pointer will make code less error prone for the race conditions.

Although I disagree with renaming `message_ptr` to the `message` where its corresponds to the shared pointer. It's still a pointer and comparison aka `message != null_ptr` looks very odd in this case.
Could you please revert such renaming?

clalancette:
> Although I disagree with renaming `message_ptr` to the `message` where its corresponds to the shared pointer. It's still a pointer and comparison aka `message != null_ptr` looks very odd in this case. Could you please revert such renaming?

I'll point out that the the current naming is following what was already there.  That is, in all cases where we were returning a `SharedPtr *` before, that was stored in a `message_ptr`, and then the dereferenced version (`SharedPtr`) was stored in a `message`.

That said, I'll go in and rename things to `message_ptr`, but I'm also going to rename the variable inside of `peek_next_message_from_queue` to `message_ptr_ptr`.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16586/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11168/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16972/)

ivanpauno (approved):
LGTM!

clalancette:
The CI failure on Linux is one of the known flaky tests in rosbag2, i.e. test_play_services sometimes just fails.  That's the next one for me to look into.

Issue #992, opened 2022-04-19
`rosbag2_cpp.TimeControllerClockTest.unpaused_sleep_returns_true` test fails on Windows


Blast545:
## Description
This test is failing in the repeated jobs of the buildfarm:
https://ci.ros2.org/view/nightly/job/nightly_win_rep/2588/testReport/junit/rosbag2_cpp/TimeControllerClockTest/unpaused_sleep_returns_true/

This is the code of the test:
https://github.com/ros2/rosbag2/blob/5fa3a67e79a03206e4e721a7c714caff72966801/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp#L154

From what I can see in the API,
https://github.com/ros2/rosbag2/blob/5fa3a67e79a03206e4e721a7c714caff72966801/rosbag2_cpp/include/rosbag2_cpp/clocks/time_controller_clock.hpp#L70
The function may return false indicating the time was not completed, is it there any way we can change the test to reflect this? or am I missing something?

### Expected Behavior
Test passes.

### Actual Behavior
Test fails in the repeated jobs buildfarm.

## System (please complete the following information)
 - OS: Windows
 - ROS 2 Distro: Rolling
 - Version: main

MichaelOrlov:
@Blast545 The test fails because we are using `std::condition_variable_any::wait_until`  in https://github.com/ros2/rosbag2/blob/0.15.1/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp#L177 and according to the https://en.cppreference.com/w/cpp/thread/condition_variable_any/wait_until 
"wait_until causes the current thread to block until the condition variable is notified, a specific time is reached, **or a spurious wakeup occurs**"  
Which is more likely to happen on `Windows` platform due to the OS and internal scheduler specific.

It's ok from design and our API point of view if our `sleep_until(rcutils_time_point_value_t until)` will periodically wake up and return false as long as it respect target sleep time.

Although we can take it in to account in test. Something like this:
```cpp
  clock.pause();
  clock.resume();
  auto sleep_until_timestamp = clock.now() + RCUTILS_S_TO_NS(1);
  auto start = std::chrono::steady_clock::now();
  bool sleep_result = clock.sleep_until(sleep_until_timestamp);
  // clock.sleep_until can periodically wake up and return false. Check for true in time constrain
  while (rclcpp::ok() && !sleep_result &&
[...]

PR #1012, opened 2022-05-12
Improve test_time_controller test

Blast545:
Should close https://github.com/ros2/rosbag2/issues/992

It's a simple fix for a test that was heavily relying on the the OS not waking up as explained here:
https://github.com/ros2/rosbag2/issues/992#issuecomment-1120111985

This test right before it makes the same assumption: https://github.com/ros2/rosbag2/blob/5fa3a67e79a03206e4e721a7c714caff72966801/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp#L150
But as it haven't failed I decided not to change it with this PR.

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
@@ -16,6 +16,7 @@
 
 #include <atomic>
 #include <thread>
+#include <rclcpp/utilities.hpp>
 #include "rosbag2_cpp/clocks/time_controller_clock.hpp"
 
 using namespace testing;  // NOLINT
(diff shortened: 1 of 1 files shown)


MichaelOrlov (changes requested):
@Blast545 Thank you for the PR.
CI build fails due to forgotten 
```cpp
#include <rclcpp/utilities.hpp>
```
Please add it.

Blast545:
Checks are passing, running full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16708/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11288/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17101/)

MichaelOrlov (approved):
Approve!

emersonknapp:
I know this is over a year old, but I'm adding some context here for future historians: This PR didn't actually do anything to the test - since `rclcpp` is not used in the test, the loop never ran, which means that any spurious wakeup still caused the test to fail. The original 100 nanosecond sleep request test likely never failed because it was very unlikely or maybe even impossible for a spurious wakeup to happen in that time frame. The new loop added in this test, on the other hand, slept for a full second (10 million times longer), which made a spurious wakeup far more likely (which is not handled because the loop is never entered), and therefore actually made this test far more likely to fail, which is why we've seen it so flaky.

PR #1014, opened 2022-05-13
Address flakiness in record_all_with_sim_time test

MichaelOrlov:
Trying to fix flaky `record_all_with_sim_time` test

Signed-off-by: Michael Orlov <michael.orlov@apex.ai>

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
@@ -77,7 +77,7 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
 
   size_t expected_messages = 10;
   auto ret = rosbag2_test_common::wait_until_shutdown(
-    std::chrono::seconds(10),
+    std::chrono::seconds(5),
     [&mock_writer, &expected_messages]() {
       return mock_writer.get_messages().size() >= expected_messages;
     });
--- a/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
@@ -86,6 +86,10 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
     "recorded messages = " << recorded_messages.size();
   stop_spinning();
 
+  auto messages_per_topic = mock_writer.messages_per_topic();
+  EXPECT_EQ(messages_per_topic[clock_topic], 5);
+  EXPECT_EQ(messages_per_topic[string_topic], 5);
+
   EXPECT_THAT(recorded_messages, SizeIs(Ge(expected_messages)));
 
   std::vector<rosbag2_storage::SerializedBagMessageSharedPtr> string_messages;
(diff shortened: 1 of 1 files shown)


MichaelOrlov:
Testing with this PR + repeated flags on linux:
[Build Status](https://ci.ros2.org/job/ci_linux/16706/)

MichaelOrlov:
cc: @Blast545

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
```diff
@@ -50,10 +51,10 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
 
   rosbag2_transport::RecordOptions record_options =
   {
-    true, true, {string_topic, clock_topic}, "rmw_format", 100ms
+    false, false, {string_topic, clock_topic}, "rmw_format", 100ms
```
Blast545, line 54: Why was it needed to change the options here?

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
```diff
@@ -70,13 +77,15 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
 
   size_t expected_messages = 10;
   auto ret = rosbag2_test_common::wait_until_shutdown(
-    std::chrono::seconds(2),
+    std::chrono::seconds(10),
```
Blast545, line 80: Changing from 2 to 10 seems a bit excessive, maybe 4-5 is enough to reduce the flakiness?

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
MichaelOrlov: The first `false` option telling recorder to not record all available topics rather to recorder only those which we specified in list {string_topic, clock_topic}.
The second one `false` telling recorder to run discovery tread if topics hasn't been visible during `Recorder` construction phase.  It could be in some cases and this why I put extra steps further on to make sure that `Recorder` ready to record those topics before we will start publishing. We need discovery to be on in general case. Discovery thread will stop automatically as soon as Recorder will subscribe to the all topics from options list.

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
MichaelOrlov: @Blast545 I can change to 5.

MichaelOrlov:
Running CI one more time
[Build Status](https://ci.ros2.org/job/ci_linux/16710/)

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
Blast545: Thanks for the explanation.

clalancette:
FYI, my latest commit fixes the warning from gcc.  Here's CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16712/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11293/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17105/)

MichaelOrlov:
@clalancette Thanks for fixing those gcc warning in test.

Blast545 (approved):
LGTM, warnings on linux job are related to know `test_play_services` issue.

PR #1018, opened 2022-05-18
Reduce message spam when topics to be recorded do not exist

### LINKED ISSUE #1011: Warnings about 'Topic ... has unknown type' spam when specifying topics to be recorded which do not exist
## Description

When asking rosbag to record a list of topics when some of those topics do not exist _and_ some of those topics have unknown types results in continuous spam of messages like this:

```
[WARN] [1652302035.004255795] [ROSBAG2_TRANSPORT]: Topic '/some/topic' has unknown type 'msg_pkg/msg/SomeMessage' . Only topics with known type are supported. Reason: 'package 'msg_pkg' not found, searching: [/some/workspace/path/some_package, ...]
```

These messages are printed over and over again, despite there being some logic to try and only print it once per type:

https://github.com/ros2/rosbag2/blob/a076560b09a6d80ce91815f4a03e113a0331a68d/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp#L154-L170

### Expected Behavior

This kind of error message is printed only once and does not spam the console while recording in this particular kind of scenario.

### Actual Behavior

The error message is printed repeatedly.

## To Reproduce

1. Run a publisher using a type not available to rosbag (either in another workspace or on another machine)
2. Ask `ros2 bag record` to record that topic and a topic that doesn't exist
3. Observe the error about the topic using a type unknown to rosbag is printed over and over

 - OS: Ubuntu 20.04
 - ROS 2 Distro: Rolling
 - Version: `0.14.1`

## Additional context

I believe the issue is this logic for when multiple topics are requested but not immediately found:

https://github.com/ros2/rosbag2/blob/a060c1da8fac91546c77c223cbabcaef85902b01/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L175-L194

This causes this function to be called in a loop:

https://github.com/ros2/rosbag2/blob/a060c1da8fac91546c77c223cbabcaef85902b01/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L196-L202

Which as you can see instantiates a new `TopicFilter` every time, and the `TopicFilter` holds the state that checks if the warning should be printed again:

https://github.com/ros2/rosbag2/blob/a076560b09a6d80ce91815f4a03e113a0331a68d/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp#L160-L161

So the likely solution is to find a way to extend the lifetime of the `TopicFilter` instance.

ihasdapie:
Commits extend lifetime of `TopicFilter` to avoid emitting warning messages on instantiation. `topic_filter.hpp` moved to make it accessible from `recorder.hpp`.

- closes #1011

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -36,6 +36,7 @@
 
 #include "rosbag2_transport/record_options.hpp"
 #include "rosbag2_transport/visibility_control.hpp"
+#include "rosbag2_transport/topic_filter.hpp"
 
 namespace rosbag2_cpp
 {
@@ -115,6 +116,7 @@ protected:
 
 private:
   void topics_discovery();
+  std::unique_ptr<TopicFilter> topic_filter;
 
   std::unordered_map<std::string, std::string>
   get_missing_topics(const std::unordered_map<std::string, std::string> & all_topics);
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -116,7 +116,6 @@ protected:
 
 private:
   void topics_discovery();
-  std::unique_ptr<TopicFilter> topic_filter;
 
   std::unordered_map<std::string, std::string>
   get_missing_topics(const std::unordered_map<std::string, std::string> & all_topics);
@@ -145,6 +144,7 @@ private:
 
   void warn_if_new_qos_for_subscribed_topic(const std::string & topic_name);
 
+  std::unique_ptr<TopicFilter> topic_filter_;
   std::shared_ptr<rosbag2_cpp::Writer> writer_;
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::RecordOptions record_options_;
--- a/rosbag2_transport/include/rosbag2_transport/topic_filter.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/topic_filter.hpp
@@ -42,7 +42,6 @@ public:
     RecordOptions record_options,
     rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph = nullptr,
     bool allow_unknown_types = false);
-
   virtual ~TopicFilter();
 
   /// Filter all topic_names_and_types via take_topic method, return the resulting filtered set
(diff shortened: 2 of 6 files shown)


rosbag2_transport/include/rosbag2_transport/recorder.hpp
```diff
@@ -115,6 +116,7 @@ class Recorder : public rclcpp::Node
 
 private:
   void topics_discovery();
+  std::unique_ptr<TopicFilter> topic_filter;
```
MichaelOrlov, line 119: Private member names shall end up with `_` underscore. Please rename `topic_filter` accordingly.
Also please move `topic_filter` declaration to the section where other private members declared. i.e. lines 148-164.

rosbag2_transport/include/rosbag2_transport/topic_filter.hpp
```diff
@@ -42,6 +42,7 @@ class ROSBAG2_TRANSPORT_PUBLIC TopicFilter
     RecordOptions record_options,
     rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph = nullptr,
     bool allow_unknown_types = false);
+
```
MichaelOrlov, line 45: Please remove redundant delta i.e. empty spaces

MichaelOrlov (changes requested):
@ihasdapie  Thank you for your contribution.
Overall looks good among a couple nitpicks.
Please see my comments.

jacobperron (approved):
LGTM with green CI

jacobperron (approved):
I think the test failures are unrelated [here's the same Linux failure appearing on last night's nightly](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/lastBuild/testReport/(root)/projectroot/test_play_services__rmw_fastrtps_cpp/) and the Windows failure appears unrelated to rosbag2.

@MichaelOrlov Could you take another look?

MichaelOrlov:
@jacobperron I confirm that test failures in [projectroot.test_play_services__rmw_fastrtps_cpp](https://ci.ros2.org/job/ci_linux/16740/testReport/junit/(root)/projectroot/test_play_services__rmw_fastrtps_cpp/) and  https://build.ros2.org/job/Rpr__rosbag2__ubuntu_jammy_amd64/93/testReport/junit/(root)/projectroot/test_play_services__rmw_fastrtps_cpp/ unrelated to the changes in this PR.

Please see my latest analysis for those failures here https://github.com/ros2/rosbag2/pull/1013#issuecomment-1126572618

ihasdapie:
@MichaelOrlov Can you approve this PR so we can get it merged? Thanks

MichaelOrlov (approved):

MichaelOrlov:
@ihasdapie Sure I forgot that I requested changes and haven't approved afterwards.

PR #1032, opened 2022-06-23
Fix incorrect boundary check for `playback_duration` and `play_until_timestamp`

### LINKED ISSUE #1031: Incorrect boundary check for `playback_duration` and `play_until_timestamp`
## Description
During implementation of the https://github.com/ros2/rosbag2/pull/1007 was discovered that our `play_for_none_are_played_due_to_duration` and `play_for_less_than_the_total_duration` successfully passes only because we have improper initialization of the bag starting time in `MockSequantialReader`. 
We initialize it to 0 instead of timestamp from first message.

It turns out that we are incorrectly handling cases when `play_options_.playback_duration = 0` i.e. still playing message.
And also incorrectly handling the case when `play_until_timestamp` less than bag starting time and more than -1.

### Expected Behavior
1. We shall not play any messages if `play_options_.playback_duration = 0` and `play_until_timestamp` less than bag starting time.
2. We shall not play any messages if `play_options_.playback_until_timestamp` less then bag starting time and still valid i.e. **> -1** and `play_options_.playback_duration` has default `-1` value.

### Actual Behavior
1. if `play_options_.playback_duration = 0` and `play_until_timestamp` less than bag starting time we actually playing back at least one message.
2. if `play_options_.playback_until_timestamp` less then bag starting time and still valid i.e. **> -1** and `play_options_.playback_duration` has default `-1` value. We are playing all messages from bag.

## To Reproduce
Steps to reproduce the behavior, e.g.
1. Properly initialize `metadata_.starting_time` in `MockSequantialReader` and see failure in `play_for` unit tests
2. Or run `ros2 bag play` with `play_options_.playback_duration` and `play_options_.playback_until_timestamp` with values as described in Expected Behavior.

 - OS: Ubuntu 20.04
 - ROS 2 Distro: Rolling
 - Version: master

## Additional context
N/A

MichaelOrlov:
- Closes #1031 

Update: Also fixed one false positive and one flaky test in `test_play_until`

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -253,22 +253,21 @@ private:
   void prepare_publishers();
   bool publish_message(rosbag2_storage::SerializedBagMessageSharedPtr message);
   static callback_handle_t get_new_on_play_msg_callback_handle();
+  void add_key_callback(
+    KeyboardHandler::KeyCode key,
+    const std::function<void()> & cb,
+    const std::string & op_name);
+  void add_keyboard_callbacks();
+  void create_control_services();
+  void configure_play_until_timestamp();
+  bool shall_stop_at_timestamp(const rcutils_time_point_value_t & msg_timestamp) const;
+
   static constexpr double read_ahead_lower_bound_percentage_ = 0.9;
   static const std::chrono::milliseconds queue_read_wait_period_;
   std::atomic_bool cancel_wait_for_next_message_{false};
 
   std::mutex reader_mutex_;
   std::unique_ptr<rosbag2_cpp::Reader> reader_ RCPPUTILS_TSA_GUARDED_BY(reader_mutex_);
-
-  void add_key_callback(
-    KeyboardHandler::KeyCode key,
-    const std::function<void()> & cb,
-    const std::string & op_name);
-  void add_keyboard_callbacks();
-
-  void create_control_services();
-
-  void configure_play_until_timestamp();
 
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::PlayOptions play_options_;
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -840,7 +840,12 @@ void Player::configure_play_until_timestamp()
 inline bool Player::shall_stop_at_timestamp(const rcutils_time_point_value_t & msg_timestamp) const
 {
   if ((play_until_timestamp_ > -1 && msg_timestamp > play_until_timestamp_) ||
-    play_until_timestamp_ == 0) {return true;} else {return false;}
+    play_until_timestamp_ == 0)
+  {
+    return true;
+  } else {
+    return false;
+  }
 }
 
 }  // namespace rosbag2_transport
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -363,16 +363,12 @@ bool Player::play_next()
   rosbag2_storage::SerializedBagMessageSharedPtr message_ptr = peek_next_message_from_queue();
 
   bool next_message_published = false;
-  while (message_ptr != nullptr && !next_message_published) {
-    if (play_until_timestamp_ >= starting_time_ &&
-      message_ptr->time_stamp > play_until_timestamp_)
-    {
-      break;
-    }
-    {
-      next_message_published = publish_message(message_ptr);
-      clock_->jump(message_ptr->time_stamp);
-    }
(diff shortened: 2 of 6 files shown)


MichaelOrlov:
cc: @agalbachicar

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -253,22 +253,21 @@ class Player : public rclcpp::Node
...
   void add_key_callback(
     KeyboardHandler::KeyCode key,
     const std::function<void()> & cb,
     const std::string & op_name);
   void add_keyboard_callbacks();
-
   void create_control_services();
-
   void configure_play_until_timestamp();
+  bool shall_stop_at_timestamp(const rcutils_time_point_value_t & msg_timestamp) const;
```
agalbachicar, line 263: Does it make sense to change this method to be: `shall_continue_at_timestamp` so as to avoid negating in every usage occurrence?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -363,16 +363,12 @@ bool Player::play_next()
...
-      message_ptr->time_stamp > play_until_timestamp_)
-    {
-      break;
-    }
-    {
-      next_message_published = publish_message(message_ptr);
-      clock_->jump(message_ptr->time_stamp);
-    }
+  while (rclcpp::ok() && !next_message_published &&
+    message_ptr != nullptr && !shall_stop_at_timestamp(message_ptr->time_stamp))
```
agalbachicar, line 367: Not sure if you considered unifying the condition in a function. This condition is pretty similar to the one in play_messages_from_queue(). Perhaps it makes sense to encapsulate it in a function because you can benefit from passing message_ptr and evaluating the timestamp conditions as well.

agalbachicar (commented):
@MichaelOrlov thanks for sending this bug and fixing the bug we introduced. I left two suggestions for your consideration which try to improve readability only.

Thanks again!

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -833,12 +826,21 @@ void Player::configure_play_until_timestamp()
...
+      std::max(play_until_from_duration, play_options_.playback_until_timestamp);
   } else {
     play_until_timestamp_ = -1;
   }
 }
 
+inline bool Player::shall_stop_at_timestamp(const rcutils_time_point_value_t & msg_timestamp) const
+{
+  if ((play_until_timestamp_ > -1 && msg_timestamp > play_until_timestamp_) ||
+    play_until_timestamp_ == 0) {return true;} else {return false;}
```
jhdcs, line 843: I _think_ the [ROS 2 Style Guide](https://docs.ros.org/en/humble/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html#open-versus-cuddled-braces) recommends the following:

```suggestion
  if (
    (play_until_timestamp_ > -1 && msg_timestamp > play_until_timestamp_) || play_until_timestamp_ == 0)
  {
    return true;
  } 
  else
  {
    return false;
  }
```

jhdcs (changes requested):
Just a style guide comment that I'm not 100% sure about.

rosbag2_transport/include/rosbag2_transport/player.hpp
MichaelOrlov: @agalbachicar Nop. I don't think so.  
I deliberately named it `shall_stop_at_timestamp` because it exactly reflects logic which is supposed to check for `play_until` and `play_for_duration` features.  
Don't worry about negation in each while loop it will not affect performance.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: @jhdcs Fixed. See new commit https://github.com/ros2/rosbag2/pull/1032/commits/cb8691ca1127de2f2589356d8d2a751d63335f3b

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/e28428938082cd90d5c8013362a8053a/raw/9900e5b8703accdc299db1280f09f1cbca688e74/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_transport rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10450
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16889/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11459/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17292/)

jhdcs (approved):
I'm not seeing any huge issues here.

PR #1033, opened 2022-06-24
Make test play services more reliable

ivanpauno:
Maybe fixes https://github.com/ros2/rosbag2/issues/862.

This is a workaround, it doesn't fix the underlying issue.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp
@@ -103,15 +103,27 @@ public:
     const char * function_name,
     int line_number)
   {
-    auto future = cli->async_send_request(request);
-    EXPECT_TRUE(future.valid());
-    EXPECT_EQ(future.wait_for(service_call_timeout_), std::future_status::ready) <<
-      function_name << ", line : " << line_number;
-    // std::cout << function_name << ", line : " << line_number << std::endl;
-    auto result = std::make_shared<typename Srv::Response>();
-    EXPECT_NO_THROW({result = future.get();});
-    EXPECT_TRUE(result);
-    return result;
+    // retry 5 times
+    // NOTE(ivanpauno): There is a discovery issue in many rmw implementatations
+    // where the first few requests can be lost because of the server not having discovered
+    // the client yet. wait_for_service() thinks the server is ready, as the client has discovered
+    // the server.
+    // By retrying a few times, we make the tests more reliable.
+    // We could also just `std::this_thread::sleep()` for some seconds after creating the clients
+    // and before making requests, which should have a similar effect.
+    for (size_t i = 0; i < 5; ++i) {
+      auto future = cli->async_send_request(request);
+      EXPECT_TRUE(future.valid());
+      if (future.wait_for(service_call_timeout_) == std::future_status::ready) {
+        auto result = std::make_shared<typename Srv::Response>();
+        EXPECT_NO_THROW({result = future.get();});
+        EXPECT_TRUE(result);
+        return result;
+      }
+      cli->remove_pending_request(future);
+    }
+    ADD_FAILURE() << function_name << ", line : " << line_number;
+    return std::make_shared<typename Srv::Response>();
   }
 
   template<typename Srv>


fujitatomoya (approved):
lgtm

MichaelOrlov (changes requested):
@ivanpauno I am sorry but this is not what I would expected to see in our tests.
Our tests exists to make sure that there are no issues rather than try to hide any issues.

Sending request multiple times in loop until we will get response is not acceptable, at least in this case.
Because we have expectations in tests and workflow in particularly in `play_next` the number or requests and corresponding responses shall match to each other.

I think that issue with discovery which you are trying to mitigate is not a root cause of the issue which we have with  failing `test_play_services` tests. Because according to the logs the failure never happened on first service request from the tests it rather happened on 3-4 request and more often on 100s and more with test_play_next scenario.

ivanpauno:
> Because according to the logs the failure never happened on first service request from the tests

are you sure?
`SetUp()` is called before each test fixture (i.e. each `TEST_F`), so there're a lot of calls that happen right after the services are creaated.

> it rather happened on 3-4 request and more often on 100s and more with test_play_next scenario

In all the CI links in https://github.com/ros2/rosbag2/issues/862, none of them seem to be failing there.

> @ivanpauno I am sorry but this is not what I would expected to see in our tests.

If it's a service discovery issue, the alternatives are doing or just waiting for some seconds after creating the services.
Fixing the real issue is a lot of work.

If it's not a discovery issue and it's another thing, I have no idea what's going on.
I would have to debug more.

MichaelOrlov:
@ivanpauno 
> are you sure?
SetUp() is called before each test fixture (i.e. each TEST_F), so there're a lot of calls that happen right after the services are creaated.

I am 100% sure that this is not a discovery issue. And I can prove it.
For instance form failure of this job https://build.ros2.org/job/Rpr__rosbag2__ubuntu_jammy_amd64/122/consoleText 
```
[INFO] [1655556476.597988951] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.598254146] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.598445490] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.598640724] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.598978766] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.599309567] [rosbag2_player_for_test_srvs]: Playing next message.
8: [INFO] [1655556476.599803300] [rosbag2_player_for_test_srvs]: Playing next message.
8: /tmp/ws/src/rosbag2/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp:108: Failure
8: Expected equality of these values:
8:   future.wait_for(service_call_timeout_)
8:     Which is: 4-byte object <01-00 00-00>
8:   std::future_status::ready
8:     Which is: 4-byte object <00-00 00-00>
8: virtual void PlaySrvsTest_play_next_Test::TestBody(), line : 378
 8/31 Test  #8: test_play_services__rmw_fastrtps_cpp ...........................***Timeout  60.06 sec
```
Test `test_play_services__rmw_fastrtps_cpp` failed and call for 
https://github.com/ros2/rosbag2/blob/164475bbfdd1d3315e8b02e9485d768c5f7bc7c8/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp#L106-L108
was called from **line : 378** 
https://github.com/ros2/rosbag2/blob/164475bbfdd1d3315e8b02e9485d768c5f7bc7c8/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp#L373-L381

From log it is very clear that we successfully called `play_next` service **186** times in loop and on **187** it got failed
and such a failure in the middle of the `play_next` test happening in **90%** cases on CI the rest **10%** mostly happened in 
`TEST_F(PlaySrvsTest, is_paused)`
https://github.com/ros2/rosbag2/blob/164475bbfdd1d3315e8b02e9485d768c5f7bc7c8/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp#L314-L332
and failure never happened on **FIRST** `ASSERT_TRUE(service_call_is_paused());` but rather on 3d or 4th call in the scope of the one unit test lines **326** and **330** respectively.

I also was able to reproduce this issue multiple times on my local machine and it never happens on first call for `auto future = cli->async_send_request(request);` `EXPECT_EQ(future.wait_for(service_call_timeout_), std::future_status::ready)`

MichaelOrlov:
Here is a fresh example from today's CI failure https://github.com/ros2/rosbag2/runs/7197760772?check_suite_focus=true
```
[ RUN      ] PlaySrvsTest.set_rate_good_values
2022-07-05T14:48:47.6799425Z 32: stdin is not a terminal device. Keyboard handling disabled.[INFO] [1657032380.610015078] [rosbag2_player_for_test_srvs]: Set rate to 1
2022-07-05T14:48:47.6799580Z 32: [INFO] [1657032380.614651567] [rosbag2_player_for_test_srvs]: Adding keyboard callbacks.
2022-07-05T14:48:47.6799737Z 32: [INFO] [1657032380.614693967] [rosbag2_player_for_test_srvs]: Press SPACE for Pause/Resume
2022-07-05T14:48:47.6799908Z 32: [INFO] [1657032380.614709067] [rosbag2_player_for_test_srvs]: Press CURSOR_RIGHT for Play Next Message
2022-07-05T14:48:47.6800075Z 32: [INFO] [1657032380.614808867] [rosbag2_player_for_test_srvs]: Press CURSOR_UP for Increase Rate 10%
2022-07-05T14:48:47.6800240Z 32: [INFO] [1657032380.614864067] [rosbag2_player_for_test_srvs]: Press CURSOR_DOWN for Decrease Rate 10%
2022-07-05T14:48:47.6800372Z 32: [INFO] [1657032380.614940767] [rosbag2_player_for_test_srvs]: Pausing play.
2022-07-05T14:48:47.6800500Z 32: [INFO] [1657032380.618946657] [rosbag2_player_for_test_srvs]: Resuming play.
2022-07-05T14:48:47.6800703Z 32: [INFO] [1657032380.619361556] [rosbag2_player_for_test_srvs]: Playback until timestamp: -1
2022-07-05T14:48:47.6800828Z 32: [INFO] [1657032380.619548056] [rosbag2_player_for_test_srvs]: Set rate to 2
2022-07-05T14:48:47.6800963Z 32: [INFO] [1657032380.620571854] [rosbag2_player_for_test_srvs]: Set rate to 0.5
2022-07-05T14:48:47.6801240Z 32: /__w/rosbag2/rosbag2/ros_ws/src/jduflpqya0l/rosbag2/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp:120: Failure
2022-07-05T14:48:47.6801341Z 32: Expected equality of these values:
2022-07-05T14:48:47.6801452Z 32:   future.wait_for(service_call_timeout_)
2022-07-05T14:48:47.6801620Z 32:     Which is: 4-byte object <01-00 00-00>
2022-07-05T14:48:47.6814957Z 32:   std::future_status::ready
2022-07-05T14:48:47.6815222Z 32:     Which is: 4-byte object <00-00 00-00>
2022-07-05T14:48:47.6815416Z 32: virtual void PlaySrvsTest_set_rate_good_values_Test::TestBody(), line : 366
2022-07-05T14:48:47.6815603Z 32/78 Test #32: test_play_services__rmw_fastrtps_cpp ...................................***Timeout  60.05 sec
```
According to the log, error happened at line **366**
https://github.com/ros2/rosbag2/blob/e1f4349f33aaef1bf0fb13625cf23d80f9af2de0/rosbag2_transport/test/rosbag2_transport/test_play_services.cpp#L350-L368
And this is second time calling `get_response = service_call_get_rate();` in this test. 
The first one was successfully called from line **360**.

ivanpauno:
This doesn't seem to be the problem, so closing

PR #1036, opened 2022-06-28
Add additional mode of publishing sim time updates

kylemarcey:
Add an additional mode for publishing sim time updates when replaying. 

Instead of publishing /clock updates at a fixed rate, this option triggers an update every time a replayed message is published. Optionally, this can be configured to trigger only for a subset of replayed topics.

This mode is most useful when the application consuming the replayed data is doing some sanity checks or otherwise correlating received messages to the clock and where the application does not need a clock that updates at a consistent rate. In these cases, using this mode can lower the overhead of having sim time enabled.

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -82,10 +82,20 @@ class PlayVerb(VerbExtension):
                  '  pragmas: [\"<setting_name>\" = <setting_value>]'
                  'Note that applicable settings are limited to read-only for ros2 bag play.'
                  'For a list of sqlite3 settings, refer to sqlite3 documentation')
-        parser.add_argument(
+        clock_args_group = parser.add_mutually_exclusive_group()
+        clock_args_group.add_argument(
             '--clock', type=positive_float, nargs='?', const=40, default=0,
             help='Publish to /clock at a specific frequency in Hz, to act as a ROS Time Source. '
                  'Value must be positive. Defaults to not publishing.')
+        clock_args_group.add_argument(
+            '--clock-topics', type=str, default=[], nargs='+',
+            help='List of topics separated by spaces that will trigger a /clock update '
+                 'when a message is published on them'
+        )
+        clock_args_group.add_argument(
+            '--clock-topics-all', default=False, action='store_true',
+            help='Publishes an update on /clock immediately before each replayed message'
+        )
         parser.add_argument(
             '-d', '--delay', type=positive_float, default=0.0,
             help='Sleep duration before play (each loop), in seconds. Negative durations invalid.')
@@ -185,6 +195,9 @@ class PlayVerb(VerbExtension):
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
+        if args.clock_topics_all or len(args.clock_topics) > 0:
+            play_options.clock_publish_on_topic_publish = True
+        play_options.clock_topics = args.clock_topics
         play_options.delay = args.delay
         play_options.playback_duration = args.playback_duration
         play_options.playback_until_timestamp = self.get_playback_until_from_arg_group(
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -53,6 +53,13 @@ public:
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
 
+  // Enable publishing to /clock when a replayed topic is published
+  bool clock_publish_on_topic_publish{false};
+
+  // If clock_publish_on_topic_publish is true, list of topics that will trigger
+  // a /clock update to be published. If list is empty, all topics will act as a trigger
+  std::vector<std::string> clock_trigger_topics = {};
+
   // Sleep before play. Negative durations invalid. Will delay at the beginning of each loop.
   rclcpp::Duration delay = rclcpp::Duration(0, 0);
 
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -600,18 +600,36 @@ void Player::prepare_publishers()
   reader_->set_filter(storage_filter);
 
   // Create /clock publisher
-  if (play_options_.clock_publish_frequency > 0.f) {
-    const auto publish_period = std::chrono::nanoseconds(
-      static_cast<uint64_t>(RCUTILS_S_TO_NS(1) / play_options_.clock_publish_frequency));
+  if (play_options_.clock_publish_frequency > 0.f || play_options_.clock_publish_on_topic_publish) {
     // NOTE: PlayerClock does not own this publisher because rosbag2_cpp
     // should not own transport-based functionality
     clock_publisher_ = create_publisher<rosgraph_msgs::msg::Clock>(
       "/clock", rclcpp::ClockQoS());
+  }
+
+  if (play_options_.clock_publish_frequency > 0.f) {
+    const auto publish_period = std::chrono::nanoseconds(
+      static_cast<uint64_t>(RCUTILS_S_TO_NS(1) / play_options_.clock_publish_frequency));
+
     clock_publish_timer_ = create_wall_timer(
       publish_period, [this]() {
-        auto msg = rosgraph_msgs::msg::Clock();
-        msg.clock = rclcpp::Time(clock_->now());
-        clock_publisher_->publish(msg);
+        publish_clock_update();
+      });
+  }
+
+  if (play_options_.clock_publish_on_topic_publish) {
+    add_on_play_message_pre_callback(
+      [this](const std::shared_ptr<rosbag2_storage::SerializedBagMessage> msg) {
+        if (play_options_.clock_trigger_topics.empty()) {
+          publish_clock_update();
+        } else {
+          for (const auto & topic : play_options_.clock_trigger_topics) {
+            if (topic == msg->topic_name) {
+              publish_clock_update();
+              break;
+            }
+          }
+        }
       });
   }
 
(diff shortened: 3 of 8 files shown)


ros2bag/ros2bag/verb/play.py
```diff
@@ -78,10 +78,20 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
             help='Publish to /clock at a specific frequency in Hz, to act as a ROS Time Source. '
                  'Value must be positive. Defaults to not publishing.')
+        clock_args_group.add_argument(
+            '--clock-topics', type=str, default=[], nargs='+',
+            help='List of topics separated by spaces that will trigger a /clock update '
+                 'when a message is published on them'
+        )
+        clock_args_group.add_argument(
+            '--clock-topics-all', default=False, action='store_true',
+            help='Publishes an update on /clock immediately before each replayed message'
```
MichaelOrlov, line 93: @kylemarcey I think `--clock-topics-all` should be mutually exclusive with '--clock-topics' and '--clock' command line options.

MichaelOrlov: Sorry I missed that it is already mutually exclusive

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -562,18 +562,35 @@ void Player::prepare_publishers()
   reader_->set_filter(storage_filter);
 
   // Create /clock publisher
-  if (play_options_.clock_publish_frequency > 0.f) {
-    const auto publish_period = std::chrono::nanoseconds(
-      static_cast<uint64_t>(RCUTILS_S_TO_NS(1) / play_options_.clock_publish_frequency));
+  if (play_options_.clock_publish_frequency > 0.f || play_options_.clock_publish_on_topic_publish) {
```
MichaelOrlov, line 565: @kylemarcey `clock_publish_on_topic_publish` is mutual exclusive with `clock_trigger_topics` and `clock_publish_on_topic_publish` is false by default at least from Python binding API.
When `clock_trigger_topics` will be not empty it seems we will not create `clcok_publisher_` isn't it?

MichaelOrlov (changes requested):
@kylemarcey  Thank for your contribution.

I see that `clock_publish_on_topic_publish` declared as mutually exclusive with `clock_trigger_topics` at least form Python binding API.  
However in Player publishing on `clock_trigger_topics` only happened if `clock_publish_on_topic_publish` is set to true.

Can you please double check this logic?

rosbag2_transport/src/rosbag2_transport/player.cpp
kylemarcey: @MichaelOrlov yes, good point. I now set `clock_publish_on_topic_publish` to true if either `--clock-topics-all` or `--clock-topics ...` are set so the publisher will be created in all cases

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -48,6 +48,14 @@ struct PlayOptions
   // 0 (or negative) means that no publisher will be created
   double clock_publish_frequency = 0.0;
 
+  // Enable publishing to /clock when a replayed topic is published
+  bool clock_publish_on_topic_publish{false};
+
+  // Topic names used to trigger a /clock publish.
+  // Only messages matching these specified topics will be trigger the /clock publish.
+  // If list is empty, all the topics are used to publish to /clock
+  std::vector<std::string> clock_trigger_topics = {};
```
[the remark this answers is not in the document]
kylemarcey: @MichaelOrlov Done, `clock_publish_on_topic_publish` will always act as the flag to enable this behavior, `clock_trigger_topics` just acts as a filter if it is enabled.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -562,18 +562,35 @@ void Player::prepare_publishers()
...
+
+  if (play_options_.clock_publish_on_topic_publish) {
+    add_on_play_message_pre_callback(
+      [this](const std::shared_ptr<rosbag2_storage::SerializedBagMessage> msg) {
+        if (play_options_.clock_trigger_topics.empty()) {
+          publish_clock_update();
+        } else {
+          for (const auto & topic : play_options_.clock_trigger_topics) {
+            if (topic == msg->topic_name) {
+              publish_clock_update();
```
MichaelOrlov, line 590: @kylemarcey It doesn't make sense to iterate further if we have already found matching topic in list.
Can you please add `"break;"` from loop after `publish_clock_update();` ?  
This way we also will guarantee to publish only once even if topic name listed twice in the  `clock_trigger_topics`

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -125,3 +127,9 @@ TEST_F(ClockPublishFixture, clock_respects_playback_rate)
   messages_to_play_ = 5;
   run_test();
 }
+
+TEST_F(ClockPublishFixture, clock_is_published_from_topic_trigger)
+{
+  play_options_.clock_publish_on_topic_publish = true;
+  run_test();
```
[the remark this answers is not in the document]
MichaelOrlov: Sorry my bad. I forgot that `play_options_.clock_publish_frequency`  will be initialized to 0 by default in test fixture in each test run.  
However in this case it will make sense to setup `messages_to_play_ = expected_clock_messages_;` and add a strict check inside `run_test()` for this scenario i.e. when `play_options_.clock_publish_on_topic_publish = true;`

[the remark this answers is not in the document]
MichaelOrlov: @kylemarcey I actually noticed that the tests from `ClockPublishFixture` test suite is tend to be flaky since `wait_for_matched` is not using there.  Also we don't need to run player in single threaded executor.   
Wouldn't you mind to rewrite `void run_test()` ?
You can refer to the right approach in   
https://github.com/ros2/rosbag2/blob/67bd7767b502eff7bac4162ff1d93e86405ff538/rosbag2_transport/test/rosbag2_transport/test_burst.cpp#L72-L87

In our case don't need to call `player->pause()` before starting playback and don't need to call `player->wait_for_playback_to_start();`

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -64,40 +59,42 @@ class ClockPublishFixture : public RosBag2PlayTestFixture
     prepared_mock_reader->prepare(messages, topic_types);
     auto reader = std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader));
 
+    auto player = std::make_shared<MockPlayer>(std::move(reader), storage_options_, play_options_);
+
+    sub_->add_subscription<test_msgs::msg::BasicTypes>("/topic1", messages.size());
```
MichaelOrlov, line 64: It turns out that for tests with timer we still need to spin player node
```suggestion
    auto player = std::make_shared<MockPlayer>(std::move(reader), storage_options_, play_options_);

    rclcpp::executors::SingleThreadedExecutor exec;
    exec.add_node(player);
    auto spin_thread = std::thread(
      [&exec]() {
        exec.spin();
      });

    sub_->add_subscription<test_msgs::msg::BasicTypes>("/topic1", messages.size());
```

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -64,40 +59,42 @@ class ClockPublishFixture : public RosBag2PlayTestFixture
...
-    auto spin_thread = std::thread(
-      [&exec]() {
-        exec.spin();
-      });
-    player->play();
+    auto player_future = std::async(std::launch::async, [&player]() -> void {player->play();});
+    player->wait_for_playback_to_start();
 
+    player_future.get();
     await_received_messages.get();
```
MichaelOrlov, line 77:
```suggestion
    await_received_messages.get();

    exec.cancel();
    spin_thread.join();
```

MichaelOrlov (changes requested):
@kylemarcey Thanks for improving tests.
It turns out that for tests with timer we still need to spin player node. 
Please see my suggestions for adding spin node for player in `run_test`.

I've tested it locally on my machine and tests passing.

Please also rebase your branch on latest master. It seems merge conflicts preventing to be able to merge this branch.

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -64,40 +59,42 @@ class ClockPublishFixture : public RosBag2PlayTestFixture
...
-        reader), storage_options_, play_options_);
-    rclcpp::executors::SingleThreadedExecutor exec;
-    exec.add_node(player);
-    auto spin_thread = std::thread(
-      [&exec]() {
-        exec.spin();
-      });
-    player->play();
+    auto player_future = std::async(std::launch::async, [&player]() -> void {player->play();});
+    player->wait_for_playback_to_start();
```
MichaelOrlov, line 74: In this case we don't need to wait for playback to start. 
```suggestion
```

kylemarcey:
@MichaelOrlov I made the suggested changes, and the tests appear to be passing. Thanks for the review!

MichaelOrlov (approved):

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/d433c4105867b2e03a160cc8517d2b1e/raw/9e899d26edeae859a3b98c3af20e507dae544afc/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10548
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16999/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11560/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17408/)

MichaelOrlov:
@kylemarcey Thank you for your contribution! 
CI passed green and I have merged this PR.

PR #1051, opened 2022-07-28
feat: add repeated transient local option

wep21:
Implement a feature equivalent to https://github.com/ros/ros_comm/pull/1850

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -175,6 +175,10 @@ class RecordVerb(VerbExtension):
             '--use-sim-time', action='store_true', default=False,
             help='Use simulation time.'
         )
+        parser.add_argument(
+            '--repeated-transient-local', action='store_true', default=False,
+            help='Repeat transient local messages at the start of each new bag file.'
+        )
         self._subparser = parser
 
     def main(self, *, args):  # noqa: D102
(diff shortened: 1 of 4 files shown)


MichaelOrlov (changes requested):
@wep21 I am sorry I can't approve this feature.
This feature doesn't align with `rosbag2` and entire DDS communication concept by design.

First of all in `rosbag2` we don't have ability to start playback from some chunk or `splitted` bag. Instead we can start from some timepoint and have jump/seek API. We also have ability to pause playback and burst some messages without respecting delay between them i.e. "fast forward" or `play_next` API.  
User can start playback, wait till `key frames` e.g. `map` or `tf` will be published then go in pause mode and quickly playback some amount of messages with `play_next` API or use `seek(time)` API to skip some amount of messages to be able to continue playback from some timepoint. There are no way to start playback or jump in playback to the next split file.

Therefore in `rosbag2` we don't have such problem with missed messages publishing only once. It's just because `rosbag2` designed differently.

Another aspect is DDS communication. With `transient local` qos settings could be and in many cases really are topics which are publishing on regular bases. Approach when saving all messages from such topics in hash map and then writing them to each new bag doesn't look like a good idea.

One more aspect is that in `rosbag2` we are trying to adhere to determinism in recording and playback. Playback should exactly repeat flow from recording as close as possible.  
It means there are **must not** be any extra magically added messages during playback or recording which was not really published when recording was happened.

MichaelOrlov:
- Closing as will not be merged due to the conflict in design and concept of Rosbag2.

PR #1054, opened 2022-08-01
[humble] Backport. Only pass one regular expression to filter topics

esteve:
This is a backport of #1053

PR #1046, opened 2022-07-19
Added support for excluding topics via regular expressions

esteve:
This PR adds `--exclude` (and `-x`) as an option to `ros2 bag play` to exclude topics from being replayed.

PR #1059, opened 2022-08-03
[humble] Backport. Added support for excluding topics via regular expressions

esteve:
Backport of #1046 

Depends on #1054


esteve:
Closing as this change would break ABI compatibility.

PR #1072, opened 2022-08-19
Readers/info can accept a single bag storage file, and detect its storage id automatically

emersonknapp:
Closes #972

API-level allowance for bare bag files as input. Main improvement is the storage factory loop over all plugins, when storage id is not provided.

To expose this new capability consistently, factor out common options into a single place for all `ros2 bag` verbs that read bags.

Issue #972, opened 2022-03-11
Readers detect storage plugin automatically


defunctzombie:
## Description

When using the `ros2 bag` CLI for info, playback, and recording it would be nice to detect the storage plugin based on the file extension.

For example if I record an [MCAP](https://github.com/foxglove/mcap) file using the [mcap storage plugin](https://github.com/ros-tooling/rosbag2_storage_mcap) I record to .mcap files. When I want to go and play those files, I'd like to use

```
ros2 bag play /path/to/file.mcap
```

Rather than having to specify the plugin manually via:

```
ros2 bag play -s mcap /path/to/file.mcap
```

Similarly, f I specify I want to record to `.mcap` files, the storage plugin should be detected and used for recording.

If there are multiple plugins installed which could support the extension that should error and require the user to manually specify the name of their desired storage plugin via the command line flag.

## Requirements

Definition: "bare bag" is a standalone file, rather than a directory containing a `metadata.yaml` file and one or more storage files.

* [ ] Readers: automatically choose the storage implementation when a bare bag is specified as input (`burst`/`convert`/`info`/`play`/`reindex` )

clalancette:
I think the idea to automatically load the plugin based on the input file is a good one.  Another way to determine the right plugin to load would be to look at the beginning of the input file, looking for particular markers.  This seems more reliable than the extension name, and is similar to how `file` works.

emersonknapp:
Agreed - though I think it doesn't hurt to use extension as a hint. Could also do a "try all storage plugins" loop and see if one is successful at opening

alluring-mushroom:
As the number of storage options grows and/or the number of users of alternate storage options grows, this becomes very valuable. This would be a worthwhile enhancement

emersonknapp:
Added concrete and comprehensive requirements for this ticket at the bottom of the original description. I think this should address all cases. If those points are completed, it opens up the non-directory ("bare") bag functionality in a way that users have been asking for basically forever.

@MichaelOrlov as current maintainer, let me know if you have any input on that design choice

MichaelOrlov:
@emersonknapp Thanks for pinging me in this issue.
The design proposals looks reasonable for me.

May be need to add consideration how to handle cases when:
1. Explicitly specified bag folder **without** file extension. For instance `ros2 bag record --output my_data /chatter` or `ros2 bag info my_data`. In former case could be multiple files inside folder due to the bag split feature.
[...]

PR #1073, opened 2022-08-19
Reader CLI argument consistency pass (coherently enables bare-file)

emersonknapp:
Depends on #1072
Part of #972

For all `ros2 bag` verbs that create a reader, factor out common options into a single place for consistency. This helps consistently expose the bare file workflow enabled in #1072


jhdcs:
The code changes look good, though test 3 appears to be failing (it can't find the metadata file, if I'm reading it correctly).

Do you think this error is unrelated to your changes? Is it another flaky test?

emersonknapp:
Need to just roll this into #1072 for end to end to work properly

PR #1081, opened 2022-09-08
Fix hangout in rosbag2 player and recorder when pressing `CTRL+C`

### LINKED ISSUE #1079: ros2 bag record and play does occasionally not properly closing when pressing Ctrl+C
## Description
When closing a recording, occasionally, the ros2 bag is apparently not closed correctly, which makes a reindex of the bag necessary. 
This happens for both the sqlite3 and the mcap plugins. Usually, this is no big issue and takes only a few seconds, but reindexing these mcap files seems to be not possible for now (which is another issue).
For our setup (recording point clouds, images, and GNSS data), this issue occurs in most cases.

### Expected Behavior
The recording should stop and write the remaining messages from the cache to the bag. A metadata.yaml file should be created.
```
...
[INFO] [1662381739.854492546] [rosbag2_recorder]: Subscribed to topic '/sensing/gnss/fixed'
[INFO] [1662381739.856257118] [rosbag2_recorder]: Subscribed to topic '/sensing/gnss/pose'
[INFO] [1662381746.062154306] [rclcpp]: signal_handler(signum=2)
[INFO] [1662381746.159286989] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
```

### Actual Behavior
ros2 bag record is stuck at `signal_handler(signum=2)`. The remaining messages are not written to the bag and killing the recording process or closing the terminal is required. No metadata.yaml file is created.
```
...
[INFO] [1662381748.073734704] [rosbag2_recorder]: Subscribed to topic '/sensing/gnss/fixed'
[INFO] [1662381748.076060419] [rosbag2_recorder]: Subscribed to topic '/sensing/gnss/pose'
[INFO] [1662381754.862466740] [rclcpp]: signal_handler(signum=2)
```

## To Reproduce
1. Start a recording with `ros2 bag record -a`. Actually, this issue also occurs when no other node is running and only the recording is started.
2. Stop recording with Ctrl+C.
3. It might need a few tries, but for me, in roughly half of the cases the recording is not stopped correctly, and killing the process is required.

 - OS: Ubuntu Jammy
 - ROS 2 Distro: Humble
 - Version: 0.15.2-1jammy.20220729 (apt release)

## RCA (Root Cause Analysis)
1. Rosbag2 player and recorder not finishing correctly because it's waiting for `Enter` key to be pressed after `CTRL+C`. If press `Enter` key after `CTRL+C` rosbag2 finishes correctly.
2. Rosbag2 expecting `Enter` key because we catching `SIGINT` signal in keyboard handler and [returning input from unbuffered to the canonical mode](https://github.com/ros-tooling/keyboard_handler/blob/70c0215c6e67dc434daadeba116e886ffd9d36c2/keyboard_handler/src/keyboard_handler_unix_impl.cpp#L43)
3.  which is imply to block next read from std input until `Enter` key will be pressed. Which normal and default behavior when we returning to the console. 
4. However we are not returning to the console right after SIGINT but rather trying to dismiss it on upper level and call `rclcpp::shutdown()` for instance  [here](https://github.com/ros2/rosbag2/blob/8a77246aa8dae36b6cf5560a7902bc49ffe18a3d/rosbag2_py/src/rosbag2_py/_transport.cpp#L200-L203) to invalidate context and force finishing current application.
5. We basically have infinite loop with check https://github.com/ros2/rosbag2/blob/8a77246aa8dae36b6cf5560a7902bc49ffe18a3d/rosbag2_py/src/rosbag2_py/_transport.cpp#L228
```
while (rclcpp::ok() {
```
inside  `executor_>spin();` call
6. We are exiting from infinite loop and going in to rosbag2 destructor when we triggering `rclcpp::shutdown()`
7. Hangout with wait for `Enter` key happening when we are trying to destruct `keyboard_handler_` instance and waiting to join main keyboard handler thread in it's own [destructor](https://github.com/ros-tooling/keyboard_handler/blob/70c0215c6e67dc434daadeba116e886ffd9d36c2/keyboard_handler/src/keyboard_handler_unix_impl.cpp#L254-L256).
```cpp
 if (key_handler_thread_.joinable()) {
    key_handler_thread_.join();
  }
```
8. Main `keyboard_handler` thread hangout for joining because it got blocked on [`read_fn`](https://github.com/ros-tooling/keyboard_handler/blob/70c0215c6e67dc434daadeba116e886ffd9d36c2/keyboard_handler/src/keyboard_handler_unix_impl.cpp#L189) system function call which became as a blocking call and waiting for `Enter` key to be pressed after we changed input to the canonical mode in signal handler.

## Proposed solution
- Do not use signal handler in `Keyboard handler` class since we are going to ignore it and correctly finish our application on upper level. There are available specific parameter in keyboard_handler constructor to avoid handling `SIGINT`.
- Trigger internal `exit` variable to true from signal handler in keyboard handler to avoid getting stuck on next read function after changing the canonical input mode.

MichaelOrlov:
- Closes #1079 
We don't need to explicitly handle `SIGINT` in `KeyboardHandler` instantiated from rosbag2 executables since we have already overriding `SIGINT` on upper python level or inside `rclcpp::init()` and correctly finishing application with proper calling of all destructors upon receiving `SIGINT` via checking `rclcpp::ok()` method in infinite loop.

Please note that it's almost impossible to cover this fix in unit tests since `gtest` using redirection of the standard input to the file and `KeyboardHandler` has detection for this special case and automatically disable keyboard handling from standard input on construction.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -116,7 +116,12 @@ Player::Player(
   const rclcpp::NodeOptions & node_options)
 : Player(std::move(reader),
     // only call KeyboardHandler when using default keyboard handler implementation
+#ifndef _WIN32
+    std::make_shared<KeyboardHandler>(false),
+#else
+    // We don't have signal handler option in constructor for windows version
     std::shared_ptr<KeyboardHandler>(new KeyboardHandler()),
+#endif
     storage_options, play_options,
     node_name, node_options)
 {}
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -59,7 +59,12 @@ Recorder::Recorder(
   const rclcpp::NodeOptions & node_options)
 : Recorder(
     std::move(writer),
-    std::make_shared<KeyboardHandler>(),
+#ifndef _WIN32
+    std::make_shared<KeyboardHandler>(false),
+#else
+    // We don't have signal handler option in constructor for windows version
+    std::shared_ptr<KeyboardHandler>(new KeyboardHandler()),
+#endif
     storage_options,
     record_options,
     node_name,


emersonknapp (approved):

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/8b1f44ec30da7cbf685d25a8290e2459/raw/571667e93bb076cef5922df3e096645f31e6cb84/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10777
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11801/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17742/)

MichaelOrlov:
Warning on Windows build is known issue existing on baseline and will be handled in https://github.com/ros2/rosbag2/pull/1084

MichaelOrlov:

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-9-15/27394/1

Issue #841, opened 2021-08-06
Bag split updating starting_time metadata field 


kielczykowski-rai:
## Description
Recording rosbag with split by duration option updates whole rosbag start timestamp after every split.

### Expected Behavior
Bag splitting doesn't update rosbag start time.

### Actual Behavior
Start time is being updated after every bag split call.

## System
 - **OS**: Ubuntu 20.04
 - **ROS** 2 Distro: Galactic/Rolling (from sources)
 - **Version**: 7407cd7364b7ce1eb2db22e2c3254eefb243f0fe (Rolling)

## To Reproduce
1. Run simple publisher `ros2 run examples_rclcpp_minimal_publisher publisher_not_composable`
2. Run rosbag record with split interval for example 30 s `ros2 bag record -a -d 30`
4. Record bag for multiple split intervals time
5. Check `duration` field in `metadata.yaml` file

## Additional context
Metadata of rosbag shows that whole bag duration equals to split time of bag even if it was being recorded for longer period of time.

Here is example of [metadata.yaml](https://github.com/ros2/rosbag2/files/6945959/metadata.txt) file for rosbag which was recorded for ~ 2.5 min.

Currently I am working on a solution to this issue. PR will be referenced here today or monday.

WJaworskiRobotec:
Additionally I discovered that checks of duration/starting_time are commented out in current tests. Together with fix of this bug we should fix existing tests and add test for recording with splitting by time and replaying the bag (I observed that it doesn't work properly with current master)

PR #1098, opened 2022-09-24
Fix: SequentialWriter incorrect metadata total duration for split bags

emersonknapp:
Fixes https://github.com/ros2/rosbag2/issues/841

This fix should get backported to Foxy/Galactic/Humble, it's been persistent for a long time. The test infra might not port well but the fix itself is very simple.

### DIFF
--- a/rosbag2_cpp/CMakeLists.txt
+++ b/rosbag2_cpp/CMakeLists.txt
@@ -223,7 +223,7 @@ if(BUILD_TESTING)
   ament_add_gmock(test_sequential_writer
     test/rosbag2_cpp/test_sequential_writer.cpp)
   if(TARGET test_sequential_writer)
-    ament_target_dependencies(test_sequential_writer rosbag2_storage)
+    ament_target_dependencies(test_sequential_writer rosbag2_storage rosbag2_test_common test_msgs)
     target_link_libraries(test_sequential_writer ${PROJECT_NAME})
   endif()
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -318,7 +318,7 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
 
   metadata_.files.back().starting_time =
     std::min(metadata_.files.back().starting_time, message_timestamp);
-  const auto duration = message_timestamp - metadata_.files.back().starting_time;
+  const auto duration = message_timestamp - metadata_.starting_time;
   metadata_.duration = std::max(metadata_.duration, duration);
 
   const auto file_duration = message_timestamp - metadata_.files.back().starting_time;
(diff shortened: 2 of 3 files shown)


emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/ba3d600536dc9333b17cb72b72bee0cf/raw/ab57c1b00397503f9dee8ac39ff6e66945f97e70/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10872
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17367/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11901/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17870/)

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/692acfa17f9a48b555e93ef4072b9302/raw/ab57c1b00397503f9dee8ac39ff6e66945f97e70/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10886
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17382/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11916/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17886/)

MichaelOrlov (commented):
@emersonknapp Thanks for fixing this issue!
LGTM.

clalancette:
@emersonknapp @MichaelOrlov I'm pretty sure this change broke building on RHEL; see https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_release/1273/console .  My guess is because RHEL-8 has an older version of gcc, and thus the version of `std::filesystem` doesn't work as well as it should.  Given that we are only using `std::filesystem` in a test, maybe we can slightly rewrite the test to be more friendly to RHEL-8?

emersonknapp:
Yeah this can be easily patched to use rcpputils filsystem helper. Just didn't catch the issue because rhel not in the standard ci suite. Maybe we should add it to ci_launcher?

Regardless, I'm AFK until tomorrow afternoon so I can't fix it until then, reverting would be ok by me if wanted.

MichaelOrlov:
@clalancette @emersonknapp I prepared fix in #1104 please review.

clalancette:
> Maybe we should add it to ci_launcher?

Yeah, we've gone back and forth on that question.  RHEL is a Tier-2 platform, which means that we don't technically guarantee that it will work.  On the other hand, we do have people using it, so we want to make sure it works as well as possible.

The problem with adding it to the `ci_launcher` is that it puts more strain on the CI workers, and we are already running kind of close to the limit.  RHEL doesn't *usually* require special handling, but did in this case.  I'm inclined to leave things as-is, and revisit this question if more RHEL-specific problems start to pop up.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-10-13/28213/1

PR #1107, opened 2022-09-30
zstd-vendor: fix find of system zstd

DasRoteSkelett:
The find_package(zstd) silently failed because the Findzstd.cmake did not get picked up. Setting the CMAKE_MODULE_PATH to look for the Findzstd.cmake in cmake/Modules and also removing the quiet from find_package, which covered the error.

Signed-off-by: Matthias Schoepfer <m.schoepfer@rethinkrobotics.com>


emersonknapp:
I don't think this is necessary (but maybe I'm wrong). In what situation was it not finding your system zstd? What operating system, ros version, code checkout. What commands did you run, and what output did you get? I'd like to make sure this fix is tied to a reproducible bug report before going forward with anything.

clalancette:
In any case, we definitely can't use PkgConfig here; that won't work on Windows.

DasRoteSkelett:
Thanks for the comments, I addressed the issues and will open an new PR

DasRoteSkelett:
> I don't think this is necessary (but maybe I'm wrong). In what situation was it not finding your system zstd? What operating system, ros version, code checkout? What commands did you run, and what output did you get? I'd like to make sure this fix is tied to a reproducible bug report before going forward with anything.

@emersonknapp Just out of curiosity: When does it work?

Issue #1110, opened 2022-10-02
zstd_vendor does not find system zstd and will always build and install 


DasRoteSkelett:
## Description
zstd_vendor does not find system zstd and will always pull and install from upstream. This is unwanted behavior. 

### Expected Behavior
System installed zstd should get found and build/install should therefore be skipped

### Actual Behavior
build zstd_vendor with zstd installed. zstd_vendor does indeed not find the installed zstd and will pull, build and install from upstream.
I could reproduce with ubuntu 22.04, gentoo and yocto. May work on Windows, but I doubt it, to be honest. 

## To Reproduce
1. `apt install zstd`
2. `source /opt/ros/humble/setup.bash`
3. clone rosbag2 and change into `zstd_vendor` subdir
4. `mkdir build && cd build && cmake ..`
5. You will *not* see message `Found Zstd, skipping build.`
6. Instead, if you build, zstd will get fetched and installed. 

## System (please complete the following information)
 - OS: Ubuntu Jammy, gentoo, yocto honister
 - ROS 2 Distro: Humble / Rolling
 - Version: rolling

## Additional context
The issue is twofold: the Findzstd.cmake is not found, which you see when removing `QUIET` from find package:
```
CMake Warning at CMakeLists.txt:12 (find_package):
  By not providing "Findzstd.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "zstd", but
  CMake did not find one.

  Could not find a package configuration file provided by "zstd" with any of
  the following names:

    zstdConfig.cmake
    zstd-config.cmake

  Add the installation prefix of "zstd" to CMAKE_PREFIX_PATH or set
  "zstd_DIR" to a directory containing one of the above files.  If "zstd"
  provides a separate development package or SDK, be sure it has been
  installed.
```

This can be fixed by adding `cmake/Modules` to the `CMAKE_MODULE_PATH`

Then, you will see: 
```
-- Found zstd: /usr/lib/x86_64-linux-gnu/libzstd.so
```
But not 
```
Found Zstd, skipping build.
```
because `${zstd_VERSION}` remains unset and therefore does not match required version comparison. 

This needs to be addressed in Findzstd.cmake

emersonknapp:
Thanks for the detailed bug report

PR #1111, opened 2022-10-02
Fix/zstd vendor does not find system zstd

DasRoteSkelett:
Closes: #1110 

by the way, what a cool issue id to have :-)


zstd_vendor/CMakeLists.txt
```diff
@@ -4,12 +4,14 @@ project(zstd_vendor)
 
 find_package(ament_cmake REQUIRED)
 
+set(CMAKE_MODULE_PATH "${CMAKE_MODFULE_PATH}" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
```
clalancette, line 7: Our usual pattern here is to use the `list` command to insert the module path:

```suggestion
list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
```

zstd_vendor/CMakeLists.txt
```diff
@@ -59,7 +61,7 @@ macro(build_zstd)
     USE_SOURCE_PERMISSIONS)
 endmacro()
 
-if (NOT zstd_FOUND OR "${zstd_VERSION}" VERSION_LESS 1.4.4)
+if (NOT zstd_FOUND)
```
clalancette, line 64: I don't think there is any reason to remove this check.  Let's just leave it:

```suggestion
if (NOT zstd_FOUND OR "${zstd_VERSION}" VERSION_LESS 1.4.4)
```

zstd_vendor/cmake/Modules/Findzstd.cmake
```diff
@@ -17,11 +17,24 @@ endif()
...
 
+if (zstd_INCLUDE_DIR)
+  file(STRINGS "${zstd_INCLUDE_DIR}/zstd.h" version-file
+    REGEX "#define[ \t]+ZSTD_VERSION_(MAJOR|MINOR|RELEASE).*")
+  list(GET version-file 0 major-line)
+  list(GET version-file 1 minor-line)
+  list(GET version-file 2 release-line)
+  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_MAJOR[ \t]+ ([0-9]+)$" "\\1" ZSTD_VERSION_MAJOR ${major-line})
+  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_MINOR[ \t] +([0-9]+)$" "\\1" ZSTD_VERSION_MINOR ${minor-line})
+  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_RELEASE[ \t]+ ([0-9]+)$" "\\1" ZSTD_VERSION_RELEASE ${release-line})
```
clalancette, line 28:
```suggestion
  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_MAJOR[ \t]+([0-9]+)$" "\\1" ZSTD_VERSION_MAJOR ${major-line})
  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_MINOR[ \t]+([0-9]+)$" "\\1" ZSTD_VERSION_MINOR ${minor-line})
  string(REGEX REPLACE "^#define[ \t]+ZSTD_VERSION_RELEASE[ \t]+([0-9]+)$" "\\1" ZSTD_VERSION_RELEASE ${release-line})
```

clalancette (changes requested):
I tested it out, and it is indeed the case that we don't detect the built-in zstd.  So thanks for this.

I left a few things inline that should clean this up more.  Once that is in, we can run CI.  I'd also like to get @emersonknapp 's opinion before merging.

zstd_vendor/CMakeLists.txt
emersonknapp: Yes, we do depend on specific APIs such that the version needs to be high enough

emersonknapp:
Thanks @clalancette for the careful pass. Like you, I think this looks mostly fine, handling the little nitpicks puts it in good shape to move forward.

DasRoteSkelett:
Thanks for the review @clalancette and @emersonknapp . I applied most suggestions, except that the version check is done with find_package, which i believe is the more standard way. I documented it a little better and hope you are okay with it

zstd_vendor/CMakeLists.txt
```diff
@@ -4,12 +4,15 @@ project(zstd_vendor)
...
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
+
 option(FORCE_BUILD_VENDOR_PKG
   "Build zstd from source, even if system-installed package is available"
   OFF)
 
 if(NOT FORCE_BUILD_VENDOR_PKG)
-  find_package(zstd QUIET)
+  # We need at least VERSION 1.4.4, version check is done in Findzstd.cmake
+  find_package(zstd 1.4.4 QUIET)
```
clalancette, line 15: But the thing is, this requests *exactly* 1.4.4, which is not what we want.  We want at least 1.4.4.  So I still think this should just be:

```suggestion
  find_package(zstd QUIET)
```

And then restore the version check below.

DasRoteSkelett: Nope. It is at least 1.4.4. I tested it *and* documentation is also on my side: 
```
The [version] argument requests a version with which the package found should be compatible. There are two possible forms in which it may be specified:

A single version with the format major[.minor[.patch[.tweak]]], where each component is a numeric value.

A version range with the format versionMin...[<]versionMax where versionMin and versionMax have the same format and constraints on components being integers as the single version. By default, both end points are included. By specifying <, the upper end point will be excluded. Version ranges are only supported with CMake 3.19 or later.

The EXACT option requests that the version be matched exactly. This option is incompatible with the specification of a version range.
```
From https://cmake.org/cmake/help/latest/command/find_package.html

clalancette: Good point.  I learned something about CMake today.

And there is precedent for this over in https://github.com/ros2/orocos_kdl_vendor/blob/rolling/orocos_kdl_vendor/CMakeLists.txt .  So OK, I'm fine with this change then.

clalancette (approved):
This looks good to me now, @emersonknapp can you take another look?

emersonknapp (approved):
yup lgtm - this needs a rebase on `rolling` branch unfortunately for the CI to pass, there was a new package added - @DasRoteSkelett could you do that? Then I'll run CI on the buildfarm

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17969/)

clalancette:
@emersonknapp Oops, I missed your edit.  You are right, this needs a rebase.  I'll let you take it from here and run CI.

DasRoteSkelett:
@emersonknapp @clalancette Thanks, I did the rebase. Guess now you need to do your magic
image

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/2a2db76258857598c12a0e6d5e6955fe/raw/a9cedca1f7f5d06066eb408baed55ab445d2fde6/ros2.repos
BUILD args:  --packages-above-and-dependencies zstd_vendor
TEST args:  --packages-above zstd_vendor
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/10952
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17458/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11986/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17987/)

clalancette:
@Mergifyio backport humble

PR #1117, opened 2022-10-06
Don't reopen file for every seek if we don't have to. Search directionally for the correct file

emersonknapp:
Use case: `rqt_bag` uses frequent seek (for every message). This might not be the best way to use rosbag, but regardless it highlights a big performance issue with `seek` - that we reopen the storage file every single time. Change this to directionally look for a file that contains the requested timestamp, then seek within that bag. This makes it so that repeated seeks within the same file keep that file open.

Discovered in testing for https://github.com/ros-visualization/rqt_bag/pull/126

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -143,8 +143,8 @@ bool SequentialReader::has_next()
     // If there's no new message, check if there's at least another file to read and update storage
     // to read from there. Otherwise, check if there's another message.
     bool current_storage_has_next = storage_->has_next();
-    if (!current_storage_has_next && has_next_file()) {
-      load_next_file();
+    if (!current_storage_has_next && has_next_file(read_order_.reverse)) {
+      load_next_file(read_order_.reverse);
       // recursively call has_next again after rollover
       return has_next();
     }
@@ -201,23 +201,33 @@ void SequentialReader::reset_filter()
 void SequentialReader::seek(const rcutils_time_point_value_t & timestamp)
 {
   seek_time_ = timestamp;
-  if (storage_) {
-    // reset to the first file
-    if (read_order_.reverse) {
-      current_file_iterator_ = std::prev(file_paths_.end());
-    } else {
-      current_file_iterator_ = file_paths_.begin();
-    }
-    load_current_file();
-    return;
+  if (!storage_) {
+    throw std::runtime_error(
+            "Bag is not open. Call open() before seeking time.");
   }
-  throw std::runtime_error(
-          "Bag is not open. Call open() before seeking time.");
+
+  auto metadata = storage_->get_metadata();
+  auto start_time = metadata.starting_time.time_since_epoch().count();
+  auto end_time = (metadata.starting_time + metadata.duration).time_since_epoch().count();
+
+  if (timestamp < start_time && has_next_file(true)) {
+    // Check back a file if the timestamp is before the beginning of the current file
+    load_next_file(true);
+    return seek(timestamp);
+  } else if (timestamp > end_time && has_next_file(false)) {
+    // Check forward a file if the timestamp is after the end of the current file
+    load_next_file(false);
+    return seek(timestamp);
+  } else {
+    // The timestamp lies in the range of this file, or there are no files left to go to
+    storage_->seek(timestamp);
+  }
+  return;
 }
 
-bool SequentialReader::has_next_file() const
+bool SequentialReader::has_next_file(bool reverse) const
 {
-  if (read_order_.reverse) {
+  if (reverse) {
     return current_file_iterator_ != file_paths_.begin();
   } else {
     return (current_file_iterator_ + 1) != file_paths_.end();
@@ -244,12 +254,12 @@ void SequentialReader::load_current_file()
   set_filter(topics_filter_);
 }
 
-void SequentialReader::load_next_file()
+void SequentialReader::load_next_file(bool reverse)
 {
   assert(current_file_iterator_ != file_paths_.end());
   auto info = std::make_shared<bag_events::BagSplitInfo>();
   info->closed_file = get_current_file();
-  if (read_order_.reverse) {
+  if (reverse) {
     current_file_iterator_--;
   } else {
     current_file_iterator_++;
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -143,10 +143,15 @@ bool SequentialReader::has_next()
     // If there's no new message, check if there's at least another file to read and update storage
     // to read from there. Otherwise, check if there's another message.
     bool current_storage_has_next = storage_->has_next();
-    if (!current_storage_has_next && has_next_file(read_order_.reverse)) {
-      load_next_file(read_order_.reverse);
-      // recursively call has_next again after rollover
-      return has_next();
+    if (!current_storage_has_next) {
+      if (!read_order_.reverse && has_next_file()) {
+        load_next_file();
+        return has_next();
+      }
+      if (read_order_.reverse && has_prev_file()) {
+        load_prev_file();
+        return has_next();
+      }
     }
     return current_storage_has_next;
   }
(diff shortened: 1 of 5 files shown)


MichaelOrlov:
@emersonknapp I've just did analysis for this problem yesterday when trying to figure out why we have strange warning in player.
```
[WARN] [rosbag2_storage]: No storage plugin found with id 'sqlite3'.
```
I came to conclusion that in
https://github.com/ros2/rosbag2/blob/e26743ea0a025b53fbaa4b02b01831216b4f4702/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L237
We need to add API to be able to open another file in current storage instead of recreating storage. Recreating storage is expensive. It is reload plugin in memory. And we call `load_current_file()` for each `seek(timestamp)` call.

Note: I've had a chance to review your changes yet.

jhdcs (approved):
These changes look good to me. I agree on removing the misleading warning - best not to confuse users by warning about normal operations.

jhdcs (approved):

rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp
```diff
@@ -116,8 +116,9 @@ get_interface_instance(
     registered_classes.begin(),
     registered_classes.end(), storage_options.storage_id);
   if (class_exists == registered_classes.end()) {
-    ROSBAG2_STORAGE_LOG_WARN_STREAM(
-      "No storage plugin found with id '" << storage_options.storage_id << "'.");
+    // This should not print a warning, because it can be used by open_read_only twice,
+    // legitimately expecting to fail for READ_ONLY but succeed for READ_WRITE
+    // The extra output is misleading to end users.
```
MichaelOrlov, line 121: @emersonknapp May be still keep logging with redirection to the debug stream? In regular usage users will not see such message.
I am worry about a case when we can fail loading plugin in second iteration with `read_write` interface and will not have any meaningful outputs or logging about it.

MichaelOrlov (changes requested):
@emersonknapp Thank you for your contribution and sorry for my late response.
This is certainly improvements comparable to how it was before. 
Although I would like to give  some "love" to the logic in implementation to make more clear.

I have some concern  in regards to the API changes in particular to the `has_next_file(bool reverse)` and `load_next_file(bool reverse)`.
After some consideration I think it will be better to create separate API for `has_previous_file()` and `load_next_file()`
You may argue that current implementation is better align with current implementation since we already have `reverse` flag on which we rely up on. 

From one hand it look like easy and more straight forward in implementation in one place 
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L146-L147

However from another hand it's not very clear in another place where logic more sophisticated 
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L213-L224
IMO it will be much cleaner to understand what is going on in logic if API will be with self described names and granuraly for instance:
```cpp
if (timestamp < start_time && has_prev_file()) {
    // Check back a file if the timestamp is before the beginning of the current file
    load_prev_file(true);
    return seek(timestamp);
  } else if (timestamp > end_time && has_next_file()) {
    // Check forward a file if the timestamp is after the end of the current file
    load_next_file();
    return seek(timestamp);
  } else {
    // The timestamp lies in the range of this file, or there are no files left to go to
    storage_->seek(timestamp);
  }
```
When someone will try to comprehend this logic and will see construction like `has_next_file(true)` it's not really obvious that this method call really checking for existence of the previous file rather than next.  And it will cause a lot of confusion and will have to keep in mind that next file is not really the next but previous even if name it as next. It's more error prone approach.

Also you can argue that it's made for the sake of the reducing code duplication in "next_/previous_" methods. I was trying to consider this argument as well and it looks like this not really the case. It will be a duplication of a few lines of code in `load_next_file()` and `load_previous_file()` 
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L259-L269
and logic there pretty much straight forward and will became eve more simple i.e. without any branching like if-else if we will duplicate it in relevant methods.  

I also was thinking about performance impact for calling `get_metadata` from storage for each seek operation.
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L209

In current implementation at least in `sqlite3` storage plugin this is pretty much expensive call. 
We do request to the database for pulling out information about all topics and counting all messages in essence internally in db engine it will translates to iterating over each message give results from such SQL request. 
I think we can optimize this place if we will readout and calculate values for metadata only once  when we are opening storage and keep metadata in private member variable ready to be returned when needed by request from `get_metadata()` method.
Although there are one caveat in such approach. We need to take in to account situation when get metadata could be called for storage which was opened with `read_write` or `append` `io_flag`. 
We need to readout metadata only if `io_flag == READ_ONLY`. 
We can built in somewhere here 
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp#L195-L197
and will need to keep this `io_flag` as private member variable to be able to check it in `get_metadata()`  to be able to decide if we need to do a fresh metadata readout or we can return just cached value saved in private member variable during storage opening.

rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp
emersonknapp: There is already an ERROR message on the second (or first) iteration in the public APIs. This function is an implementation internal 

https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp#L199

emersonknapp:
I'm fine with `has_prev_file` and `load_prev_file` API instead of parameter. I'll make that change

rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp
MichaelOrlov: Ok good. :+1:

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -244,12 +254,12 @@ void SequentialReader::load_current_file()
   set_filter(topics_filter_);
 }
 
-void SequentialReader::load_next_file()
+void SequentialReader::load_next_file(bool reverse)
 {
   assert(current_file_iterator_ != file_paths_.end());
```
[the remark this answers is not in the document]
emersonknapp: This assert is `load_current_file` - assert is still a valid error check, `current_file_iterator_ == file_paths_.end()` will always be wrong

emersonknapp:
@MichaelOrlov updated based on review - re-review requested (sorry for slow turnaround, didn't work during japan roscon trip)

MichaelOrlov (approved):
@emersonknapp Looks good enough to be merged with green CI. 

Although it could be even better if address performance impact for calling `get_metadata` from storage for each seek operation.
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp#L209

In current implementation at least in `sqlite3` storage plugin this is pretty much expensive call. 
We do request to the database for pulling out information about all topics and counting all messages in essence internally in db engine it will translates to iterating over each message give results from such SQL request. 
I think we can optimize this place if we will readout and calculate values for metadata only once  when we are opening storage and keep metadata in private member variable ready to be returned when needed by request from `get_metadata()` method.
Although there are one caveat in such approach. We need to take in to account situation when get metadata could be called for storage which was opened with `read_write` or `append` `io_flag`. 
We need to readout metadata only if `io_flag == READ_ONLY`. 
We can built in somewhere here 
https://github.com/ros2/rosbag2/blob/e948e954700335b326daee3152f81221e4b7d42a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp#L195-L197
and will need to keep this `io_flag` as private member variable to be able to check it in `get_metadata()`  to be able to decide if we need to do a fresh metadata readout or we can return just cached value saved in private member variable during storage opening.

@emersonknapp Thoughts? May be implement in follow up issue?

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/310884637693cd81b243857191413684/raw/cb6b5965073a9c3c99085a1d4f69258479aa7998/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_storage rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_cpp rosbag2_storage rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11087
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17604/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18174/)

MichaelOrlov:
@emersonknapp Ups I occasionally merged this PR without running CI. 
I've restored branch and trying to run CI.  
I hope we haven't broke anything.
- If CI will fail most probably will go ahead with #1151

MichaelOrlov:
As regards my suggestions about improvements with caching metadata in `get_metadata()` I have addressed it in #1149

MichaelOrlov:
All fine. I am deleting brunch and closing #1151

emersonknapp:
Thank you for the follow through! I meant to say that yes, I think other changes should be followups to this one, keep PRs small

PR #1118, opened 2022-10-08
Bump zstd to 1.5.2 in zstd_vendor package

MichaelOrlov:
- Follow up from discussion https://github.com/ros-tooling/rosbag2_storage_mcap/pull/56#issuecomment-1272101956

Note: We need zstd at least version 1.5.2 for mcap_vendor package.

Also removed cmake patches, since they became obsolete for new version.


MichaelOrlov:
For the context of this PR I am re-posting relevant discussion from https://github.com/ros-tooling/rosbag2_storage_mcap/pull/56 thread to continue discussion here:

[MichaelOrlov](https://github.com/MichaelOrlov) commented [3 days ago](https://github.com/ros-tooling/rosbag2_storage_mcap/pull/56#issuecomment-1272217508)
> @clalancette I figured out how to compile mcap_vendor package with zstd_vendor package as dependencies without pulling zstd from git repo and compiling it from source. Perhaps we will be able to do the same with lz4.
Long story short it will require fix inside mcap cpp implementation and multiple adjustments in mcap_vendor cmake file.
There are one caveat that we will need to bump zstd version in zstd_vendor to the 1.5.2. This is the latest version of the zstd version and not available by default in Ubuntu 22.04.
We still can have the rule to use locally installed version without mandatory downloading it. But the problem is that it's going to cause downloading it since this 1.5.2 version is not available yet by default.
In my Ubuntu 22.04 installation zstd with version 1.4.8.

> Is it ok to go with this approach and bump zstd version to 1.5.2?

[amacneil](https://github.com/amacneil) commented [yesterday](https://github.com/ros-tooling/rosbag2_storage_mcap/pull/56#issuecomment-1272589268)
> What's the reason that mcap depends on such a new version of zstd? Can it be updated to support older versions?

[clalancette](https://github.com/clalancette) commented [yesterday](https://github.com/ros-tooling/rosbag2_storage_mcap/pull/56#issuecomment-1272671488)
> What's the reason that mcap depends on such a new version of zstd? Can it be updated to support older versions?

> We've been trying to reduce the number of vendored packages that we actually build[1] on our primary platform, Ubuntu 22.04. This reduces compile times and integrates much more nicely with the base system.

> Thus, if there is any way we can make mcap use zstd 0.4.4 (the version current in Ubuntu 22.04), that would be ideal.

> [1] In the interest of full disclosure, we aren't there yet. We unconditionally vendor at least OGRE and libyaml currently, but ideally we'd make both of those conditional as well. Most of our other vendored packages are conditional.

MichaelOrlov:
@clalancette as regards to the 
> What's the reason that mcap depends on such a new version of zstd? Can it be updated to support older versions?

@amacneil Please correct me if I am wrong. mcap depends from API from new version `zstd` to achieve outstanding performance when using compression. And this is a good and fair reason to depend from newer `zstd` version.

As far as understand we can't downgrade `zstd` version to 1.4.8 or 1.4.4 without sacrificing some functionality in mcap. At least build fails on my local machine with `zstd` v1.4.8.

The major motivation to start using zstd from `zstd_vendor` package is to avoid downloading and usage of different versions of `zstd` library in ROS2 ecosystem.

If you will see to the proposed changes more closely the situation is that even without bumping `zstd` version to 1.5.2 we are currently downloading older **1.4.4** version on CI since Ubuntu 22.04 has **1.4.8** version.
https://github.com/ros2/rosbag2/blob/e26743ea0a025b53fbaa4b02b01831216b4f4702/zstd_vendor/CMakeLists.txt#L15

I would propose to move forward with upgrading to the 1.5.2 version.
And if need to get rid from downloading it on CI I think the proper solution would be to pre-install zstd version 1.5.2 to the base docker image with Ubuntu 22.04 on CI.

amacneil:
👍 it might be worth benchmarking the feature we use in v1.5.2 vs doing it the v1.4.4 way so that we have some actual data.

But I agree upgrading `mcap_vendor` seems like the best path forward if we require that version.

clalancette:
> And if need to get rid from downloading it on CI I think the proper solution would be to pre-install zstd version 1.5.2 to the base docker image with Ubuntu 22.04 on CI.

This is not an option; while that would fix CI, that would not make it work for end-users without installing additional repositories, which we don't want to do.

> If you will take a look to the proposed changes more closely the situation is that even without bumping `zstd` version to 1.5.2 we are currently downloading older **1.4.4** version on CI since Ubuntu 22.04 has **1.4.8** version.

I got the version numbers wrong, sorry.  The goal for vendor packages is always to use the package unmodified from our reference platform (Ubuntu 22.04), and then build it for the platforms where it is not that easy to do that (mostly Windows, but Ubuntu 20.04 would also fall into that category).  So in this case, the best situation is that we use zstd 1.4.8 from Ubuntu 22.04, and then build version 1.4.8 from source on all other platforms.

I know you said that mcap doesn't compile with zstd 1.4.8, but did you look into the reasons why?  Could we make it work with that version?

MichaelOrlov:
@clalancette as regards to the your question
> I know you said that mcap doesn't compile with zstd 1.4.8, but did you look into the reasons why? Could we make it work with that version?

It would be better to ask folks from Foxglove. I don't know details of the mcap C++ implementation.
cc: @amacneil

amacneil:
Pinging Foxglove to answer why the newer version was used @jhurliman @james-rms

clalancette:
See https://github.com/ros-tooling/rosbag2_storage_mcap/pull/59 , which is the direction I think we should go in.

emersonknapp:
I am noticing that it is implicitly, but not explicitly mentioned above - the `mcap` C++ library uses a new API from zstd 1.5 that is not present in 1.4. However it hasn't been benchmarked how much benefit is gained by using this new API

clalancette:
> I am noticing that it is implicitly, but not explicitly mentioned above - the `mcap` C++ library uses a new API from zstd 1.5 that is not present in 1.4. However it hasn't been benchmarked how much benefit is gained by using this new API

Hm, are you sure?  I was able to build successfully with https://github.com/ros-tooling/rosbag2_storage_mcap/pull/59 and my zstd 1.4.8 (though I did not run any tests).

jhurliman:
The https://github.com/foxglove/mcap C++ library uses the `ZSTD_compress2()` API for reusing a compression context across multiple compression sessions. The context includes internal zstd allocations and should only be initialized once during startup, not repeatedly in the recording loop. The `ZSTD_compress2()` API stabilized in zstd 1.4.7. Ubuntu Bionic and Focal are currently shipping zstd <= 1.4.4, so compilation will fail on those platforms when linking against the system zstd library. Jammy is fine. I can't speak to whether dropping support for pre-Jammy is ok for rosbag2.

In addition to the compilation failure, reverting from zstd 1.5.x to 1.4.8 will result in lower compression and decompression performance and worse compression ratios according to the release notes at <https://github.com/facebook/zstd/blob/dev/CHANGELOG>. We have a benchmark suite for the mcap rosbag2 plugin now, but it needs to be run on the same hardware before + after to measure the impact on write throughput and resulting file size. If the benchmarks do show a >5% drop in throughput, I can't imagine the typical ROS2 user would be happy with a change that results in a poorer-performing robot in exchange for a more pristine packaging approach. In that situation, should we consider copying the zstd source code into the zstd_vendor ROS2 package?

clalancette:
> The https://github.com/foxglove/mcap C++ library uses the `ZSTD_compress2()` API for reusing a compression context across multiple compression sessions. The context includes internal zstd allocations and should only be initialized once during startup, not repeatedly in the recording loop. The `ZSTD_compress2()` API stabilized in zstd 1.4.7. Ubuntu Bionic and Focal are currently shipping zstd <= 1.4.4, so compilation will fail on those platforms when linking against the system zstd library. Jammy is fine. I can't speak to whether dropping support for pre-Jammy is ok for rosbag2.

I don't think we should drop support for pre-Jammy, but that's what the vendor package is for.  If we make the vendor package look for at least 1.4.7, then anything that has less than that will compile it from source (Ubuntu 20.04, Windows, etc).  Ubuntu Jammy and newer will then use the system package.

> In addition to the compilation failure, reverting from zstd 1.5.x to 1.4.8 will result in lower compression and decompression performance and worse compression ratios according to the release notes at https://github.com/facebook/zstd/blob/dev/CHANGELOG. We have a benchmark suite for the mcap rosbag2 plugin now, but it needs to be run on the same hardware before + after to measure the impact on write throughput and resulting file size. If the benchmarks do show a >5% drop in throughput, I can't imagine the typical ROS2 user would be happy with a change that results in a poorer-performing robot in exchange for a more pristine packaging approach. In that situation, should we consider copying the zstd source code into the zstd_vendor ROS2 package?

It's not just about some pristine packaging approach.  It's that 1) vendoring packages can very easily lead to hard-to-debug ABI problems, and that 2) you can make the same argument about pretty much every single library that we use.  Yes, it would be better to have newer OGRE.  Yes, it would be better to have newer OpenCV.  But the fact is that we cannot vendor everything we depend on, and we don't want to impose the ABI problems on users.

So with all of that said, my preferred approach here is to require at least zstd 1.4.8, which will solve basically all of these issues except for the better performance.  When Ubuntu 24.04 rolls around, we'll get the updated performance.

clalancette:
I think we can close this one, right?  We've updated to 1.4.8, which matches what is in Ubuntu 22.04.

MichaelOrlov:
@clalancette Thank you for remind. I almost forgot about this PR.
- Closing this PR as will not done in favor #1132

The rational for that:
1. Both `mcap` and `rosbag2` compile without errors with `zstd` version 1.4.8
2. There are not a big difference in performance between 1.4.8 and 1.5.2(1) versions
Performance comparison taken from https://github.com/facebook/zstd/tree/v1.5.2 and https://github.com/facebook/zstd/tree/v1.4.8

Compressor name | Ratio | Compression | Decompress.
-- | -- | -- | --
zstd 1.5.1 -1 | 2.887 | 530 MB/s | 1700 MB/s
zstd 1.4.5 -1 | 2.884 | 500 MB/s | 1660 MB/s
zstd 1.4.0 -1 | 2.884 | 530 MB/s | 1360 MB/s

Issue #1086, opened 2022-09-12
Store ROS_DISTRO as metadata.yaml and in the storage file


wkalt:
## Description
I recently discovered that ros2 db3 bag files recorded by different versions of ROS can have different database schemas. For example, bags recorded on eloquent do not have the offered_qos_profiles column in the topics table, which caused our (Foxglove's) db3 to mcap converter to fail on eloquent bags.

It would be helpful if the ros2 bag recorder wrote its ROS version to the db3 file, to enable consumers to cleanly dispatch to different parsing behavior.

## Implementation Notes / Suggestions
Perhaps a new table like "metadata" or some similar concept.

MichaelOrlov:
@wkalt Could you please provide some rosbag2 db3 bag file which is failing converter to process it? 
It will significantly help for issue analysis and possible fixes.

wkalt:
@MichaelOrlov thanks, here's one @amacneil recorded on eloquent:
[eloquent-twist.db3.zip](https://github.com/ros2/rosbag2/files/9549856/eloquent-twist.db3.zip)

MichaelOrlov:
@wkalt In `rosbag2_storage` we already storing metadata version and taking it in to account when parsing metadata from `yaml` file
https://github.com/ros2/rosbag2/blob/e179c79b0d9acee0be58c676ae8b124fd42a3bc6/rosbag2_storage/include/rosbag2_storage/yaml.hpp#L246-L260
In particular `offered_qos_profiles` was introduced since metadata.version = 4
https://github.com/ros2/rosbag2/blob/e179c79b0d9acee0be58c676ae8b124fd42a3bc6/rosbag2_storage/include/rosbag2_storage/yaml.hpp#L109-L113

MichaelOrlov:
However we are not handling medata.version properly when trying to read out metadata from database directly.

amacneil:
@MichaelOrlov to clarify, the specific request here is that this metadata (the ROS 2 release) is stored in the db3 file when using the sqlite plugin.

We have users who open a sqlite file directly (i.e. no `metadata.yml` file available to us) and having the metadata fully contained within the db3 file makes it easier to reason about. More generally I think it would be best if there is no information in `metadata.yml` that is not also stored in the db3 file directly.

MichaelOrlov:
@amacneil I agree it would be nice if all metadata information from `metadata.yaml` file would be duplicated in db3 file directly.
But currently it is not. 
[...]

PR #1149, opened 2022-11-06
Add `update_metadata(BagMetadata)` API for storage plugin interface

MichaelOrlov:
- We need `update_metadata(BagMetadata)` API to be able to save `BagMetadata` directly inside each bag file, in SQLite DB or MCAP file.
- update_metadata(BagMetadata) calling ones after creating storage object and second time just before destructing it.
- Also added caching for `SqliteStorage::get_metadata()` as follow up for #1117

- Note: Real saving of bag metadata in SQLite DB will be done in separate PR after #1108. We need to bring this API first to unblock #1144

Relates:
  - #1109
  - #1144 
  - #1117

Signed-off-by: Michael Orlov <michael.orlov@apex.ai>

Issue #1109, opened 2022-10-02
Store metadata in db3 files directly


MichaelOrlov:
## Description
We have a proposal to make rosbag2 .db3 files self-contained. To achieve this goal we need to store metadata in .db3 files directly.

## Related Issues
Relates: 
- https://github.com/ros2/rosbag2/issues/1074
- https://github.com/ros2/rosbag2/issues/1108
- https://github.com/ros2/rosbag2/issues/782

## Completion Criteria
All `metadata` fields stored in .db3 files directly.

## Implementation Notes / Suggestions
On "Rosbag2 and Tooling WG meeting" from 2022-09-23 we discussed possible implementation and came to the following design decisions and proposals:
1. We decided to store metadata as a string in yaml format in newly created `schema`  table from https://github.com/ros2/rosbag2/issues/1108. 
It will be relatively easily to do since we already have all necessary methods for creating and parsing metadata from string read out from yaml file.
2. We will keep writing `metadata.yaml` file as before and rely on metadata from it. 
3. We will fallback on metadata from .db3 file directly if `metadata.yaml` absent or our parser failed to read or parse data from yaml file. 
The rational for this is that it's very convenient to operate with metadata stored in yaml file. i.e. It's very convenient to open yaml file in simple text editor to see what inside or edit it if necessary.

## Testing Notes / Suggestions
- Will need to add unit tests to cover read out and parsing metadata from .db3 files directly
- Will need to add unit tests to make sure that old versions of bag files supported. Need to make sure that we support backward compatibility.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-10-13/28213/1

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-11-17/28457/1

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-12-15/28840/1

PR #1156, opened 2022-11-10
Store db schema version and ROS_DISTRO name in db3 files

### LINKED ISSUE #1108: Store DB schema versioning in .db3 files
## Description
Some times we need to change DB schema. For instance add new field or even a new table connected by some index with another table.
Currently we are not storing schema version in .db3 file and it's very difficult to support backward compatibility in case of changing DB schema. 
We need to store DB schema version in .db3 file directly to be able properly read out bag files with different versions.

## Related Issues
Related to the #1086 

## Completion Criteria
1. - [ ] Database schema version stored in .db3 files directly.
2. - [ ] Backward compatibility for old db schema supported.

On "Rosbag2 and Tooling WG meeting" from 2022-09-23 we discussed possible implementation and considered 2 options for storing db schema.
1. Using builtin sqlite pragma for db schema version https://www.sqlite.org/pragma.html#pragma_schema_version 
3. Creating new table in database for storing DB schema version in it.

We decided to go with second option because builtin pragma schema version is intended for internal usage by sqlite engine and supposed to be incremented by sqlite engine when doing internal operations on database. There are some notes in description that changing those value could lead to the undefined behavior.
One more argument towards to the second option is that we can store metadata version in the same table as well.

Presumably will need to add new table in [SqliteStorage::Initialize()](https://github.com/ros2/rosbag2/blob/6a07c16c8b094d0dda47da15ac2498010c1a4dc8/rosbag2_storage_default_plugins/src/rosbag2_storage_default_plugins/sqlite/sqlite_storage.cpp#L337-L354) method.
```cpp
 std::string create_stmt = "CREATE TABLE schema(" \
    "schema_version INTEGER PRIMARY KEY,"
    "metadata_version INTEGER NOT NULL);";
 database_->prepare_statement(create_stmt)->execute_and_reset();
```

Will need to add coverage in unit tests to make sure that bag files created with previous version of the schema layout supported. i.e. possible to create, open reader and read messages from it.

MichaelOrlov:
- Closes #1108 
- Relates #1086

- Add new tables for `schema` and `metadata`

### DIFF
--- a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
+++ b/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
@@ -213,7 +213,10 @@ void SqliteStorage::open(
 
   // initialize only for READ_WRITE since the DB is already initialized if in APPEND.
   if (is_read_write(io_flag)) {
+    db_schema_version_ = kDBSchemaVersion_;
     initialize();
+  } else {
+    db_schema_version_ = read_db_schema_version();
   }
 
   // Reset the read and write statements in case the database changed.
@@ -640,11 +643,47 @@ SqliteWrapper & SqliteStorage::get_sqlite_database_wrapper()
   return *database_;
 }
 
+int SqliteStorage::get_db_schema_version() const
+{
+  return db_schema_version_;
+}
+
+std::string SqliteStorage::get_recorded_ros_distro() const
+{
+  std::string ros_distro;
+  if (db_schema_version_ >= 3 && database_->table_exists("schema")) {
+    // Read schema version
+    auto statement = database_->prepare_statement("SELECT ros_distro from schema;");
+    auto query_results = statement->execute_query<std::string>();
+    ros_distro = std::get<0>(*query_results.begin());
+  }
+  return ros_distro;
+}
+
 int SqliteStorage::get_last_rowid()
 {
   auto statement = database_->prepare_statement("SELECT max(rowid) from messages;");
   auto query_results = statement->execute_query<int>();
   return std::get<0>(*query_results.begin());
+}
+
+int SqliteStorage::read_db_schema_version()
+{
+  int schema_version = -1;
+  if (database_->table_exists("schema")) {
+    // Read schema version
+    auto statement = database_->prepare_statement("SELECT schema_version from schema;");
+    auto query_results = statement->execute_query<int>();
+    schema_version = std::get<0>(*query_results.begin());
+  } else {
+    if (database_->field_exists("topics", "offered_qos_profiles")) {
+      schema_version = 2;
+    } else {
+      schema_version = 1;
+    }
+  }
+
+  return schema_version;
 }
 
 }  // namespace rosbag2_storage_plugins
--- a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
+++ b/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
@@ -27,6 +27,7 @@
 #include <utility>
 #include <vector>
 
+#include "rcpputils/env.hpp"
 #include "rcpputils/filesystem_helper.hpp"
 
 #include "rosbag2_storage/metadata_io.hpp"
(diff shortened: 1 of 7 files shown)


MichaelOrlov:
@clalancette Could you please review this PR for **ABI** and **API** stability to be able to backport this PR to Humble.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -368,21 +372,41 @@ uint64_t SqliteStorage::get_bagfile_size() const
 
 void SqliteStorage::initialize()
 {
-  std::string create_stmt = "CREATE TABLE topics(" \
+  std::string create_stmt = "CREATE TABLE schema(" \
+    "schema_version INTEGER PRIMARY KEY," \
+    "ros_distro TEXT NOT NULL);";
```
amacneil, line 377: Why not store these two keys in the metadata table?

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -368,21 +372,41 @@ uint64_t SqliteStorage::get_bagfile_size() const
...
+  database_->prepare_statement(create_stmt)->execute_and_reset();
+
+  create_stmt = "CREATE TABLE metadata(" \
+    "id INTEGER PRIMARY KEY," \
+    "metadata_version INTEGER NOT NULL," \
+    "metadata TEXT NOT NULL);";
```
amacneil, line 383: What are the rows of this table intended to be? What would `id INTEGER` be set to?

Since the database schema should be changed as infrequently as possible, it seems more "future proof" to have a general purpose metadata table, like:
| `key TEXT` | `value TEXT` |
| -- | -- |
| schema_version | 1 |
| ros_distro | GALACTIC |
| duration | 11250 |
| compression_format | |
| compression_mode | |

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: There are multiple reasons to not store these two keys in the metadata table.
1. To not make a mess and not store everything in one table.
2.  Schema version should be first what should be read from table and upon of it value we could make a decision what and in how to read. For instance even in current implementation we have a logic that if `schema_version < 3` we will not even try to read metadata since on those versions metadata table doesn't exists.
3. According to the `update_metadata` API https://github.com/ros2/rosbag2/pull/1149. We will call `update_metadata` twice one time right after opening storage and second time just before closing bag. Since currently we don't have update API for fields in DB perhaps we will store 2 records for metadata and store another fields repeatedly looks odd.
4. To be able to propagate changes related for storing `ROS_DISTRO` name to the Humble. If we would store it inside metadata table we would likely need to make changes which is breaking ABI and API from #1149. Note that we don't have valid instance of metadata object during bag initialization and `storage::open()` call.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: As regards to the question about "What are the rows of this table intended to be? What would id INTEGER be set to?"
See my explanation from previous comment in item 3

> 3. According to the update_metadata API https://github.com/ros2/rosbag2/pull/1149. We will call update_metadata twice one time right after opening storage and second time just before closing bag. Since currently we don't have update API for fields in DB perhaps we will store 2 records for metadata and store another fields repeatedly looks odd.

- Hence `id` will be a sequence number for metadata records. I expect that it will be at least 2 records of them. And when we will open storage for read we are going to read the latest one.
- `metadata_version` will be needed to be able to know how to parse serialized metadata object stored in string. According to the design proposal in #1109 we are going to serialize metadata to string in the same form as we are preparing it to store in `metadata.yaml` file. But who knows, we might change this approach in future or have some specific for deserialization based on the metadata version.

emersonknapp: I kind of like the key->value approach as well, but I also don't care at the end of the day specifically at this level how it's stored. As long as the rosbag2-side APIs are clean, to me this is an implementation detail of this particular storage format.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
emersonknapp: `ROS_DISTRO` does seem like an appropriate addition to `BagMetadata`, to me. The `schema_version` on the other hand is unique to the storage implementation and so shouldn't be part of generic metadata, so it makes sense for it to live somewhere else.

emersonknapp (approved):
I'm going to approve this as fine, I don't think we need to go back and forth too much on this part of it since it doesn't affect how other storage implementations have to do anything, there are no core APIs.

I have noted in the comment stream that I agree with Adrian that I might approach it a little differently, but those opinions are non-blocking for me.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/43376b9180367dfdbc6f83ac9e0b571f/raw/9de6d51acca2cd167442395b576ed59daa25d077/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_storage rosbag2_storage_sqlite3 rosbag2_tests
TEST args:  --packages-above rosbag2_storage rosbag2_storage_sqlite3 rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11164
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17689/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12235/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18280/)

MichaelOrlov:

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2022-12-15/28840/1

Issue #794, opened 2021-06-25
relative_file_paths issue when replaying compressed bag file


nedo99:
## Description
When replaying the bag file, recorded with compression mode zstd, one must hardcode folder path plus file under `relative_file_paths` parameter in `metadata.yaml`.

### Expected Behavior
The `relative_file_paths` parameter should contain only the name of the bag file by default, the same as when one records the bag file in normal mode. ROS bag playing behavior should be the same for both compressed mode and non-compressed mode.

### Actual Behavior
When recording with compression mode, `relative_file_paths` by default contains the folder name plus the file name (e.g. `rosbag2_2021_06_25-11_59_52_0/rosbag2_2021_06_25-11_59_52_0.db3.zstd`). If we rename the recorded folder, playing of the bag file will not happen as the folder name is hardcoded (one will get errno(2)). This limits execution folder location since every time `metadata.yaml` file needs to be updated.

## To Reproduce
1. Record a bag file: `ros2 bag record --compression-mode file --compression-format zstd /ros/topic`
2. Rename recorded folder or execute `ros2 bag play <path/to/bag/folder>` from non-relative folder
3. You will get message
```
[ERROR] [1624628638.536812180] [rosbag2_transport]: Failed to play: Failed to open file: "<path>/rosbag2_2021_06_25-11_59_52_0.db3.zstd" for binary reading! errno(2)
```

## System (please complete the following information)
 - OS: Ubuntu 20.04
 - ROS 2 Distro: Foxy
 - Version: 0.3.7-1focal.20210415.042502 (ros-foxy-ros2bag)

llapx:
It works well on rolling and galactic, maybe fixed in somewhere.

DLu:
Note: This is fixed in `foxy-future`, `galactic`, `humble` and `rolling` since https://github.com/ros2/rosbag2/pull/612

I opened #1161 to start to address the issue in `foxy`, and I have a fix for it, but I'm not sure what the intent is w.r.t. `foxy-future`. Should I still make the fix for this in a non-api breaking way?

MichaelOrlov:
- Closing as fixed in the #345

PR #1126, opened 2022-10-15
Fix for ros2 bag play exit with non-zero code on SIGINT

MichaelOrlov:
Fix for ros2 bag play exit with non-zero code on SIGINT
- Closes #1123
- Don't allow to propagate `KeyboardInterrupt` exception on upper level from python wrapper for `play` verb.
- Fixed incorrect error checks in process_execution infrastructure when `SIGINT` not processed in child process.

Issue #1123, opened 2022-10-13
`ros2 bag play` exiting with code 2 by SIGINT or CTRL+C


MichaelOrlov:
## Description
`ros2 bag play` exiting with code 2 by `SIGINT` or `CTRL+C`

### Expected Behavior
`ros2 bag play` shall exiting with code 0 by `SIGINT` or `CTRL+C`

### Actual Behavior
`ros2 bag play` exiting with code 2 by `SIGINT` or `CTRL+C`

## To Reproduce
** Steps to reproduce the behavior, e.g.
1. Go to 'ros2_rolling' cloned folder
2. `source ./install/setup.bash`
3. Run tmux
4. In tmux panel 1 run commands: 
```console
ros2 bag play --loop src/ros2/rosbag2/rosbag2_tests/resources/cdr_test
```
5. In tmux panel 2 run command
```console
ps -aux | grep "ros2 bag play"
```
6. Find `PID` for `ros2 bag play`. For instance `195930`
7. In tmux panel 2 run command
```console
kill -s 2 195930
```
or alternatively switch to the tmux panel 1 and pres `CTRL+C`
8. In tmux panel 1 run command
```console
echo $?
```
`echo $?` will output exit code from last finished process.

## System (please complete the following information)
 - OS: Ubuntu 22.04 LTS
 - ROS 2 Distro: Rolling
 - Version: rolling branch

## RCA
1. rosbag2 player successfully and correctly finishing after processing `SIGINT` signal.
2. The problem in python part of the `ros2 bag play` verb.

## Proposed solution 
Don't propagate `KeyboardInterrupt` exception from python code  in 
https://github.com/ros2/rosbag2/blob/5b8b658b1f70a9f6184e4f6f31a42d50267f0b47/ros2bag/ros2bag/verb/play.py#L210
use instead following try catch block.
```
 try:
     player.play(storage_options, play_options)
        except KeyboardInterrupt:
     pass
```

PR #1161, opened 2022-11-15
Failing test for #794 foxy

DLu:
This hardly seems like the best way to perform the test (I would have expected a non-success return code) but it works. Open to feedback.


MichaelOrlov:
@DLu Most probably for the same reason as it was passing for `SIG_INT` in #1123 i.e. incorrectly handling `WEXITSTATUS(exitcode)`
I fixed it for `start_execution()` and `stop_execution` in https://github.com/ros2/rosbag2/pull/1126 but it seems missed for `execute_and_wait_until_completion(const std::string & command, const std::string & path)`

MichaelOrlov:
- Closing as wouldn't be done since Foxy reached its EOL.

PR #1153, opened 2022-11-08
Add option to specify a message type

carlossvg:
This PR adds an option to the Performance benchmarking  package to use a predefined message type instead of generic type. The motivation for this option is to allow to replicate scenarios with exactly the same message types as the original case.

Issue #688, opened 2021-03-22
Benchmark is performed incorrectly while transport is enabled and the count of publisher is big (such as 200).


Barry-Xu-2018:
## Description
Benchmark is performed incorrectly while transport is enabled and the count of publisher is big (such as 200).

### Expected Behavior
Benchmark is performed correctly while transport is enabled and the count of publisher is big (such as 200).

### Actual Behavior
Benchmark is performed incorrectly 

Get below error:  
```
[benchmark_publishers-2] 1616403871.076306 [] benchmark_: create_thread: benchmark_publi: no free slot
```
And execution is terminated.

## To Reproduce
1. benchmark config file --- 'test.yaml'
    ```yaml
   rosbag2_performance_benchmarking:
   benchmark_node:
    ros__parameters:
      benchmark:
        summary_result_file:  "results.csv"
        db_root_folder:       "rosbag2_performance_test_results"
        repeat_each:          1     # How many times to run each configurations (to average results)
        no_transport:         False  # Whether to run storage-only or end-to-end (including transport) benchmark
        preserve_bags:        False # Whether to leave bag files after experiment (and between runs). Some configurations can take lots of space!
        parameters:                 # Each combination of parameters in this section will be benchmarked
          max_cache_size:      [100000000]
          max_bag_size:         [0]
          compression:            [""]
          compression_queue_size: [1]
          compression_threads:    [0]
          storage_config_file:    [""] 
    ```
2. producers config file  --- '10k_300inst_100hz.yaml'
    ```yaml
   rosbag2_performance_benchmarking_node:
   ros__parameters:
    publishers: # publisher_groups parameter needs to include all the subsequent groups
      publisher_groups: [ "10k_300inst" ] 
      wait_for_subscriptions: True
      10k_300inst:
        publishers_count:   300
        topic_root:         "benchmarking_10k_300inst"
        msg_size_bytes:     10000
        msg_count_each:     2000
        rate_hz:            100
    ```

3. Perform benchmark
    ```shell
    $ ros2 launch rosbag2_performance_benchmarking benchmark_launch.py benchmark:=/PATH/TO/test.yaml producers:=/PATH/TO/10k_300inst_100hz.ymal
     ...
     [ros2-1] [INFO] [1616403871.061948147] [rosbag2_transport]: Subscribed to topic '/rosbag2_performance_benchmarking_node/benchmarking_10k_300inst_165'
[...]

PR #1171, opened 2022-11-21
Use thread pool to run benchmark publishers

carlossvg:
Related to: https://github.com/ros2/rosbag2/issues/688

Depends on https://github.com/ros2/rosbag2/pull/1153

This PR changes `benchmark_publishers` to use a thread pool instead of running each publisher in separate threads. The motivation for this change is to overcome a limitation on the number of threads imposed by Cyclone DDS (limited to 128 threads).

### DIFF
--- a/rosbag2_compression/CMakeLists.txt
+++ b/rosbag2_compression/CMakeLists.txt
@@ -37,12 +37,13 @@ target_include_directories(${PROJECT_NAME}
   PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/${PROJECT_NAME}>)
-ament_target_dependencies(${PROJECT_NAME}
-  pluginlib
-  rcpputils
-  rcutils
-  rosbag2_cpp
-  rosbag2_storage)
+target_link_libraries(${PROJECT_NAME}
+  pluginlib::pluginlib
+  rcpputils::rcpputils
+  rcutils::rcutils
+  rosbag2_cpp::rosbag2_cpp
+  rosbag2_storage::rosbag2_storage
+)
 target_compile_definitions(${PROJECT_NAME} PRIVATE ROSBAG2_COMPRESSION_BUILDING_DLL)
 
 install(
@@ -94,13 +95,19 @@ if(BUILD_TESTING)
 
   ament_add_gmock(test_sequential_compression_reader
     test/rosbag2_compression/test_sequential_compression_reader.cpp)
-  target_link_libraries(test_sequential_compression_reader ${PROJECT_NAME})
-  ament_target_dependencies(test_sequential_compression_reader rosbag2_cpp rosbag2_storage)
+  target_link_libraries(test_sequential_compression_reader
+    ${PROJECT_NAME}
+    rosbag2_cpp::rosbag2_cpp
+    rosbag2_storage::rosbag2_storage
+  )
 
   ament_add_gmock(test_sequential_compression_writer
     test/rosbag2_compression/test_sequential_compression_writer.cpp)
-  target_link_libraries(test_sequential_compression_writer ${PROJECT_NAME})
-  ament_target_dependencies(test_sequential_compression_writer rosbag2_cpp rosbag2_storage)
+  target_link_libraries(test_sequential_compression_writer
+    ${PROJECT_NAME}
+    rosbag2_cpp::rosbag2_cpp
+    rosbag2_storage::rosbag2_storage
+  )
 endif()
 
 ament_package()
--- a/rosbag2_compression_zstd/CMakeLists.txt
+++ b/rosbag2_compression_zstd/CMakeLists.txt
@@ -35,10 +35,11 @@ target_include_directories(${PROJECT_NAME}
   PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/${PROJECT_NAME}>)
-ament_target_dependencies(${PROJECT_NAME}
-  rcpputils
-  rosbag2_compression
-  zstd)
+target_link_libraries(${PROJECT_NAME}
+  rcpputils::rcpputils
+  rosbag2_compression::rosbag2_compression
+  zstd::zstd
+)
 target_compile_definitions(${PROJECT_NAME} PRIVATE ROSBAG2_COMPRESSION_ZSTD_BUILDING_DLL)
 pluginlib_export_plugin_description_file(rosbag2_compression plugin_description.xml)
 
@@ -73,8 +74,11 @@ if(BUILD_TESTING)
 
   ament_add_gmock(test_zstd_compressor
     test/rosbag2_compression_zstd/test_zstd_compressor.cpp)
-  target_link_libraries(test_zstd_compressor ${PROJECT_NAME})
-  ament_target_dependencies(test_zstd_compressor rclcpp rosbag2_test_common)
+  target_link_libraries(test_zstd_compressor
+    ${PROJECT_NAME}
+    rclcpp::rclcpp
+    rosbag2_test_common::rosbag2_test_common
+  )
 endif()
 
 ament_package()
--- a/rosbag2_cpp/CMakeLists.txt
+++ b/rosbag2_cpp/CMakeLists.txt
@@ -72,20 +72,20 @@ add_library(${PROJECT_NAME} SHARED
   src/rosbag2_cpp/writers/sequential_writer.cpp
   src/rosbag2_cpp/reindexer.cpp)
 
-ament_target_dependencies(${PROJECT_NAME}
+target_link_libraries(${PROJECT_NAME}
   PUBLIC
-  ament_index_cpp
-  pluginlib
-  rclcpp
-  rcpputils
-  rcutils
-  rmw
-  rmw_implementation
-  rosbag2_storage
-  rosidl_runtime_c
-  rosidl_runtime_cpp
-  rosidl_typesupport_cpp
-  rosidl_typesupport_introspection_cpp
+  ament_index_cpp::ament_index_cpp
+  pluginlib::pluginlib
+  rclcpp::rclcpp
+  rcpputils::rcpputils
+  rcutils::rcutils
+  rmw::rmw
+  rmw_implementation::rmw_implementation
+  rosbag2_storage::rosbag2_storage
+  rosidl_runtime_c::rosidl_runtime_c
+  rosidl_runtime_cpp::rosidl_runtime_cpp
+  rosidl_typesupport_cpp::rosidl_typesupport_cpp
+  rosidl_typesupport_introspection_cpp::rosidl_typesupport_introspection_cpp
 )
 
 target_include_directories(${PROJECT_NAME}
@@ -139,8 +139,10 @@ if(BUILD_TESTING)
     SHARED
     test/rosbag2_cpp/serializer_test_plugin.cpp
     test/rosbag2_cpp/converter_test_plugin.cpp)
-  ament_target_dependencies(converter_test_plugins rosbag2_storage)
-  target_link_libraries(converter_test_plugins ${PROJECT_NAME})
+  target_link_libraries(converter_test_plugins
+    ${PROJECT_NAME}
+    rosbag2_storage::rosbag2_storage
+  )
   install(
     TARGETS converter_test_plugins
     ARCHIVE DESTINATION lib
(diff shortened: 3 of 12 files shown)


MichaelOrlov:
cc: @james-rms

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -20,79 +20,118 @@
...
-
-    create_benchmark_publishers_and_producers();
+    create_benchmark_producers();
+    wait_for_subscriptions();
   }
 
   void run()
   {
-    std::vector<std::thread> producer_threads;
+    thread_pool_.start(std::thread::hardware_concurrency());
```
james-rms, line 43: This is a hidden variable within the benchmark apparatus which is not easily inspected or controlled for.

james-rms:
This PR adds a source of confusion - previously the number of threads was controlled by the producers configuration, now it's controlled by `hardware_concurrency`. From the description it's not clear why this is a neccessary feature.

Why run so many publishers, rather than publishing more messages from each individual publisher?

carlossvg:
Thanks for the review. 

> This PR adds a source of confusion - previously the number of threads was controlled by the producers configuration, now it's controlled by `hardware_concurrency`. From the description it's not clear why this is a neccessary feature.

We could expose this parameter to make it configurable. For the default value I see tow options:
1. `hardware_concurrency`: this is the same option used by  [`MultiThreadedExecutor`](https://github.com/ros2/rclcpp/blob/85c0af4fa06264a6b92a519197774c0e10b9606e/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L40)
2. Number of thread equal to the number of publishers: This would be similar to what we have now

> Why run so many publishers, rather than publishing more messages from each individual publisher?

The motivation for us is to be able to benchmark exactly a specific real setup. Publishing more messages from each individual publisher would be a different scenario in terms of resources, queues, concurrency, etc.

MichaelOrlov:
@carlossvg Could you please reiterate on this PR and rebase it on top of latest rolling?

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -87,19 +85,29 @@ class BenchmarkPublishers : public rclcpp::Node
     }
 
     const std::string node_name(get_fully_qualified_name());
+    const auto initial_time = std::chrono::high_resolution_clock::now() + std::chrono::seconds(1);
 
+    size_t total_producer_number = 0U;
```
MichaelOrlov, line 90:
```suggestion
    size_t total_producers_number = 0U;
```

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -109,20 +117,34 @@ class BenchmarkPublishers : public rclcpp::Node
     producer->period = std::chrono::milliseconds(
       producer_config.frequency ? 1000 / producer_config.frequency : 1);
 
+    thread_pool_.queue(
```
MichaelOrlov, line 120: It would be nice to check that `producer->max_messages > 0` before adding producer_job to the thread pool queue.

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -109,20 +117,34 @@ class BenchmarkPublishers : public rclcpp::Node
...
+    std::shared_ptr<BenchmarkProducer> producer)
   {
-    for (auto i = 0u; i < producer->max_messages; ++i) {
-      if (!rclcpp::ok()) {
-        break;
-      }
-      std::this_thread::sleep_for(producer->period);
-      producer->produce();
+    std::this_thread::sleep_until(when);
+    producer->produce();
```
MichaelOrlov, line 133: I think we need to add check for 
```
      if (!rclcpp::ok()) {
        producer->promise_finished.set_value();
        return;
      }
```
Before calling `producer->produce();` or at least before `std::this_thread::sleep_until(when);`

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -109,20 +117,34 @@ class BenchmarkPublishers : public rclcpp::Node
...
+
+    if (producer->produced_messages < producer->max_messages) {
+      thread_pool_.queue(
+        [this, next = std::chrono::high_resolution_clock::now() + producer->period,
+        producer] {
+          producer_job(next, producer);
+        });
```
MichaelOrlov, line 140:
```suggestion
        [this, next_timestamp = when + producer->period, producer] {
          producer_job(next_timestamp, producer);
        });
```
This is interesting finding! 
Producer period means interval between messages. Call for the `producer->produce();` could take some substantial amount of time which we shouldn't count when calculating next timestamp.

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
```diff
@@ -87,19 +85,29 @@ class BenchmarkPublishers : public rclcpp::Node
     }
 
     const std::string node_name(get_fully_qualified_name());
+    const auto initial_time = std::chrono::high_resolution_clock::now() + std::chrono::seconds(1);
```
MichaelOrlov, line 88: @carlossvg I would advise to rename `initial_time` to `when_to_start` for clarity

MichaelOrlov (changes requested):
@carlossvg Thanks for the PR and rebasing.  I like the idea with tread pool.  
However I have a number of suggestions and findings for improvements.

Side note: It was a conflicts in CMake file, I've tried to resolve it but git somehow managed to add one commit from baseline `Rolling` branch above your changes.  Sorry for mess.
Feel free to rebase.

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
carlossvg: You are right, we can do it as you suggests, use an initial time and increase period each time. One relevant change is that now if a job last longer than one period the next time will be in the past and there will be no wait. This is probably better than the previous behavior but I wonder if we should warn the user this is happening.

rosbag2_performance/rosbag2_performance_benchmarking/include/rosbag2_performance_benchmarking/thread_pool.hpp
```diff
@@ -0,0 +1,95 @@
...
+        job = jobs_.front();
+        jobs_.pop();
+      }
+      job();
+    }
+  }
+
+  bool terminate_ = false;
+  std::mutex mutex_;
+  std::queue<job_type> jobs_;
```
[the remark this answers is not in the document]
carlossvg: done

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
carlossvg: done

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
carlossvg: done

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
carlossvg: done

carlossvg:
@MichaelOrlov I think there are a couple of things left to do:
1. Document the new added option `number_of_threads`
2. Updated `writer_benchmark` to use the thread pool so the behavior is consistent with `benchmark_publishers`
3. Compare benchmark results before and after the changes to double-check there are no big differences

MichaelOrlov:
@carlossvg As regards to the 
> Updated writer_benchmark to use the thread pool so the behavior is consistent with benchmark_publishers

I have been thinking about it yesterday.  Although come to the conclusion that it is not necessary and would be useless by design.  
Because:
1. We can configure number of writers in config file and there are no differences in behavior if it would be less threads or less writers in config file. It make sense to have current approach when always one thread per one writer to it's own queue.
2. Allowing to have less threads than number of writers will not give us any advantages in testing or simulation. It will introduce some uncertainty where we want to avoid it. We usually  using `writer_benchmark` scenario when we need to measure if rosbag2 can keep up in speed with some data stream but without transport layer. Reducing number of threads will impose uncertainty in this case for the volume of the data written to the rosbag2 backend per seconds. Which we are trying to strictly define via number of writers and message size with  frame rate in config file.

rosbag2_performance/rosbag2_performance_benchmarking/src/benchmark_publishers.cpp
MichaelOrlov: I think this is better approach since it is more close to the real use case.
In reality most system will behave this way since also have some internal buffering on upper layers.
If we will keep it as is we can't guaranty that if we define message rate 100 frames per second that it will be published exact 100 frames in one second. Most probably it will be published 99 or less.

rosbag2_performance/rosbag2_performance_benchmarking/include/rosbag2_performance_benchmarking/thread_pool.hpp
```diff
@@ -26,7 +26,10 @@ class ThreadPool
 {
 public:
   using job_type = std::function<void ()>;
-
+  ~ThreadPool()
+  {
+    this->terminate();
+  }
   void start(size_t size)
```
MichaelOrlov, line 33: Please add empty line after closing `}`

MichaelOrlov (changes requested):
@carlossvg Semantically PR looks good for me. 
Although I tried to test performance benchmarking with your branch and found a few issues.
Could you please fix them?
1. Need remove  `std::this_thread::sleep_for(1ms);` in `void WriterBenchmark::start_benchmark()` it basically usles and could affect reproduction rate for figh frequency topics.
2. Need change topic.type from ` topic.type = "std_msgs::msgs::ByteMultiArray";` to the "rosbag2_performance_benchmarking_msgs/msg/ByteArray";  The `writer_benchmark` crashes without this fix if start benchmarking with `no_transport = True`.
3. It would be nice to add `number_of_threads: 16` or 50 to some of the producers config files to show users in what exact place it's expected to be.

carlossvg:
> 1. Need remove  `std::this_thread::sleep_for(1ms);` in `void WriterBenchmark::start_benchmark()` it basically usles and could affect reproduction rate for figh frequency topics.

If we do that the loop becomes a busy wait loop while waiting on new messages, right?

As pointed out by this comment the proper fix would be to use a conditional variable to implement the wait.

>// TODO(adamdbrw) Performance can be improved. Use conditional variables

I would implement that change in a separate MR.

>2. Need change topic.type from  topic.type = "std_msgs::msgs::ByteMultiArray"; to the >"rosbag2_performance_benchmarking_msgs/msg/ByteArray"; The writer_benchmark crashes without this fix if start >benchmarking with no_transport = True.

Good catch, fixed

>3. It would be nice to add number_of_threads: 16 or 50 to some of the producers config files to show users in what exact place it's expected to be.

I order to make more explicit this parameter exists I added `# number_of_threads: 16` commented in all producer config files. If we prefer not to include them as a comment we can create a new config file with this option enabled.

rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp
```diff
@@ -188,7 +193,7 @@ void WriterBenchmark::create_writer()
     rosbag2_storage::TopicMetadata topic;
     topic.name = queue->topic_name();
     // TODO(adamdbrw) - replace with something more general if needed
-    topic.type = "std_msgs::msgs::ByteMultiArray";
+    topic.type = "rosbag2_performance_benchmarking_msgs::msgs::ByteArray";
```
MichaelOrlov: @carlossvg Delimiters with double colon doesn't work.
It is expected to have `/`  slash as delimiter.  
i.e. **"rosbag2_performance_benchmarking_msgs/msg/ByteArray"**

MichaelOrlov (changes requested):
@carlossvg We can fix issue with delay in separate PR - I agree.
However delimiters in topic type is expected to be with backslashes instead of the `::`

carlossvg:
> 1. Need remove  `std::this_thread::sleep_for(1ms);` in `void WriterBenchmark::start_benchmark()` it basically usles and could affect reproduction rate for figh frequency topics.

If we do that the loop becomes a busy wait loop while waiting on new messages, right?

As pointed out by this comment the proper fix would be to use a conditional variable to implement the wait.

>// TODO(adamdbrw) Performance can be improved. Use conditional variables

I would implement that change in a separate MR.

>2. Need change topic.type from  topic.type = "std_msgs::msgs::ByteMultiArray"; to the >"rosbag2_performance_benchmarking_msgs/msg/ByteArray"; The writer_benchmark crashes without this fix if start >benchmarking with no_transport = True.

Good catch, fixed

>3. It would be nice to add number_of_threads: 16 or 50 to some of the producers config files to show users in what exact place it's expected to be.

I order to make more explicit this parameter exists I added `# number_of_threads: 16` commented in all producer config files. If we prefer not to include them as a comment we can create a new config file with this option enabled.

rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp
carlossvg: done

MichaelOrlov (approved):

MichaelOrlov:
@carlossvg Could you please rebase your branch? There are merge conflicts.

MichaelOrlov:
@carlossvg It seems something went wrong with your rebase.  
I see **zero** commits on your branch and no delta was merged.
Also please note that before merging to the parent `ros2::rolling` we need to run CI to make sure that we will not break anything.
I found reference to your latest commit in remote git log https://github.com/carlossvg/rosbag2/tree/eb1fd354fc3d4c5a484fb6891e1a7be50448a2b7 
Although I am not able to check it out or clone it.  
May be you will be able to do it on your local git repo by doing `git checkout eb1fd354fc3d4c5a484fb6891e1a7be50448a2b7` or `git reset --hard eb1fd354fc3d4c5a484fb6891e1a7be50448a2b7`

carlossvg:
@MichaelOrlov It looks like the PR was merged automatically when syncing the fork. I created a new PR here https://github.com/ros2/rosbag2/pull/1250

PR #1173, opened 2022-11-21
Remove storage_evaluation directory

emersonknapp:
As discussed in WG meeting.
The directory `rosbag2_storage_evaluation` is an artifact from the original exploration leading to the current rosbag2 code. It is explicit evaluations around SQLite3's performance, and was written before any of the rosbag2 APIs. It is no longer useful, especially as we explore other storage formats. The `rosbag2_performance` directory contains tests that go through rosbag2 APIs and can therefore evaluate different plugins.


MichaelOrlov (approved):
LGTM

PR #1177, opened 2022-11-23
set_read_order: return success

james-rms:
This PR makes these changes:

1. Notes to client code that `set_read_order` should only be called after `open()`. There are cases where a storage plugin actually needs to read the underlying bag file in order to determine whether a given read order is supported. For example, if a timestamp index is required to read in timestamp order, the storage plugin needs to know whether that is present before accepting a request to read in timestamp order.
1. Throw an exception from within `SequentialReader` if read order is set before opening the underlying storage.
1. changes `set_read_order` to return a boolean for success or failure. This makes the flow for client code attempting to set a read order more ergonomic. Previously a client would have to check for a `std::runtime_error` and possibly inspect the exception text to ensure that the exception is about whether the read order is supported, rather than something unrelated.

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -57,7 +57,7 @@ std::vector<std::string> resolve_relative_paths(
 }
 }  // namespace details
 
-const static rosbag2_storage::ReadOrder kFallbackOrder(rosbag2_storage::ReadOrder::SortBy::File,
+static const rosbag2_storage::ReadOrder kFallbackOrder(rosbag2_storage::ReadOrder::SortBy::File,
   false);
 
 SequentialReader::SequentialReader(
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -57,6 +57,9 @@ std::vector<std::string> resolve_relative_paths(
 }
 }  // namespace details
 
+const static rosbag2_storage::ReadOrder kFallbackOrder(rosbag2_storage::ReadOrder::SortBy::File,
+  false);
+
 SequentialReader::SequentialReader(
   std::unique_ptr<rosbag2_storage::StorageFactoryInterface> storage_factory,
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory,
@@ -108,7 +111,10 @@ void SequentialReader::open(
     }
     if (!set_read_order(read_order_)) {
       ROSBAG2_CPP_LOG_WARN(
-        "Could not set read order on open(), defaulting to storage plugin's default read order");
+        "Could not set read order on open(), falling back to file order");
+      if (!set_read_order(kFallbackOrder)) {
+        throw std::runtime_error("Could not set read order on open()");
+      }
     }
     metadata_ = storage_->get_metadata();
     if (metadata_.relative_file_paths.empty()) {
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -106,7 +106,10 @@ void SequentialReader::open(
     if (!storage_) {
       throw std::runtime_error{"No storage could be initialized from the inputs."};
     }
-    storage_->set_read_order(read_order_);
+    if (!set_read_order(read_order_)) {
+      ROSBAG2_CPP_LOG_WARN(
+        "Could not set read order on open(), defaulting to storage plugin's default read order");
+    }
     metadata_ = storage_->get_metadata();
     if (metadata_.relative_file_paths.empty()) {
       ROSBAG2_CPP_LOG_WARN("No file paths were found in metadata.");
@@ -129,12 +132,13 @@ void SequentialReader::open(
     topics[0].topic_metadata.serialization_format);
 }
 
-void SequentialReader::set_read_order(const rosbag2_storage::ReadOrder & order)
+bool SequentialReader::set_read_order(const rosbag2_storage::ReadOrder & order)
 {
-  if (storage_) {
-    storage_->set_read_order(order);
+  if (!storage_) {
+    throw std::runtime_error("read order can only be set after open()");
   }
   read_order_ = order;
+  return storage_->set_read_order(read_order_);
 }
 
 bool SequentialReader::has_next()
--- a/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_read_interface.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_read_interface.hpp
@@ -62,8 +62,8 @@ public:
   ///   This affects the outcome of has_next and read_next.
   ///   Note that when setting to reverse order, this will not change the read head, so user
   ///   must first seek() to the end in order to read messages from the end.
-  ///   Also note that set_read_order should only be called after open() has been called.
   /// @param read_order The order in which to return messages.
+  /// @throws runtime_error if the reader is not open.
   /// @return true if the requested read order has been successfully set.
   virtual bool set_read_order(const ReadOrder & read_order) = 0;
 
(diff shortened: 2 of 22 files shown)


rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_read_interface.hpp
```diff
@@ -62,8 +62,10 @@ class ROSBAG2_STORAGE_PUBLIC BaseReadInterface
   ///   This affects the outcome of has_next and read_next.
   ///   Note that when setting to reverse order, this will not change the read head, so user
   ///   must first seek() to the end in order to read messages from the end.
+  ///   Also note that set_read_order should only be called after open() has been called.
```
emersonknapp, line 65: Maybe specify that `set_read_order` should always return false before `open`?

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -320,17 +320,18 @@ void SqliteStorage::write(
   commit_transaction();
 }
 
-void SqliteStorage::set_read_order(const rosbag2_storage::ReadOrder & read_order)
+bool SqliteStorage::set_read_order(const rosbag2_storage::ReadOrder & read_order)
 {
   if (read_order.sort_by == rosbag2_storage::ReadOrder::PublishedTimestamp) {
-    throw std::runtime_error("Not Implemented - PublishedTimestamp read order.");
```
emersonknapp, line 326: Maybe worth logging warnings or at least debug?

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -129,12 +132,13 @@ void SequentialReader::open(
     topics[0].topic_metadata.serialization_format);
 }
 
-void SequentialReader::set_read_order(const rosbag2_storage::ReadOrder & order)
+bool SequentialReader::set_read_order(const rosbag2_storage::ReadOrder & order)
 {
-  if (storage_) {
-    storage_->set_read_order(order);
+  if (!storage_) {
```
emersonknapp, line 137: One might expect this to `return false` given the new API? I think that'll work OK - you have to set_read_order after open on the `Reader`s as well.

emersonknapp (changes requested):

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
emersonknapp: log debug? I'm pretty up on log statements in general - in my experience from having to debug complex (dozens of processes) robot applications on devices running in the field, detailed text logs are often the only way to even get a sense of what happened on somebody else's thing after the fact, especially if it's not 100% reproducible. That's a much deeper general software engineering philosophy and my preference is based on my specific experiences/projects, but regardless, a debug log doesn't hurt especially if it uses macros that can be compiled out.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_read_interface.hpp
james-rms: i think that might mislead callers - there's a difference between "this read order is not supported" and "you have called these functions in the wrong order, go change your code".

emersonknapp: So `if !open: throw; elif open && unsupported: return false`? 

I just want the exact behavior expectation to be specified here. If ^ then should add `/// \throws` to docstring

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
james-rms: If `storage_` is null at this point that indicates that open hasn't been called, which is programmer error rather than something we can derive read order support from.

emersonknapp (approved):

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/e5759b670fc0136412f5d739932ca559/raw/aea5e00d3f23132b22e229fe8ebed53517a0969e/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_storage rosbag2_storage_sqlite3 rosbag2_transport
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_storage rosbag2_storage_sqlite3 rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11172
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17698/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12243/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18302/)

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -106,7 +106,10 @@ void SequentialReader::open(
     if (!storage_) {
       throw std::runtime_error{"No storage could be initialized from the inputs."};
     }
-    storage_->set_read_order(read_order_);
+    if (!set_read_order(read_order_)) {
+      ROSBAG2_CPP_LOG_WARN(
+        "Could not set read order on open(), defaulting to storage plugin's default read order");
+    }
```
MichaelOrlov, line 112: @james-rms @emersonknapp I believe we shall throw exception here.
We can't fall back on the storage plugin's default read order since we don't know what is default for it.
We need to be assure that `read_order_` in `SequantialReader` corresponds to the read order in storage plugin.
If we can't be assure - this is undefined behavior and we shall not proceed - in this case.

MichaelOrlov (changes requested):
I have a concern about not throwing exception in open if we fail to setup read order.

MichaelOrlov (approved):
Thank you @james-rms. Now LGTM if CI will pass.

james-rms:
@MichaelOrlov I believe the failures in https://build.ros2.org/job/Rpr__rosbag2__ubuntu_jammy_amd64/431/ are not related to this PR - would you be able to kick off a full CI run on the build farm?

MichaelOrlov:
[@ros-pull-request-builder](https://github.com/ros-pull-request-builder) retest this please

MichaelOrlov:
Ok re-running full CI since Rpr passed
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11187
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17716/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12262/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18338/)

MichaelOrlov:
It seems Windows CI build is broken on build farm. I have the same issue in another PR https://github.com/ros2/rosbag2/pull/1175#issuecomment-1329993811
@clalancette Is it known issue?

clalancette:
> @clalancette Is it known issue?

This time of year, the AWS spot prices fluctuate wildly, so our workers sometimes get unceremoniously yanked from underneath us.  Just try again; it will eventually work.

MichaelOrlov:
Re-running Windows build
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18354/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/db3e068b715188676e7d53c6ee250379/raw/aea5e00d3f23132b22e229fe8ebed53517a0969e/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_tests
TEST args:  --packages-above rosbag2_cpp rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11192
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17724/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12267/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18358/)

PR #1163, opened 2022-11-17
rosbag2_storage_mcap: merge into rosbag2 repo

james-rms:
Merges `rosbag2_storage_mcap` into the `rosbag2` monorepo.

This PR merges the entire `main` branch of rosbag2_storage_mcap into `rolling`. Once this is approved, subsequent PRs into `galactic`, `foxy`, and `humble` will follow with the exact same code. The `rosbag2_storage_mcap` codebase is designed to be compatible with all active distributions, and we intend to keep the copies of the code in each release branch exactly the same for as long as possible.

Fixes https://github.com/ros-tooling/rosbag2_storage_mcap/issues/63

PR #1190, opened 2022-11-29
[backport, galactic] rosbag2_storage_mcap: merge into rosbag2 repo (#1163)

james-rms:


clalancette:
I'm not totally convinced that we should do this for the older distributions.

It makes maintenance easier, but because it is not the default in the older distributions (and never will be), it seems like this will just make the situation murkier for those who want to use it.

That said, I don't have any fundamental objection here, so if the maintainers think this is better, go ahead with it.

james-rms:
I think the code for the humble distribution of `rosbag2_storage_mcap` should live somewhere, and now https://github.com/ros-tooling/rosbag2_storage_mcap is archived :) I think it's best for the code to be all discoverable in one place.

MichaelOrlov:

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/acde50a4d3aa8093b4e96cd1cf613c86/raw/2cfa5c77ccc57baed0f2f321ed6edb538407a5ca/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11196
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17729/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12271/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18370/)

PR #1209, opened 2022-12-12
CLI: Get storage-specific values from plugin

emersonknapp:
Have plugins register an optional Python entrypoint providing plugin-specific CLI argument values, instead of hardcoding the values for a single plugin, or skipping input validation at the argparse level.

This first pass introduces:
* Python entrypoint registration and  loading
* A single extension API `get_preset_profiles` for storage preset profile names
* Update plugin development documentation showing how to use this feature

Following changes expanding on this framework will cover:
* Compression modes/formats
* Fully custom CLI arguments

### DIFF
--- a/docs/storage_plugin_development.md
+++ b/docs/storage_plugin_development.md
@@ -137,7 +137,10 @@ This registers an entrypoint in group `ros2bag.storage_plugin_cli_extension`, fo
 
 The exposed entrypoint can be installed as a Python module by any method, for example via `ament_cmake_python`'s `ament_python_install_package` macro, or by having a pure-python `ament_python` package with a `setup.py`.
 
-The functions this entrypoint may provide:
+The functions this entry point may provide:
 
-* `get_preset_profiles(): List[str]` - provide a list of names of preset profiles for writing storage files. The first item will be used as default, consider making this 'none'
+* `get_preset_profiles(): List[str]` - provide a list of names of preset profiles for writing storage files. The first item will be used as default. Consider returning 'none' as the first element.
+
 NOTE: For each of these lists, the string literal 'none' will be used to indicate the feature is disable/not used.
+
+NOTE: Any entry point may exclude any of the extension functions, and a warning will be printed for each extention point omitted. When the function for a list of values is not provided, or returns `None`, by default `'none'` will be provided as the only option.
--- a/docs/storage_plugin_development.md
+++ b/docs/storage_plugin_development.md
@@ -1,55 +1,145 @@
-# Writing storage plugins
+# Writing storage plugins for Rosbag2
 
-There are different interfaces for storage plugins depending on your need: The general `ReadWriteStorage` and the more specific `ReadableStorage`.
+Storage plugins provide the actual underlying storage format for Rosbag2.
 
-## Writing a general plugin
+Plugins can implement the following APIs to provide a storage plugin, specified in `rosbag2_storage::storage_interfaces`:
+* `ReadOnlyInterface` which covers only reading files
+* `ReadWriteInterface` can both write new files and read existing ones. It is a superset of ReadOnly
 
-Assume you write a plugin `MyStorage` which can both save messages and read messages.
-Its header file could be `my_storage.hpp` and `MyStorage` will derive from `rosbag2_storage::storage_interfaces::ReadWriteInterface`.
-**Important:** While implementing the interface provided by `rosbag2_storage::storage_interfaces::ReadWriteInterface`, make sure that all resources such as file handles or database connections are closed or destroyed in the destructor, no additional `close` call should be necessary.
+## Creating a ReadWrite plugin
 
-In order to find the plugin at runtime, it needs to be exported to the pluginlib.
-Add the following lines to `my_storage.cpp`:
+Goal: Create a plugin named `my_storage`, in package `rosbag2_storage_my_storage`, implemented by class `my_namespace::MyStorage`.
+
+The following code snippets shows the necessary pieces to provide this plugin.
+
 
 ```
+// my_storage.cpp
+#include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
+
+namespace my_namespace {
+
+class MyStorage : public rosbag2_storage::storage_interfaces::ReadWriteInterface
+{
+public:
+  MyStorage();
+  ~MyStorage() override;  // IMPORTANT: All cleanup must happen in the destructor, such as closing  file handles or database connections
+
+  // ReadWriteInterface's virtual overrides here
+};
+
+// Implementations
+
+}  // namespace my_namespace
+
+// The following block exposes our class to pluginlib so that it can be discovered at runtime.
 #include "pluginlib/class_list_macros.hpp"
-PLUGINLIB_EXPORT_CLASS(MyStorage, rosbag2_storage::storage_interfaces::ReadWriteInterface)
+PLUGINLIB_EXPORT_CLASS(my_namespace::MyStorage,
+                       rosbag2_storage::storage_interfaces::ReadWriteInterface)
 ```
 
-Furthermore, we need some meta-information in the form of a `plugin_description.xml` file.
+Next, our package must provide a file named `plugin_description.xml`.
 Here, it contains
 
 ```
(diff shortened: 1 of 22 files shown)


ros2bag/ros2bag/api/__init__.py
```diff
@@ -129,3 +130,65 @@ def add_standard_reader_args(parser: ArgumentParser) -> None:
...
+        choices=plugin_choices,
+        help='Storage implementation of bag. '
+            'By default attempts to detect automatically - use this argument to override.')
+    storage_parsed_args, _ = storage_parser.parse_known_args()
+    plugin_id = storage_parsed_args.storage
+
+    if plugin_id not in plugin_choices:
+        raise ValueError(f'No storage plugin found with ID {plugin_id}')
+
+    pkg_name = f'rosbag2_storage_{plugin_id}'
```
james-rms, line 150: do we have no way to extract the package name from a registered plugin otherwise?

emersonknapp: Yes, I've found a way. The only catch is that the python module must be equal to the containing package name (e.g. has to be called `rosbag2_storage_mcap`). I think this is a reasonable requirement, you agree?

emersonknapp:
Moving back to draft. I realized that the CLI registration works fine and works for preset profile, but the compression values  e.g. `compression_mode: chunk` are not propagated down to the storage plugin.

emersonknapp:
This is now ready for review, with a slightly reduced scope and a straightforward smoke test. See PR description for latest info

ros2bag/ros2bag/api/__init__.py
emersonknapp: Second update: i've changed it to use setuptools' entry_points, so no a priori knowledge is required here at all, any package can register an extension on  its side.

docs/storage_plugin_development.md
```diff
@@ -1,55 +1,143 @@
...
+[options.entry_points]
+ros2bag.storage_plugin_cli_extension =
+  mcap = ros2bag_mcap_cli
+```
+
+This registers an entrypoint in group `ros2bag.storage_plugin_cli_extension`, for the plugin named `mcap`, that is implemented by a Python module called `rosbag2_mcap_cli`.
+
+The exposed entrypoint can be installed as a Python module by any method, for example via `ament_cmake_python`'s `ament_python_install_package` macro, or by having a pure-python `ament_python` package with a `setup.py`.
+
+The functions this entrypoint may provide:
```
james-rms, line 140: would be good to include what happens by default if this entrypoint exists but does not expose any functions.

docs/storage_plugin_development.md
```diff
@@ -1,55 +1,143 @@
...
+
+* `get_preset_profiles(): List[str]` - provide a list of names of preset profiles for writing storage files. The first item will be used as default, consider making this 'none'
+NOTE: For each of these lists, the string literal 'none' will be used to indicate the feature is disable/not used.
```
james-rms, line 143: add an extra newline here for the HTML to render a line break.

docs/storage_plugin_development.md
```diff
@@ -1,55 +1,143 @@
...
+  mcap = ros2bag_mcap_cli
+```
+
+This registers an entrypoint in group `ros2bag.storage_plugin_cli_extension`, for the plugin named `mcap`, that is implemented by a Python module called `rosbag2_mcap_cli`.
+
+The exposed entrypoint can be installed as a Python module by any method, for example via `ament_cmake_python`'s `ament_python_install_package` macro, or by having a pure-python `ament_python` package with a `setup.py`.
+
+The functions this entrypoint may provide:
+
+* `get_preset_profiles(): List[str]` - provide a list of names of preset profiles for writing storage files. The first item will be used as default, consider making this 'none'
```
james-rms, line 142:
```suggestion
* `get_preset_profiles(): List[str]` - provide a list of names of preset profiles for writing storage files. The first item will be used as default. Consider returning `'none'` as the first element.
```

ros2bag/ros2bag/verb/play.py
```diff
@@ -99,29 +95,29 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--playback-duration', type=float, default=-1.0,
             help='Playback duration, in seconds. Negative durations mark an infinite playback. '
-                 'Default is -1.0. When positive, the maximum of `playback-until-*` and the one '
-                 'that this attribute yields will be used to determine which one stops playback '
-                 'execution.')
+                 'Default is %(default)d. '
```
james-rms, line 98: not sure if we enforce a single string interpolation style in ROS 2, but I always assume f-strings are the way to go.

james-rms (approved):
I  like the direction of this PR. One improvement might be for `get_preset_profiles` to return a list of (name, description) tuples, rather than asking the user to dig into the docs themselves.

ros2bag/ros2bag/verb/play.py
clalancette: FYI, one thing to be careful of is RHEL-8 support.  RHEL-8 only has Python 3.6, and we've had some problems with certain features because it is so old.  Not sure if that is the case here, just pointing it out.

emersonknapp: f-strings are good for many things! But they don't encode the builtin info that this argparse special token `%(default)` does. This way we only specify `default=` in the parameter description and the helpstring can use that same value, instead of duplicating it.

I'll double check whether it's supported in 3.6

emersonknapp: Update:  the argparse help tokens do work in python3.6.

However, the signature we are using for `importlib.metadata.entry_points` is not. @clalancette is there any CI job I can run to make sure I don't break RHEL-8?

clalancette: > @clalancette is there any CI job I can run to make sure I don't break RHEL-8?

Yep!  https://ci.ros2.org/job/ci_linux-rhel/

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/7d0336d8d9ea79b5142a24e7f40d9906/raw/3bb8259f060ddba5cec06157296c95da1d8bcf43/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_storage_mcap rosbag2_storage_sqlite3 ros2bag
TEST args:  --packages-above rosbag2_storage_mcap rosbag2_storage_sqlite3 ros2bag
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11391
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17980/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12513/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18618/)

emersonknapp:
@james-rms Updated to return a pair of (name, description) for preset profiles. Required a little bit of monkeying around to get the helptext formatting right, but nothing crazy, it works now.

emersonknapp:
Hmm `launch_testing` is not finding `rosbag2_storage_mcap` for some reason. Will have to figure that out before we can merge (Windows doesn't run that specific test, is why it's green). Other than that everything looks good.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17983/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12514/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18619/)
* RHEL8 [Build Status](https://ci.ros2.org/job/ci_linux-rhel/297/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17984/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12515/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18620/)
* [Build Status](https://ci.ros2.org/job/ci_linux-rhel/298/)

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-01-19/29423/1

PR #1220, opened 2022-12-20
Expose more Writer methods in python interface

emersonknapp:
Just missing a few.

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_writer.cpp
+++ b/rosbag2_py/src/rosbag2_py/_writer.cpp
@@ -35,30 +35,12 @@ namespace rosbag2_py
 {
 
 template<typename T>
-class Writer
+class Writer : public rosbag2_cpp::Writer
 {
 public:
   Writer()
-  : writer_(std::make_unique<rosbag2_cpp::Writer>(std::make_unique<T>()))
-  {
-  }
-
-  void open(
-    rosbag2_storage::StorageOptions & storage_options,
-    rosbag2_cpp::ConverterOptions & converter_options)
-  {
-    writer_->open(storage_options, converter_options);
-  }
-
-  void create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
-  {
-    writer_->create_topic(topic_with_type);
-  }
-
-  void remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
-  {
-    writer_->remove_topic(topic_with_type);
-  }
+  : rosbag2_cpp::Writer(std::make_unique<T>())
+  {}
 
   /// Write a serialized message to a bag file
   void write(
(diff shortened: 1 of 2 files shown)


rosbag2_py/src/rosbag2_py/_writer.cpp
```diff
@@ -104,28 +83,39 @@ std::unordered_set<std::string> get_registered_serializers()
...
   .def(pybind11::init())
-  .def("open", &rosbag2_py::Writer<rosbag2_compression::SequentialCompressionWriter>::open)
-  .def("write", &rosbag2_py::Writer<rosbag2_compression::SequentialCompressionWriter>::write)
   .def(
-    "remove_topic",
-    &rosbag2_py::Writer<rosbag2_compression::SequentialCompressionWriter>::remove_topic)
+    "open",
+    pybind11::overload_cast<
+      const rosbag2_storage::StorageOptions &, const rosbag2_cpp::ConverterOptions &
+    >(&PyWriter::open))
```
MichaelOrlov, line 98: @emersonknapp Out of curiosity why do we need `pybind11::overload_cast<..>` for open method here?

MichaelOrlov (approved):
Overall looks good to me.  
Although I have a question in regards to the `pybind11::overload_cast<..>` usage.

rosbag2_py/src/rosbag2_py/_writer.cpp
emersonknapp: It's because `rosbag2_cpp::Writer` has two overloaded `::open` methods. (`open(const std::string & uri)` and `open(const StorageOptions &, const ConverterOptions &)`) Python doesn't allow for function overloading, so we have to specify which of the two signatures to expose as the `.open` method here. Without the `overload_cast`, compilation fails due to the ambiguous signature. 

We didn't need this before because we had defined `rosbag2_py::Writer::open` method explicitly. Now, by changing to inherit from `rosbag2_cpp::Writer` instead, both `open` methods are available and we have to specify which one we mean.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/6cc4a0b28a01cde178eaa2a96144562f/raw/813c41abadd7e8e5c595e49c99a16eed8c9bfcd6/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_py
TEST args:  --packages-above rosbag2_cpp rosbag2_py
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11318
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18537/)

rosbag2_py/src/rosbag2_py/_writer.cpp
MichaelOrlov: @emersonknapp Thanks for explanation. Now it became clear for me.

PR #1230, opened 2023-01-09
mcap_vendor: add readme with versioning procedure

james-rms:


emersonknapp (approved):
Fine by me. We have the ABI checker in place within stable distros, and Rolling is explicitly no-promises. This seems good and keeps the burden off of `mcap` non-ros core infrastructure

PR #1231, opened 2023-01-10
Bump to 0.15.4

audrow:
I bumped the patch version.

I see that the mcap vendor has a different version than the rest of the packages. To my knowledge, this is the only package in the ROS 2 core that does this. It doesn't seem to work well with the tooling.

I'd suggest keeping the mcap_vendor package version in sync with the other packages in this repo. Also it looks like MCap vendor is using v0.8.0, not 0.6.0, which is the vendored version. I've split this PR into two commits if we want to keep the MCap vendor at 0.6.0, we can easily undo that change.


emersonknapp (approved):
This is fine, no need for the vendor to be out of step. This is probably an artifact from when the package was copied over to this repo

PR #1238, opened 2023-02-01
Update rosbag2 to C++17.

clalancette:
The main reason to do this is so that we can compile rosbag2 with the clang static analyzer.  As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version.  So set this as well.


james-rms (approved):

clalancette:
Just as an FYI, I haven't moved on this one because this change seemed to cause some new warnings in CI.  I haven't had time to go back and look at it yet.

clalancette:
I put just rosbag2_storage_sqlite3 back to C++14 for now, as Windows doesn't seem to like that.  I'll run CI with this in place just to make sure everything else is happy, and we can do a followup for that package later on:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18191/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18851/)

clalancette:
Since the only change here was to move `rosbag2_storage_sqlite3` back to C++14, I'm going to still consider the previous approvals valid.  With green CI, I'm going to go ahead and merge this.  Thanks!

Issue #1239, opened 2023-02-01
Store ROS_DISTRO in mcap storage file as extra metadata


emersonknapp:
## Description

Followup #1086 

Do the same storage, but in the `mcap` storage file as well.

## Related Issues

#1086

## Completion Criteria

`ROS_DISTRO` can be gleaned from a standalone `.mcap` file created by `ros2 bag record`

## Implementation Notes / Suggestions

None

## Testing Notes / Suggestions

None

PR #1241, opened 2023-02-02
Implement storing and loading ROS_DISTRO from metadata.yaml and mcap files

emersonknapp:
Closes https://github.com/ros2/rosbag2/issues/1239
Depends on https://github.com/foxglove/mcap/pull/898

Also finishes the letter of the title of #1086, where #1156 didn't add the info to metadata.yaml

Stores `ros_distro` as a member of `metadata.yaml`, by adding it to `rosbag2_storage::BagMetadata` as a string field. Implements storage and retrieval of this metadata in mcap storage plugin and modifies sqlite3 to match the new pattern.


rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -77,6 +78,7 @@ void SequentialWriter::init_metadata()
   file_info.message_count = 0;
   metadata_.custom_data = storage_options_.custom_data;
   metadata_.files = {file_info};
+  metadata_.ros_distro = rcpputils::get_env_var("ROS_DISTRO");
```
james-rms, line 81: Could we print a warning (or possibly exit) if ROS_DISTRO is not set? Users might try to launch a recorder outside of `ros2 launch` or a ros env, and may not be aware of the dependency here.

MichaelOrlov:
@emersonknapp @james-rms @defunctzombie I am sort of confused about this PR in terms of adding new field in to the metadata and request to add `ROS_DISTRO` name to the MCAP.

As far as I remember original request to adding `ROS_DISTRO` name was described and motivated as https://github.com/ros2/rosbag2/issues/1086#issuecomment-1306397786 
> Storing ROS_DISTRO specifically would be useful for Foxglove, because we try to open db3 files using a bundled copy of the standard ROS schemas, but those schemas change from release to release (e.g. Marker message is different from Galactic to Humble). Storing ROS_DISTRO would allow us to use the correct version of Marker to deserialize that file.

And in following up discussion we come to the conclusion that we need solution which would be portable to the Humble and will not break ABI and API compatibility.

This is why we ended up with solution for reading `ROS_DISTRO` env variable inside `rosbag2_storage_sqlite3` plugin without changing existing API and without adding new fields to the data structures.

Do we still need to save `ROS_DISTRO` in mcap? 
I was under impression that since mcap store message definitions aka "schema" inside, the problem with different message definitions between ROS2 distros has been solved by design.
Am I wrong about it? Or maybe I missed something?

If we still need to save `ROS_DISTRO` in mcap, does it make sense to do it the same way as it was done in sqlite3 plugin without altering metadata?

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -225,7 +226,8 @@ PYBIND11_MODULE(_storage, m) {
           topics_with_message_count,
           compression_format,
           compression_mode,
-          custom_data
+          custom_data,
+          ros_distro,
         };
       }),
     pybind11::arg("version") = 6,
```
MichaelOrlov, line 233: @emersonknapp Just in case. It seems forgotten to increment version number Python code.

defunctzombie:
> Do we still need to save `ROS_DISTRO` in mcap?
> I was under impression that since mcap store message definitions aka "schema" inside, the problem with different message definitions between ROS2 distros has been solved by design.
> Am I wrong about it? Or maybe I missed something?

It is true that the schemas are in the file so the _exact_ the reasons for adding the distro to mcap are not exactly the same as for sqlite where you need to know which message types to use to deserialize core messages. However, it is useful debugging information to capture the distro into the mcap file regardless. It can help consumers of mcap files understand which distro the file did come from. More importantly, if you `ros2 bag play` a mcap file from an older distro on a new one or vice versa some things might not work (due to the same core message incompatibilities) again making it nice to be able to identify the origin distro for debugging.

emersonknapp:
Just a quick note on accounting: #1086 never changed its title/description to reflect that we would _not_ store info in metadata.yaml, so the issue should not have been closed as it was written.

As for whether we actually want to do this - I think it's just a useful bit of information and doesn't hurt anything to add. More context is better! It doesn't backport to Humble but that's fine, future users can benefit.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
emersonknapp: I definitely don't want it to exit (especially if everything else would work properly!). But since unset ROS_DISTRO will be a minority case, I see no problem with a warning message

MichaelOrlov:
I agree with @emersonknapp that if it could be useful let's store it.  And that "More context is better!"
Another question which will affect implementation is, if we need to backport these changes to the Humble?

@defunctzombie Could you please clarify if we will need to backport it to the Humble? Or, it is Ok to keep it starting from Rolling only and move forward?

defunctzombie:
> Could you please clarify if we will need to backport it to the Humble?

From @emersonknapp statement it sounds like it doesn't backport? Or is that in reference to changes in the metadata.yaml file and not adding the mcap metadata record? Personally I think it would be useful to have this information in the file so if a backport is possible we should do it for the sake of having this information when debugging. From the tooling side it isn't strictly "required" to figure out the correct schemas since tools like Studio and PlotJuggler can read the schemas from the file.

emersonknapp:
The way this is written (most likely) isn't ABI stable because of the change to the Bag Metadata struct. But we could mod up a humble backport patch that adds the mcap metadata record without that breaking portion of the change. So it just wouldn't go into metadata.yaml, same as the prior sqlite change.

That's generally my preference - write it the way we want it for the future on rolling, and more carefully backport where necessary.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/e0c83113557dae878c326a2f26bb38cb/raw/7d11e4fc17825377d83de076520cc8ec2ba0d83d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3
TEST args:  --packages-above rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11588
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18212/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12743/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18881/)

emersonknapp:
rerun after rebasing the branch

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18220/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12751/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18889/)

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -324,6 +325,20 @@ void MCAPStorage::open_impl(const std::string & uri, const std::string & preset_
...
+        if (metadata.name == "rosbag2") {
+          metadata_.ros_distro = metadata.metadata.at("ROS_DISTRO");
+          done = true;
+        }
+      };
+      while (!done && typedReader.next()) {
+        if (!typedReader.status().ok()) {
+          throw std::runtime_error("Failed to read mcap metadata");
+        }
+      }
```
MichaelOrlov: @emersonknapp I think that this block for reading out ROS_DISTRO should reside at least in `rosbag2_storage::BagMetadata MCAPStorage::get_metadata()` function. And even better would be to move it in separate function to be called from `MCAPStorage::get_metadata()`. 
Wouldn't you mind to move it there in the scope of this PR?

Also current implementation is a bit "sub-optimal" in general case. Since it is trying to parse whole file to find metadata. We can try to speed up this process by finding direct offset to the metadata records by reading out summary `ensure_summary_read();` and if index offset on the metadata exists use it to fast jump to file offset for metadata readout. We can use `McapReader::ReadRecord(IReadable& reader, uint64_t offset, Record* record);`  and `McapReader::ParseMetadata(const Record& record, Metadata* metadata);` if we will have correct offset on metadata.

For first implementation I think it is ok to proceed with current "sub-optimal" implementation and add TODO for future improvements.

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -352,6 +367,14 @@ void MCAPStorage::open_impl(const std::string & uri, const std::string & preset_
...
         throw std::runtime_error(status.message);
       }
+
+      mcap::Metadata metadata;
+      metadata.name = "rosbag2";
+      metadata.metadata = {{"ROS_DISTRO", rcpputils::get_env_var("ROS_DISTRO")}};
+      status = mcap_writer_->write(metadata);
+      if (!status.ok()) {
+        throw std::runtime_error(status.message);
+      }
```
MichaelOrlov: @emersonknapp For this part writing `ROS_DISTRO` I think it would be better to move it inside `MCAPStorage::update_metadata(const rosbag2_storage::BagMetadata & bag_metadata)` and call it here with some pre-filled data for `bag_metadata` as much as we can on creation.  
The normal flow would be to call `update_metadata(bag_metadata) twice. Once on mcap writer open and once on writer close or destruction. Hence we will have two metadata in file to avoid loosing data and jumping back during recording. On readout we can figure out which one is more up to date by offset of by the content.

I think for first implementation it would be ok to just move `ROS_DISTRO` writing inside `update_metadata(..)` .

MichaelOrlov (changes requested):
@emersonknapp Thank you for trying to move forward with this PR. 
Overall implementation looks good to me, although I have couple suggestions for implementation in `mcap_storage.cpp` which would be good to address. Please see them in dedicated comments.

rosbag2_py/src/rosbag2_py/_storage.cpp
MichaelOrlov: @emersonknapp Please don't forget to increment default metadata version. 
i.e. pybind11::arg("version") = **7**,

rosbag2_storage_mcap/src/mcap_storage.cpp
emersonknapp: OK - I have updated to move code to `get_metadata` and `update_metadata`, plus used indices for retrieving metadata, but that depends on https://github.com/foxglove/mcap/pull/898

rosbag2_py/src/rosbag2_py/_storage.cpp
emersonknapp: Done (incremented to 8 now) - and changed this code to use the version from `BagMetadata` rather than duplicating

rosbag2_storage_mcap/src/mcap_storage.cpp
emersonknapp: Done, including using MetadataIndex entries

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
```diff
@@ -147,6 +147,7 @@ std::string format_bag_meta_data(const rosbag2_storage::BagMetadata & metadata)
   info_stream << "Bag size:          " << format_file_size(
     metadata.bag_size) << std::endl;
   info_stream << "Storage id:        " << metadata.storage_identifier << std::endl;
+  info_stream << "ROS Distro:        " << metadata.ros_distro << std::endl;
```
MichaelOrlov: Need to change expectation in `InfoEndToEndTestFixture::info_end_to_end_test` accordingly to the changes.
Currently, the test fails with an error message
```
/tmp/ws/src/rosbag2/rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.cpp:49
Value of: output
Expected: contains regular expression "\nFiles:             cdr_test_0.mcap\nBag size:          .*B\nStorage id:        mcap\nDuration:          0\\.151s\nStart:             Apr  .+ 2020 .*:.*:36.763 \\(1586406456\\.763\\)\nEnd:               Apr  .+ 2020 .*:.*:36.914 \\(1586406456\\.914\\)\nMessages:          7\nTopic information: "
  Actual: "\nFiles:             cdr_test_0.mcap\nBag size:          11.8 KiB\nStorage id:        mcap\nROS Distro:        \nDuration:          0.151s\nStart:             Apr  8 2020 20:27:36.763 (1586406456.763)\nEnd:               Apr  8 2020 20:27:36.914 (1586406456.914)\nMessages:          7\nTopic information: Topic: /array_topic | Type: test_msgs/msg/Arrays | Count: 4 | Serialization Format: cdr\n                   Topic: /test_topic | Type: test_msgs/msg/BasicTypes | Count: 3 | Serialization Format: cdr\n\n"
```

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -424,6 +424,31 @@ rosbag2_storage::BagMetadata MCAPStorage::get_metadata()
     metadata_.topics_with_message_count.push_back(topic_info);
   }
 
+  const auto & mcap_metadatas = mcap_reader_->metadataIndexes();
+  auto range = mcap_metadatas.equal_range("rosbag2");
+  mcap::Status status;
+  mcap::Record mcap_record;
```
MichaelOrlov:
```suggestion
  mcap::Record mcap_record{};
```
Nitpick. For the sake of the defensive programming.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
```diff
@@ -118,6 +118,9 @@ TEST_F(TemporaryDirectoryFixture, can_write_and_read_basic_mcap_file)
                   1u}}));
     EXPECT_THAT(metadata.message_count, Eq(1u));
 
+    const auto current_distro = rcpputils::get_env_var("ROS_DISTRO");
+    EXPECT_EQ(metadata.ros_distro, current_distro);
```
MichaelOrlov, line 122: @emersonknapp Am I understand correctly that in `Humble` and `Iron` backport it will be added `std::string get_recorded_ros_distro() const;` API in the `rosbag2_storage_mcap` ?

MichaelOrlov (changes requested):
@emersonknapp Thanks for re-iterating with this PR and addressing previous comments and suggestions.
Changes look good to me among one nitpick and failing test which needs to be addressed.
See details in my comments.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
emersonknapp: `get_recorded_ros_distro` is not part of the `ReadWriteInterface` that readers/writers use to interact with the plugin. Can we even backport an update to the `BagMetadata` structure? This change feels un-backportable, and the fact that there is a function on `sqlite_storage` doesn't seem like it actually does anything in Humble and Iron, unless I'm missing something - there is nothing that can call it.

rosbag2_storage_mcap/src/mcap_storage.cpp
emersonknapp: I'm adding this - _but_ I want to register my rebuttal that I don't know of a recommendation to always use an empty-brace initializer for default initialization. There is nothing recommending this in the Google C++ style guide, and there is no ambiguity in the syntax. Maybe you are thinking of preferring `Foo bar{}` to the "most vexing parse" `Foo bar()`. However, syntactically `Foo bar;` is equal to `Foo bar{};`, they both are unambiguous ways to call the default constructor (as far as I am aware)

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
emersonknapp: Done

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
emersonknapp: The one thing that I think I could do with no problem is backport the change to add the ROS_DISTRO metadata in the mcap. That way, newer versions of rosbag2 could read the value out and understand it.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/fbe3dfb0e62fbfe0929bc954bdf9a5aa/raw/a5522d325cec8e3223175883b545278f94a6c8e9/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_tests
TEST args:  --packages-above rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12145
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18840/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13376/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19592/)


EDIT: as per note in internal slack, ARM and Windows CI agents are down right now with issues - so we will have to wait for those to be fixed and rerun

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
MichaelOrlov: @emersonknapp Right we can't add new member in BagMetadata structure on Humble and Iron.
And as regards to the 
> and the fact that there is a function on sqlite_storage doesn't seem like it actually does anything in Humble and Iron, unless I'm missing something - there is nothing that can call it.

Though the `std::string get_recorded_ros_distro() const;` really does things by getting ros distro from the sqlite storage plugin. For `Humble` and `Iron` we don't expect that `get_recorded_ros_distro()` will be used somewhere else rather than in tests. 
However, we need to store ros distro name in mcap and sqlite3 storages on those ros2 distros to be able to parse it in third-party tools like `Foxglove Studio` and future ros2 releases for backward compatibility to be able to determine on the latter releases (Rolling or "Jazy") that some bag was recorded on Humble or Iron.

rosbag2_storage_mcap/src/mcap_storage.cpp
MichaelOrlov: Ok, here the things why I am asking to put `{}` for this particular variable.
The `mcap::record` is defined as 
```
struct MCAP_PUBLIC Record {
  OpCode opcode;
  uint64_t dataSize;
  std::byte* data;

  uint64_t recordSize() const {
    return sizeof(opcode) + sizeof(dataSize) + dataSize;
  }
};
```
The statement that `"syntactically Foo bar; is equal to Foo bar{};"` is incorrect in this particular case. Because this is a structure POD type without a default constructor. 

If we will write just `mcap::Record mcap_record` usually on `arm` platforms in release mode the value of the 
`mcap_record.dataSize` and `mcap_record.data` will be un-initialized and will contain garbage.  
Compilers don't have obligations to init this POD at all in this case. e.g. Some compilers versions in some cases still do this default initialization but this fully up to the compiler.  
If define `mcap_record` as `mcap::Record mcap_record{};` with `{}` - the compiler will be forced to call either a default constructor or explicitly initialize each member of the POD structure with default value i.e `0`.

MichaelOrlov: Also, clang-tidy inspection highlighting this particular variable
image

MichaelOrlov (approved):
LGTM! Thanks.

rosbag2_storage_mcap/src/mcap_storage.cpp
emersonknapp: Interesting subtle point - thanks for the info

MichaelOrlov:
CI got back to normal. Re-running `Linux-aarch64` and `Windows` jobs
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19597/)

MichaelOrlov:
The Windows CI job failed with error message
```
C:\ci\ws\src\ros2\rosbag2\rosbag2_cpp\test\rosbag2_cpp\test_time_controller_clock.cpp:170
Value of: sleep_result
  Actual: false
Expected: true
```
Which I don't know how to explain rather than if it was a situation when `TimeControllerClock` was not resumed or for some reason `rclcpp::ok()` became not true.  
I created a follow-up PR https://github.com/ros2/rosbag2/pull/1367 to add extra checks to be able to clarify the root cause of the failure if it will happen next time.

cc: @emersonknapp

MichaelOrlov:
Re-running Windows CI job one more time
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19599/)

PR #1242, opened 2023-02-02
Skip install when not building rosbag2_performance_benchmarking

sloretz:
This fixes a bug where `rosbag2_performance_benchmarking` gets installed to the ament resource index when it's not built, and that causes rosdep to fail to find `rosbag2_performance_benchmarking_msgs`.

Example:
```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies (ROS distro is not set. Make sure `ROS_DISTRO` environment variable is set, or use `--rosdistro` option to specify the distro, e.g. `--rosdistro indigo`):
rosbag2_performance_benchmarking: Cannot locate rosdep definition for [rosbag2_performance_benchmarking_msgs]
```

The problem is the `ament_package()` call is still run when `rosbag2_performance_benchmarking` is not built, however it's dependency `rosbag2_performance_benchmarking_msgs` does not run `ament_package()`. This causes rosdep to think `rosbag2_performance_benchmarking` is installed and needs it's dependencies, but the dependency `rosbag2_performance_benchmarking_msgs` can't be found.

I fixed it by inverting the logic and putting it at the top of the file with a `return()` statement to make sure none of the logic is run when skipping this package.

### DIFF
--- a/rosbag2_performance/rosbag2_performance_benchmarking/CMakeLists.txt
+++ b/rosbag2_performance/rosbag2_performance_benchmarking/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(rosbag2_performance_benchmarking)
 
+if (NOT BUILD_ROSBAG2_BENCHMARKS)
+  return()
+endif()
+
 # Default to C++14
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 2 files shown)


emersonknapp (approved):
No  concerns here

sloretz:
CI (build: `--packages-up-to rosbag2_performance_benchmarking` test: `rosbag2_performance_benchmarking_msgs`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18025/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12550/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18663/)

PR #1260, opened 2023-03-10
Enable document generation using rosdoc2 for ament_python pkgs

Yadunund:
Fix missing `exec_depends` in `package.xmls` and docstrings to generate documentation without any warnings using `autodoc`

### DIFF
--- a/ros2bag/package.xml
+++ b/ros2bag/package.xml
@@ -13,12 +13,10 @@
   <maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>
   <license>Apache License 2.0</license>
 
-  <depend>ros2cli</depend>
-  <depend>rosbag2_py</depend>
-  <depend>rosbag2_transport</depend>
-
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>rclpy</exec_depend>
+  <exec_depend>ros2cli</exec_depend>
+  <exec_depend>rosbag2_py</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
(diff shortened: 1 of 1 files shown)


MichaelOrlov (approved):
@Yadunund LGTM.
I am curious if we need to add the same dependencies to the https://github.com/ros2/rosbag2/tree/rolling/rosbag2_py package?

clalancette (commented):
These are obviously correct, so that's good.

While we are in here, though, we should probably clean up the other depends right above.  In particular, both of `ros2cli` and `rosbag2_py` should be `exec_depend`, not `depend`.  And this package doesn't directly depend on `rosbag2_transport` at all, so that should be removed.

Yadunund:
Thanks @clalancette for pointing out correction to the dependencies!

@MichaelOrlov I think `sphinx` can generate documentation for pybind11 bindings but we'll need to add some custom `rosdoc2.yaml` and `conf.py` files. I'm currently focussing on `ament_python` pkgs and can take a look at `rosbag2_py` in the future.

clalancette (approved):
This looks good to me now, thanks.

I'm going to run CI on it next, but I'll wait for another approval from @MichaelOrlov before merging.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18214/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12745/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18883/)

MichaelOrlov (approved):
LGTM.

MichaelOrlov:

PR #1265, opened 2023-03-19
rosbag2_cpp: move local message definition source out of MCAP plugin

james-rms:
The intention of this PR is to move the message-definition-finding capability outside of rosbag2_storage_mcap, and allow any rosbag2 storage plugin to store message definitions.


rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -169,6 +169,24 @@ void SequentialWriter::close()
...
+      "Bag is not open. Call open() before registering a message definition.");
+  }
+
+  storage_->register_message_definition(message_definition);
+
+  type_names_to_message_definitions_.insert(std::make_pair(message_definition.name, message_definition));
+  // TODO something with converters
+}
+
 void SequentialWriter::create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
```
james-rms: @MichaelOrlov I got quite confused around here, was wondering if you could clarify. There's a mutex that locks access to `topic_names_to_info_`, and it's used in `create_topic` when writing to it (but not reading from it?). Also, in `Recorder` all calls to the sequential writer are wrapped in a mutex anyway, so I'm not sure if it's neccessary at all. With that in mind I didn't bother adding a new mutex to lock access to `type_names_to_message_definitions_` here.

rosbag2_storage/include/rosbag2_storage/storage_interfaces/base_write_interface.hpp
```diff
@@ -40,6 +41,8 @@ class ROSBAG2_STORAGE_PUBLIC BaseWriteInterface
 
   virtual void update_metadata(const BagMetadata & bag_metadata) = 0;
 
+  virtual void register_message_definition(const rosbag2_storage::MessageDefinition & message_definition) = 0;
```
emersonknapp: It's good that this is a struct type. I figure we will add the `type_description_interfaces::msg::TypeDescription` information eventually (or could even add it now?) so that storage implementation could do something with the inspectible description information.

james-rms: I was going to add `type_hash` after getting the bare minimum working, but i'm not sure what rosbag2 would want to do with TypeDescription content. I was going to retrieve just the raw definitions from the `GetTypeDescription` service and insert them here.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: Hi @james-rms, Yes it is a bit messy in regards to the usage of the `topics_info_mutex_`.  
As far as I remember @adamdbrw Tried to protect access to the message's counter in `topics_names_to_info_`
https://github.com/ros2/rosbag2/blob/76665ad8b977721c059b95d307697b8424fed07d/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L423-L428
which are calling from cache consumer as call back from separate thread.
https://github.com/ros2/rosbag2/blob/76665ad8b977721c059b95d307697b8424fed07d/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L143-L145
Since those mesages counter increment was calling from inner thread they were not protected under the `writer::writer_mutex_` and we have externally exposed API for removing topic 
https://github.com/ros2/rosbag2/blob/76665ad8b977721c059b95d307697b8424fed07d/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L210-L220
Considering the flow how we are using rosbag2_cpp API from rosbag2_transport layer, we don't need this `topics_info_mutex_`. Since we are not calling `remove_topic(..)` API if subscription with write(..)  method in callback has been successfully created. 
https://github.com/ros2/rosbag2/blob/76665ad8b977721c059b95d307697b8424fed07d/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L343-L355

However since we expose `Writer::remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type)` API in theory if someone will create their own recorder they could potentially call it if valid subscription has been created. I believe that `topics_info_mutex_` was added for such case.

In your case as you mentioned above you don't have ability to access `type_names_to_message_definitions_` without  locking `writer::writer_mutex_`. 
I agree that you don't need additional mutex for it.

rosbag2_cpp/include/rosbag2_cpp/message_definitions/message_definition_cache.hpp
```diff
@@ -72,16 +49,39 @@ class MessageDefinitionCache final
...
+
+  struct DefinitionIdentifier
+  {
+    Format format;
+    std::string package_resource_name;
+
+    bool operator==(const DefinitionIdentifier & di) const
+    {
+      return (format == di.format) && (package_resource_name == di.package_resource_name);
+    }
```
MichaelOrlov: I think it would be more appropriate and efficient if  
```
    Format format;
    std::string package_resource_name;
```
will be a private members and we will have getters for them.
The rational for that is that they are settling up only once on object construction and using then as read only.
First of all it is not safe to have direct access to member variables which are supposed to be read only.
We can occasionally overwrite them.
Another point it would be more efficient if we will add one more optimization and cache hash calculation. i.e. calculate hash once on construction and store it in private member variable. In this case
`struct DefinitionIdentifierHash` will be simplified to 
```
  struct DefinitionIdentifierHash
  {
    std::size_t operator()(const DefinitionIdentifier & di) const
    {
       return di.hash()
    }
  }
```
It will be more efficient operations in unorderd_maps and sets

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -721,25 +734,8 @@ void MCAPStorage::create_topic(const rosbag2_storage::TopicMetadata & topic)
...
-      schema.encoding = "";
-    }
-    mcap_writer_->addSchema(schema);
-    schema_ids_.emplace(datatype, schema.id);
-    schema_id = schema.id;
+    throw std::runtime_error{"schema not registered for topic " + topic_info.topic_metadata.name +
+                             "\n"};
   } else {
     schema_id = schema_it->second;
   }
```
MichaelOrlov: This part could be a bit simplified 
```
  const auto schema_it = schema_ids_.find(datatype);
  if (schema_it == schema_ids_.end()) {
    throw std::runtime_error{"schema not registered for topic " + topic_info.topic_metadata.name + 
                                            "\n"};
  }
  mcap::SchemaId schema_id = schema_it->second;
```

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -110,7 +112,10 @@ void Writer::write(
   tm.name = topic_name;
   tm.type = type_name;
   tm.serialization_format = serialization_format;
-  create_topic(tm);
+  rosbag2_storage::MessageDefinition md;  // TODO(morlov) We need either add message difinition
```
james-rms: remove this comment - this is a reasonable action to take for this API.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -185,22 +187,25 @@ void SequentialWriter::create_topic(const rosbag2_storage::TopicMetadata & topic
...
-  if (!insert_succeded) {
+  if (!insert_succeeded) {
     std::stringstream errmsg;
     errmsg << "Failed to insert topic \"" << topic_with_type.name << "\"!";
 
     throw std::runtime_error(errmsg.str());
   }
 
-  storage_->create_topic(topic_with_type);
+  topic_names_to_message_definitions_.insert(
```
james-rms, line 224: unit-test this codepath

rosbag2_storage/include/rosbag2_storage/message_definition.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+struct MessageDefinition
+{
+  /// @brief  The name of the type.
+  ///
+  /// Should match the `name` in TopicMetadata for all topics using this message definition.
+  std::string type_name;
+  /// @brief  The type description hash of the type described by this MessageDefinition.
+  /// an empty string indicates that no type hash could be calculated for this definition.
+  std::string type_hash;
```
james-rms: rename to type_description_hash

rosbag2_cpp/CMakeLists.txt
```diff
@@ -133,6 +134,7 @@ if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
   find_package(ament_lint_auto REQUIRED)
   find_package(test_msgs REQUIRED)
+  find_package(rosbag2_storage_mcap_testdata REQUIRED)
```
james-rms: rename this package

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -340,7 +340,15 @@ void Recorder::subscribe_topic(const rosbag2_storage::TopicMetadata & topic)
   // Need to create topic in writer before we are trying to create subscription. Since in
   // callback for subscription we are calling writer_->write(bag_message); and it could happened
   // that callback called before we reached out the line: writer_->create_topic(topic)
-  writer_->create_topic(topic);
+  try {
+    auto message_definition = local_message_definition_source_.get_full_text(topic.type);
+    writer_->create_topic(topic, message_definition);
+  } catch (const rosbag2_cpp::DefinitionNotFoundError &) {
+    // no definition available locally - leave the schema and definition encoding as empty strings.
+    rosbag2_storage::MessageDefinition message_definition;
```
james-rms: unit-test this code path

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -122,7 +122,8 @@ setup_topic_filtering(
       std::stringstream qos_profiles;
       qos_profiles << input_topics_qos_profiles[topic_name];
       topic_metadata.offered_qos_profiles = qos_profiles.str();
-      writer->create_topic(topic_metadata);
+      // TODO(morlov:) Get message definition for topic from input storage
```
james-rms, line 125: yes

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
james-rms: maybe formalize this with a special constructor taking only type name

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
james-rms: I want to leave this for a second PR.

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/862ded3efcfa3e4156c6273481ed065f/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11704
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18350/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12878/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19018/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/0005c26246f5d4387fb1471965f60c18/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11712
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18363/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12897/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19033/)

docs/message_definition_encoding.md
```diff
@@ -0,0 +1,98 @@
+# Message Definition Encoding
+
+Message definitions can be stored in ROS 2 bags to facilitate decoding messages. This document
+describes how message definitions can be encoded within a bag.
+
+## Background
+
+When playing a bag with `ros2 bag play`, messages are not deserialized by the player. Instead, they
+are passed directly to subscribers, which deserialize them using code generated at build time.
+Therefore the message definitions in the bag are not used for bag playback. However, some
```
MichaelOrlov: This is a bit not true. 
In cases when we are calling `rclcpp::GenericPublisher::publish_as_loaned_msg(msg)`  https://github.com/ros2/rosbag2/blob/d926a617c12831b422f4eb364e5e886aec988461/rosbag2_transport/include/rosbag2_transport/player.hpp#L221-L227
message will be deserialized on publisher side. However on `rclcpp` layer.

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
```diff
@@ -0,0 +1,138 @@
...
+   * The format is different for MSG and IDL definitions, and is described fully in
+   * docs/message_definition_encoding.md
+   * Throws DefinitionNotFoundError if one or more definition files are missing for the given
+   * package resource name.
+   */
+  rosbag2_storage::MessageDefinition get_full_text(const std::string & type_name);
+
+  enum struct Format
+  {
+    MSG,
```
MichaelOrlov: @james-rms Could you please add `UNKNOWN` as first item in the enum?
```suggestion
   UNKNOWN = 0,
    MSG,
```

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
```diff
@@ -0,0 +1,138 @@
...
+    IDL,
+  };
+
+private:
+  struct MessageSpec
+  {
+    MessageSpec(Format format, std::string text, const std::string & package_context);
+    const std::set<std::string> dependencies;
+    const std::string text;
+    Format format;
```
MichaelOrlov: It will be much safely to add default value for format variable. 
```suggestion
    Format format{Format::UNKNOWN};
```
also it could be `Format::MSG`

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
```diff
@@ -0,0 +1,138 @@
...
+  struct DefinitionIdentifier
+  {
+    DefinitionIdentifier(std::string type_name, Format format)
+    : type_name_(type_name)
+      , format_(format)
+    {
+      size_t h1 = std::hash<Format>()(format_);
+      size_t h2 = std::hash<std::string>()(type_name_);
+      hash_ = h1 ^ h2;
+    }
```
MichaelOrlov: Need to disable default constructor or even better add default initialization for `hash_` and `format_` variables to avoid surprises in future.

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -183,15 +193,27 @@ std::pair<Format, std::string> MessageDefinitionCache::get_full_text(
...
-  return std::make_pair(format, result);
+  rosbag2_storage::MessageDefinition out;
+  switch (format) {
+    case Format::MSG:
+      out.encoding = "ros2msg";
+      break;
+    case Format::IDL:
+      out.encoding = "ros2idl";
+      break;
+  }
```
MichaelOrlov: Please add `default:` and `Format::UNKNOWN` section to switch.  
We can assign empty string in this case.
I know in a few lines upper we are assigning `format` variable to either `MSG` or `IDL`, although I am asking this modification for the sake of defensive programming and to be less error prone in case of code modifications or refactoring in future.

Also please consider moving this switch in 
```
std::string msg_definition_format_to_string(const LocalMessageDefinitionSource::Format & format)
```
conversion function in `rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp`.  And perhaps function for opposite conversion from string.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: @james-rms We also need to add removal of  topic definition from `topic_names_to_message_definitions_` when doing `void SequentialWriter::remove_topic(const rosbag2_storage::TopicMetadata & topic_with_type)`

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -263,10 +268,9 @@ void SequentialWriter::switch_to_next_storage()
 
     throw std::runtime_error(errmsg.str());
   }
-
-  // Re-register all topics since we rolled-over to a new bagfile.
   for (const auto & topic : topics_names_to_info_) {
-    storage_->create_topic(topic.second.topic_metadata);
+    auto const & md = topic_names_to_message_definitions_[topic.first];
```
MichaelOrlov, line 293: I liked the  comment `// Re-register all topics since we rolled-over to a new bagfile.` as it was before.
Please add it back.

rosbag2_cpp/test/rosbag2_cpp/test_storage_without_metadata_file.cpp
```diff
@@ -56,7 +56,7 @@ TEST_F(StorageWithoutMetadataFileTest, open_uses_storage_id_from_storage_options
       "topic",
       "test_msgs/BasicTypes",
       kRmwFormat,
-      ""
+      "",
```
MichaelOrlov: forgotten extra `,`

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
```diff
@@ -0,0 +1,138 @@
...
+    MessageSpec(Format format, std::string text, const std::string & package_context);
+    const std::set<std::string> dependencies;
+    const std::string text;
+    Format format;
+  };
+
+  struct DefinitionIdentifier
+  {
+    DefinitionIdentifier(std::string type_name, Format format)
+    : type_name_(type_name)
```
MichaelOrlov: Small optimization from clang-tiny
```suggestion
    : type_name_(std::move(type_name))
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -28,7 +25,11 @@
 #include <unordered_set>
 #include <utility>
 
-namespace rosbag2_storage_mcap::internal
+#include <ament_index_cpp/get_package_share_directory.hpp>
+#include <ament_index_cpp/get_resource.hpp>
+#include <ament_index_cpp/get_resources.hpp>
```
MichaelOrlov: Small hint from CLion code inspection. The last two includes seems redundant.
```suggestion
#include <ament_index_cpp/get_package_share_directory.hpp>
```

MichaelOrlov (changes requested):
@james-rms Thanks for reiterating and addressing comments from previous round of code review.
I like your renames and replacing tuples with structures.
I mostly have comments about small nitpicks and improvements related to the defensive programming or some misses.
However I have one big concern about API and usage of the Writer::write(..) API when we are substituting `rosbag2_storage::MessageDefinition::empty_message_definition_for(type_name)` instead of the real message definition. This is default API which we are promoting for the users and using in rosbag2 cpp examples.

I think we can handle it a better way.
Here is my proposal:
- Will need to add `create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)` API as it was before to `rosbag2_cpp::Writer` and down to the `rosbag2_cpp::SequentialWriter` in addition to the create_topic(..) with message definition. 
- Also will need to move LocalMessageDefinitionSource from recorder.cpp down to the SequentialWriter class.
- In `SequentialWriter::create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)` we can check if  if topic_name not found in local `topics_names_to_info_ ` then do similar stuff as we currently doing in recorder.cpp
```
try {
    auto message_definition = local_message_definition_source_.get_full_text(topic.type);
    writer_->create_topic(topic, message_definition);
  } catch (const rosbag2_cpp::DefinitionNotFoundError &) {
    // no definition available locally - leave the schema and definition encoding as empty strings.
    writer_->create_topic(
      topic, rosbag2_storage::MessageDefinition::empty_message_definition_for(topic.type));
  }
```
- at the same time we need to keep create_topic(..) with separate message definition parameter to be able to use it from `bag_rewrite.cpp` 

If implement my proposal we can simplify a lot in most of the unit tests (i.e. we can keep calling old API) and in generic usage of the create topic. 
Message definition cache will always persists only in one place i.e. SequentialWriter and upper level API will be abstract from it and don't need to taking care about it.

rosbag2_storage/include/rosbag2_storage/message_definition.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace rosbag2_storage
+{
+
+struct MessageDefinition
+{
+  /// @brief  The name of the type.
+  ///
+  /// Should match the `name` in TopicMetadata for all topics using this message definition.
+  std::string type_name;
```
MichaelOrlov: @james-rms It seems there are a big confusion about this `type_name` fileld.
After some investigation I think that it really shall be **topic_type** and shall correspond to the field `type` from `TopicMetada` struct.  
We don't use this field on storage plugin level since we need `TopicMetada::type` instead. 
Although on all upper levels we treat it as it mapped to the `TopicMetada::type` instead of the `TopicMetada::name`. Take a look on all tests and even in recorder.cpp 
https://github.com/ros2/rosbag2/blob/f1074b908573fda94bc68ee571b28a4fcc93ed7f/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L344
where we are calling get_full_text(..)

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
```diff
@@ -0,0 +1,138 @@
...
+{
+public:
+  /**
+   * Concatenate the message definition with its dependencies into a self-contained schema.
+   * The format is different for MSG and IDL definitions, and is described fully in
+   * docs/message_definition_encoding.md
+   * Throws DefinitionNotFoundError if one or more definition files are missing for the given
+   * package resource name.
+   */
+  rosbag2_storage::MessageDefinition get_full_text(const std::string & type_name);
```
MichaelOrlov: Please rename to avoid confusion.
```suggestion
  rosbag2_storage::MessageDefinition get_full_text(const std::string & root_topic_type);
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -114,53 +121,56 @@ static std::string delimiter(const DefinitionIdentifier & definition_identifier)
...
 
   // "References and pointers to data stored in the container are only invalidated by erasing that
   // element, even when the corresponding iterator is invalidated."
   return spec;
 }
 
-std::pair<Format, std::string> MessageDefinitionCache::get_full_text(
-  const std::string & root_package_resource_name)
+rosbag2_storage::MessageDefinition LocalMessageDefinitionSource::get_full_text(
+  const std::string & root_type_name)
```
MichaelOrlov: @james-rms Please rename parameter to avoid misleading in API
```suggestion
rosbag2_storage::MessageDefinition LocalMessageDefinitionSource::get_full_text(
  const std::string & root_topic_type)
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -183,15 +193,27 @@ std::pair<Format, std::string> MessageDefinitionCache::get_full_text(
   std::string result;
   Format format = Format::MSG;
   try {
-    result = append_recursive(DefinitionIdentifier{format, root_package_resource_name});
+    result = append_recursive(DefinitionIdentifier(root_type_name, format));
```
MichaelOrlov:
```suggestion
    result = append_recursive(DefinitionIdentifier(root_topic_type, format));
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -183,15 +193,27 @@ std::pair<Format, std::string> MessageDefinitionCache::get_full_text(
...
   } catch (const DefinitionNotFoundError & err) {
     // log that we've fallen back
-    RCUTILS_LOG_WARN_NAMED("rosbag2_storage_mcap", "no .msg definition for %s, falling back to IDL",
-                           err.what());
+    RCUTILS_LOG_WARN_NAMED(
+      "rosbag2_cpp", "no .msg definition for %s, falling back to IDL",
+      err.what());
     format = Format::IDL;
-    DefinitionIdentifier root_definition_identifier{format, root_package_resource_name};
+    DefinitionIdentifier root_definition_identifier(root_type_name, format);
```
MichaelOrlov:
```suggestion
    DefinitionIdentifier root_definition_identifier(root_topic_type, format);
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -183,15 +193,27 @@ std::pair<Format, std::string> MessageDefinitionCache::get_full_text(
...
+  switch (format) {
+    case Format::MSG:
+      out.encoding = "ros2msg";
+      break;
+    case Format::IDL:
+      out.encoding = "ros2idl";
+      break;
+  }
+  out.encoded_message_definition = result;
+  out.type_name = root_type_name;
```
MichaelOrlov:
```suggestion
  out.type_name = root_topic_type;
```

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
MichaelOrlov: I don't really like unbounded recursion here.  
Recursion is evil in embedded systems. Especially when it tied to the input from file system.
Attacker can easily create cyclic link on message datatype and cause stackoverflow. 
Need to consider how to get rid from recursion here.

MichaelOrlov (changes requested):

rosbag2_cpp/include/rosbag2_cpp/message_definitions/local_message_definition_source.hpp
james-rms: resolved by taking the argument as a const reference

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
james-rms: Inserted a hard-check on recursion depth.

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -114,84 +121,107 @@ static std::string delimiter(const DefinitionIdentifier & definition_identifier)
...
+  const std::string & root_topic_type)
 {
   std::unordered_set<DefinitionIdentifier, DefinitionIdentifierHash> seen_deps;
 
-  std::function<std::string(const DefinitionIdentifier &)> append_recursive =
-    [&](const DefinitionIdentifier & definition_identifier) {
+  std::function<std::string(const DefinitionIdentifier &, int32_t)> append_recursive =
+    [&](const DefinitionIdentifier & definition_identifier, int32_t depth) {
+      if (depth <= 0) {
+        throw DefinitionNotFoundError(definition_identifier.topic_type());
```
MichaelOrlov: I think we shall raise some `runtime_error` exception.  
This is faulty situation happened in runtime not due to the really missing message definition but rather because our internal hard coded limitation on recursion depth.   
We need upper levels and users know about it explicitly. 
Throwing `DefinitionNotFoundError` exception in this case will be misleading and it will be difficult to understand why it fails to find some message definition while it is obviously available.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -170,6 +170,20 @@ void SequentialWriter::close()
 }
 
 void SequentialWriter::create_topic(const rosbag2_storage::TopicMetadata & topic_with_type)
+{
+  rosbag2_storage::MessageDefinition definition;
```
MichaelOrlov: @james-rms Could you please add preliminary check ?
```
if (topics_names_to_info_.find(topic_with_type.name) !=
    topics_names_to_info_.end())
  {
    // nothing to do, topic already created
    return;
  }
```
Before we will try to look at message definition local cache.
(or use `topic_names_to_message_definitions_` instead of `topics_names_to_info_`)
I know we do the same check in another inner `create_topic(tm, md)` with message definition.
Although this particular `create_topic(tm)` will be called for each data write from upper writer::write(msg) level
https://github.com/ros2/rosbag2/blob/3e9b610e56030135e32950f33b206a95695e74b1/rosbag2_cpp/src/rosbag2_cpp/writer.cpp#L121-L122
And we will have to do incure in cache and composing full message definition and copy past a lot of strings. Need to avoid extra operations especially related to the memory copy on this data path.

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -122,6 +122,7 @@ setup_topic_filtering(
       std::stringstream qos_profiles;
       qos_profiles << input_topics_qos_profiles[topic_name];
       topic_metadata.offered_qos_profiles = qos_profiles.str();
+      // TODO(morlov:) Get message definition for topic from input storage
       writer->create_topic(topic_metadata);
```
MichaelOrlov, line 126: Nitpick. I would prefer to use `writer->create_topic(topic_metadata, {});` here to show that we need to use here version with message definition.

rosbag2_compression/include/rosbag2_compression/sequential_compression_writer.hpp
```diff
@@ -78,7 +78,9 @@ class ROSBAG2_COMPRESSION_PUBLIC SequentialCompressionWriter
    * \param topic_with_type name and type identifier of topic to be created
    * \throws runtime_error if the Writer is not open.
    */
-  void create_topic(const rosbag2_storage::TopicMetadata & topic_with_type) override;
+  void create_topic(
+    const rosbag2_storage::TopicMetadata & topic_with_type,
+    const rosbag2_storage::MessageDefinition & message_definition) override;
```
MichaelOrlov, line 93: Forgotten original API 
```
void create_topic(const rosbag2_storage::TopicMetadata & topic_with_type) override;
```
Need to keep it as well

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -221,10 +221,11 @@ void SequentialCompressionWriter::close()
 }
 
 void SequentialCompressionWriter::create_topic(
-  const rosbag2_storage::TopicMetadata & topic_with_type)
+  const rosbag2_storage::TopicMetadata & topic_with_type,
+  const rosbag2_storage::MessageDefinition & message_definition)
 {
   std::lock_guard<std::recursive_mutex> lock(storage_mutex_);
-  SequentialWriter::create_topic(topic_with_type);
+  SequentialWriter::create_topic(topic_with_type, message_definition);
```
MichaelOrlov: Please add original API as well
```
void SequentialCompressionWriter::create_topic(
  const rosbag2_storage::TopicMetadata & topic_with_type)
{
  std::lock_guard<std::recursive_mutex> lock(storage_mutex_);
  SequentialWriter::create_topic(topic_with_type);
}
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -243,7 +243,7 @@ TEST_F(SequentialCompressionWriterTest, writer_creates_correct_metadata_relative
 
   tmp_dir_storage_options_.max_bagfile_size = 1;
   writer_->open(tmp_dir_storage_options_);
-  writer_->create_topic({test_topic_name, test_topic_type, "", ""});
+  writer_->create_topic({test_topic_name, test_topic_type, "", ""}, {});
```
MichaelOrlov: @james-rms I would recommend to revert to use original API without message definition in tests to reduce amount of unrelated delta.
This is also relates to all other occurrences of usage create_topic(..) with empty message_definition in this and all other files with old unit tests.

rosbag2_cpp/src/rosbag2_cpp/message_definitions/local_message_definition_source.cpp
```diff
@@ -114,84 +121,107 @@ static std::string delimiter(const DefinitionIdentifier & definition_identifier)
...
   }
 
   std::string contents{std::istreambuf_iterator(file), {}};
   const MessageSpec & spec =
     msg_specs_by_definition_identifier_
-      .emplace(definition_identifier,
-               MessageSpec(definition_identifier.format, std::move(contents), package))
-      .first->second;
+    .emplace(
+    definition_identifier,
```
MichaelOrlov: Tentative suggestion for beautification after ~~uglifier~~ uncrustify
```
const MessageSpec & spec = msg_specs_by_definition_identifier_.emplace(
    definition_identifier,
    MessageSpec(definition_identifier.format(), std::move(contents), package)).first->second;
```
Note. In this form uncrustify also passes.

rosbag2_examples/rosbag2_examples_cpp/src/data_generator_node.cpp
```diff
@@ -39,7 +39,7 @@ class DataGenerator : public rclcpp::Node
       "example_interfaces/msg/Int32",
       rmw_get_serialization_format(),
       ""
-    });
+    }, {});
```
MichaelOrlov: Please use old style API without message definition in examples package

rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp
```diff
@@ -195,7 +195,7 @@ void WriterBenchmark::create_writer()
     // TODO(adamdbrw) - replace with something more general if needed
     topic.type = "rosbag2_performance_benchmarking_msgs/msg/ByteArray";
     topic.serialization_format = serialization_format;
-    writer_->create_topic(topic);
+    writer_->create_topic(topic, {});
```
MichaelOrlov: The same comment as above
Please revert to use old style API without message definition in examples package

rosbag2_py/test/test_sequential_writer.py
```diff
@@ -36,7 +36,6 @@ def create_topic(writer, topic_name, topic_type, serialization_format='cdr'):
     topic_name = topic_name
     topic = rosbag2_py.TopicMetadata(name=topic_name, type=topic_type,
                                      serialization_format=serialization_format)
-
```
MichaelOrlov: Nitpick. After multiple editions appear redundant delta. (tentative)

rosbag2_examples/rosbag2_examples_cpp/src/data_generator_executable.cpp
```diff
@@ -38,7 +38,7 @@ int main(int, char **)
     "example_interfaces/msg/Int32",
     rmw_get_serialization_format(),
     ""
-  });
+  }, {});
```
MichaelOrlov: Please use old style API without message definition in examples package

MichaelOrlov (changes requested):
@james-rms Thank you fixing comments from previous round of review.
I think we are pretty close to finish.
In this new round of review I found mostly nitpicks and small improvements. Please see them in dedicated comments.
Also I have generic recommendation to rollback to use original create_topic(..) API without message definition in old tests in places where message definition substituted with `{}` to the sake of reducing amount of unrelated delta.

MichaelOrlov (approved):
@james-rms Thanks for addressing my review comments and suggestions.
We nailed it down!
Now it looks good to me. Approving.
Please don't forget to run CI.

MichaelOrlov:
@james-rms I noticed that `rosbag2_compression` start failing and pretty consistently.
Although there are not to much details from the log about reasoning.
```
023-03-31T19:59:04.3603209Z The following tests FAILED:
2023-03-31T19:59:04.3603574Z 	  4 - test_sequential_compression_writer (Failed)
2023-03-31T19:59:04.3603840Z Errors while running CTest
2023-03-31T19:59:04.3604226Z Output from these tests are in: /__w/rosbag2/rosbag2/ros_ws/build/rosbag2_compression/Testing/Temporary/LastTest.log
2023-03-31T19:59:04.3604762Z Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
2023-03-31T19:59:04.3605088Z ---
2023-03-31T19:59:04.3605348Z Finished <<< rosbag2_compression [5.92s]	[ with test failures ]
2023-03-31T19:59:04.3605666Z Starting >>> rosbag2_compression_zstd
2023-03-31T19:59:04.3605979Z --- stderr: rosbag2_compression
2023-03-31T19:59:04.3606222Z Errors while running CTest
2023-03-31T19:59:04.3606600Z Output from these tests are in: /__w/rosbag2/rosbag2/ros_ws/build/rosbag2_compression/Testing/Temporary/LastTest.log
2023-03-31T19:59:04.3607472Z Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```
Unfortunately I was not able to reproduce this failure locally.
I have a suspicious that it might be related to the missing 
`MOCK_METHOD1(create_topic, void(const rosbag2_storage::TopicMetadata &));` in
https://github.com/ros2/rosbag2/blob/4e87dd1195e2bcf080f8f2418a9445f3e03e23e2/rosbag2_compression/test/rosbag2_compression/mock_storage.hpp#L36-L40
which it had before.
Could you please try to add this mock method for the original `create_topic(tm)` API?

james-rms:
@MichaelOrlov that can't be it, the `rosbag2_storage` write API doesn't have a single-argument `create_topic(tm)` method. Only the Writer API does.

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/acca7459da55f451285325a2104aae7d/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11732
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12925/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19055/)

MichaelOrlov:
@james-rms You are right "the rosbag2_storage write API doesn't have a single-argument create_topic(tm) method"
However It seems I figured out why tests was failing. See my fix proposal and description in #1271

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/1da96a28cbb84b065c9be1628fb017fa/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11741
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18389/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12935/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19068/)

MichaelOrlov:
@james-rms There are weird failure on windows with message 
```
 Building Custom Rule C:/ci/ws/src/ros2/rosbag2/rosbag2_storage_mcap/CMakeLists.txt

  test_mcap_storage.cpp

LINK : fatal error LNK1181: cannot open input file 'Release\rosbag2_storage_mcap.lib' [C:\ci\ws\build\rosbag2_storage_mcap\test_mcap_storage.vcxproj]

---
Failed   <<< rosbag2_storage_mcap [1min 3s, exited with code 1]
``` 
Which need to investigate.

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/48357b3e104237c2ca27b86257e26b56/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11779
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18428/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12974/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19112/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/ed3b67fc54cfec8020eef1d94432724e/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11783
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18432/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12978/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19116/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/18f3874492d72c1f4b7d045b0bfebe7f/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11784
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18433/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19117/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/a55fb2b413f67ef7ae02a13c5c59ec5e/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11787
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18436/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12982/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19120/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/b00cebebac10d81e6c7e4abc14fe5075/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11799
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18448/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12994/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19137/)

james-rms:
Gist: https://gist.githubusercontent.com/james-rms/73e2d46529262ec2bd0444083b97adbe/raw/c0b2ed4d5edd741dc2b2004f8c9cd99ef142890d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_storage_mcap rosbag2_test_msgdefs rosbag2_compression rosbag2_storage
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11803
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18452/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12998/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19143/)

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-4-20-2023/31087/1

PR #1266, opened 2023-03-22
DRAFT: Add message definition to TopicMetadata

MichaelOrlov:
Intention of this PR is to prototype one of the possible solution of how to pass `message_definition` to the storage and read it back for bag rewriter.

### DIFF
--- a/rosbag2_storage/include/rosbag2_storage/topic_metadata.hpp
+++ b/rosbag2_storage/include/rosbag2_storage/topic_metadata.hpp
@@ -16,6 +16,7 @@
 #define ROSBAG2_STORAGE__TOPIC_METADATA_HPP_
 
 #include <string>
+#include "rosbag2_storage/message_definition.hpp"
 
 namespace rosbag2_storage
 {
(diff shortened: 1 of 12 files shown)


rosbag2_storage/include/rosbag2_storage/topic_metadata.hpp
```diff
@@ -27,12 +28,13 @@ struct TopicMetadata
   std::string serialization_format;
   // Serialized std::vector<rclcpp::QoS> as a YAML string
   std::string offered_qos_profiles;
-  // type hash, retrieved from subscription QoS metadata
-  std::string type_hash;
+  rosbag2_storage::MessageDefinition message_definition;
```
emersonknapp, line 31: Maybe this could be a reference or `shared_ptr`, so that we don't have to store several copies of complex topics?

MichaelOrlov: @emersonknapp I agree and I don't like neither that it will be a lot of redundant and useless for human reader information in BagMetadata.  `TopicMetadata` uses in `TopicInformation`
https://github.com/ros2/rosbag2/blob/d926a617c12831b422f4eb364e5e886aec988461/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L29-L33
and further on   `TopicInformation` included in `BagMetadata` as a vector with elements for each topic 
https://github.com/ros2/rosbag2/blob/d926a617c12831b422f4eb364e5e886aec988461/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L53

It is unclear for me how to make a reference or `shared_ptr` in this case. We will need to serialize and desrialize it on some stage.  If you have some concrete idea or suggestions - could you please elaborate on it?

MichaelOrlov:
- Closing as will not be merged since we opted out for solution from [DRAFT: Add message_definition to create_topic(..) API #1269](https://github.com/ros2/rosbag2/pull/1269)

PR #1273, opened 2023-04-05
zstd_vendor: update to v1.5.5

james-rms:
Facebook recently published a patch update to fix a data corruption bug in their zstd compression library. See the release notes here:

https://github.com/facebook/zstd/releases/tag/v1.5.5

See also the bug-fix patch here:

https://github.com/facebook/zstd/pull/3517

### DIFF
--- a/zstd_vendor/CMakeLists.txt
+++ b/zstd_vendor/CMakeLists.txt
@@ -35,7 +35,7 @@ macro(build_zstd)
   # We need to configure the CMake command to build from there instead.
   externalproject_add(zstd-${zstd_version}
     GIT_REPOSITORY https://github.com/facebook/zstd.git
-    GIT_TAG 97a3da1df009d4dc67251de0c4b1c9d7fe286fc1 # v1.4.8
+    GIT_TAG 63779c798237346c2b245c546c40b72a5a5913fe # v1.5.5
     GIT_CONFIG advice.detachedHead=false
     # Suppress git update due to https://gitlab.kitware.com/cmake/cmake/-/issues/16419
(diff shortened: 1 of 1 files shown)


zstd_vendor/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ macro(build_zstd)
   # We need to configure the CMake command to build from there instead.
   externalproject_add(zstd-${zstd_version}
     GIT_REPOSITORY https://github.com/facebook/zstd.git
-    GIT_TAG 97a3da1df009d4dc67251de0c4b1c9d7fe286fc1 # v1.4.8
+    GIT_TAG 63779c798237346c2b245c546c40b72a5a5913fe # v1.5.5
```
clalancette, line 38: So I don't have time to think about this tonight, but please don't merge this too quickly.  We have to think further about how this interacts with the packages on Ubuntu, for instance.

clalancette: I'll think about it and comment further tomorrow.

james-rms:
[Note that this bug was only introduced in **v1.5.0**](https://github.com/facebook/zstd/pull/3517#issuecomment-1441020493), and current `rolling` vendors version 1.4.8, so we are not vulnerable to this bug at this stage.

### How does this interact with Ubuntu zstd package versions?

See the versions shipped with various ubuntu distros here: https://packages.ubuntu.com/search?keywords=libzstd-dev

Right now the only distros that ship a vulnerable version are Kinetic and Lunar, which we don't support.

In conclusion, this patch isn't neccessary. We can revisit when it comes time to support the next ubuntu LTS (24.04).

PR #1277, opened 2023-04-11
Switch the actions to use the 'rolling' branch.

clalancette:


mjcarroll (approved):

clalancette:
I'm not going to run CI on this one, because there is no point; we've only changed GitHub actions, not any code.  It looks like the jobs are happy, so I'll go ahead and merge.

emersonknapp:
Yup looks good, thanks for the patch

PR #1279, opened 2023-04-11
Switch the actions to use the 'humble' branch.

MichaelOrlov:
We were using master branch and rolling distro for `humble` branch. The 'master' branch was replaced with distro specific branches and shall not be used any more.
Follow up from https://github.com/ros2/rosbag2/pull/1275
Addressing failure in CI actions build and test with error in` rosidl_dynamic_typesupport_fastrtps` package https://github.com/ros2/rosbag2/actions/runs/4661768584/jobs/8251418933?pr=1142

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-latest
     container:
-      image: rostooling/setup-ros-docker:ubuntu-focal-latest
+      image: rostooling/setup-ros-docker:ubuntu-jammy-latest
     steps:
     - name: Build and run tests
       id: action-ros-ci
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - humble
   schedule:
     # Run every hour. This helps detect flakiness,
     # and broken external dependencies.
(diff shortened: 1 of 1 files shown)


clalancette (approved):
I just realized that the container image being used here is also incorrect, and should be Jammy.  I'm going to approve anyway because this is an obvious improvement, but we should update that as well (also on the `rolling` branch).

emersonknapp:
@clalancette good news, rolling is already using `jammy` images  as  of https://github.com/ros2/rosbag2/pull/1209 - surprisingly late in the release cycle, the focal builds kept working just fine for some reason.

Might as well update this PR to use the correct images rather than requiring a separate followup

MichaelOrlov:
@clalancette @emersonknapp I was considering to switch to the Jammy but after looking at the https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027 I was thinking that it uses focal on purpose since humble is declared to support  it.
Do you still think that I should switch it to the `ubuntu-jammy-latest`?

clalancette:
> Do you still think that I should switch it to the `ubuntu-jammy-latest`?

I think so, yes.  While we do support Focal in Humble, it is only from-source and Tier 3 support.  The Tier 1 platform is Jammy, so I'd think we'd want to do most of our testing there.

I guess you could also add in a matrix to test them both.

emersonknapp (approved):

PR #1281, opened 2023-04-12
rosbag2_transport: move recorder implementation into pImpl

james-rms:
* Moves the implementation details of `Recorder` out of the public API, allowing us to iterate on its members without being held to ABI stability requirements.
* Removes the unexported but public methods `topics_using_fallback_qos()` and `subscriptions()`, which were not being used anywhere in the `rosbag2` repo. Searching all code on github for `topics_using_fallback_qos` returns no code using this API.

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -109,72 +109,9 @@ public:
 protected:
   ROSBAG2_TRANSPORT_EXPORT
   std::unordered_map<std::string, std::string> get_requested_or_available_topics();
+
 private:
   std::unique_ptr<RecorderImpl> impl_;
-
-  std::shared_ptr<rosbag2_cpp::Writer> writer_;
-  rosbag2_storage::StorageOptions storage_options_;
-  rosbag2_transport::RecordOptions record_options_;
-  std::atomic<bool> stop_discovery_;
-  void topics_discovery();
-
-  std::unordered_map<std::string, std::string>
-  get_missing_topics(const std::unordered_map<std::string, std::string> & all_topics);
-
-  void subscribe_topics(
-    const std::unordered_map<std::string, std::string> & topics_and_types);
-
-  void subscribe_topic(const rosbag2_storage::TopicMetadata & topic);
-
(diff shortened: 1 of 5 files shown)


rosbag2_transport/include/rosbag2_transport/recorder.hpp
```diff
@@ -119,12 +119,12 @@ class Recorder : public rclcpp::Node
 protected:
   ROSBAG2_TRANSPORT_EXPORT
   std::unordered_map<std::string, std::string> get_requested_or_available_topics();
+private:
+
```
MichaelOrlov: @james-rms What was the rational for making those data members private?  
As far as I remember they were made protected on purpose to be able to create composable recorder via inheritance. 
i.e. to have a default constructor without parameters in derived class.

MichaelOrlov (commented):
@james-rms I have one concern that in the way how this PR currently organized we will lost all git history in `recorder` since content was copy pasted from `recorder.cpp` to the `recorder_impl.cpp`
It will be more appropriate to use rename `recorder.cpp` to the `recorder_impl.cpp` and then creating `recorder.cpp` again as a new file to keep git history for `recorder_impl.cpp`
The same in regards to the `recorder.hpp`.

emersonknapp (changes requested):
I agree with the idea of course, but I think this implementation is a bit over-agressive, I have two bits of structural feedback I'd like to address:
* Maintain the current `public` and `protected` APIs unchanged
* Keep the `impl` implementation in-place in `recorder.cpp` without introducing new file indirection to follow to read the code, this should cut down the size of the diff considerably.

Edit: that means that the pimpl might need to hold a `friend`ed pointer back to its owner so it can have access to protected members. Bit of a tradeoff, open to thoughts on that.

MichaelOrlov:
Protected  members was made for the sake of the ability to create default constructor without any parameters and have ability to provision recording configuration in latter separate calls or for the sake of using it from unit tests.  
I have suspicious that we don't have dependencies on protected members currently in our unit tests. 
Therefore a proper workaround would be to make all of them private and provide public setters methods or kind of `recorder::init(..)` method along side with default constructor.

PR #1284, opened 2023-04-12
add message definition reading API

james-rms:


MichaelOrlov (commented):
@james-rms I don't see any benefits returning message_definition as pair for TopicMetadata in `get_all_topics_and_types()` API. 
Any way we will need to form `unordered_map<topic_type, message_definition> message_definitions_map` in 
https://github.com/ros2/rosbag2/blob/f7bd8bde0cc448173dc89849694f5f1883f6ab3b/rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp#L87-L88
before or after parsing `bag_topics_and_types`
since we finally need to find correspondence of the filtered topics to the type_definition by name and by topic type down here 
https://github.com/ros2/rosbag2/blob/f7bd8bde0cc448173dc89849694f5f1883f6ab3b/rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp#L110-L126
e.g. `writer->create_topic(topic_metadata, message_definitions_map[topic_type]);`

It will be much less delta and more optimal to add new API 
```cpp
std::vector<rosbag2_storage::MessageDefinition> reader->get_message_definitions() const;
```
instead of changing existent `get_all_topics_and_types()`
We can easily iterate through vector and add elements to the  `message_definitions_map` by using `rosbag2_storage::MessageDefinition::topic_type` as a key for the map.

PR #1289, opened 2023-04-13
docs: add ros2_type_description encoding

james-rms:


docs/schemas/ros2_type_description.schema.json
```diff
@@ -0,0 +1,84 @@
+{
```
MichaelOrlov, line 1: @james-rms I am bit confused with json file. As far as I recall in  one of your original gist comments you mention that you have changed format to yaml.

PR #1300, opened 2023-04-20
Add recorder stop() API

MichaelOrlov:
- Relates  #1213
- Replaces #1216

- Expose Writer::close() as public API and add Recorder::stop() API to be able to use in future for gracefully handle SIGINT and SIGTERM and send bagsplit() or bagclose() event from Recorder::stop() API.

PR #1213, opened 2022-12-14
Add `BagSplitInfo` on bag close

aditya2592:

Issue #926, opened 2021-11-30
Windows CI can't find ``metadata.yaml`` file when using ``ros2 bag record``


adityapande-1995:
## Description
This came up in https://github.com/ros2/examples/pull/327, in the script ``rosbag_recrd_launch_test.py``. The test runs ``ros2 bag record -a -o <temp_directory>`` with a ``talker`` node, waits for 3 secs for the data to be recorded then checks for ``metadata.yaml`` file in the ``<temp_directory>``

### Expected Behavior
The Windows CI should find the ``metadata.yaml`` file. Linux and MacOS CI work just fine.

### Actual Behavior
* Failing build : https://ci.ros2.org/job/ci_windows/15711/testReport/junit/launch_testing_examples.launch_testing_examples/record_rosbag_launch_test/launch_testing_examples_record_rosbag_launch_test/
* CI error log: 
```
Error Message
launch_testing_examples.record_rosbag_launch_test.TestFixtureAfterShutdown.test_rosbag_record failed
Stacktrace
===============================================================================================================================================================
FAIL: launch_testing_examples.record_rosbag_launch_test.TestFixtureAfterShutdown.test_rosbag_record
---------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\ci\ws\src\ros2\examples\launch_testing\launch_testing_examples\launch_testing_examples\record_rosbag_launch_test.py", line 66, in test_rosbag_record
    with open(rosbag_dir + '/metadata.yaml', 'r') as file:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmp60w8hd1r/test_bag/metadata.yaml'
```

## System (please complete the following information)
 - OS: CI windows
 - ROS 2 Distro: Rolling

## Additional context
The example script has been disabled for windows for now.

emersonknapp:
Can't remember exactly - but I think this is a problem we've run into on Windows in the past - that windows can't send a proper SIGINT. So, the ExecuteProcess does not receive the SIGINT and does not have a clean shutdown, which is how the `metadata.yaml` file is created at the end of the `ros2 bag record` command.

Not 100% positive that's what is happening, but it is ringing that bell. I don't know how it is solved because Windows doesn't have the same concept of signals. How could we provide a clean shutdown for the process on Windows?

ivanpauno:
[...]

PR #1301, opened 2023-04-20
Gracefully handle SIGINT and SIGTERM in rosbag2 recorder

MichaelOrlov:
- Followup from #1300 
- Relates #1213
- Relates #926 

- Call recorder->stop() and exec_->cancel() instead of rclcpp::shutdown in signal handlers.

Note: Will address Windows specific signals handling in a separate PR.

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -189,11 +189,6 @@ public:
 
 class Recorder
 {
-private:
-  using ExecutorT = rclcpp::executors::SingleThreadedExecutor;
-  std::unique_ptr<ExecutorT> exec_;
-  std::shared_ptr<rosbag2_transport::Recorder> recorder_;
-
 public:
   Recorder()
   {
@@ -201,7 +196,6 @@ public:
     init_options.shutdown_on_signal = false;
     rclcpp::init(0, nullptr, init_options, rclcpp::SignalHandlerOptions::None);
 
-    exec_ = std::make_unique<ExecutorT>();
     std::signal(
       SIGTERM, [](int /* signal */) {
         rosbag2_py::Recorder::cancel();
@@ -222,42 +216,52 @@ public:
     RecordOptions & record_options,
     std::string & node_name)
   {
+    exit_ = false;
+    auto exec = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
     if (record_options.rmw_serialization_format.empty()) {
       record_options.rmw_serialization_format = std::string(rmw_get_serialization_format());
     }
 
     auto writer = rosbag2_transport::ReaderWriterFactory::make_writer(record_options);
-    recorder_ = std::make_shared<rosbag2_transport::Recorder>(
+    auto recorder = std::make_shared<rosbag2_transport::Recorder>(
       std::move(writer), storage_options, record_options, node_name);
-    recorder_->record();
+    recorder->record();
 
-    exec_->add_node(recorder_);
-    try {
+    exec->add_node(recorder);
+    // Run exec->spin() in a separate thread, because we need to call exec->cancel() after
+    // recorder->stop() to be able to send notifications about bag split and close.
+    auto spin_thread = std::thread(
+      [&exec]() {
+        exec->spin();
+      });
+    {
       // Release the GIL for long-running record, so that calling Python code can use other threads
       py::gil_scoped_release release;
-      while (!exit_) {
-        exec_->spin_all(std::chrono::milliseconds(10));
-      }
-      if (recorder_) {
-        recorder_->stop();
-        // Need to spin node to be able to send notifications about bag split and close
-        exec_->spin_all(std::chrono::milliseconds(10));
-      }
-    } catch (...) {
-      exec_->remove_node(recorder_);
+      std::unique_lock<std::mutex> lock(wait_for_exit_mutex_);
+      wait_for_exit_cv_.wait(lock, [] {return rosbag2_py::Recorder::exit_.load();});
+      recorder->stop();
     }
-    exec_->remove_node(recorder_);
+    exec->cancel();
+    if (spin_thread.joinable()) {
+      spin_thread.join();
+    }
+    exec->remove_node(recorder);
   }
 
   static void cancel()
   {
     exit_ = true;
+    wait_for_exit_cv_.notify_all();
   }
 
+protected:
   static std::atomic_bool exit_;
+  static std::condition_variable wait_for_exit_cv_;
+  std::mutex wait_for_exit_mutex_;
 };
 
 std::atomic_bool Recorder::exit_{false};
+std::condition_variable Recorder::wait_for_exit_cv_{};
 
 // Return a RecordOptions struct with defaults set for rewriting bags.
 rosbag2_transport::RecordOptions bag_rewrite_default_record_options()
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -190,7 +190,18 @@ public:
 class Recorder
 {
 private:
-  static std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_;
+  using ExecutorT = rclcpp::executors::SingleThreadedExecutor;
+  static std::unique_ptr<ExecutorT, void (*)(ExecutorT *)> & get_executor_instance()
+  {
+    static std::unique_ptr<ExecutorT, void (*)(ExecutorT *)> executor(
+      new ExecutorT(),
+      [](ExecutorT * exec_ptr) {
+        delete exec_ptr;
+        rclcpp::shutdown();
+      }
+    );
+    return executor;
+  }
   std::shared_ptr<rosbag2_transport::Recorder> recorder_;
 
 public:
@@ -198,8 +209,9 @@ public:
   {
     auto init_options = rclcpp::InitOptions();
     init_options.shutdown_on_signal = false;
-    rclcpp::init(0, nullptr, init_options, rclcpp::SignalHandlerOptions::None);
-    exec_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
+    if (!rclcpp::ok()) {
+      rclcpp::init(0, nullptr, init_options, rclcpp::SignalHandlerOptions::None);
+    }
     std::signal(
       SIGTERM, [](int /* signal */) {
         rosbag2_py::Recorder::cancel();
@@ -210,10 +222,7 @@ public:
       });
   }
 
-  virtual ~Recorder()
-  {
-    rclcpp::shutdown();
-  }
+  virtual ~Recorder() = default;
 
   void record(
     const rosbag2_storage::StorageOptions & storage_options,
@@ -229,33 +238,35 @@ public:
       std::move(writer), storage_options, record_options, node_name);
     recorder_->record();
 
-    exec_->add_node(recorder_);
-    // Release the GIL for long-running record, so that calling Python code can use other threads
-    {
+    get_executor_instance()->add_node(recorder_);
+    try {
+      // Release the GIL for long-running record, so that calling Python code can use other threads
       py::gil_scoped_release release;
       while (!exit_) {
-        exec_->spin_all(std::chrono::milliseconds(30));
+        get_executor_instance()->spin_all(std::chrono::milliseconds(30));
       }
       if (recorder_) {
         recorder_->stop();
         // Need to spin once to be able to send notifications about bag split and close
-        exec_->spin_some(std::chrono::milliseconds(30));
+        get_executor_instance()->spin_some(std::chrono::milliseconds(30));
       }
+    } catch (...) {
+      get_executor_instance()->remove_node(recorder_);
     }
+    get_executor_instance()->remove_node(recorder_);
   }
 
   static void cancel()
   {
     exit_ = true;
-    if (exec_) {
-      exec_->cancel();
+    if (get_executor_instance()) {
+      get_executor_instance()->cancel();
     }
   }
 
   static std::atomic_bool exit_;
 };
 
-std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> Recorder::exec_;
 std::atomic_bool Recorder::exit_{false};
 
 // Return a RecordOptions struct with defaults set for rewriting bags.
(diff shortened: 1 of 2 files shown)


MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/95900a255259b630779ca19eff1f12d4/raw/db8ac815a5d0ab97586d75360a4a7b13d9f2786c/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py rosbag2_tests
TEST args:  --packages-above rosbag2_py rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12090
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18772/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13295/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19513/)

MichaelOrlov:
@emersonknapp @james-rms This PR is relatively small and ready for review

MichaelOrlov:
@clalancette @emersonknapp @james-rms Kindly ping for review. 
This PR is blocking another https://github.com/ros2/rosbag2/pull/1342 PR

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,16 +190,23 @@ class Player
 class Recorder
 {
 private:
-  std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_;
+  static std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_;
```
emersonknapp: Why is this member static now? Because of the `rclcpp::init/shutdown` in here, it doesn't  really make sense to have multiple `rclpy::Recorder` instances, but the semantics of this global variable don't really make sense to me. Would these be better as just a global variable, or a singleton object?

MichaelOrlov: @emersonknapp Hmm, I didn't think about the situation when multiple instances of the `rclpy::Recorder` and one static executor. This is a good point.
I made the executor static to be able to call its static `exec_->cancel();` method from signal handlers to interrupt execution.  
In the case of the situation when multiple instances of the `rclpy::Recorder` exists it looks reasonable I would say even must have to make the executor as static to convey signal interruption event to all instances. 

However, I agree that it would be better to wrap the static executor in the singleton pattern. I am sorry I missed it, - will fix.
Good catch!

MichaelOrlov:
The build_and_test job failed with exactly the same error messages as I have seen recently in https://github.com/ros2/rosbag2/pull/1342
```
FAIL: test_record.TestRecordAfterShutdown.test_exit_code[sqlite3]
2023-06-03T19:45:38.7299964Z ----------------------------------------------------------------------------------------------------------------------------------
2023-06-03T19:45:38.7300321Z Traceback (most recent call last):
2023-06-03T19:45:38.7300670Z   File "/__w/rosbag2/rosbag2/ros_ws/src/s1oktrgw0j/rosbag2/ros2bag/test/test_record.py", line 73, in test_exit_code
2023-06-03T19:45:38.7301027Z     launch_testing.asserts.assertExitCodes(
2023-06-03T19:45:38.7301467Z   File "/__w/rosbag2/rosbag2/ros_ws/build/launch_testing/launch_testing/asserts/assert_exit_codes.py", line 62, in assertExitCodes
2023-06-03T19:45:38.7301967Z     assert info.returncode in allowable_exit_codes, 'Proc {} exited with code {}'.format(
2023-06-03T19:45:38.7302418Z AssertionError: Proc ros2bag-cli-7 exited with code -9
2023-06-03T19:45:38.7302839Z ----------------------------- Captured stdout call -----------------------------
2023-06-03T19:45:38.7303047Z 
2023-06-03T19:45:38.7303143Z Starting test run test_record[sqlite3]
2023-06-03T19:45:38.7303591Z [INFO] [launch]: All log files can be found below /github/home/.ros/log/2023-06-03-19-45-17-529104-0dd7204ee0ae-126136
2023-06-03T19:45:38.7303940Z [INFO] [launch]: Default logging verbosity is set to INFO
2023-06-03T19:45:38.7304291Z [INFO] [ros2bag-cli-7]: process started with pid [126220]
2023-06-03T19:45:38.7304840Z [ros2bag-cli-7] stdin is not a terminal device. Keyboard handling disabled.[INFO] [1685821526.257381155] [rosbag2_recorder]: Press SPACE for pausing/resuming
2023-06-03T19:45:38.7305542Z [ros2bag-cli-7] [INFO] [1685821526.260584248] [rosbag2_storage]: Opened database '/tmp/tmpcx5wk9g_/ros2bag_test_record/ros2bag_test_record_0.db3' for READ_WRITE.
2023-06-03T19:45:38.7306069Z [ros2bag-cli-7] [INFO] [1685821526.264542138] [rosbag2_recorder]: Event publisher thread: Starting
2023-06-03T19:45:38.7306561Z [ros2bag-cli-7] [INFO] [1685821526.265532736] [rosbag2_recorder]: Listening for topics...
2023-06-03T19:45:38.7307011Z [ros2bag-cli-7] [INFO] [1685821526.270718724] [rosbag2_recorder]: Subscribed to topic '/rosout'
2023-06-03T19:45:38.7307436Z [INFO] [ros2bag-cli-7]: sending signal 'SIGINT' to process[ros2bag-cli-7]
2023-06-03T19:45:38.7307965Z [ERROR] [ros2bag-cli-7]: process[ros2bag-cli-7] failed to terminate '5' seconds after receiving 'SIGINT', escalating to 'SIGTERM'
2023-06-03T19:45:38.7308438Z [INFO] [ros2bag-cli-7]: sending signal 'SIGTERM' to process[ros2bag-cli-7]
2023-06-03T19:45:38.7308974Z [ERROR] [ros2bag-cli-7]: process[ros2bag-cli-7] failed to terminate '10.0' seconds after receiving 'SIGTERM', escalating to 'SIGKILL'
2023-06-03T19:45:38.7309436Z [INFO] [ros2bag-cli-7]: sending signal 'SIGKILL' to process[ros2bag-cli-7]
2023-06-03T19:45:38.7310002Z [ERROR] [ros2bag-cli-7]: process has died [pid 126220, exit code -9, cmd 'ros2 bag record -a -s sqlite3 --output /tmp/tmpcx5wk9g_/ros2bag_test_record'].
2023-06-03T19:45:38.7310263Z 
2023-06-03T19:45:38.7310369Z Starting test run test_record[mcap]
2023-06-03T19:45:38.7310818Z [INFO] [launch]: All log files can be found below /github/home/.ros/log/2023-06-03-19-45-17-529104-0dd7204ee0ae-126136
2023-06-03T19:45:38.7311165Z [INFO] [launch]: Default logging verbosity is set to INFO
2023-06-03T19:45:38.7311511Z [INFO] [ros2bag-cli-8]: process started with pid [126237]
2023-06-03T19:45:38.7312041Z [ros2bag-cli-8] stdin is not a terminal device. Keyboard handling disabled.[INFO] [1685821536.633422675] [rosbag2_recorder]: Press SPACE for pausing/resuming
2023-06-03T19:45:38.7312563Z [ros2bag-cli-8] [INFO] [1685821536.638780051] [rosbag2_recorder]: Event publisher thread: Starting
2023-06-03T19:45:38.7312999Z [ros2bag-cli-8] [INFO] [1685821536.640277344] [rosbag2_recorder]: Listening for topics...
2023-06-03T19:45:38.7313438Z [ros2bag-cli-8] [INFO] [1685821536.645916818] [rosbag2_recorder]: Subscribed to topic '/rosout'
2023-06-03T19:45:38.7313859Z [INFO] [ros2bag-cli-8]: sending signal 'SIGINT' to process[ros2bag-cli-8]
2023-06-03T19:45:38.7314332Z [ros2bag-cli-8] [INFO] [1685821536.649572001] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
2023-06-03T19:45:38.7314847Z [ros2bag-cli-8] [INFO] [1685821536.650399197] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
2023-06-03T19:45:38.7315323Z [ros2bag-cli-8] [INFO] [1685821536.651652191] [rosbag2_recorder]: Event publisher thread: Exiting
2023-06-03T19:45:38.7315982Z [INFO] [ros2bag-cli-8]: process has finished cleanly [pid 126237]
```
It looks like something got changed recently in underlying sources which is leading to those error.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,23 +190,39 @@ class Player
...
 private:
-  std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_;
+  using ExecutorT = rclcpp::executors::SingleThreadedExecutor;
+  static std::unique_ptr<ExecutorT, void (*)(ExecutorT *)> & get_executor_instance()
+  {
+    static std::unique_ptr<ExecutorT, void (*)(ExecutorT *)> executor(
+      new ExecutorT(),
+      [](ExecutorT * exec_ptr) {
+        delete exec_ptr;
+        rclcpp::shutdown();
```
emersonknapp: This destructor will only be called on program shutdown, correct?

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,23 +190,39 @@ class Player
...
+    auto init_options = rclcpp::InitOptions();
+    init_options.shutdown_on_signal = false;
+    if (!rclcpp::ok()) {
+      rclcpp::init(0, nullptr, init_options, rclcpp::SignalHandlerOptions::None);
+    }
     std::signal(
       SIGTERM, [](int /* signal */) {
-        rclcpp::shutdown();
+        rosbag2_py::Recorder::cancel();
+      });
```
emersonknapp, line 202: I guess I still don't understand why this has to be static, singleton globals. Why not e.g.

```
std::signal(
  SIGTERM, [this](int) {
    this->cancel();
  });
```

Or, does the instance get destroyed before the signal handler is called?

MichaelOrlov: Well, just because 
```
std::signal(
  SIGTERM, [this](int) {
    this->cancel();
  });
```
will not compile.
We can't pass `this` pointer in this case. `std::signal(..)`  expecting a pointer to the static function.
i.e. `this->cancel();` shall be static.

After making a deeper analysis and hefty debug session of the failure on CI referred in the upper comment https://github.com/ros2/rosbag2/pull/1301#issuecomment-1577729842 I found out that there sometimes appear deadlock inside `SingleThreadedExecutor` when calling `executor->cancel()` method from another thread.

I will create a follow-up issue for the executor, however for a while considering to rewrite this PR without a static instance of the executor. As a workaround, I am considering triggering my static `exit_` variable inside signal handlers and just waiting in the main thread when the executor will finish its `spin_all(std::chrono::milliseconds(10));` operation.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: @emersonknapp Yes. Do you think it could cause any problems?

MichaelOrlov:
@emersonknapp Well after a more detailed analysis and debug session, I wasn't able to prove that freeze happening inside `executor->cancel()` call. However `executor->cancel()` can throw an exception.
- See details in the #1380

I rolled back to the non-static executor. And the rationale for that is that it seems the previous implementation with static executor member looks a bit overengineering and doesn't align well with the design of the class since we have context init/shutdown in the constructor and destructor. But the existence of the executor doesn't make sense without a valid context. 
In case when the signal will arrive we will trigger our internal static `exit_` variable and wait while the current `exec_>spin_all(10msec)` will exit.

I think we can turn back to the version with static executor later on if we will find out that the current implementation doesn't satisfy our needs.

Thoughts?

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -218,24 +227,38 @@ class Recorder
...
+        exec_->spin_all(std::chrono::milliseconds(10));
+      }
+    } catch (...) {
+      exec_->remove_node(recorder_);
     }
+    exec_->remove_node(recorder_);
   }
 
-  void cancel()
+  static void cancel()
```
emersonknapp, line 251: This still means that we can't have multiple `Recorder` in a single process, correct? I'm not sure if that was really supported before, but this decision pushes away from that possibility.

MichaelOrlov: Well, multiple recorders from Python are unlikely since we have `rclcpp::init()` in the constructor and `rclcpp::shutdown()` in the destructor. And it was not able before. 
I only changed the `cancel()` method to be static to be able to call it from signal handlers and it does make sense to me.

If the user presses `CTRL+C` we will stop all recorders in the current ~thread~ process to be able to gracefully finish and exit from the process.

emersonknapp (approved):
This is fine for now. If we do want to support multiple Recorders from python we will need to do some other changes anyways.

MichaelOrlov:
@emersonknapp I found a better way to handle executor spin() interruption via spinning it in a separate thread and wait on the conditional variable before triggering `recorder->stop()` method.  
I realized that we need spin() executor until the end of the  `recorder->stop()` call to avoid failure on publishing events about bag close or split. Also waiting on the conditional variable for exit will bring more deterministic behavior.
Please review my changes in the new commit https://github.com/ros2/rosbag2/pull/1301/commits/c0b75eb80e740e8370f5db884a016657c7de1615

MichaelOrlov:
Re-run CI
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12218
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18927/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13463/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19680/)

emersonknapp (approved):

MichaelOrlov:

MichaelOrlov:

tonynajjar:
Hey, I'm on Ubuntu, Iron and this doesn't seem to work. You can easily reproduce it with:


```python
import rclpy

from rclpy.node import Node
from rosbag2_py import Recorder

class RosBagRecorder(Node):
    def __init__(self):
        super().__init__("rosbag_recorder")
        self.recorder = Recorder()

def main():
    rclpy.init()
    node = RosBagRecorder()
    rclpy.spin(node)


if __name__ == "__main__":
    main()
```

The process never exits with CTRL-C

opened https://github.com/ros2/rosbag2/issues/1458

PR #1303, opened 2023-04-23
Add config option to use storage_id parameter in benchmark_launch.py

MichaelOrlov:
- Add ability to run performance benchmarking with multiple storage plugins

### DIFF
--- a/rosbag2_performance/rosbag2_performance_benchmarking/config/benchmarks/default_benchmark_producers.yaml
+++ b/rosbag2_performance/rosbag2_performance_benchmarking/config/benchmarks/default_benchmark_producers.yaml
@@ -8,6 +8,7 @@ rosbag2_performance_benchmarking:
         no_transport:         False # Whether to run storage-only or end-to-end (including transport) benchmark
         preserve_bags:        False # Whether to leave bag files after experiment (and between runs). Some configurations can take lots of space!
         parameters:                 # Each combination of parameters in this section will be benchmarked
+          storage_id:             ["mcap", "sqlite3"]
           max_cache_size:         [100000000, 500000000]
           max_bag_size:           [0, 1000000000]
           compression:            [""]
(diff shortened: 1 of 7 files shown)


MichaelOrlov:
@carlossvg Could you please review this PR?

carlossvg (approved):
LGTM!

MichaelOrlov:
@emersonknapp @james-rms This PR has been reviewd by @carlossvg. Please review additionally or formally approve to be able to merge this PR.

james-rms (approved):

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/a4393fca076a51e8cd16018c93fd0a1c/raw/2258d6943445538b10abdebbb9c0771b21bcec69/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_performance_benchmarking
TEST args:  --packages-above rosbag2_performance_benchmarking
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11948
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18613/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13149/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19349/)

MichaelOrlov:
@clalancette Just to double-check. Am I understand correctly that it's ok to backport such PRs when changes only in python and cpp code?  
I am pretty sure about cpp, but how about python?  
It should be ok too since it is scripting language and there shouldn't be any problems with API/ABI compatibility right?

MichaelOrlov:

MichaelOrlov:

clalancette:
> @clalancette Just to double-check. Am I understand correctly that it's ok to backport such PRs when changes only in python and cpp code?
> I am pretty sure about cpp, but how about python?
> It should be ok too since it is scripting language and there shouldn't be any problems with API/ABI compatibility right?

The general goal of the freeze is to prevent too many changes from entering Iron before the release and potentially destabilizing it.  However, we also want to allow bugfixes in to fix problems found while testing.

So we should consider each change and what the benefits vs. rewards of the change.  If it fixes a bug and doesn't break API, then that is fine, it can go into Iron.  If it introduces a new feature or breaks an API, it should not go into Iron at this time.  Note that ABI breaks are permitted right now, but they will not be permitted in Iron after May 23rd.

For this particular PR, my thinking is that we shouldn't backport it to Iron.  It isn't in code that users are likely to use or see, and it doesn't seem like a bugfix to me.  But please correct me if I'm wrong.  Also, FYI @Yadunund .

MichaelOrlov:
@clalancette Thank you for clarification I will hold on on merging it to iron until Iron's final release after May 23th.

PR #1309, opened 2023-04-28
Bump version to 0.23.0

Yadunund:
Bumping minor version so that `rolling` remains ahead of `iron`


clalancette (approved):
This looks good to me, but I'd like another review from the rosbag2 maintainers before we merge this.

MichaelOrlov (approved):

Issue #927, opened 2021-12-02
``.db3`` file access error in windows CI


adityapande-1995:
## Description
This error appears consistently in vanilla foxy CI in the test case ``TestRos2BagRecord.test_incomplete_qos_profile,
TestRos2BagRecord.test_qos_simple``. First discovered in PR : https://github.com/ros2/launch_ros/pull/289

### Expected Behavior
The db3 file should be found and CI should pass 

### Actual Behavior
The following error is thrown: https://ci.ros2.org/job/ci_windows/15875/testReport/junit/ros2bag.test/test_record_qos_profiles/test_record_qos_profiles/
```
FAIL: TestRos2BagRecord.test_qos_simple
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python38\lib\shutil.py", line 613, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmplmfz5xs0\\ros2bag_test_basic_0.db3'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python38\lib\tempfile.py", line 803, in onerror
    _os.unlink(path)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmplmfz5xs0\\ros2bag_test_basic_0.db3'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\ci\ws\src\ros2\rosbag2\ros2bag\test\test_record_qos_profiles.py", line 75, in test_qos_simple
    assert not matches, print('ros2bag CLI did not produce the expected output')
  File "C:\Python38\lib\tempfile.py", line 828, in __exit__
    self.cleanup()
  File "C:\Python38\lib\tempfile.py", line 832, in cleanup
    self._rmtree(self.name)
  File "C:\Python38\lib\tempfile.py", line 814, in _rmtree
    _shutil.rmtree(name, onerror=onerror)
  File "C:\Python38\lib\shutil.py", line 737, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Python38\lib\shutil.py", line 615, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
[...]

PR #1342, opened 2023-05-22
Gracefully handle CTRL+C and CTR+Break events on Windows

### LINKED ISSUE #926: Windows CI can't find ``metadata.yaml`` file when using ``ros2 bag record``

### LINKED ISSUE #1270: Consistently failing test in nightly_win_xfail: rosbag2_tests.TestRecordEndToEnd/RecordFixture
## Description
The https://ci.ros2.org/view/nightly/job/nightly_win_xfail have consistently failing tests in the `RecordFixture` test since approximately forever (at least back into November 2022).

The most recent failures are here: https://ci.ros2.org/view/nightly/job/nightly_win_xfail/1095/#showFailuresLink

### Expected Behavior
Tests pass on the nightly builds.

### Actual Behavior
Full log is at https://ci.ros2.org/view/nightly/job/nightly_win_xfail/1095/testReport/ , but choosing one of the tests at random shows:

```
unknown file
C++ exception with description "Bag is not open. Call open() before setting filter." thrown in the test body.
```

## To Reproduce
1. Run a job on Windows equivalent to the nightly windows xfail tests.
2. See the failure.

 - OS: Windows 10
 - ROS 2 Distro: Rolling
 - Version: d926a617c12831b422f4eb364e5e886aec988461 (but this has been happening for a long time)

## Additional context

### LINKED ISSUE #1326: Windows Signal Handling
I have spent some time investigating the cause of test failures on Windows due to unclean shutdown issues.  Here are some notes that I captured so that I can come back and look at this after the Iron release (or so someone else can pick it up).

I started with a candidate disabled test: https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L214-L272

This test does a simple recording and verifies that the command line tooling shuts down cleanly and writes all necessary data on the way down.  Currently, on Windows, this test fails.

The recording process is started via the `CreateProcess` call on Windows [(rosbag2 code)](https://github.com/ros2/rosbag2/blob/65c889e1fa55dd85a148b27b8c27dadc73238e67/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_windows.hpp#L40-L53) [(msdn ref)](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa).

The recording process is stopped via `GenerateConsoleCtrlEvent` [(rosbag2 code)](https://github.com/ros2/rosbag2/blob/65c889e1fa55dd85a148b27b8c27dadc73238e67/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_windows.hpp#L108-L120) [(msdn ref)](https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent).

* GenerateConsoleCtrlEvent fires either a `CTRL_C_EVENT` (equivalent to sigint) or a `CTRL_BREAK_EVENT` (closer to sigterm).
  * We must also specify the process group to send the event to, otherwise it is sent to everything attached to the console (including the ros nodes used by the test).
  * The only way to get the process group is to set the  `CREATE_NEW_PROCESS_GROUP` flag when we do `CreateProcess` and store it somewhere.
  * When a new process group is created, the Ctrl-C event handler is *disabled* by default in the new process:
   > When a process is created with CREATE_NEW_PROCESS_GROUP specified, an implicit call to SetConsoleCtrlHandler(NULL,TRUE) is made on behalf of the new process; this means that the new process has CTRL+C disabled. This lets shells handle CTRL+C themselves, and selectively pass that signal on to sub-processes. CTRL+BREAK is not disabled, and may be used to interrupt the process/process group.

  * Interestingly, using `std::signal` does _not_ automatically re-enable the sigint handler, you must explicitly handle the windows signals.

The best approach would be to add a parallel set of handlers in `rclcpp::SignalHandlers` and potentially in `test_rosbag2` to make sure we are explicitly handling windows signals, and I believe that will clear up some of these shutdown issues.  There is an example on how to do this here (https://learn.microsoft.com/en-us/windows/console/registering-a-control-handler-function)

MichaelOrlov:
- Add handler for the native Windows `CTRL_C_EVENT`, `CTRL_BREAK_EVENT` and `CTRL_CLOSE_EVENT` in rosbag2 recorder and player.
- Map SIGINT signal to the native Windows `CTRL_C_EVENT` in the `stop_execution(handle, signum)` version for Windows to be able correctly use start and stop execution routines from unit tests.
- Uses `CREATE_NEW_PROCESS_GROUP` when creating new process in process execution helper functions to be able to properly send native Windows `CTRL_C_EVENT` to the newly spawned child process.
- Creates new process suspended and resumes it after adding to the newly created job. To avoid a potential race condition where a newly created process starts a subprocess before we've called `AssignProcessToJobObject();`
- Enable integration tests that were disabled for Windows due to the incorrect sending and handling of the SIGINT event.
- Got rid of the `finalize_metadata_kludge(..)` helper function.

- ~~Depends from #1301~~
- Relates #927 
- Closes #926
- Closes #1326
- Closes #1270

**Note:** 
According to the https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170
> SIGINT is not supported for any Win32 application. When a CTRL+C interrupt occurs, Win32 operating systems generate a new thread to specifically handle that interrupt.

> The SIGILL and SIGTERM signals aren't generated under Windows. They're included for ANSI compatibility. Therefore, you can set signal handlers for these signals by using signal, and you can also explicitly generate these signals by calling [raise](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=msvc-170).

The `CTRL_CLOSE_EVENT` is an analog of the `SIGTERM` from POSIX. Windows sends `CTRL_CLOSE_EVENT`
to all processes attached to a console when the user closes the console (either by clicking Close on the console window's window menu, or by clicking the End Task button command from Task Manager).
Although according to the https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent the
`GenerateConsoleCtrlEvent(..)` doesn't support sending `CTRL_CLOSE_EVENT`. There is no way to directly send `CTRL_CLOSE_EVENT` to the process in the same console application.
Therefore, added `SIGTERM` to the unsupported events in the `stop_execution(process_handle, signum)` API.

- **Update:** Removed dependencies from #1301 and rebased on top of the latest Rolling to be able to move forward with this PR and easily backport it to the Iron and Humble.

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -31,6 +31,11 @@
 
 #include "./pybind11.hpp"
 
+#ifdef _WIN32
+#include <windows.h>
+#include <iostream>
+#endif
+
 namespace py = pybind11;
 typedef std::unordered_map<std::string, rclcpp::QoS> QoSMap;
 
@@ -132,7 +137,49 @@ class Player
 public:
   Player()
   {
-    rclcpp::init(0, nullptr);
+    auto init_options = rclcpp::InitOptions();
+    init_options.shutdown_on_signal = false;
+    rclcpp::init(0, nullptr, init_options, rclcpp::SignalHandlerOptions::None);
+    std::signal(
+      SIGTERM, [](int /* signal */) {
+        rclcpp::shutdown();
+      });
+    std::signal(
+      SIGINT, [](int /* signal */) {
+        rclcpp::shutdown();
+      });
+#ifdef _WIN32
+    // According to the
+    // https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170
+    // SIGINT is not supported for any Win32 application. When a CTRL+C interrupt occurs,
+    // Win32 operating systems generate a new thread to specifically handle that interrupt.
+    // Therefore, need to use native Windows control event handler as analog for the SIGINT.
+
+    // Enable default CTRL+C handler first. This is workaround and needed for the cases when
+    // process created with CREATE_NEW_PROCESS_GROUP flag. Without it, installing custom Ctrl+C
+    // handler will not work.
+    if (!SetConsoleCtrlHandler(nullptr, false)) {
+      std::cerr << "Rosbag2 player error: Failed to enable default CTL+C handler. \n";
+    }
+    // Installing our own control handler
+    auto CtrlHandler = [](DWORD fdwCtrlType) -> BOOL {
+        switch (fdwCtrlType) {
+          case CTRL_C_EVENT:
+          case CTRL_BREAK_EVENT:
+          case CTRL_CLOSE_EVENT:
+            rclcpp::shutdown();
+            return TRUE;
+          case CTRL_SHUTDOWN_EVENT:
+            rclcpp::shutdown();
+            return FALSE;
+          default:
+            return FALSE;
+        }
+      };
+    if (!SetConsoleCtrlHandler(CtrlHandler, TRUE)) {
+      std::cerr << "Rosbag2 player error: Could not set control handler.\n";
+    }
+#endif  // #ifdef _WIN32
   }
 
   virtual ~Player()
@@ -158,7 +205,7 @@ public:
     player->wait_for_playback_to_finish();
 
     exec.cancel();
-    spin_thread.join();
+    if (spin_thread.joinable()) {spin_thread.join();}
   }
 
   void burst(
@@ -180,7 +227,8 @@ public:
     player->burst(num_messages);
 
     exec.cancel();
-    spin_thread.join();
+    if (spin_thread.joinable()) {spin_thread.join();}
+    if (play_thread.joinable()) {play_thread.join();}
   }
 };
 
@@ -210,6 +258,39 @@ public:
       SIGINT, [](int /* signal */) {
         rosbag2_py::Recorder::cancel();
       });
+
+#ifdef _WIN32
+    // According to the
+    // https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170
+    // SIGINT is not supported for any Win32 application. When a CTRL+C interrupt occurs,
+    // Win32 operating systems generate a new thread to specifically handle that interrupt.
+    // Therefore, need to use native Windows control event handler as analog for the SIGINT.
+
+    // Enable default CTRL+C handler first. This is workaround and needed for the cases when
+    // process created with CREATE_NEW_PROCESS_GROUP flag. Without it, installing custom Ctrl+C
+    // handler will not work.
+    if (!SetConsoleCtrlHandler(nullptr, false)) {
+      std::cerr << "Rosbag2 recorder error: Failed to enable default CTL+C handler.\n";
+    }
+    // Installing our own control handler
+    auto CtrlHandler = [](DWORD fdwCtrlType) -> BOOL {
+        switch (fdwCtrlType) {
+          case CTRL_C_EVENT:
+          case CTRL_BREAK_EVENT:
+          case CTRL_CLOSE_EVENT:
+            rosbag2_py::Recorder::cancel();
+            return TRUE;
+          case CTRL_SHUTDOWN_EVENT:
+            rosbag2_py::Recorder::cancel();
+            return TRUE;
+          default:
+            return FALSE;
+        }
+      };
+    if (!SetConsoleCtrlHandler(CtrlHandler, TRUE)) {
+      std::cerr << "Rosbag2 recorder error: Could not set control handler.\n";
+    }
+#endif  // #ifdef _WIN32
 
     try {
       exit_ = false;
@@ -251,6 +332,13 @@ public:
       if (old_sigint_handler != SIG_ERR) {
         std::signal(SIGTERM, old_sigint_handler);
       }
+#ifdef _WIN32
+      // Remove CtrlHandler from ConsoleCtrl handlers
+      SetConsoleCtrlHandler(CtrlHandler, FALSE);
+      if (!SetConsoleCtrlHandler(nullptr, false)) {
+        std::cerr << "Rosbag2 recorder error: Failed to enable default CTL+C handler.\n";
+      }
+#endif  // #ifdef _WIN32
       throw;
     }
     // Return old signal handlers
@@ -260,6 +348,13 @@ public:
     if (old_sigint_handler != SIG_ERR) {
       std::signal(SIGTERM, old_sigint_handler);
     }
+#ifdef _WIN32
+    // Remove CtrlHandler from ConsoleCtrl handlers
+    SetConsoleCtrlHandler(CtrlHandler, FALSE);
+    if (!SetConsoleCtrlHandler(nullptr, false)) {
+      std::cerr << "Rosbag2 recorder error: Failed to enable default CTL+C handler.\n";
+    }
+#endif  // #ifdef _WIN32
   }
 
   static void cancel()
--- a/ros2bag/test/test_record.py
+++ b/ros2bag/test/test_record.py
@@ -53,13 +53,10 @@ class TestRecord(unittest.TestCase):
 
     def test_output(self, record_all_process, proc_output):
         proc_output.assertWaitFor(
-            'Listening for topics...',
+            "Subscribed to topic '/events/write_split'",
             process=record_all_process
         )
-        proc_output.assertWaitFor(
-            "Subscribed to topic '/rosout'",
-            process=record_all_process
-        )
+        proc_output.assertWaitFor('Recording...', process=record_all_process)
 
 
 @launch_testing.post_shutdown_test()
--- a/ros2bag/test/test_record_qos_profiles.py
+++ b/ros2bag/test/test_record_qos_profiles.py
@@ -84,7 +84,7 @@ class TestRos2BagRecord(unittest.TestCase):
         output_path = Path(self.tmpdir.name) / 'ros2bag_test_basic'
         arguments = ['record', '-a', '--qos-profile-overrides-path', profile_path.as_posix(),
                      '--output', output_path.as_posix()]
-        expected_output = 'Listening for topics...'
+        expected_output = 'Recording...'
         with self.launch_bag_command(arguments=arguments) as bag_command:
             bag_command.wait_for_output(
                 condition=lambda output: expected_output in output,
@@ -99,7 +99,7 @@ class TestRos2BagRecord(unittest.TestCase):
         output_path = Path(self.tmpdir.name) / 'ros2bag_test_incomplete'
         arguments = ['record', '-a', '--qos-profile-overrides-path', profile_path.as_posix(),
                      '--output', output_path.as_posix()]
-        expected_output = 'Listening for topics...'
+        expected_output = 'Recording...'
         with self.launch_bag_command(arguments=arguments) as bag_command:
             bag_command.wait_for_output(
                 condition=lambda output: expected_output in output,
--- a/rosbag2_test_common/CMakeLists.txt
+++ b/rosbag2_test_common/CMakeLists.txt
@@ -42,8 +42,26 @@ install(
   DESTINATION include/${PROJECT_NAME})
 
 if(BUILD_TESTING)
+  find_package(ament_cmake_gmock REQUIRED)
   find_package(ament_lint_auto REQUIRED)
+  find_package(rcpputils REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  add_executable(loop_with_ctrl_c_handler test/rosbag2_test_common/loop_with_ctrl_c_handler.cpp)
+  install(
+      TARGETS loop_with_ctrl_c_handler
+      EXPORT export_loop_with_ctrl_c_handler
+      ARCHIVE DESTINATION lib
+      LIBRARY DESTINATION lib
+      RUNTIME DESTINATION bin
+  )
+
+  ament_add_gmock(test_process_execution_helpers
+    test/rosbag2_test_common/test_process_execution_helpers.cpp
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+  if(TARGET test_process_execution_helpers)
+    target_link_libraries(test_process_execution_helpers ${PROJECT_NAME})
+  endif()
 endif()
 
 ament_python_install_package(${PROJECT_NAME})
--- a/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_windows.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_windows.hpp
@@ -18,7 +18,7 @@
 #include <gmock/gmock.h>
 
 #include <direct.h>
-#include <Windows.h>
+#include <windows.h>
 
 #include <chrono>
 #include <csignal>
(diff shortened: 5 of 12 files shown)


clalancette (commented):
I wonder if this is something that we should put into `rclcpp` rather than here.  That is, it seems to me that all ROS 2 binaries would benefit from this kind of fix.

@MichaelOrlov @mjcarroll what do you think?

mjcarroll:
> I wonder if this is something that we should put into rclcpp rather than here. That is, it seems to me that all ROS 2 binaries would benefit from this kind of fix.

I think it makes sense to put a version of this in ROS 2, but I think @MichaelOrlov wants to get this fix in for Iron's version of rosbag2, so it seems reasonable to do it here as well for now.  Maybe in the longer term if we get something landed in rolling, he can deprecate it here to reduce redundancy?

MichaelOrlov:
@clalancette,  I agree with @mjcarroll and yes I would like to keep it in rosbag2 to be able to backport it to the iron branch. And will get rid of redundancy when it will be available in the `rclcpp` layer.

emersonknapp:
> And will get rid of redundancy when it will be available in the rclcpp layer.

@MichaelOrlov would you be willing to make the patch to add the helpers to `rclcpp`?

MichaelOrlov:
> @MichaelOrlov would you be willing to make the patch to add the helpers to `rclcpp`?

@emersonknapp Yes. I can try to make the changes in `rcpcpp` layer but a bit later on, it might need to make changes on the `rclpy` side as well.
I think after merging this PR.

MichaelOrlov:
- Update: Removed dependencies from https://github.com/ros2/rosbag2/pull/1301 and rebased on top of the latest Rolling to be able to move forward with this PR and easily backport it to the Iron and Humble.
@clalancette @emersonknapp Please review when you will have time.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/55adb4c4d5dac5435adace5c095d45b2/raw/09cc03757951ddc3c5ec8885a4d766a84f26207f/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py rosbag2_test_common ros2bag rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_py rosbag2_test_common ros2bag rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12152
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18847/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13388/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19601/)

MichaelOrlov:
Returning back to draft due to some unexpected test failures on CI which are not reproduced locally and need to be investigated.

MichaelOrlov:
Re-run CI after rebase and adjusting expectations in flaky ros2bag record tests
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12197
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18901/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19658/)

MichaelOrlov:
It seems that CI failure on windows in [rosbag2_tests.TestPlayEndToEnd/PlayEndToEndTestFixture.](https://ci.ros2.org/job/ci_windows/19658/testReport/junit/rosbag2_tests/TestPlayEndToEnd_PlayEndToEndTestFixture/play_end_to_end_exits_gracefully_on_sigint_0___sqlite3_______________________/) relates to the way how we are running tests on Windows CI. 
It fails with the error messages
```
C:\ci\ws\install\include\rosbag2_test_common\rosbag2_test_common\process_execution_helpers_windows.hpp(177): error: Value of: GenerateConsoleCtrlEvent(CTRL_C_EVENT, handle.process_info.dwProcessId)
  Actual: false
Expected: true
Error code:1. Error message: Incorrect function.
```
The error code is very weird and I can't reproduce this failure on my local Windows + ROS2 Rolling setup.
However, I found a discussion on the docker-related forum [Can't call GenerateConsoleCtrlEvent system call in Windows containers](https://github.com/docker/for-win/issues/3173#top)](https://github.com/docker/for-win/issues/3173) which is describes exactly the same problem with the `GenerateConsoleCtrlEvent` system call and its `Error code:1. Error message: Incorrect function.`
According to one of the reply 
> @lox have you tried Windows Server 2019? It is fixed in the ltsc2019 image.
> I can confirm this works in ltsc2019 images, closing!

It seems this issue has been fixed on the `Windows Server 2019` 
@mjcarroll @clalancette What version of the Windows and docker containers are we using for running our Windows CI?

MichaelOrlov:
@nuclearsandwich If I am not mistaken I recall that we had a plan to move to the `Windows Server 2019` on our Windows CI builds.  
Could you please comment about it if it is true? 
I came across with an issue that calling to `GenerateConsoleCtrlEvent` returns `Incorrect function` error when running it on Windows 10 under the docker image.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -204,6 +251,38 @@ class Recorder
       SIGINT, [](int /* signal */) {
         rosbag2_py::Recorder::cancel();
       });
+#ifdef _WIN32
```
emersonknapp, line 262: If we were to repeat this block a third time - I'd suggest that we pull it out somehow but for now in just 2 locations, it's OK

emersonknapp (approved):

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: Yeah, I had the same thoughts. But decided to keep it in place. It is just 1 line inside and with lambda, it looks more clear what is going on inside the signal handler.

MichaelOrlov:
Re-running CI after rebase since on today's ROS2 weekly synch meeting was mentioned that buildfarm CI already migrated to the Windows Server 2020.  
Will see if the issue still persists.
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12301
Building on the built-in node in workspace /var/lib/jenkins/jobs/ci_launcher/workspace
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19040/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13558/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19785/)

MichaelOrlov:
Re-run Windows CI job after adding `ctrl_c_event_can_be_send_and_received` sanity test 
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19807/)

MichaelOrlov:
Hi @claraberendsen, 
You mentioned to poing you if the issue with Windows not being able to send the `GenerateConsoleCtrlEvent` will reproduce after re-run on the fresh Windows build.
It is still failing https://ci.ros2.org/job/ci_windows/19807/
I have added the "minimum valuable reproducer"  `ctrl_c_event_can_be_send_and_received` test to show that issue reproduces even with a simple application with an endless loop without dependencies from other packages. Please refer to my latest commit [Add basic ctrl_c_event_can_be_send_and_received test](https://github.com/ros2/rosbag2/pull/1342/commits/0bcf9040f840840f287e36b7aa1697146bb9646a) for details.

The tests which are failing on the CI are passes on my local Windows setup without docker container.
Here is the link to a similar issue related to Windows under the docker [Can't call GenerateConsoleCtrlEvent system call in Windows containers](https://github.com/docker/for-win/issues/3173#top)](https://github.com/docker/for-win/issues/3173) use case.

It seems someone needs to contact Microsoft Support for this issue. Since I have no more ideas on how to fix this issue.
Could you please handle this issue or forward it to someone who can handle it?

claraberendsen:
> It seems someone needs to contact Microsoft Support for this issue. Since I have no more ideas on how to fix this issue. Could you please handle this issue or forward it to someone who can handle it?

@MichaelOrlov Thanks for pinging me I'm going to take a look at this and see what I find and if needed be forward it.

claraberendsen:
@MichaelOrlov I can confirm that we are using the new WS2022 image. In the logs of the [build](https://ci.ros2.org/job/ci_windows/19807/) you referenced above you can see the change on the url for the windows image to point to `/windows/server` which is a new built image for WS2022 ([See](https://hub.docker.com/_/microsoft-windows) as reference to the previous 2019 image).

MichaelOrlov:
@claraberendsen As regards
> I can confirm that we are using the new WS2022 image

I know. I have seen it from logs. The problem is that issue reproduces even on the new WS2022 image.  
This is why I mentioned that 
> It seems someone needs to contact Microsoft Support for this issue. Since I have no more ideas on how to fix this issue.

claraberendsen:
@MichaelOrlov We do not currently have the bandwidth to investigate this further. I see you pointed that in your local machine you are not seeing this error only when running the CI because we utilize docker. I don't have clarity that the issue comes from  using docker or the docker image of our setup.  To verify that the issue comes from it being ran inside docker further testing should be conducted by running this on a WS2019 image and a WS2022 image.  I'm pinging @clalancette to see if he can help unblock this.

MichaelOrlov:
@claraberendsen I don't see a rationale in trying to run the test on the different WS2019 and a WS2022 images under the docker.  
The fact that on our latest WS2022 image in the docker container, any call to `GenerateConsoleCtrlEvent` returns an Incorrect function error is enough.  
IMO the next step would be to reach out to Microsoft and say: "We are using your licensed latest WS2022 image under the docker container version ... and any call to the `GenerateConsoleCtrlEvent` returns an Incorrect function error". This behavior doesn't correspond to your spec written here https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent"

clalancette:
> I'm pinging @clalancette to see if he can help unblock this.

I don't have any particular knowledge about what is going on with Windows.  Maybe @ooeygui can help get us in touch with the right people?

ooeygui:
@clalancette I'm looking into it.

clalancette:
> @clalancette I'm looking into it.

Thank you, much appreciated!

clalancette:
@ooeygui Friendly ping.  Any thoughts here?

ooeygui:
@clalancette I've been trying to track this down internally - it looks like there was a fix, but it was backed out due to compatibility issues. I am trying to find an alternative method, but there does not appear to be a supported graceful solution.

ooeygui:
@clalancette I found the owner of the API. It looks like my statement about a reverted fix was incorrect, that this should work.

I'm working on a small repro to root cause.

ooeygui:
Both the owner of the Windows API and I have been attempting to reproduce this, but haven't been able to.
Does this still repro for you?

MichaelOrlov:
@ooeygui Originally this issue was not reproducible on my local setup. But as you can see CI jobs fail.

ooeygui:
Thanks @MichaelOrlov. @clalancette Do you know where I can find the windows container spec? I'd like to try building it locally.

nuclearsandwich-ai:
> Do you know where I can find the windows container spec? I'd like to try building it locally.

Our Windows containers are currently running on Windows Server 2022 hosts running in AWS EC2. We're using the Docker engine for Windows in the non-virtualized mode so we can take advantage of shared mounts between the container and host system (this forces our containers and the host Windows version to be the same, unlike the virtualized container engine which does not support shared filesystem mounts between host and container) on [ci.ros2.org](http://ci.ros2.org/) you can see our dockerfile template here: https://github.com/ros2/ci/blob/master/windows_docker_resources/Dockerfile and you can inspect the logs to see the build arguments that a given job is run with. Within the container most of the heavy lifting for setup and configuration is managed by this chef cookbook https://github.com/ros-infrastructure/ros2-cookbooks/tree/latest/cookbooks/ros2_windows

MichaelOrlov:
@ooeygui @nuclearsandwich-ai Any updates on this issue?

cc: @clalancette

ooeygui:
Hi @MichaelOrlov,
I spoke with the owner of the API which exhibits this issue. We have not been able to repduce this issue, so cannot debug it. Discussing the runtime environment, we believe the root cause is the version difference between the container and the host (Windows containers behave differently than linux containers; which requires that they be the same version of the OS)

In a future version of Windows, the API has been refactored so it should not exhibit this problem, but this depends on migrating the host and clients CI to a more current version of Windows. 

I'd recommend disabling the test on Windows until this migration takes place.

MichaelOrlov:
@ooeygui Do we have an issue to track when the migration of the host and clients CI to a more current version of Windows is going to happen?  
If not could you please create it?

I need to refer to something in the code when will be disabling tests for Windows.

BTW can we disable tests for CI and keep them running by default for the local setup?

MichaelOrlov:
@ooeygui @nuclearsandwich-ai @clalancette May I have an issue number to track when the migration of the host and clients CI to a more current version of Windows is going to happen? As from discussion above ^^^.
Long story short:
On our latest WS2022 image in the docker container, any call to `GenerateConsoleCtrlEvent` returns an Incorrect function error.
The actual behaviour doesn't correspond to the official spec found here https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent. It was promised here https://github.com/ros2/rosbag2/pull/1342#issuecomment-1876269502 that the issue was fixed in the latter versions of Windows.

If we will not have a follow-up issue for infrastructure - the fix will never happen. 

Also please note that disabling tests will defeat the purpose and fixes in this PR.

claraberendsen:
@MichaelOrlov @ooeygui do we know from what version this should be resolved? I cannot assess or give timing of a potential migration without this information.

MichaelOrlov:
@claraberendsen I don't know. It seems only @ooeygui knows.
@ooeygui Could you please elaborate on what Windows version this issue should have already been resolved?

PR #1354, opened 2023-05-26
Don't record sim-time messages before first /clock

### LINKED ISSUE #1276: Messages published before /clock recorded with timestamp 0 (--use-sim-time)
## Description

When recording the data with `--use-sim-time` the messages published before the `/clock` are recorded with timestamp 0 this results in false statistics and so on. See the graph from rqt_graph 
image

## Related Issues

-

## Completion Criteria

Data recording with consecutive timestamps

Add the option to ignore or some buffer or restamp. I don't know if the behavior is expected or not, please share your thoughts. This issue is more like a question than a Feature request

1. pub to `/clock` 
2. pause publishing
3. begin record some other topics and resume publishing time

emersonknapp:
Fixes https://github.com/ros2/rosbag2/issues/1276

When using sim-time, check that `now != 0` before writing any messages into the bag, to avoid the time jump that then happens on first `/clock` message. This potentially drops some messages that it could have written, but is more correct, in that those messages are "out of time" and have no temporal standing in simulated time.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -364,6 +364,18 @@ bool RecorderImpl::is_paused()
 
 void RecorderImpl::topics_discovery()
 {
+  // If using sim time - wait until /clock topic received before even creating subscriptions
+  if (record_options_.use_sim_time) {
+    RCLCPP_INFO(
+      node->get_logger(),
+      "use_sim_time set, waiting for /clock before starting recording...");
+    while (rclcpp::ok()) {
+      if (node->get_clock()->wait_until_started(record_options_.topic_polling_interval)) {
+        break;
+      }
+    }
+    RCLCPP_INFO(node->get_logger(), "Sim time /clock found, starting recording.");
+  }
   while (rclcpp::ok() && stop_discovery_ == false) {
     auto topics_to_subscribe =
       get_requested_or_available_topics();
@@ -449,11 +461,7 @@ RecorderImpl::create_subscription(
     qos,
     [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message) {
       if (!paused_.load()) {
-        auto now = node->get_clock()->now();
-        // When using sim time, do not record messages before receiving first /clock
-        if (!(record_options_.use_sim_time && now.nanoseconds() == 0)) {
-          writer_->write(message, topic_name, topic_type, now);
-        }
+        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
       }
     });
   return subscription;
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -449,7 +449,11 @@ RecorderImpl::create_subscription(
     qos,
     [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message) {
       if (!paused_.load()) {
-        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
+        auto now = node->get_clock()->now();
+        // When using sim time, do not record messages before receiving first /clock
+        if (!(record_options_.use_sim_time && now.nanoseconds() == 0)) {
+          writer_->write(message, topic_name, topic_type, now);
+        }
       }
     });
   return subscription;
--- a/README.md
+++ b/README.md
@@ -93,6 +93,15 @@ In the same fashion, this auto discovery can be disabled with `--no-discovery`.
 If not further specified, `ros2 bag record` will create a new folder named to the current time stamp and stores all data within this folder.
 A user defined name can be given with `-o, --output`.
 
+#### Simulation time
+
+In ROS 2, "simulation time" refers to publishing a clock value on the `/clock` topic, instead of using the system clock to tell time.
+By passing `--use-sim-time` argument to `ros2 bag record`, we turn on this option for the recording node.
+Messages written to the bag will use the latest received value of `/clock` for the timestamp of the recorded message.
+
+Note: Until the first `/clock` message is received, the recorder will not write any messages.
+Before that message is received, the time is 0, which leads to a significant time jump once simulation time begins, making the bag essentially unplayable if messages are written first with time 0 and then time N from `/clock`.
+
 #### Splitting recorded bag files
 
 rosbag2 offers the capability to split bag files when they reach a maximum size or after a specified duration. By default rosbag2 will record all data into a single bag file, but this can be changed using the CLI options.
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -129,7 +129,8 @@ class RecordVerb(VerbExtension):
             help='Start the recorder in a paused state.')
         parser.add_argument(
             '--use-sim-time', action='store_true', default=False,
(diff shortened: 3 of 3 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -449,7 +449,11 @@ RecorderImpl::create_subscription(
     qos,
     [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message) {
       if (!paused_.load()) {
-        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
+        auto now = node->get_clock()->now();
+        // When using sim time, do not record messages before receiving first /clock
+        if (!(record_options_.use_sim_time && now.nanoseconds() == 0)) {
```
MichaelOrlov:
```suggestion
        if (!(record_options_.use_sim_time && now.seconds() == 0 && now.nanoseconds() == 0)) {
```

emersonknapp: This is not necessary, it does not work the same as a `struct time_t`. `double seconds()` returns "seconds since epoch", and `int64_t nanoseconds()` returns "nanoseconds since epoch" - they are just different views into the same piece of information, and I prefer to use the integer value instead of incurring a division and trying to compare floats.

MichaelOrlov: Ok. this makes sense. Thanks for the explanation.

MichaelOrlov (approved):
Thanks for the PR.

MichaelOrlov (commented):
@emersonknapp A bunch of tests for use-sim-time start failing
```
 build/rosbag2_transport/test_results/rosbag2_transport/test_record_all_use_sim_time__rmw_fastrtps_cpp.gtest.xml: 1 test, 0 errors, 1 failure, 0 skipped
2023-05-26T01:21:28.7555631Z - rosbag2_transport.RecordIntegrationTestFixture record_all_with_sim_time
2023-05-26T01:21:28.7555951Z   <<< failure message
2023-05-26T01:21:28.7556306Z     /__w/rosbag2/rosbag2/ros_ws/src/nn632imx3xs/rosbag2/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp:85
2023-05-26T01:21:28.7556696Z     Value of: ret
2023-05-26T01:21:28.7556882Z       Actual: false
2023-05-26T01:21:28.7557085Z     Expected: true
2023-05-26T01:21:28.7557361Z     failed to capture expected messages in time. recorded messages = 9
2023-05-26T01:21:28.7557817Z     /__w/rosbag2/rosbag2/ros_ws/src/nn632imx3xs/rosbag2/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp:90
2023-05-26T01:21:28.7558171Z     Expected equality of these values:
2023-05-26T01:21:28.7558424Z       messages_per_topic[clock_topic]
2023-05-26T01:21:28.7558635Z         Which is: 4
2023-05-26T01:21:28.7558827Z       5u
2023-05-26T01:21:28.7559016Z         Which is: 5
2023-05-26T01:21:28.7559350Z     /__w/rosbag2/rosbag2/ros_ws/src/nn632imx3xs/rosbag2/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp:93
2023-05-26T01:21:28.7559698Z     Value of: recorded_messages
2023-05-26T01:21:28.7559927Z     Expected: size is >= 10
2023-05-26T01:21:28.7561306Z       Actual: { 16-byte object <10-7E D0-B8 3D-56 00-00 00-7E D0-B8 3D-56 00-00>, 16-byte object <60-50 D0-B8 3D-56 00-00 50-50 D0-B8 3D-56 00-00>, 16-byte object <80-55 D0-B8 3D-56 00-00 70-55 D0-B8 3D-56 00-00>, 16-byte object <40-47 D0-B8 3D-56 00-00 30-47 D0-B8 3D-56 00-00>, 16-byte object <A0-3F CD-B8 3D-56 00-00 90-3F CD-B8 3D-56 00-00>, 16-byte object <90-40 D0-B8 3D-56 00-00 80-40 D0-B8 3D-56 00-00>, 16-byte object <20-41 D0-B8 3D-56 00-00 10-41 D0-B8 3D-56 00-00>, 16-byte object <30-43 D0-B8 3D-56 00-00 20-43 D0-B8 3D-56 00-00>, 16-byte object <C0-43 D0-B8 3D-56 00-00 B0-43 D0-B8 3D-56 00-00> }, whose size 9 doesn't match
2023-05-26T01:21:28.7561953Z   >>>
```
Could you please investigate?

MichaelOrlov (changes requested):
Test fails need investigation

MichaelOrlov (approved):
LGTM with green CI if failing tests will be addressed.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -449,7 +449,11 @@ RecorderImpl::create_subscription(
...
+          writer_->write(message, topic_name, topic_type, now);
```
MichaelOrlov:
```suggestion
        // When using sim time, do not record messages before receiving the first message on the 
        // "/clock" topic
        if (!(record_options_.use_sim_time && !node->get_clock()->started())) {
          writer_->write(message, topic_name, topic_type, node->get_clock()->now());
```

MichaelOrlov (commented):
@emersonknapp I found out that node's clock has method `started()` - which is indicating if first message on the `"/clock"` topic has arrived It looks like a better way to determine if clock was primed rather than relying on the `clock()->now().nanoseconds == 0` condition.

Also, I found out that the node's clock has  `wait_until_started(timeout)` method
```
RCLCPP_PUBLIC
  bool
  wait_until_started(
    const rclcpp::Duration & timeout,
    Context::SharedPtr context = contexts::get_global_default_context(),
    const rclcpp::Duration & wait_tick_ns = rclcpp::Duration(0, static_cast<uint32_t>(1e7)));
```
Perhaps it would be even better to wait until clock will be started before creating subscriptions if using `use_sim_time` option. We can wait in the loop and check for `rclcpp::ok` in the `Recorder::record()` method and in the discovery thread just skip trying to subscribe on the topics and continue to sleep in the discovery thread.

What do you think?

emersonknapp:
Thanks for looking into that - I think that waiting for clock before starting discovery is a cleaner approach, I am rewriting to do that, and will address test issues

emersonknapp:
Pulls: ros2/rosbag2#1354
Gist: https://gist.githubusercontent.com/emersonknapp/e213a931a39a2f287543f266a7a9fdcc/raw/4f65cca6c2058ec607b06da56eef3d4da03c62eb/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12167
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18865/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19619/)

emersonknapp:
@mergifyio backport iron humble

emersonknapp:
Oops, that was premature - I thought of a couple edge cases that this implementation did not cover, while doing the backports. Reverting for now (https://github.com/ros2/rosbag2/pull/1377) - and reopened PR as https://github.com/ros2/rosbag2/pull/1378

Issue #773, opened 2021-05-28
Record service requests


kunaltyagi:
## Description
In ROS1, recording services was not implemented. However, actions were implemented on top of messages.

This made it possible to debug/enhance/test a system by recording just messages and replaying them. In ROS2, server client handshake is done over services. Would it be possible to implement a rosbag record of the sent goals?

## Related Issues
ROS1: https://github.com/ros/ros_comm/issues/250

## Completion Criteria
Successful trigger of action by following this sequence:
1. ros2 run action_tutorials_py fibonacci_action_server
2. ros2 bag record -a --include-hidden # or any other flag
3. ros2 run action_tutorials_py fibonacci_action_client
4. Close all nodes
5. ros2 run action_tutorials_py fibonacci_action_server
6. ros2 bag play rosbag2_<timestamp>
7. Wait to see if server responds

This sequence with ROS1 works if the goal messages are recorded. If this also works with ROS2, we can consider this feature as "ready"

## Implementation Notes / Suggestions
On the interface, we can have 3 different behaviors, all of which make sense from a certain perspective:
1. `ros2 bag record --include-actions` or `ros2 bag record --include-services` to allow replaying of service requests (most sensible, keeps concerns separate)
2. Make it default behavior to record action goal/cancellations but not services (same behavior from ROS1)
3. Extend the existing `ros2 bag record --include-hidden-topics` (better if not done this way)

Since the services and topics are implemented similarly at a RMW level, it should be possible to record the data sent via services.

On the other hand, just because all current RMW implementations use the same mechanism for topic and service communication, this might not hold true in future.

## Testing Notes / Suggestions
The integrated testing for this would be quite simple:
1. start service/action server
2. start recording
3. call service/action
4. stop recording and play the bag
5. expect a call for the service/action

I don't know how it'd be unit tested

emersonknapp:
Given that this is a feature request - I would appreciate if you could reformat using the [feature request template](https://github.com/ros2/rosbag2/blob/master/.github/ISSUE_TEMPLATE/feature_request.md) instead of the [bug report template](https://github.com/ros2/rosbag2/tree/master/.github/ISSUE_TEMPLATE) - it will be easier to understand.

Karsten1987:
@kunaltyagi while it's a valid feature request, it actually goes beyond the scope of rosbag2. For this to happen, we actually need support in ROS2 for serialized service (and action) calls. Unfortunately, up to today, we only have serialized messages.

[...]

PR #1359, opened 2023-05-29
[Design] Add design on recording and replay service

Barry-Xu-2018:
Related to #773

### DIFF
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -62,6 +62,7 @@ The description of the relevant parameter behavior.
 | --all | Record all topics and service event topics. Other hidden topics are excluded. |
 | --all-topics | Record all topics. Hidden topics including service event topics are excluded. |
 | --all-services | Only record all service event topics. |
+| --include-hidden-topics | Record all hidden topics. Include service event topic. |
 | --exclude-topics | Exclude topics containing provided regular expression. |
 | --exclude-services | Exclude services containing provided regular expression. |
 | -e REGEX, --regex REGEX | Record only topics and service containing provided regular expression. |
@@ -118,9 +119,25 @@ Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Request Count: XX |
 
 ### Expand the 'play' command
 
-Add 1 parameters.
+Add 3 parameters.
 - `--services ServiceName1 [[ServiceName2 ...]`
 
     Determine which service's requests are played.
 
-Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.
+- `--exclude-topics`
+
+    Rename from `--exclude`. Exclude topics containing provided regular expression.
+
+- `--exclude-services`
+
+    Exclude services containing provided regular expression.
+
+`-e REGEX, --regex REGEX` affects both topics and services.
+
+## Implementation Staging
+
+Implementation is going to be split into two phases.  
+On phase 1 is expected to be implemented all functionality which is related to the service events recording.  
+On phase 2 the rest functionality related to the service requests playback.  
+The service requests playback is not straightforward to implement since we don't have API from rclcpp layer to be able to publish service requests with dynamically created type support. We will need to create an analog of the `GenericPublisher` i.e. `GenericClient` on the `rclcpp` layer to be able to do this. To facilitate implementation of the `GenericClient` we likely will need to update get_typesupport_handle() to support service [ros2/rclcpp#2209](https://github.com/ros2/rclcpp/pull/2209).  
+Therefore, phase 2 will be implemented when the required new rclcpp API will be ready.
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -1,0 +1,73 @@
+# Rosbag2 record and replay service
+
+## Context
+
+In [ros2/rosbag2#773](https://github.com/ros2/rosbag2/issues/773), it was recommended that record service requests feature should be added to rosbag2.  
+
+This design describe how to implement this feature based on current architecture and how to extend current command parameter.  
+
+## How to save record service message
+
+Service introspection ([ros2/ros2#1285](https://github.com/ros2/ros2/issues/1285)) had been implemented. All service messages (requests and responses) are published to service introspection topic which is hidden topic named as `_service_event` under service name (e.g. /add_two_ints/_service_event). So all messages in this topic can be recorded like general topic. 
+
+## How to distinguish service introspection topic in all recorded topics 
+
+Service introspection is saved like general topic. But they should use different process after reading from recording file. So they should be distinguished by topic name.  
+
+- General topic  
+  
+  [NameSpace/]TopicName
+
+- Hidden topic  
+  
+  [NameSpace/]_TopicName
+
+- Service introspection topic
+  
+  [NameSpace/]ServiceName/_service_event
+
+So if the string ends with `/_service_event` in topic name, it should be service introspection topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service introspection must ends with "_Event" (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service introspection topic can be identified.
+
+### Expand the 'record' command
+
+Add 1 parameter.
+
+- `-S [SerivceName1 ServiceName2 ...]` or `--services [SerivceName1 ServiceName2 ...]`
+  
+    If the user doesn't specify service names, all service introspection topics are recorded.
+
+### Change output of 'Info' command
+
+Topic for service introspection should be listed distinctively, as below example.
+
+```
+Files:             123_0.mcap
+Bag size:          28.2 KiB
+Storage id:        mcap
+Duration:          15.59s
+Start:             May 19 2023 13:22:25.340 (1684473745.340)
+End:               May 19 2023 13:22:40.400 (1684473760.400)
+Messages:          60
+Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 16 | Serialization Format: cdr
+                   Topic: /events/write_split | Type: rosbag2_interfaces/msg/WriteSplitEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /parameter_events | Type: rcl_interfaces/msg/ParameterEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /rosout | Type: rcl_interfaces/msg/Log | Count: 44 | Serialization Format: cdr
+
+>>>> The below is new added items for service introspection <<<<<<<
+Service Events: XX  
+Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
+                     Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
+```
+
+### Expand the 'play' command
+
+Add 2 parameters.  
+- `-S [ServiceName1 ...]` or `--services [ServiceName1 ...]` (TBD: Not sure whether this is needed.)
+    
+    Replay service events. If there is no service name set, all recorded service events are played.  User can use `ros2 service echo ServiceName` to check different service events. 
+
+- `--services-requests [ServiceName1 ...]`
+
+    Replay request with recorded sequence and interval. If there is no service name set, all recorded service request are sent.
+
+Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.  
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -40,7 +40,7 @@ Add 3 parameters.
 
 - `--all-topics`
 
-    Record all topics. Hidden topic are excluded (include service event topics).
+    Record all topics. Hidden topics including service event topics are excluded.
 
 - `--all-services`
 
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -26,15 +26,17 @@ Service event topic is saved like general topic. But they should use different p
   
   [NameSpace/]ServiceName/_service_event
 
-So if the string ends with `/_service_event` in topic name, it should be service event topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service event topic must include `/srv/` and ends with `_Event` (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service event topic can be identified.
+The following criteria will be used to determine a service event topic:
+1. The topic ends with `/_service_event`
+2. The topic type name matches  `*/srv/*_Event` (e.g. `example_interfaces/srv/AddTwoInts_Event`)
 
 ### Expand the 'record' command
 
 Add 3 parameters.
 
-- `-S SerivceName1 [ServiceName2 ...]` or `--services SerivceName1 [ServiceName2 ...]`
+- `-S ServiceName1 [ServiceName2 ...]` or `--services ServiceName1 [ServiceName2 ...]`
   
-    If the user doesn't specify service names, all service event topics are recorded.
+    Specify which services will be recorded. User must specify at least one service.
 
 - `--all-topics`
 
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -73,7 +73,7 @@ A new parameter is added.
 
 - `-v` or `--verbose`
 
-    This parameter only affect service event topic. if `-v` or `--verbose` is set, info command shows the number of request and response for each services based on service event. Otherwise, only show the number of service event. Note that parsing the number of request and response need spent time. The duration of the parsing is related to the number of recorded service events.
+    This parameter only affects the service event topic. if `-v` or `--verbose` is set, the info command shows the number of requests and responses for each service based on the service event. Otherwise, only show the number of service events. Note that parsing the number of request and response need spent time. The duration of the parsing is related to the number of recorded service events.
 
 
 Without `-v` or `--verbose` parameter, info command shows as below example.
@@ -120,12 +120,12 @@ Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Request Count: XX |
 ### Expand the 'play' command
 
 Add 2 parameters.  
-- `-S [ServiceName1 ...]` or `--services [ServiceName1 ...]` (TBD: Not sure whether this is needed.)
-    
-    Replay service events. If there is no service name set, all recorded service events are played.  User can use `ros2 service echo ServiceName` to check different service events. 
+- `--services ServiceName1 [[ServiceName2 ...]`
 
-- `--services-requests [ServiceName1 ...]`
+    Decide which services will be played.
 
-    Replay request with recorded sequence and interval. If there is no service name set, all recorded service request are sent.
+- `--service-request-mode`
 
-Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.  
+    Replay service request with recorded sequence and interval instead of playing service event message.
+
+Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -8,11 +8,11 @@ This design describe how to implement this feature based on current architecture
 
 ## How to save record service message
 
-Service introspection ([ros2/ros2#1285](https://github.com/ros2/ros2/issues/1285)) had been implemented. All service messages (requests and responses) are published to service introspection topic which is hidden topic named as `_service_event` under service name (e.g. /add_two_ints/_service_event). So all messages in this topic can be recorded like general topic. 
+Service introspection ([ros2/ros2#1285](https://github.com/ros2/ros2/issues/1285)) had been implemented. All service messages (requests and responses) are published to service event topic which is hidden topic named as `_service_event` under service name (e.g. /add_two_ints/_service_event). So all messages in this topic can be recorded like general topic.
 
-## How to distinguish service introspection topic in all recorded topics 
+## How to distinguish service event topic in all recorded topics
 
-Service introspection is saved like general topic. But they should use different process after reading from recording file. So they should be distinguished by topic name.  
+Service event topic is saved like general topic. But they should use different process after reading from recording file. So they should be distinguished by topic name.
 
 - General topic  
   
@@ -22,23 +22,48 @@ Service introspection is saved like general topic. But they should use different
   
   [NameSpace/]_TopicName
 
-- Service introspection topic
+- Service event topic
   
   [NameSpace/]ServiceName/_service_event
 
-So if the string ends with `/_service_event` in topic name, it should be service introspection topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service introspection must ends with "_Event" (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service introspection topic can be identified.
+So if the string ends with `/_service_event` in topic name, it should be service event topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service event topic must include `/srv/` and ends with `_Event` (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service event topic can be identified.
 
 ### Expand the 'record' command
 
-Add 1 parameter.
+Add 3 parameters.
 
-- `-S [SerivceName1 ServiceName2 ...]` or `--services [SerivceName1 ServiceName2 ...]`
+- `-S SerivceName1 [ServiceName2 ...]` or `--services SerivceName1 [ServiceName2 ...]`
   
-    If the user doesn't specify service names, all service introspection topics are recorded.
+    If the user doesn't specify service names, all service event topics are recorded.
+
+- `--all-topics`
+
+    Record all topics. Hidden topic are excluded (include service event topics).
+
+- `--all-services`
+
+    Record all service event topics.
+
+The description of the relevant parameter behavior.
+
+| Parameter | Description |
+| :-- | :--|
+|--include-hidden-topics| Record all hidden topics. Include service event topic. |
+|--services ServiceName1 [ServiceName2 ...] | Record services (service event topics) with specified service names. |
+| --all | Record all topics and service event topics. Other hidden topics are excluded. |
+| --all-topics | Record all topics. Hidden topic are excluded (include service event topic). |
+| --all-services | Only record all service event topics. |
 
 ### Change output of 'Info' command
 
-Topic for service introspection should be listed distinctively, as below example.
+A new parameter is added.
+
+- `-v` or `--verbose`
+
+    This parameter only affect service event topic. if `-v` or `--verbose` is set, info command shows the number of request and response for each services based on service event. Otherwise, only show the number of service event. Note that parsing the number of request and response need spent time. The duration of the parsing is related to the number of recorded service events.
+
+
+Without `-v` or `--verbose` parameter, info command shows as below example.
 
 ```
 Files:             123_0.mcap
@@ -53,10 +78,30 @@ Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 16 | Ser
                    Topic: /parameter_events | Type: rcl_interfaces/msg/ParameterEvent | Count: 0 | Serialization Format: cdr
                    Topic: /rosout | Type: rcl_interfaces/msg/Log | Count: 44 | Serialization Format: cdr
 
->>>> The below is new added items for service introspection <<<<<<<
-Service Events: XX  
+>>>> The below is new added items for service event topic <<<<<<<
+Service : XX  <== The number of service
 Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
                      Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
+```
+
+With `-v` or `--verbose` parameter, info command shows as below example.
+```
+Files:             123_0.mcap
+Bag size:          28.2 KiB
+Storage id:        mcap
+Duration:          15.59s
+Start:             May 19 2023 13:22:25.340 (1684473745.340)
+End:               May 19 2023 13:22:40.400 (1684473760.400)
+Messages:          60
+Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 16 | Serialization Format: cdr
+                   Topic: /events/write_split | Type: rosbag2_interfaces/msg/WriteSplitEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /parameter_events | Type: rcl_interfaces/msg/ParameterEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /rosout | Type: rcl_interfaces/msg/Log | Count: 44 | Serialization Format: cdr
+
+>>>> The below is new added items for service <<<<<<<
+Service : XX  <== The number of service
+Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Request Count: XX | Response Count: XX | Serialization Format: XX
+                     Service: /xxx/xxx | Type: xxx/xxx/xxx | Request Count: XX | Response Count: XX | Serialization Format: XX
 ```
 
 ### Expand the 'play' command
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -32,9 +32,9 @@ The following criteria will be used to determine a service event topic:
 
 ### Expand the 'record' command
 
-Add 3 parameters.
+Add 5 parameters.
 
-- `-S ServiceName1 [ServiceName2 ...]` or `--services ServiceName1 [ServiceName2 ...]`
+- `--services ServiceName1 [ServiceName2 ...]`
   
     Specify which services will be recorded. User must specify at least one service.
 
@@ -46,6 +46,14 @@ Add 3 parameters.
 
     Record all service event topics.
 
+- `--exclude-topics`
+
+    Rename from `--exclude`. Exclude topics containing provided regular expression.
+
+- `--exclude-services`
+
+    Exclude services containing provided regular expression.
+
 The description of the relevant parameter behavior.
 
 | Parameter | Description |
@@ -53,8 +61,11 @@ The description of the relevant parameter behavior.
 |--include-hidden-topics| Record all hidden topics. Include service event topic. |
 |--services ServiceName1 [ServiceName2 ...] | Record services (service event topics) with specified service names. |
 | --all | Record all topics and service event topics. Other hidden topics are excluded. |
-| --all-topics | Record all topics. Hidden topic are excluded (include service event topic). |
+| --all-topics | Record all topics. Hidden topics including service event topics are excluded. |
 | --all-services | Only record all service event topics. |
+| --exclude-topics | Exclude topics containing provided regular expression. |
+| --exclude-services | Exclude services containing provided regular expression. |
+| -e REGEX, --regex REGEX | Record only topics and service containing provided regular expression. |
 
 ### Change output of 'Info' command
 
--- a/docs/design/rosbag2_record_replay_service.md
+++ b/docs/design/rosbag2_record_replay_service.md
@@ -58,7 +58,6 @@ The description of the relevant parameter behavior.
 
 | Parameter | Description |
 | :-- | :--|
-|--include-hidden-topics| Record all hidden topics. Include service event topic. |
 |--services ServiceName1 [ServiceName2 ...] | Record services (service event topics) with specified service names. |
 | --all | Record all topics and service event topics. Other hidden topics are excluded. |
 | --all-topics | Record all topics. Hidden topics including service event topics are excluded. |
@@ -119,13 +118,9 @@ Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Request Count: XX |
 
 ### Expand the 'play' command
 
-Add 2 parameters.  
+Add 1 parameters.
 - `--services ServiceName1 [[ServiceName2 ...]`
 
-    Decide which services will be played.
-
-- `--service-request-mode`
-
-    Replay service request with recorded sequence and interval instead of playing service event message.
+    Determine which service's requests are played.
 
 Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.


docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+  
+  [NameSpace/]TopicName
+
+- Hidden topic  
+  
+  [NameSpace/]_TopicName
+
+- Service introspection topic
+  
+  [NameSpace/]ServiceName/_service_event
```
fujitatomoya, line 27: if we specify `ros2 bag record --include-hidden-topics`, this will be included as hidden topic currently? if that is so, `/_service_event` should be excluded from topic category for rosbag2 by default?

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+
+So if the string ends with `/_service_event` in topic name, it should be service introspection topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service introspection must ends with "_Event" (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service introspection topic can be identified.
+
+### Expand the 'record' command
+
+Add 1 parameter.
+
+- `-S [SerivceName1 ServiceName2 ...]` or `--services [SerivceName1 ServiceName2 ...]`
+  
+    If the user doesn't specify service names, all service introspection topics are recorded.
```
fujitatomoya: as user experience, this could be opposite? i think most likely that user wants to record the topics but service. so having this option as `opt-in` would be better to avoid unnecessary topics to be recorded?

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+Messages:          60
+Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 16 | Serialization Format: cdr
+                   Topic: /events/write_split | Type: rosbag2_interfaces/msg/WriteSplitEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /parameter_events | Type: rcl_interfaces/msg/ParameterEvent | Count: 0 | Serialization Format: cdr
+                   Topic: /rosout | Type: rcl_interfaces/msg/Log | Count: 44 | Serialization Format: cdr
+
+>>>> The below is new added items for service introspection <<<<<<<
+Service Events: XX  
+Service information: Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
+                     Service: /xxx/xxx | Type: xxx/xxx/xxx | Event Count: XX | Serialization Format: XX
```
fujitatomoya, line 97: this incudes everything as `event`? i would expect that it can show `request` and `response` separately, so that user can understand how many requests and responses are recorded via `info` command?

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+### Expand the 'play' command
+
+Add 2 parameters.  
+- `-S [ServiceName1 ...]` or `--services [ServiceName1 ...]` (TBD: Not sure whether this is needed.)
+    
+    Replay service events. If there is no service name set, all recorded service events are played.  User can use `ros2 service echo ServiceName` to check different service events.
```
fujitatomoya: I believe that this is related to `motivation` section, i am not sure the actual use case for this.

one think i can think of is, publishing all service events recorded in rosbag2 to transmit to the another remote node running over the network to share. i do not think this is major use case for this... any thoughts? i am not even sure if we need to have this option.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+### Expand the 'play' command
+
+Add 2 parameters.  
+- `-S [ServiceName1 ...]` or `--services [ServiceName1 ...]` (TBD: Not sure whether this is needed.)
+    
+    Replay service events. If there is no service name set, all recorded service events are played.  User can use `ros2 service echo ServiceName` to check different service events. 
+
+- `--services-requests [ServiceName1 ...]`
+
+    Replay request with recorded sequence and interval. If there is no service name set, all recorded service request are sent.
```
fujitatomoya: i think user likely wants to do is to debug the service server in this case.
`rosbag2` can only replay service requests in the bag file but not print the response result from the service server.
so user needs to start `ros2 service echo` in another terminal to see the service request and response to evaluate.
which is not related to design or implementation for rosbag2, but i think that would be better to describe as practice?

fujitatomoya (commented):
@Barry-Xu-2018 i have several comments.

@ihasdapie @emersonknapp thanks for sharing the status, since we have been away from rosbag2 activity for a while, it would be really appreciated if you can share details and thoughts on this.

CC: @MichaelOrlov

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Yes. If the user doesn't want to record service, service event topic ([NameSpace/]ServiceName/_service_event) is ignored even if the user set `--include-hidden-topics`.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Yes. 
Users can decide whether to set `-S` or `--services` or not.   
If user doesn't set `-S or --services`, rosbag doesn't record service (service event topic).

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: When retrieving information from requests and responses, it requires deserializing the recorded message and parsing the data inside.   
This process can slow down the "info" command when dealing with large recording files (record many service events).   
However, we can consider adding a new parameter that allows users to specify whether they want to retrieve detailed service information. What do you think?

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Yes.  I'm not sure if this is indeed an actual use case.  
Let's listen to the opinions of others.

docs/design/rosbag2_record_replay_service.md
ihasdapie: IMO it would be more consistent to include service events in `--include-hidden-topics` since it is, after all, a hidden topic and not something special

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Hidden topic  
+  
+  [NameSpace/]_TopicName
+
+- Service introspection topic
+  
+  [NameSpace/]ServiceName/_service_event
+
+So if the string ends with `/_service_event` in topic name, it should be service introspection topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service introspection must ends with "_Event" (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service introspection topic can be identified.
```
ihasdapie: Checking for the `/srv`/ would be more robust

docs/design/rosbag2_record_replay_service.md
ihasdapie: What will the default behaviour with --all? I take that it will include service events too?

docs/design/rosbag2_record_replay_service.md
ihasdapie: I think it makes sense to include the number of request/responses by default. I can see it being useful for users who might want to check if a bag includes requests or reponses, or if they want to investigate dropped responses.  Plus, `ros2 bag info` isn't the most performance critical command anyways

docs/design/rosbag2_record_replay_service.md
ihasdapie: I think this is already implictly handled via `--topics` so there's no real need for this.

ihasdapie (commented):
Few comments. I think some additional opinions on playback behaviour should be considered before doing anything concrete

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > IMO it would be more consistent to include service events in --include-hidden-topics since it is, after all, a hidden topic and not something special

In `rosbag2 info`, service information will be listed separately. So I prefer not to categorize service event topics as general hidden topics.  From the user's perspective, service event topics are hidden.  
What do you think ?

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: I don't want to change previous behavior. So '--all' only means all topic (Not include special service event topic).  
To make it clearer, '--all' may be changed to '--all-topics'.   
Or like service parameter, '--all' and specified topics are combined to '--topics [TopicName ...]' (If no topic name is specified, all topics are recorded). But this change is a little big for user.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > I think it makes sense to include the number of request/responses by default. 

Okay. 

> This process can slow down the "info" command when dealing with large recording files (record many service events).
However, we can consider adding a new parameter that allows users to specify whether they want to retrieve detailed service information. What do you think?

Sorry. I forgot that metadata file was created after recording. Typically, this file exists, which eliminates the need for re-analyzing the recording file. That is, info can be parsed from this file.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > I think this is already implictly handled via --topics so there's no real need for this.

In this design, my thought is that the service event topic is not a regular topic. Therefore, users should not be aware of the existence of this topic (for example, the 'info' command will not display information about this topic, instead users can only see the information related to service request/response).

Barry-Xu-2018:
While recording finish, a YAML file will be created to save some information for rosbag info.     
After the above discussion, we want to provide the number of request/response for service. To get this information, we must deserialize the service event message and parse the content to get request/response info.  
In order to keep record performance, I think deserialization should not be processed during record phase.  

So, after recording, created YAML file only save service name without request/response info.  While executing rosbag info command, record file will be scanned (deserialization and parse content for service event message) and get the number of request/response. And then update YAML file. That is, the next time you execute "rosbag info," you will have complete information without the need to scan the recording file again.  

Of course, this approach would result in users not being able to obtain the complete information of services when they immediately view the YAML file after recording.   

If you have any better suggestions, please share them with me.

MichaelOrlov:
@Barry-Xu-2018 Thanks for initiating discussion about service introspection recording and replay.
I have added item about it in agneda of the next Rosbag2 WG meeteng https://docs.google.com/document/d/1Dsg_9XZQPhihpKQGQWMYTz2doGH4P2cAaNqr60cuNgw/edit#heading=h.v4jyd0vn7d0k which is scheduled on tomorrow `Friday 9:00 PST`. 
I would be glad to have a live discussion about it. 
If timing is not good for you please let me know, perhaps we can reschedule meeting on the Monday afternoon PST.

Meanwhile I have a few comments in regards to the design in general.
1. I agree that deserilizing and trying to parse any messages during recording is likely will cost in the performance penalty and eventually may lead to the highher probability of the messages drops.
2. Updating YAML file after calling `ros2 bag2 info` command looks doable but not very nice by design. First of all it might be turns out that there are no message definition for the service type which is going to be deserialized on the machine where calling `ros2 bag info` command. We can do writing a new yaml file and deleting the old one. But the tricky part is that `metadata.yaml` file duplicatied inside bag file and we can't change it at least for MCAP backend.
3. Many questions arise for replay case. How to replay servcies? As far as I understand from the design doc it will be desirable to replay service requests. But how to implement this? 
We simply don't have API for this, because service requests was wrapped in to the `service event` and we store it in the serialized form.

Barry-Xu-2018:
@MichaelOrlov 

> If timing is not good for you please let me know, perhaps we can reschedule meeting on the Monday afternoon PST.

It is 12:00 am Saturday in my local time.   
I think I can attend this meeting.

Barry-Xu-2018:
> 2. Updating YAML file after calling ros2 bag2 info command looks doable but not very nice by design. First of all it might be turns out that there are no message definition for the service type which is going to be deserialized on the machine where calling ros2 bag info command. 

Service type can be gotten while recording phase (according to service event type). So it can be stored to metadata.yaml file like service name.  Besides, service type can be gotten from MCAP file. 

> We can do writing a new yaml file and deleting the old one. But the tricky part is that metadata.yaml file duplicatied inside bag file and we can't change it at least for MCAP backend.

Do you think the information in the metadata.yaml file and the mcap file is different after only modifying metadata.yaml file?  
In mcap file, it recorded service event message. In metadata.yaml, we want to show the request/response information which is gotten by analyzing service event message. So I think the information isn't different.   

> 3. Many questions arise for replay case. How to replay servcies? As far as I understand from the design doc it will be desirable to replay service requests. But how to implement this?
We simply don't have API for this, because service requests was wrapped in to the service event and we store it in the serialized form.

Yes. There is no API.   
Now I am considering to use type introspection information to pickup request contents from service event.

MichaelOrlov:
@Barry-Xu-2018 Instead of modifying `metadata.yaml` file after parsing service event messages in `ros2 bag info` I want to suggest adding `--verbose` CLI option.
The idea is that regular `ros2 bag info` will not parse messages and will output statistics about service events. 
While `ros2 bag info` with `--verbose` key will do parsing of the service events and provide an extended version of the statistics with a breakout on the number of service requests/responses. And we will not update `metadata.yaml` for consistency.

Barry-Xu-2018:
@MichaelOrlov 

> Instead of modifying metadata.yaml file after parsing service event messages in ros2 bag info I want to suggest adding --verbose CLI option.
The idea is that regular ros2 bag info will not parse messages and will output statistics about service events.
While ros2 bag info with --verbose key will do parsing of the service events and provide an extended version of the statistics with a breakout on the number of service requests/responses. And we will not update metadata.yaml for consistency.

Totally agree. Thanks.
There is a little bad user experience for this scenario. User want to list detailed info more than once for a big record file. Parsing the number of request/response take a little more time when dealing with large recording file. Due to the absence of saved parsing information, it will take time each time to display detailed information.

Barry-Xu-2018:
BTW, there is a quick way to parsing the number of request/response information.  

Use AddTwoInts demo as an example.  
About AddTwoInts_Event
```
struct AddTwoInts_Event_ {
...
  using _info_type =
    service_msgs::msg::ServiceEventInfo_<ContainerAllocator>;
  _info_type info;
  using _request_type =
    rosidl_runtime_cpp::BoundedVector<example_interfaces::srv::AddTwoInts_Request_<ContainerAllocator>, 1, typename std::allocator_traits<ContainerAllocator>::template rebind_alloc<example_interfaces::srv::AddTwoInts_Request_<ContainerAllocator>>>;
  _request_type request;
  using _response_type =
    rosidl_runtime_cpp::BoundedVector<example_interfaces::srv::AddTwoInts_Response_<ContainerAllocator>, 1, typename std::allocator_traits<ContainerAllocator>::template rebind_alloc<example_interfaces::srv::AddTwoInts_Response_<ContainerAllocator>>>;
  _response_type response;
...
}
```

The information regarding the request and response can be obtained from `service_msgs::msg::ServiceEventInfo_`.   
Deserializing `request` and `response` is unnecessary.   
`ServiceEventInfo` is the first item in AddTwoInts_Event.  
So, regardless of how the request and response are defined, we only need to deserialize ServiceEventInfo instead of all event message.  
I had tried the below code, it can correctly get event type from service event topic.  
```c++
auto message_info = service_msgs::msg::ServiceEventInfo();
const rosidl_message_type_support_t * type_support_info =
    rosidl_typesupport_cpp::get_message_type_support_handle<service_msgs::msg::ServiceEventInfo>();
// `info` member is located in the first item in `example_interfaces::srv::AddTwoInts_Event`
ret = rmw_deserialize(
   &(serialized_message->get_rcl_serialized_message()),
   type_support_info,
   &message_info);
 if (ret == RMW_RET_OK) {
          RCLCPP_INFO(this->get_logger(), "Receive an event message, rmw_deserialize with `ServiceEventInfo`, type is: '%d'", message_info.event_type);
  } else {
     RCLCPP_INFO(this->get_logger(), "failed to rmw_deserialize");
  }
```

Since ServiceEventInfo has a fewer items (that is, Deserialization is fast), it is possible to get the number of request/response while recording.  
```
# The type of event this message represents
uint8 event_type

# Timestamp for when the event occurred (sent or received time)
builtin_interfaces/Time stamp

# Unique identifier for the client that sent the service request
# Note, this is only unique for the current session
unique_identifier_msgs/UUID client_id

# Sequence number for the request
# Combined with the client ID, this creates a unique ID for the service transaction
int64 sequence_number
```

fujitatomoya:
@MichaelOrlov @Barry-Xu-2018 I cannot make it today's call, so if this takes place i will catch up with you later, thanks!

MichaelOrlov:
@Barry-Xu-2018 Am i understand correctly that in
```
ret = rmw_deserialize(
   &(serialized_message->get_rcl_serialized_message()),
   type_support_info,
   &message_info);
```
`serialized_message->get_rcl_serialized_message())` - is supposed to be a recorded message with service event?

Barry-Xu-2018:
> serialized_message->get_rcl_serialized_message()) - is supposed to be a recorded message with service event?

Yes. You are right.

MichaelOrlov:
@fujitatomoya @emersonknapp @ihasdapie  We met with @Barry-Xu-2018 on today's Rosbag2 and Tooling WG meeting and come to the conclusion about the following:
1. Decided to provide `--verbose` parameter for `ros2 bag info` and don’t update `metadata.yaml` file in regards to the number of service requests and responses when running info with the verbose option. Will parse service event messages each time when calling info with the `--verbose` option. It shouldn't take significant time even for the large bags since usually there are not too many service events and we will read out and skip most of the recorded messages and it's going to be pretty fast.
2. Decided that `--include-hidden-topics` will record service events as well, since it is just a hidden topic.
3. To be decided yet about if  `ros2 record --service` without service names should record all service events. On one hand it would be user-friendly to record all services if no service names are mentioned explicitly, however from another hand it's not straightforward to implement since the parameter for the services is going to be a vector of strings with names - and an empty vector will be equal to the situation when no `--service` parameter at all. 
I am more inclined to not record all services if there are no service names after the `--service` parameter. Users will be able to record all service events with the `--include-hidden-topics` parameter or perhaps `--all` parameter.
5. `--all` CLI ? To be decided how `--all` and `--service` and `--include-hidden-topics` will coexist together. Need some time for consideration and discussion.
On one hand from a consistency point of view I am inclined to the option when `--all` will record only topics as before and will not record service events since they are just a hidden topics and we have `--include-hidden-topics` for them. However from another hand, if look from the user-friendliness point of view probably the user will expect that `--all` will record service events since there exists `--service` parameter and the user may not be aware that "service" in this case is just a specific hidden topic.
Open to feedback and opinion about this item.
7. We are not able to replay dedicated service requests due to the lack of the API currently because service requests are wrapped inside the service event and they are serialized. We decided that possibility of playback service requests needs to be further investigated and prototyped before we will make a final decision about including this option in the design.  
@Barry-Xu-2018 volunteer to try prototyping this option.

P.S. Your feedback and opinion are welcome.

fujitatomoya:
@MichaelOrlov @Barry-Xu-2018  thanks for leading this.

> Decided to provide --verbose parameter for ros2 bag info and don’t update metadata.yaml

agree on not having it update metadata.yaml, and this behavior.

(just a comment) `--verbose` seems to be general option for topics as well, so it should be described what it is for with `help message`.

> Decided that --include-hidden-topics will record service events as well

agree. if this option applies, it should collect all hidden topics.

> I am more inclined to not record all services if there are no service names after the --service parameter.
> Users will be able to record all service events with the --include-hidden-topics parameter or perhaps --all parameter.

IMO,

- `--service` is the option to explicitly specify the target service. if that is NULL, that returns as `invalid argument`.
- `--all` to record all topics and services. as developer point of view, it is understandable that `service_events` is hidden topic, but user. probably user expects that service is just different `normal` entity as well as topics when we support `recording services` feature.
- if `--all` is specified, that prevails to `--service` to record all services as well.

> We decided that possibility of playback service requests needs to be further investigated and prototyped before we will make a final decision about including this option in the design.

sounds good.

Barry-Xu-2018:
@fujitatomoya @MichaelOrlov 

About record parameter for service,I list them based on current discussion result.
| Parameter | Description |
| :-- | :--|
|--include-hidden-topics| Record all hidden topics. Include service event topic. |
| --all | Record all topics and service event topic. Other hidden topics are excluded. |
| --services ServiceName [ServiceName ...] | Record service (service event topic) with specified service name. |

If users want to only record all services, there is no single parameter that can fulfill this requirement. The closest parameter to the recent requirement is "--include-hidden-topics."

MichaelOrlov:
In this case, it will arise the same question if the user will want to record only all topics without services, there is no single parameter that can fulfill this requirement.

Barry-Xu-2018:
> In this case, it will arise the same question if the user will want to record only all topics without services, there is no single parameter that can fulfill this requirement.

Yes.  
Maybe should add 2 new parameters for all. What do you think ?
| Parameter | Description |
| :-- | :--|
| --all | Record all topics and service event topic. Other hidden topics are excluded. |
| --all-topics | Only record all topics. Hidden topic are excluded. |
| --all-services | Only record all service event topics. |

MichaelOrlov:
> Maybe should add 2 new parameters for all. What do you think ?

Adding `--all-topics` and `--all-services` along with `--all` sounds logically reasonable. :+1:

Barry-Xu-2018:
> We are not able to replay dedicated service requests due to the lack of the API currently because service requests are wrapped inside the service event and they are serialized. We decided that possibility of playback service requests needs to be further investigated and prototyped before we will make a final decision about including this option in the design.

How to unwrap request part in service event message.   
(I had written code to verify this way.)  

Assume we know the below information
1. service event topic  --- `/add_two_ints/_service_event`
2. service event type --- `example_interfaces/srv/AddTwoInts_Event`

```c++
// Use type support introspection 
std::shared_ptr<rcpputils::SharedLibrary>    ts_lib = rclcpp::get_typesupport_library(
      "example_interfaces/srv/AddTwoInts_Event", "rosidl_typesupport_cpp");
    
const rosidl_message_type_support_t *   ts = rclcpp::get_typesupport_handle(
      "example_interfaces/srv/AddTwoInts_Event",
      "rosidl_typesupport_cpp",
      *ts_lib);

const rosidl_message_type_support_t *  ts_handle = get_message_typesupport_handle(
      ts,
      rosidl_typesupport_introspection_cpp::typesupport_identifier);
    
const rosidl_typesupport_introspection_cpp::MessageMembers *  message_members = (rosidl_typesupport_introspection_cpp::MessageMembers *)ts_handle->data;

// Allocate memory for deseriliazed message
void * ros_message = (void *)malloc(message_members->size_of_);
      message_members->init_function(
ros_message, rosidl_runtime_cpp::MessageInitialization::ZERO);

// serialized_msg_ptr (std::shared_ptr<rcutils_uint8_array_t>) points to one service event seriliazed message

// Deserialize service event message
rmw_deserialize(serialized_msg_ptr.get(), ts, ros_message);

// Now ros_message point to all deseriliazed service topic event

// Get the offset part of request
// Now the size of ServiceEventInfo is fixed value. So the offset can be only calculated once.
//
// The strcut of service event message 
// {
//    ServiceEventInfo info
//    example_interfaces::srv::AddTwoInts_Request_ request
//    example_interfaces::srv::AddTwoInts_Response_ response
// }
// 
// Service Event Info structure
// {
//   uint8 event_type
//   builtin_interfaces/Time stamp
//   char[16] client_gid
//   int64 sequence_number
// }

size_t request_offset = 0;
size_t request_offset_tmp = 0;
  for(size_t i=0; i < message_members->member_count_; i++) {
    auto message_member = message_members->members_[i];
    if (message_member.name_ == std::string("request")) {
      request_offset = request_offset_tmp;
    } else {
      auto sub_members = (rosidl_typesupport_introspection_cpp::MessageMembers *)message_member.members_->data;
      request_offset_tmp += sub_members->size_of_;
    }
   if (request_offset != 0) {
      break;
    }
  }

// Create node  
auto node = std::make_shared<rclcpp::Node>("send_request_topic");

rclcpp::QoS qos = rclcpp::ServicesQoS();
rcl_client_options_t options = rcl_client_get_default_options();
options.qos = qos.get_rmw_qos_profile();

// Create client
// GenericClient should be new created. It can accept service type string to create client
// async_send_request() can accept void * as input message.
auto client = std::make_shared<rclcpp::GenericClient>(
    &*node->get_node_base_interface(),
     node->get_node_graph_interface(),
    "add_two_ints",   // Get from service event topic
    "example_interfaces/srv/AddTwoInts",  // Get from service event type
    options);

 // Check if service is ready
 while (!client->service_is_ready()) {
     std::printf("wait for service ready !\n");
     std::this_thread::sleep_for(std::chrono::seconds(1));
 }

// Send request message
// Input data will be serilized according to type example_interfaces::srv::AddTwoInts_Request_. 
 client->async_send_request((void *)((uint8_t *)ros_message + request_offset));
```

The definition of class GenericClient likes
```c++
namespace rclcpp {
class GenericClient : public ClientBase {
public:  
  GenericClient(
    rclcpp::node_interfaces::NodeBaseInterface * node_base,
    rclcpp::node_interfaces::NodeGraphInterface::SharedPtr node_graph,
    const std::string & service_name,
    const std::string & service_type,
    rcl_client_options_t & client_options)
  : ClientBase(node_base, node_graph)
  {
    ts_lib_ = rclcpp::get_typesupport_library(
      service_type, "rosidl_typesupport_cpp");
    
    // Now rclcpp::get_typesupport_handle only support message.  
    // We should make it support service
    auto srv_type_support = rclcpp::get_typesupport_handle_srv(
      service_type,
      "rosidl_typesupport_cpp",
      *ts_lib_);
    
    srv_type_support_handle_ = get_service_typesupport_handle(srv_type_support, "rosidl_typesupport_cpp");

    auto request_typesupport = srv_type_support_handle_->request_typesupport;
    request_members_ = (rosidl_typesupport_introspection_cpp::MessageMembers *)request_typesupport->data;

    auto response_typesupport = srv_type_support_handle_->response_typesupport;
    response_members_ = (rosidl_typesupport_introspection_cpp::MessageMembers *)response_typesupport->data;

    rcl_ret_t ret = rcl_client_init(
      this->get_client_handle().get(),
      this->get_rcl_node_handle(),
      srv_type_support_handle_,
      service_name.c_str(),
      &client_options);
    if (ret != RCL_RET_OK) {
      if (ret == RCL_RET_SERVICE_NAME_INVALID) {
        auto rcl_node_handle = this->get_rcl_node_handle();
        // this will throw on any validation problem
        rcl_reset_error();
        expand_topic_or_service_name(
          service_name,
          rcl_node_get_name(rcl_node_handle),
          rcl_node_get_namespace(rcl_node_handle),
          true);
      }
      rclcpp::exceptions::throw_from_rcl_error(ret, "could not create generic client");
    }
  }

  std::shared_ptr<void>
  create_response() override
  {
    auto response = (void *)malloc(response_members_->size_of_);
    response_members_->init_function(response, rosidl_runtime_cpp::MessageInitialization::ZERO);
    return std::shared_ptr<void>(response, [](void *p){free(p);});
  }

  std::shared_ptr<rmw_request_id_t>
  create_request_header() override
  {
    // TODO(wjwwood): This should probably use rmw_request_id's allocator.
    //                (since it is a C type)
    return std::shared_ptr<rmw_request_id_t>(new rmw_request_id_t);
  }

  void
  handle_response(
    std::shared_ptr<rmw_request_id_t> request_header,
    std::shared_ptr<void> response) override
  {
    erase_pending_request(request_header->sequence_number);
  }

  int64_t
  async_send_request(const void * request)
  {
    int64_t sequence_number;
    std::lock_guard<std::mutex> lock(pending_requests_mutex_);
    rcl_ret_t ret = rcl_send_request(get_client_handle().get(), request, &sequence_number);
    if (RCL_RET_OK != ret) {
      rclcpp::exceptions::throw_from_rcl_error(ret, "failed to send request");
    }
    pending_requests_.emplace(sequence_number);
    return sequence_number;
  }

protected:
  void
  erase_pending_request(int64_t request_number)
  {
    std::unique_lock<std::mutex> lock(pending_requests_mutex_);
    auto it = pending_requests_.find(request_number);
    if (it == this->pending_requests_.end()) {
      RCUTILS_LOG_DEBUG_NAMED(
        "rclcpp",
        "Received invalid sequence number. Ignoring...");
    } else {
      pending_requests_.erase(request_number);
    }
  }


  RCLCPP_DISABLE_COPY(GenericClient)

  std::set<int64_t> pending_requests_;
  std::mutex pending_requests_mutex_;

private:
  std::shared_ptr<rcpputils::SharedLibrary> ts_lib_;
  const rosidl_service_type_support_t * srv_type_support_handle_;
  const rosidl_typesupport_introspection_cpp::MessageMembers * request_members_;
  const rosidl_typesupport_introspection_cpp::MessageMembers * response_members_;
};
}  // rclcpp
```

rclcpp::get_typesupport_handle_srv() should be wrapped to rclcpp::get_typesupport_handle(). 
For verification, I made a new function 
```c++
const rosidl_service_type_support_t *
get_typesupport_handle_srv(
  const std::string & type,
  const std::string & typesupport_identifier,
  rcpputils::SharedLibrary & library)
{
  std::string package_name;
  std::string middle_module;
  std::string type_name;
  std::tie(package_name, middle_module, type_name) = extract_type_identifier(type);

  auto mk_error = [&package_name, &type_name](auto reason) {
      std::stringstream rcutils_dynamic_loading_error;
      rcutils_dynamic_loading_error <<
        "Something went wrong loading the typesupport library for service type " << package_name <<
        "/" << type_name << ". " << reason;
      return rcutils_dynamic_loading_error.str();
    };

  try {
    std::string symbol_name = typesupport_identifier + "__get_service_type_support_handle__" +
      package_name + "__" + (middle_module.empty() ? "srv" : middle_module) + "__" + type_name;

    const rosidl_service_type_support_t * (* get_ts)() = nullptr;
    // This will throw runtime_error if the symbol was not found.
    get_ts = reinterpret_cast<decltype(get_ts)>(library.get_symbol(symbol_name));
    return get_ts();
  } catch (std::runtime_error &) {
    throw std::runtime_error{mk_error("Library could not be found.")};
  }
}
```

Barry-Xu-2018:
@MichaelOrlov Friendly ping. Please look at above description of property on how to send request according to recorded service event message.

Barry-Xu-2018:
According to above discussion, I updated the design. Please refer to https://github.com/ros2/rosbag2/pull/1359/commits/13effd3ec507299242a14026bbdd8d8bd9bd1185

docs/design/rosbag2_record_replay_service.md
emersonknapp: I agree that service introspection should be a separate feature from hidden topics. They have a similar implementation, but are different ideas, I think.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Hidden topic  
+  
+  [NameSpace/]_TopicName
+
+- Service event topic
+  
+  [NameSpace/]ServiceName/_service_event
+
+So if the string ends with `/_service_event` in topic name, it should be service event topic. Of course, it is possible that hidden topic has the same name. So message type is used for further check. The message type for service event topic must include `/srv/` and ends with `_Event` (e.g. example_interfaces/srv/AddTwoInts_Event). Therefore, service event topic can be identified.
```
emersonknapp:
```suggestion
The following criteria will be used to determine a service introspection topic:
1. The topic ends with `/_service_event`
1. The topic type name matches  `*/srv/*_Event` (e.g. `example_interfaces/srv/AddTwoInts_Event`)
```

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Expand the 'record' command
+
+Add 3 parameters.
+
+- `-S SerivceName1 [ServiceName2 ...]` or `--services SerivceName1 [ServiceName2 ...]`
```
emersonknapp:
```suggestion
- `-S ServiceName1 [ServiceName2 ...]` or `--services ServiceName1 [ServiceName2 ...]`
```

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,118 @@
...
+  
+    If the user doesn't specify service names, all service event topics are recorded.
```
emersonknapp: This is overloaded with `--all-services`. I would prefer that `-S,--services` takes `nargs='+'` so the user _must_ specify some service if they use this option.

docs/design/rosbag2_record_replay_service.md
emersonknapp: Yes, we can keep metadata of `num_requests, num_responses` at recording time, and store that as metadata about the topic

docs/design/rosbag2_record_replay_service.md
emersonknapp: Is this feature suggesting that `ros2 bag play` creates service clients and sends requests with original timing if service is available?

docs/design/rosbag2_record_replay_service.md
emersonknapp: I think we should not have a `-S,--services` option for `play`. The user can `--exclude */_service_event` if they want to skip these topics.

emersonknapp:
Sorry for being slow, but I've left some small feedback

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Thanks. This makes the expression clearer.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Fixed

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Thanks. I forgot to update this description. User must specify at least one service.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > Yes, we can keep metadata of num_requests, num_responses at recording time, and store that as metadata about the topic

If we want to get num_requests and num_responses, we have to deserialize service event message to parse this information during recording. In order to avoid affecting performance, we plan to not get num_requests and num_responses while recording. Instead, if user execute info command with `-v`', num_requests and num_responses will be parsed from recorded file.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > I think we should not have a -S,--services option for play. The user can --exclude */_service_event if they want to skip these topics.

Yes.  
The inclusion of the '-S' parameter was initially intended to exclude service event topics from being played by default.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > Is this feature suggesting that ros2 bag play creates service clients and sends requests with original timing if service is available?

Yes.

Barry-Xu-2018 (commented):
Thanks for your comments. @emersonknapp

docs/design/rosbag2_record_replay_service.md
```diff
@@ -22,23 +22,50 @@ Service introspection is saved like general topic. But they should use different
...
 
-- `-S [SerivceName1 ServiceName2 ...]` or `--services [SerivceName1 ServiceName2 ...]`
+- `-S ServiceName1 [ServiceName2 ...]` or `--services ServiceName1 [ServiceName2 ...]`
   
-    If the user doesn't specify service names, all service introspection topics are recorded.
+    Specify which services will be recorded. User must specify at least one service.
+
+- `--all-topics`
+
+    Record all topics. Hidden topic are excluded (include service event topics).
```
MichaelOrlov: It doesn't make sense to include service event topics in `--all-topics` if we will have `--all` command which will be doing exactly the same.
It will be more reasonable to have `--all-topics` options when only regular topics without service events would be recorded. 
I was under the impression that we clarified this case in some of the comments above and after the last WG meeting.

MichaelOrlov: After some consideration, it seems I misunderstood the phrase 
> Hidden topic are excluded (include service event topic)  

To avoid confusion for others I would suggest rewording it to be more clear. 
Something like. **"Hidden topics, including service event topics are excluded."**

Barry-Xu-2018: > To avoid confusion for others I would suggest rewording it to be more clear.
> Something like. "Hidden topics, including service event topics are excluded."

Sorry to make you confused.  
I will update description as your suggestion.

MichaelOrlov:
> > We are not able to replay dedicated service requests due to the lack of the API currently because service requests are wrapped inside the service event and they are serialized. We decided that possibility of playback service requests needs to be further investigated and prototyped before we will make a final decision about including this option in the design.
> 
> How to unwrap request part in service event message. (I had written code to verify this way.)
> 
> Assume we know the below information
> 
> 1. service event topic  --- `/add_two_ints/_service_event`
> 2. service event type --- `example_interfaces/srv/AddTwoInts_Event`
> 
> ```c++
> // Use type support introspection 
> std::shared_ptr<rcpputils::SharedLibrary>    ts_lib = rclcpp::get_typesupport_library(
>       "example_interfaces/srv/AddTwoInts_Event", "rosidl_typesupport_cpp");
> ```
@Barry-Xu-2018 Thank you for spending your time prototyping possible implementation for the service playback.
It will give us a sense of the level of work required to get it done.   
We discussed it on today's Rosbag2 and Tooling WG meeting and it seems everybody agreed that it would be reasonable to split support for the services recording and replay on the two phases.
- In the first phase, we will add support for the service events recording.
- In the second phase we will add support for the service request publishing during playback.

And the rationale for that is that publishing service requests require support from the rclcpp layer and the level of the changes looks like a significant feature that will require some time to get it done. 
We have tried to consider the option when we would implement rclcpp parts like `GenericClient` and `get_typesupport_handle_srv` temporarily inside rosbag2 packages. But we don't like this scenario for various reasons and would prefer to make it on the rclcpp layer from the very beginning. 

As regards the implementation.
I think the following part 
```cpp
size_t request_offset = 0;
size_t request_offset_tmp = 0;
  for(size_t i=0; i < message_members->member_count_; i++) {
    auto message_member = message_members->members_[i];
    if (message_member.name_ == std::string("request")) {
      request_offset = request_offset_tmp;
    } else {
      auto sub_members = (rosidl_typesupport_introspection_cpp::MessageMembers *)message_member.members_->data;
      request_offset_tmp += sub_members->size_of_;
    }
   if (request_offset != 0) {
      break;
    }
  }
```
Has to be rewritten in some better way. And actually, I think there is some mistake in calculating the offset. Basically, we need to check the service event type and based on it. Perhaps we can cast the beginning of the deserialized ros message to the Service Event Info structure to be able to get the `event_type`. And the service request structure is going to be right after the `ros_message + sizeof(Service Event Info structure)`.

As regards the `GenericClient` implementation it looks like a bare minimum implementation and in real life it will need a bit more code to support pruning of the sent or stale service requests and other helper stuff similar to what is already done in the similar templated service client class.

Barry-Xu-2018:
@MichaelOrlov 

Thank you for sharing the results of your discussion

> We discussed it on today's Rosbag2 and Tooling WG meeting and it seems everybody agreed that it would be reasonable to split support for the services recording and replay on the two phases.
>
> - In the first phase, we will add support for the service events recording.
> - In the second phase we will add support for the service request publishing during playback.

Totally agree. The second phase relies on modifications to the rclcpp for implementation.

> We have tried to consider the option when we would implement rclcpp parts like GenericClient and get_typesupport_handle_srv temporarily inside rosbag2 packages. But we don't like this scenario for various reasons and would prefer to make it on the rclcpp layer from the very beginning.

Yes. I also prefer to place them in rclcpp.

> I think the following part
> ...
> Has to be rewritten in some better way. And actually, I think there is some mistake in calculating the offset.  

Yes. I will update it. It can be simplified as your suggestion `ros_message + sizeof(Service Event Info structure)`.

> we need to check the service event type and based on it. Perhaps we can cast the beginning of the deserialized ros message to the Service Event Info structure to be able to get the event_type. And the service request structure is going to be right after the ros_message + sizeof(Service Event Info structure).

Yes. We should check if the type of event message is `REQUEST_SENT` or `REQUEST_RECEIVED`. And then send it. 

> As regards the GenericClient implementation it looks like a bare minimum implementation and in real life it will need a bit more code to support pruning of the sent or stale service requests and other helper stuff similar to what is already done in the similar templated service client class.

Yes. I only do simply implementation for sending request. Finally, GenericClient should be implemented as you suggested.

ihasdapie:
Small comment: I'm personally not a big fan of all the new options for selecting topics/services, after this lands we will have `--services`, `--all-services`, `--topics`, `--all-topics`, `-S` .... and so forth. It may make sense to support (and document/advertise!) wildcards instead, i.e. `--all-services` can be `--services=*`, etc. This has the added bonus of supporting queries like `--services=/planner/*` for free.

During initial prototyping for service introspection I recall having to do something similar with loading the typesupports at runtime, but for getting the request/response msg typesupports given a service. Not sure if it'll be useful, but here it is anyways: https://gist.github.com/ihasdapie/e3950d0802a23d23419947c3496980ca. Another caveat I recall is that there are differences between the `_cpp` and `_c` typesupports, and IIRC the `_c` typesupports are usable for python & c, but the `_cpp` ones are only good for cpp -- which made it difficult to implement at the `rcl` level. But then again rosbag2 lives above the client library level, so it shouldn't be impacted by this.

I second the two-phase implementation, playback is a lot more rather unrelated work and it'd quite annoying to block what is essentially a rosbag2 cli syntactic sugar PR because of it.

Barry-Xu-2018:
Thanks for your comments.

> Small comment: I'm personally not a big fan of all the new options for selecting topics/services, after this lands we will have --services, --all-services, --topics, --all-topics, -S .... and so forth. It may make sense to support (and document/advertise!) wildcards instead, i.e. --all-services can be --services=*, etc. This has the added bonus of supporting queries like --services=/planner/* for free.

Yes, that would reduce the number of parameters. However, the style and usage of topic are different (topic uses '-e REGEX').
Except the missing '-e REGEX' parameter, the service currently lacks another parameter '--exclude'. Whether do we add new parameter for service ?  @MichaelOrlov Do you have any suggestion?  

> During initial prototyping for service introspection I recall having to do something similar with loading the typesupports at runtime, but for getting the request/response msg typesupports given a service. Not sure if it'll be useful, but here it is anyways: https://gist.github.com/ihasdapie/e3950d0802a23d23419947c3496980ca. 

Thanks, Your example provides another way to get request/response msg typesupports. 

> I second the two-phase implementation, playback is a lot more rather unrelated work and it'd quite annoying to block what is essentially a rosbag2 cli syntactic sugar PR because of it.

Yes.

ros-discourse:

MichaelOrlov:
@Barry-Xu-2018 Sorry for my late response.
> Yes, that would reduce the number of parameters. However, the style and usage of topic are different (topic uses '-e REGEX').
Except the missing '-e REGEX' parameter, the service currently lacks another parameter '--exclude'. Whether do we add new parameter for service ? @MichaelOrlov Do you have any suggestion?

On one hand it would be good to have support for wild cards, although on another hand it is going to be non-trivial in implementation. I would go with dedicated CLI options first.
As regards to the '-e REGEX' parameter it looks like it would be logically correct if it would be a more generic regex and will be applied for the both topics and services if they were enabled. However, it would be better to have separate exclude options for service and topics. But in this case we need to add a one more parameter and likely rename existent `-exclude` parameter.  
In this case it would be challenging to create short cuts for those new "exclude" paramers. Probably the compromise solution would be to get rid from short names for those new CLI options.

Barry-Xu-2018:
> As regards to the '-e REGEX' parameter it looks like it would be logically correct if it would be a more generic regex and will be applied for the both topics and services if they were enabled.

Agree.  

> Probably the compromise solution would be to get rid from short names for those new CLI options.

Agree. 

So in order to keep the same style, we have to add new parameter for the service  
For the record command, I list all parameters to clearly describe whether it affects the record of topic or service or both.  

| Parameter name |  Topic | Service |
| :-- | :-- | :-- |
| -s {my_test_plugin,sqlite3,mcap}, --storage {my_test_plugin,sqlite3,mcap} | :white_check_mark: | :white_check_mark: |
| -a, --all | :white_check_mark: | :white_check_mark: |
| --all-topics | :white_check_mark: | |
| --all-services | |  :white_check_mark: |
|  -e REGEX, --regex REGEX | :white_check_mark: | :white_check_mark: |
| --exclude-topics EXCLUDE | :white_check_mark: | |
| --exclude-services EXCLUDE | |:white_check_mark: | 
| -S [ServiceName ...], --services [ServiceName ...]  |  | :white_check_mark: |
| --include-unpublished-topics | :white_check_mark: | :white_check_mark: |
| --include-hidden-topics | :white_check_mark: | :white_check_mark: |
| --no-discovery | :white_check_mark: | :white_check_mark: |
| -p POLLING_INTERVAL, --polling-interval POLLING_INTERVAL | :white_check_mark: | :white_check_mark: |
| --ignore-leaf-topics | :white_check_mark: | |
| --qos-profile-overrides-path | :white_check_mark: |  :white_check_mark: |
| -f {a,s}, --serialization-format {a,s} | :white_check_mark: |  :white_check_mark: |
| -b MAX_BAG_SIZE, --max-bag-size MAX_BAG_SIZE | :white_check_mark: |  :white_check_mark: |
| -d MAX_BAG_DURATION, --max-bag-duration MAX_BAG_DURATION | :white_check_mark: |  :white_check_mark: |
| --max-cache-size MAX_CACHE_SIZE | :white_check_mark: |  :white_check_mark: |
| --start-paused |  :white_check_mark: |  :white_check_mark: |
| --node-name NODE_NAME |  :white_check_mark: |  :white_check_mark: | 
| --custom-data [KEY=VALUE ...] | :white_check_mark: |  :white_check_mark: | 
| --snapshot-mode | :white_check_mark: |  :white_check_mark: | 
|  --storage-config-file STORAGE_CONFIG_FILE | :white_check_mark: |  :white_check_mark: | 
| --storage-preset-profile {none,fastwrite,zstd_fast,zstd_small} | :white_check_mark: |  :white_check_mark: | 
| --compression-queue-size COMPRESSION_QUEUE_SIZE | :white_check_mark: |  :white_check_mark: | 
| --compression-threads COMPRESSION_THREADS | :white_check_mark: |  :white_check_mark: | 
| --compression-mode {none,file,message} | :white_check_mark: |  :white_check_mark: | 
| --compression-format {zstd,fake_comp} | :white_check_mark: |  :white_check_mark: | 

About `--ignore-leaf-topics`, service event topic has no subscription normally. So I think this parameter is only for recording topic. 

Do you any comments on the above table? @MichaelOrlov @ihasdapie

MichaelOrlov:
@Barry-Xu-2018 Looks good to me.
I am only thinking about removing the shortcut for the `-S [ServiceName ...]` to avoid a mismatch with the `-s {storage_plugin, sqlite3, mcap}`

Barry-Xu-2018:
> I am only thinking about removing the shortcut for the -S [ServiceName ...] to avoid a mismatch with the -s {storage_plugin, sqlite3, mcap}

Okay.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,131 @@
...
+| --exclude-services | Exclude services containing provided regular expression. |
+| -e REGEX, --regex REGEX | Record only topics and service containing provided regular expression. |
+
+### Change output of 'Info' command
+
+A new parameter is added.
+
+- `-v` or `--verbose`
+
+    This parameter only affect service event topic. if `-v` or `--verbose` is set, info command shows the number of request and response for each services based on service event. Otherwise, only show the number of service event. Note that parsing the number of request and response need spent time. The duration of the parsing is related to the number of recorded service events.
```
MichaelOrlov:
```suggestion
This parameter only affects the service event topic. if `-v` or `--verbose` is set, the info command shows the number of requests and responses for each service based on the service event. Otherwise, only show the number of service events. Note that parsing the number of request and response need spent time. The duration of the parsing is related to the number of recorded service events.
```
Grammar fixes

docs/design/rosbag2_record_replay_service.md
MichaelOrlov: @Barry-Xu-2018 I agree with @emersonknapp adding `-s` and `--service` for playback to act as service exclusion will very confuse users because we have exactly the same parameters on the recording where they act as the service inclusion which is semantically opposite.  
The user can `--exclude */_service_event` if they want to skip these topics.

MichaelOrlov (commented):
@Barry-Xu-2018 I think we are pretty close to being good to go with this design document.
Need to mention that services playback is gonna implemented as phase 2 after implementing relevant API on rclcpp layer (list API which we needed), also remove `-s` and `--services` from playback.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Sorry. Maybe I misunderstand emersonknapp's comments.   

`--services ServiceName1 [ServiceName2]` isn't used for exclusion.  It is used for inclusion. Like `-topics TOPICS [TOPICS ...]`.  
About exclusion, record option includes `--exclude-topics` and `--exclude-services`. Taking consistency into consideration, I think we should also have 2 parameters in play option for exclusion.   

About another play option in design 
```
- `--services-requests [ServiceName1 ...]`

    Replay request with recorded sequence and interval. If there is no service name set, all recorded service request are sent.
```

I will remove it since play only sends the request for service (Not send service event message)

Summary for play option
- `-services ServiceName1 [ServiceName2]`
     User can choose which services' requests are played.  
- `-e REGEX, --regex REGEX`
     Affect both topic and service
- `--exclude-topics`
     Exclude topics
- `--exlude-services`
     Exclude services

What do you think ? @MichaelOrlov

Barry-Xu-2018:
@MichaelOrlov 

> I think we are pretty close to being good to go with this design document.
Need to mention that services playback is gonna implemented as phase 2 after implementing relevant API on rclcpp layer (list API which we needed), also remove -s and --services from playback.

Yes. I have started implementing the main functionality of phase 1. Once the new relevant API is added to rclcpp, I will start the implementation of phase 2.

Barry-Xu-2018:
https://github.com/ros2/rosbag2/pull/1359/commits/f0f8dddd62c7aba7ec3286a097c49e662b99e10b will be updated for play option part based on the disccusion result of https://github.com/ros2/rosbag2/pull/1359/files#r1244556452

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,131 @@
...
+
+- `--exclude-services`
+
+    Exclude services containing provided regular expression.
+
+The description of the relevant parameter behavior.
+
+| Parameter | Description |
+| :-- | :--|
+|--include-hidden-topics| Record all hidden topics. Include service event topic. |
```
fujitatomoya: this seems to be inconsistency for me.

as described above, service introspection topic is not one of the hidden topics, but this option includes the service event topic? as far as i see from the current discussion, this should not be included? just checking.

docs/design/rosbag2_record_replay_service.md
fujitatomoya: the option sounds okay, but i am included to have this information in default. i am with https://github.com/ros2/rosbag2/pull/1359/files#r1211126103 for default behavior. after all, either is fine to me.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,131 @@
...
+### Expand the 'play' command
+
+Add 2 parameters.  
+- `--services ServiceName1 [[ServiceName2 ...]`
+
+    Decide which services will be played.
+
+- `--service-request-mode`
+
+    Replay service request with recorded sequence and interval instead of playing service event message.
```
fujitatomoya: I am not sure when non-service-request-mode is required though? do we have any use cases for this mode?

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: You mean that description should not specifically emphasize “Include service event topic.” ？
Agree. I should remove it.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: I think I misunderstand phase 1 (https://github.com/ros2/rosbag2/pull/1359#issuecomment-1595530464). It should only do recording service event (Not want to play it. Phase 2 includes play. So only play request).  I will update it.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Currently, we have some special options for service, such as `--all-services, --services`. So whether service should not be affected by `--include-hidden-topics`.  That is, service record cannot be enabled by `--include-hidden-topics`.  Users must explicitly indicate their intention to record services via e.g. `--all-services, --services`.   
What do you think ? @MichaelOrlov @ihasdapie

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Getting the number of request and response need to spend time to parse bag file.   
If bag file is big and users only want to check topic information, users have to wait parsing service information.   
So not set as default mode.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: After giving it some thought, I've decided to further discuss this point at https://github.com/ros2/rosbag2/pull/1359#discussion_r1246072371

docs/design/rosbag2_record_replay_service.md
MichaelOrlov: I tend to agree with @ihasdapie 
> IMO it would be more consistent to include service events in --include-hidden-topics since it is, after all, a hidden topic and not something special

It is what it is. Service events are hidden topics.

docs/design/rosbag2_record_replay_service.md
MichaelOrlov: @Barry-Xu-2018 Summary for the play option looks good to me :+1:

docs/design/rosbag2_record_replay_service.md
MichaelOrlov: @Barry-Xu-2018 As regards to the 
> I don't want to change previous behavior. So '--all' only means all topic (Not include special service event topic).

As far as I remember we agreed on the version 
https://github.com/ros2/rosbag2/pull/1359/files#diff-cc475e77d25d214937d96658f496944c9dbf1b0dd6e01fc776528672300841bcR62
```
| --all | Record all topics and service event topics. Other hidden topics are excluded. |
```
I am closing this thread as resolved. If the proposed resolution is not complete or not as expected please feel free to mark it as unresolved again.

docs/design/rosbag2_record_replay_service.md
MichaelOrlov: As far as I recall we have made the decision about `--include-hidden-topics` at one of the `Rosbag2 and Tooling WG` meetings last month. 
> 2. Decided that --include-hidden-topics will record service events as well, since it is just a hidden topic.

Here is the full summary from the meeting https://github.com/ros2/rosbag2/pull/1359#issuecomment-1574575492

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,126 @@
...
+
+The description of the relevant parameter behavior.
+
+| Parameter | Description |
+| :-- | :--|
+|--services ServiceName1 [ServiceName2 ...] | Record services (service event topics) with specified service names. |
+| --all | Record all topics and service event topics. Other hidden topics are excluded. |
+| --all-topics | Record all topics. Hidden topics including service event topics are excluded. |
+| --all-services | Only record all service event topics. |
+| --exclude-topics | Exclude topics containing provided regular expression. |
```
MichaelOrlov, line 66: @Barry-Xu-2018 Please add one more row about 
```
|--include-hidden-topics| Record all hidden topics. Include service event topic. |
```
To avoid confusion and explicitly define expectations.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+### Expand the 'play' command
+
+Add 1 parameters.
+- `--services ServiceName1 [[ServiceName2 ...]`
+
+    Determine which service's requests are played.
+
+Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.
```
MichaelOrlov:
```suggestion
Other parameters need to be checked one by one. If it is unsuitable for playing service event, the description of parameter should be updated to mention this limitation.

## Implementation Staging
Implementation is going to be split into two phases.
On phase 1 is expected to be implemented all functionality which is related to the service events recording.
On phase 2 the rest functionality related to the service requests playback.
The service requests playback is not straightforward to implement since we don't have API from rclcpp layer to be able to publish service requests with dynamically created type support. We will need to create an analog of the `GenericPublisher` i.e. `GenericClient` on the `rclcpp` layer to be able to do this. To facilitate implementation of the `GenericClient` we likely will need to add one more missing API:   
const rosidl_service_type_support_t * get_typesupport_handle_srv(
  const std::string & type,
  const std::string & typesupport_identifier,
  rcpputils::SharedLibrary & library)
Therefore, phase 2 will be implemented when the required new rclcpp API will be ready.
```

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: @MichaelOrlov Thanks for your confirmation. 

> As far as I recall we have made the decision about --include-hidden-topics at one of the Rosbag2 and Tooling WG meetings last month.

Yes. It is because new record options `--all-services` and `--all-topics` were added later. There is an option used to enable recording services. Now `--include-hidden-topics` also has the same effect. If users doesn't know service introspection, maybe they confuse the behavior of `--include-hidden-topics`.   
The behavior of  `--include-hidden-topics` will be described as you suggestion https://github.com/ros2/rosbag2/pull/1359#discussion_r1248317190

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: > As far as I remember we agreed on the version

Yes. https://github.com/ros2/rosbag2/pull/1359#discussion_r1211203374 is old comments.   
We have reached a consensus on this.

docs/design/rosbag2_record_replay_service.md
Barry-Xu-2018: Okay. I supplement the description of `--include-hidden-topics`.

MichaelOrlov (approved):
@Barry-Xu-2018 LGTM.
Thanks for your patients and versioning.

Barry-Xu-2018:
@MichaelOrlov Thanks.

I share the current status.  
The implementation for phase 1 had been complete. Now I am adding tests.

MichaelOrlov:
@Barry-Xu-2018 As regards
> The implementation for phase 1 had been complete. Now I am adding tests.

The draft PR will be appreciated.

Barry-Xu-2018:
@MichaelOrlov 

> The draft PR will be appreciated.

Okay. I will create draft PR.

Barry-Xu-2018:
@MichaelOrlov 
I created a draft PR #1414.   
Recording service and showing service info can work correctly.

MichaelOrlov:
Thanks, much appreciated.

PR #1367, opened 2023-06-03
Refine root cause of the `unpaused_sleep_returns_true` test failure

MichaelOrlov:
- Follow up to the failure in https://ci.ros2.org/job/ci_windows/19597/

### DIFF
--- a/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
@@ -165,6 +165,8 @@ TEST_F(TimeControllerClockTest, unpaused_sleep_returns_true)
     sleep_result = clock.sleep_until(sleep_until_timestamp);
   }
   auto end = std::chrono::steady_clock::now();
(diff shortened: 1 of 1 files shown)


emersonknapp (approved):
:+1: I think we should merge this, and get the extra info for future test failures.

emersonknapp:
Oh ... `rclcpp::ok()` was _never_ true, ever. Nothing ever calls `rclcpp::init()` in this test suite! In fact, `rclcpp::ok()` is explicitly not a requirement - `TimeControllerClock` only calls a `now_fn`, it doesn't care about `rclcpp` status. The check that was added to the test in https://github.com/ros2/rosbag2/pull/1012/files#diff-ea9ff4b69cd29013d4599d8e20313ab1f1dd918e82e810755dad97abfd9f4a51R161 did exactly nothing, because if the clock has a spurious wakeup then `rclcpp::ok() == false` and the loop exits anyways.

emersonknapp (changes requested):
Revoking approval since this will never be true

MichaelOrlov:
- Closing as will not be merged in favor of the #1384

PR #1370, opened 2023-06-05
Fix a clang warning about uninitialized variable.

clalancette:
In the tests for the type description hash, we were forgetting to initialize the endpoint_type, which is necessary when constructing an rclcpp::TopicEndpointInfo object.  Just set it to INVALID here, as we don't care whether this is a publisher or subscriber.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18854/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13393/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19606/)

MichaelOrlov (approved):
LGTM.

PR #1290, opened 2023-04-13
Fix for flaky `TimeControllerClockTest::unpaused_sleep_returns_true`

MichaelOrlov:
- Closes https://github.com/ros2/rosbag2/issues/1282
- Addressing failure on nightly job https://ci.ros2.org/view/nightly/job/nightly_win_rep/2955/testReport/(root)/projectroot/test_time_controller_clock/
It was a problem in test since we were not able to properly get the same time source for `now()` for measuring start point and delay for `sleep_until(timepoint)` API.
- Brought up `ros_to_steady(timepoint)` API in `TimeControllerClock` class from inner implementation layer to be able to transform `TimeControllerClock::now()` in to the steady clock domain.

PR #1371, opened 2023-06-05
[iron] Add ROS_DISTRO metadata record to mcap file when opening for writing

emersonknapp:
Related to #1239
This is an API/ABI-stable patch to add some of the functionality from #1241 to Iron

Adds ROS_DISTRO metadata record to mcap storage file when opened for writing. `ros2 bag info` will not know about this information, but its presence in the file will be readable by other mcap-reading tools

When complete - will backport to Humble

### DIFF
--- a/rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
+++ b/rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
@@ -14,6 +14,7 @@
 
 #include "rclcpp/serialization.hpp"
 #include "rclcpp/serialized_message.hpp"
+#include "rcpputils/env.hpp"
 #include "rcpputils/filesystem_helper.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
 #ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
@@ -21,6 +22,8 @@
 #endif
 #include "rosbag2_test_common/temporary_directory_fixture.hpp"
 #include "std_msgs/msg/string.hpp"
+
+#include <mcap/mcap.hpp>
 
 #include <gmock/gmock.h>
 
@@ -200,3 +203,40 @@ TEST_F(TemporaryDirectoryFixture, can_write_mcap_with_zstd_configured_from_yaml)
   }
 }
 #endif  // #ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
+
+TEST_F(TemporaryDirectoryFixture, mcap_contains_ros_distro)
+{
+  // const auto expected_file = rcpputils::fs::path(temporary_dir_path_) / "rosdistro_bag.mcap";
+  const auto expected_file = rcpputils::fs::path("rosdistro_bag.mcap");
+  const auto uri = rcpputils::fs::remove_extension(expected_file);
+  const std::string storage_id = "mcap";
+  const std::string current_ros_distro = rcpputils::get_env_var("ROS_DISTRO");
+  ASSERT_FALSE(current_ros_distro.empty());
+  std::string read_metadata_ros_distro = "";
+
+  // Open writer to create no-data file and then delete the writer to close
+  rosbag2_storage::StorageFactory factory;
+  rosbag2_storage::StorageOptions options;
+  options.uri = uri.string();
+  options.storage_id = storage_id;
+  auto writer = factory.open_read_write(options);
+  writer.reset();
+  ASSERT_TRUE(expected_file.is_regular_file());
+
+  // Open created mcap file, read all metadata records to find rosbag2.ROS_DISTRO value
+  mcap::Status status{};
+  std::ifstream input{expected_file.string(), std::ios::binary};
+  mcap::FileStreamReader data_source{input};
+  mcap::TypedRecordReader typed_reader(data_source, 8);
+  bool done = false;
+  typed_reader.onMetadata = [&](const mcap::Metadata & metadata, mcap::ByteOffset) {
+    if (metadata.name == "rosbag2") {
+      read_metadata_ros_distro = metadata.metadata.at("ROS_DISTRO");
+      done = true;
+    }
+  };
+  while (!done && typed_reader.next()) {
+    EXPECT_TRUE(typed_reader.status().ok());
+  }
+  EXPECT_EQ(read_metadata_ros_distro, current_ros_distro);
+}
--- a/rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
+++ b/rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
@@ -206,12 +206,13 @@ TEST_F(TemporaryDirectoryFixture, can_write_mcap_with_zstd_configured_from_yaml)
 
 TEST_F(TemporaryDirectoryFixture, mcap_contains_ros_distro)
 {
-  // const auto expected_file = rcpputils::fs::path(temporary_dir_path_) / "rosdistro_bag.mcap";
-  const auto expected_file = rcpputils::fs::path("rosdistro_bag.mcap");
+  // Guarantee env var set for mcap to use - in a full-build testing environment it may not be.
+  const std::string current_ros_distro = "rolling";
+  ASSERT_TRUE(rcpputils::set_env_var("ROS_DISTRO", current_ros_distro.c_str()));
+
+  const auto expected_file = rcpputils::fs::path(temporary_dir_path_) / "rosdistro_bag.mcap";
   const auto uri = rcpputils::fs::remove_extension(expected_file);
   const std::string storage_id = "mcap";
-  const std::string current_ros_distro = rcpputils::get_env_var("ROS_DISTRO");
-  ASSERT_FALSE(current_ros_distro.empty());
   std::string read_metadata_ros_distro = "";
 
   // Open writer to create no-data file and then delete the writer to close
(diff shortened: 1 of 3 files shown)


james-rms (approved):

MichaelOrlov (commented):
@emersonknapp In the future I would like to add functionality for saving full serialized metadata and it would be more reasonable to add it at the same place as ROS_DISTRO.
Wouldn't you mind considering moving functionality by saving ROS_DISTRO in the `update_metada(metadata)` method as it was made originally on Rolling?

And simple unit tests for getting saved ROS_DISTRO as in SQLite3 backend on Iron with `get_ros_distro()` API would be much appreciated.

emersonknapp:
If we do `update_metadata` - that means we can create multiple metadata records in the output mcap. I think `SequentialWriter` calls `update_metadata` twice in normal operation. Maybe that's fine - if so do we just consider the latest Metadata record named "rosbag2" to be the correct values?

emersonknapp:
Update - by using the flag I only write this particular metadata value once - so it's fine. I added a unit test but since `MCAPStorage` headers are not exported I just used mcap library directly to inspect the bag for the test.

emersonknapp:
Pulls: ros2/rosbag2#1371
Gist: https://gist.githubusercontent.com/emersonknapp/4bde126f5e1dfaf0c5fb9699d7d2ffe4/raw/00461a95d0a7f269e7fdc3d234551779b3fea7c5/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_storage_mcap
TEST args:  --packages-above rosbag2_storage_mcap
ROS Distro: iron
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12168
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13406/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19620/)

MichaelOrlov:
@emersonknapp As regards
>  I think SequentialWriter calls update_metadata twice in normal operation. Maybe that's fine - if so do we just consider the latest Metadata record named "rosbag2" to be the correct values?

Yes, this is correct I imply we do similar things as on Rolling. i.e. Look out for the first record of the Metadata record named "rosbag2" and take `ROS_DISTRO` value if it is non-empty. If it is empty try to lookout for the next Metadata record.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
```diff
@@ -200,3 +203,40 @@ TEST_F(TemporaryDirectoryFixture, can_write_mcap_with_zstd_configured_from_yaml)
...
+  rosbag2_storage::StorageFactory factory;
+  rosbag2_storage::StorageOptions options;
+  options.uri = uri.string();
+  options.storage_id = storage_id;
+  auto writer = factory.open_read_write(options);
+  writer.reset();
+  ASSERT_TRUE(expected_file.is_regular_file());
+
+  // Open created mcap file, read all metadata records to find rosbag2.ROS_DISTRO value
+  mcap::Status status{};
```
MichaelOrlov, line 228: @emersonknapp I was expecting that you will add a new public method to the mcap_storage.cpp(hpp) aka `get_ros_distro_name()` for calling it from unit test. 
According to the https://www.acodersjourney.com/20-abi-breaking-changes/ 
> 3. Add new non-virtual methods to an existing class

These are ABI-compatible changes.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
```diff
@@ -200,3 +203,40 @@ TEST_F(TemporaryDirectoryFixture, can_write_mcap_with_zstd_configured_from_yaml)
   }
 }
 #endif  // #ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
+
+TEST_F(TemporaryDirectoryFixture, mcap_contains_ros_distro)
+{
+  // const auto expected_file = rcpputils::fs::path(temporary_dir_path_) / "rosdistro_bag.mcap";
+  const auto expected_file = rcpputils::fs::path("rosdistro_bag.mcap");
```
MichaelOrlov: It looks like it is supposed to be a path to the temporary folder but was forgotten to change back after debugging.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
emersonknapp: There is no `.hpp` file, `rosbag2_storage_mcap` doesn't export any headers - there's not really a point in adding such a new method since the only exposed API is via `ReadWriteInterface`

emersonknapp:
Rebuilds build #[12168](https://ci.ros2.org/job/ci_launcher/12168)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18871/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19625/)

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
```diff
@@ -200,3 +203,41 @@ TEST_F(TemporaryDirectoryFixture, can_write_mcap_with_zstd_configured_from_yaml)
   }
 }
 #endif  // #ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
+
+TEST_F(TemporaryDirectoryFixture, mcap_contains_ros_distro)
+{
+  // Guarantee env var set for mcap to use - in a full-build testing environment it may not be.
+  const std::string current_ros_distro = "rolling";
```
MichaelOrlov, line 210:
```suggestion
  const std::string current_ros_distro = "iron";
```

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
MichaelOrlov: Hmm, indeed ~"Yes, we have no bananas!" :)~  there is no `.hpp` file.
Ok, for now. However, we might need to consider adding it on the Rolling at least, for the purpose of serving unit tests in the future at least.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
emersonknapp: This is a testing value and I want to just backport it verbatim - it's a process-scoped environment variable and therefore the actual value doesn't matter.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
emersonknapp: I personally believe in the school of thought that we shouldn't have to add to feature code to serve unit tests (whenever possible - strong preference not unbreakable rule). If the user-facing API is not testable, then it's maybe a bad API. In this case, we produce an mcap file as output and I'd rather just inspect that to see if it's correct than change the writer somehow to have it tell me that it's doing something correctly - that's testing the testing code, not testing the true outcome.

rosbag2_storage_mcap/test/rosbag2_storage_mcap/test_mcap_storage.cpp
MichaelOrlov: Ok, never mind in this case.

emersonknapp:
I expect that the Windows error is a totally unrelated flaky test (which has had attempts to make it reliable in https://github.com/ros2/rosbag2/pull/1012 and https://github.com/ros2/rosbag2/pull/1290 - but I don't think either of them actually fixed it). Rerunning to check, and not block this PR on that issue

Windows (just build-up-to rosbag2_cpp and test that package) [Build Status](https://ci.ros2.org/job/ci_windows/19627/)

MichaelOrlov:
@emersonknapp As regards to the test failure in the **rosbag2_cpp.TimeControllerClockTest.unpaused_sleep_returns_true** on Windows CI. 
Take a look on my PR https://github.com/ros2/rosbag2/pull/1367 and more interesting a failed rpr job https://build.ros2.org/job/Rpr__rosbag2__ubuntu_jammy_amd64/849/testReport/junit/rosbag2_cpp/TimeControllerClockTest/unpaused_sleep_returns_true/ with those changes.

It seems the root cause of the issue is that `rclcpp::ok()` became false somewhere in the middle of the test. 
But how it could be?
cc: @clalancette

emersonknapp:
Note https://github.com/ros2/rosbag2/pull/1384 fixing the test - I'll hold off on merging this one until that is backported to Iron so we can have good happy green CI

emersonknapp:
OK - just waiting on https://github.com/ros2/rosbag2/pull/1389 to merge, then will rebase and CI

emersonknapp:
Pulls: ros2/rosbag2#1371
Gist: https://gist.githubusercontent.com/emersonknapp/bf5714f5aed1a9461b42b0c74dff1116/raw/00461a95d0a7f269e7fdc3d234551779b3fea7c5/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_storage_mcap
TEST args:  --packages-above rosbag2_storage_mcap
ROS Distro: iron
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12213
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18922/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13452/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19675/)

emersonknapp:
@mergifyio backport humble

PR #1378, opened 2023-06-07
When using sim time, wait for /clock before beginning recording

### LINKED ISSUE #1276: Messages published before /clock recorded with timestamp 0 (--use-sim-time)

emersonknapp:
Fixes https://github.com/ros2/rosbag2/issues/1276
Retry at #1354 after reverting

When using sim-time, check that `/clock` has actually been received before writing any messages into the bag, to avoid the time jump that then happens on first `/clock` message, given that initial messages are written with a timestamp of 0. 
This potentially drops some messages that it could have written, but is more correct, in that those messages are "out of time" and have no temporal standing in simulated time.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -161,6 +161,12 @@ RecorderImpl::RecorderImpl(
   paused_(record_options.start_paused),
   keyboard_handler_(std::move(keyboard_handler))
 {
+  if (record_options_.use_sim_time && record_options_.is_discovery_disabled) {
+    throw std::runtime_error(
+      "use_sim_time and is_discovery_disabled both set, but are incompatible settings. "
+      "The /clock topic needs to be discovered to record with sim time.");
+  }
+
   std::string key_str = enum_key_code_to_str(Recorder::kPauseResumeToggleKey);
   toggle_paused_key_callback_handle_ =
     keyboard_handler_->add_key_press_callback(
@@ -223,6 +229,7 @@ void RecorderImpl::record()
     return;
   }
   paused_ = record_options_.start_paused;
+  stop_discovery_ = record_options_.is_discovery_disabled;
   topic_qos_profile_overrides_ = record_options_.topic_qos_profile_overrides;
   if (record_options_.rmw_serialization_format.empty()) {
     throw std::runtime_error("No serialization format specified!");
@@ -304,9 +311,9 @@ void RecorderImpl::record()
 
   serialization_format_ = record_options_.rmw_serialization_format;
   RCLCPP_INFO(node->get_logger(), "Listening for topics...");
-  subscribe_topics(get_requested_or_available_topics());
-
-  if (!record_options_.is_discovery_disabled) {
+  if (record_options_.is_discovery_disabled) {
+    subscribe_topics(get_requested_or_available_topics());
+  } else {
     discovery_future_ =
       std::async(std::launch::async, std::bind(&RecorderImpl::topics_discovery, this));
   }
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -372,6 +372,20 @@ bool RecorderImpl::is_paused()
 
 void RecorderImpl::topics_discovery()
 {
+  // If using sim time - wait until /clock topic received before even creating subscriptions
+  if (record_options_.use_sim_time) {
+    RCLCPP_INFO(
+      node->get_logger(),
+      "use_sim_time set, waiting for /clock before starting recording...");
+    while (rclcpp::ok() && stop_discovery_ == false) {
+      if (node->get_clock()->wait_until_started(record_options_.topic_polling_interval)) {
+        break;
+      }
+    }
+    if (node->get_clock()->started()) {
+      RCLCPP_INFO(node->get_logger(), "Sim time /clock found, starting recording.");
+    }
+  }
   while (rclcpp::ok() && stop_discovery_ == false) {
     auto topics_to_subscribe =
       get_requested_or_available_topics();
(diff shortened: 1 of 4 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -364,6 +364,20 @@ bool RecorderImpl::is_paused()
 
 void RecorderImpl::topics_discovery()
 {
+  // If using sim time - wait until /clock topic received before even creating subscriptions
+  if (record_options_.use_sim_time) {
```
emersonknapp, line 384: The problem here is the initial batch of potential subscriptions in `record()`, before starting the `topics_discovery` thread. We need `record` to return synchronously so I think I will need to put the clock waiting loop in its own `std::async` call from `record`. This block of code is approximately correct, but it is in the wrong place since some subscriptions can be created before it is called.

MichaelOrlov: @emersonknapp I had a bit more simple fix in my mind to add a check 
```
if (!record_options_.use_sim_time) {
    subscribe_topics(get_requested_or_available_topics());
  }
```
in the `Recorder::record()` method. i.e. not subscribe to the specified topics on record and let spawned discovery thread do this with the delay as you did in the original PR.
It will have a side effect that `record_options_.is_discovery_disabled` and `record_options_.use_sim_time` will became incompatible if they are set to true together. 
However, we can add another check and throw a warning with an exception for invalid parameters settings.

emersonknapp: So you're saying `RecordOptions::is_discovery_disabled && RecordOptions::use_sim_time` should be an invalid setting. OK, that sounds like a good idea to me. It is reasonable to say something like "we have to discover the /clock topic" (even if this doesn't 100% capture the nuance of the implementation). I will add an error check for that configuration case.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -291,9 +298,9 @@ void RecorderImpl::record()
 
   serialization_format_ = record_options_.rmw_serialization_format;
   RCLCPP_INFO(node->get_logger(), "Listening for topics...");
-  subscribe_topics(get_requested_or_available_topics());
-
-  if (!record_options_.is_discovery_disabled) {
+  if (record_options_.is_discovery_disabled) {
+    subscribe_topics(get_requested_or_available_topics());
+  } else {
```
MichaelOrlov: Nitpick. I would prefer to keep the original logic and add 
```
if (!record_options_.use_sim_time) {
  subscribe_topics(get_requested_or_available_topics());
}
```
For the sake of the case, if someone will decide to simplify or rewrite a logic and for the sake to have a more explicit logic it will be explicitly visible that we need to handle `if (!record_options_.use_sim_time)` as a special case.
In the current implementation the logic about `if (!record_options_.use_sim_time)` is hidden in the upper-level check and via throwing exception.

MichaelOrlov (commented):
@emersonknapp While implementation looks good to me, the failure on the `record_all_with_sim_time` Is a big surprise for me.
```
20: [ RUN      ] RecordIntegrationTestFixture.record_all_with_sim_time
20: stdin is not a terminal device. Keyboard handling disabled.[INFO] [1686360701.266425264] [rosbag2_recorder]: Press SPACE for pausing/resuming
20: [INFO] [1686360701.268967062] [rosbag2_recorder]: Event publisher thread: Starting
20: [INFO] [1686360701.270181670] [rosbag2_recorder]: Listening for topics...
20: [INFO] [1686360701.270365093] [rosbag2_recorder]: use_sim_time set, waiting for /clock before starting recording...
20: /tmp/ws/src/rosbag2/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp:64: Failure
20: Value of: pub_manager.wait_for_matched(string_topic.c_str())
20:   Actual: false
20: Expected: true
20: [INFO] [1686360711.281632076] [rosbag2_recorder]: Event publisher thread: Exiting
20: [  FAILED  ] RecordIntegrationTestFixture.record_all_with_sim_time (10034 ms)
20: [----------] 1 test from RecordIntegrationTestFixture (10034 ms total)
```
It looks like our check for the
```
if (node->get_clock()->wait_until_started(record_options_.topic_polling_interval))
```
doesn't work as we expected.

emersonknapp:
The test failure was a side effect of changing the discovery behavior. The test was treating `/clock` like any other topic to record.
* Recorder was waiting for first `/clock` message before creating subscriptions
* Test was waiting for both `/clock` and `/string_topic` publishers to have matched subscriptions before publishing messages

The above two conditions created a deadlock. I've rewritten the test to publish `/clock` constantly, and only worry about recording `/string_topic`.

MichaelOrlov (approved):
@emersonknapp LGTM. Thanks for the PR!

emersonknapp:
Pulls: ros2/rosbag2#1378
Gist: https://gist.githubusercontent.com/emersonknapp/36579bb645001b58e67310f066d1f600/raw/a112d00e3d040cabccbf7c01f10c5bade79bd4bb/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12208
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18913/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13447/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19669/)

emersonknapp:
@mergifyio backport iron humble

ros-discourse:

Issue #1380, opened 2023-06-09
The test_record.TestRecordAfterShutdown.test_exit_code is tend to be flaky


MichaelOrlov:
## Description
The `test_record.TestRecordAfterShutdown.test_exit_code` from `ros2bag` is tend to be flaky since it appears to have failed multiple times recently on CI jobs:
http://ci.ros2.org/job/ci_linux/18847/
http://ci.ros2.org/job/ci_linux-aarch64/13388/
- in the PR https://github.com/ros2/rosbag2/pull/1342
and in one [build_and_test](https://github.com/ros2/rosbag2/actions/runs/5165136507/jobs/9304466822?pr=1301) run in https://github.com/ros2/rosbag2/pull/1301

### Expected Behavior
The `test_record.TestRecordAfterShutdown.test_exit_code` shall not be flaky and shall not fail randomly on the CI.

### Actual Behavior
The `test_record.TestRecordAfterShutdown.test_exit_code` randomly fails with the error message:
```
FAIL: test_record.TestRecordAfterShutdown.test_exit_code[sqlite3]
2023-06-03T19:45:38.7299964Z ----------------------------------------------------------------------------------------------------------------------------------
2023-06-03T19:45:38.7300321Z Traceback (most recent call last):
2023-06-03T19:45:38.7300670Z   File "/__w/rosbag2/rosbag2/ros_ws/src/s1oktrgw0j/rosbag2/ros2bag/test/test_record.py", line 73, in test_exit_code
2023-06-03T19:45:38.7301027Z     launch_testing.asserts.assertExitCodes(
2023-06-03T19:45:38.7301467Z   File "/__w/rosbag2/rosbag2/ros_ws/build/launch_testing/launch_testing/asserts/assert_exit_codes.py", line 62, in assertExitCodes
2023-06-03T19:45:38.7301967Z     assert info.returncode in allowable_exit_codes, 'Proc {} exited with code {}'.format(
2023-06-03T19:45:38.7302418Z AssertionError: Proc ros2bag-cli-7 exited with code -9
2023-06-03T19:45:38.7302839Z ----------------------------- Captured stdout call -----------------------------
2023-06-03T19:45:38.7303047Z 
2023-06-03T19:45:38.7303143Z Starting test run test_record[sqlite3]
2023-06-03T19:45:38.7303591Z [INFO] [launch]: All log files can be found below /github/home/.ros/log/2023-06-03-19-45-17-529104-0dd7204ee0ae-126136
[...]

PR #1381, opened 2023-06-09
Fix for possible freeze in Recorder::stop()

MichaelOrlov:
- Relates #1380
- Relates #1301
- Relates #1342

It was possible a freeze in the recorder due to the race condition when calling `Recorder::stop()` while the event publisher thread hasn't been fully started yet.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -82,7 +82,7 @@ public:
   std::shared_ptr<rosbag2_cpp::Writer> writer_;
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::RecordOptions record_options_;
-  std::atomic<bool> stop_discovery_;
+  std::atomic<bool> stop_discovery_ = false;
   std::unordered_map<std::string, std::shared_ptr<rclcpp::SubscriptionBase>> subscriptions_;
 
 private:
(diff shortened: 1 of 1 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -199,15 +205,23 @@ void RecorderImpl::stop()
   {
     std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
     event_publisher_thread_should_exit_ = true;
+    event_publisher_thread_wake_cv_.notify_all();
```
emersonknapp: I don't think this change is necessary. According to [the docs](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all):
> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s). Doing so may be a pessimization,

emersonknapp (approved):
One nitpick, otherwise looks fine

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: You are right. It seems I missed a fact that we are checking for `event_publisher_thread_should_exit_` in the `condition_variable::wait(..)` under the mutex. 
And if it is true even before we will call for the `wait(..)` for the first time. The `wait()..` will exit immediately even without `notify_all()` call.
I will move `event_publisher_thread_wake_cv_.notify_all();` out of the mutex lock.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/41a006aa45ebe367963c3e7156a1f583/raw/baee2f29cb956adfb2b348ce59d7c978a7266da7/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12191
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18894/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13430/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19651/)

MichaelOrlov:

MichaelOrlov:

ros-discourse:

PR #1382, opened 2023-06-09
Fix for rosbag2_transport::Recorder failures due to the unhandled exceptions

MichaelOrlov:
- Relates #1380
- Relates #1301
- Relates #1342

When Record exits by signal interrupt we are invalidating context by calling `rclcpp::shutdown()` it could potentially cause raising exceptions in inquiries related to the node's operations and node_graph which we are not handling. 

Unhandled exceptions from the underlying node could potentially lead to the unclean exit of the Recorder i.e. abnormal termination with a non-zero exit code.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -328,13 +328,7 @@ void RecorderImpl::event_publisher_thread_main()
       auto message = rosbag2_interfaces::msg::WriteSplitEvent();
       message.closed_file = bag_split_info_.closed_file;
       message.opened_file = bag_split_info_.opened_file;
-      try {
-        split_event_pub_->publish(message);
-      } catch (const std::exception & e) {
-        RCLCPP_ERROR_STREAM(
-          node->get_logger(),
-          "Failed to publish message on '/events/write_split' topic. \nError: " << e.what());
-      }
+      split_event_pub_->publish(message);
     }
   }
   RCLCPP_INFO(node->get_logger(), "Event publisher thread: Exiting");
@@ -379,7 +373,8 @@ bool RecorderImpl::is_paused()
 void RecorderImpl::topics_discovery()
 {
   while (rclcpp::ok() && stop_discovery_ == false) {
-    auto topics_to_subscribe = get_requested_or_available_topics();
+    auto topics_to_subscribe =
+      get_requested_or_available_topics();
     for (const auto & topic_and_type : topics_to_subscribe) {
       warn_if_new_qos_for_subscribed_topic(topic_and_type.first);
     }
@@ -399,13 +394,7 @@ void RecorderImpl::topics_discovery()
 std::unordered_map<std::string, std::string>
 RecorderImpl::get_requested_or_available_topics()
 {
-  std::map<std::string, std::vector<std::string>> all_topics_and_types;
-  try {
-    all_topics_and_types = node->get_topic_names_and_types();
-  } catch (const std::exception & e) {
-    RCLCPP_ERROR_STREAM(
-      node->get_logger(), "Failed to get topic names and types from node \nError: " << e.what());
-  }
+  auto all_topics_and_types = node->get_topic_names_and_types();
   return topic_filter_->filter_topics(all_topics_and_types);
 }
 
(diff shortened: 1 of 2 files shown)


emersonknapp:
I don't fully understand the motivation for this change. Is the problem that we are calling APIs after `rclcpp::ok() == false`?

As a pattern, I don't think it's a good practice to put catch-all `catch (std::exception)` blocks in a lot of places. If we are going to handle literally any possible exception, then it should probably be in one place up high, in the main control function or something.

One place I see you catching exceptions is in the ` event_publisher_thread` - because an unhandled exception in the thread can halt the whole program. Maybe this means that it should be an `std::async` that provides us an `std::future` instead - that way it can have an exception safely without crashing the program, and we can handle it up top in `close()`

MichaelOrlov:
Yes, the motivation was that we are calling node's APIs after `rclcpp::shutdown()` i.e. when `rclcpp::ok() == false`?

On one hand, I agree that this is not a good practice to put catch-all catch (std::exception) blocks in a lot of places, but from another hand, I have to do this since 1) our rclcpp API very often doesn't specify what exact exception could be thrown and 2) we don't have the `main` control function or something like that to be able to handle all of them in one place. 
And here is the really matter to consider in each concrete case if an exception from the underlying node is critical for us and we should abnormally finish execution or not so critical and we can continue. 

As regards to the catching exceptions is in the `event_publisher_thread` I don't think that makes sense to rewrite it with `std::async`. The only difference is that `future::get()`  will rethrow the exception that happened inside the async task. 
But do we really want this behavior?
I don't think so. At least for a while I do not see reasons for that.
And we can achieve the same result if pass `exception_ptr` to the `event_publisher_thread` and process it after the `join()` command.

@emersonknapp In general I understand your concern that catching all exceptions in multiple places is a bad practice. 
I will try to reconsider my changes if I can combine or get rid of some of them or make handlers more specific.  
If you have a concrete proposals or suggestions - I would be glad to consider them.

MichaelOrlov:
@emersonknapp I've rewritten exception handling from the underlying node in my new commit https://github.com/ros2/rosbag2/pull/1382/commits/1c79e8a31b2389dff4e4b7861518e3994e73309c. Please review it again.
Now I am handling exceptions only in two places in `event publisher` and `discovery` threads which are acting like our "main" function analogs. Since during recording, we are running only two of those threads.

I've tried to consider rethrowing exceptions or not handling them in relevant threads and processing them inside `close()` method. Although it doesn't look like the right approach IMO.  
The questions arise for instance:
- What if we got failure in discovery for some reason? 
- What if we occasionally failed to subscribe to one of the topics? 
- **Shall we terminate the recording or discovery thread in these cases?**
I think it would be more reasonable to not interrupt the recording or discovery thread in case of an error on discovery or an attempt to subscribe to one topic. We may run discovery and fail on some topic occasionally which is `\rosout` or some stray topic that might be not very interesting for us. However, it would be valuable to continue recording on already subscribed topics or try to subscribe to other available topics.  
- As regards the `event publisher` thread, the situation is similar. We likely don't want to interrupt the recording if we failed to publish a notification about the bag split event.

If considering handling deferred exceptions from discovery and event publisher threads in `recorder::close()` method. 
Let's assume we will get them there - but what we can do with them? 
We can't really do much with them since errors in `close()` usually will be happening during exit and we are just before or already in the destructor.

MichaelOrlov:
@emersonknapp friendly ping for review.

emersonknapp (approved):
:+1:

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/4b2e711c40c326977285ea111ede124c/raw/3e4519e1f68ff0a741131a0f145c9e62ec617050/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12258
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18985/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13509/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19729/)

MichaelOrlov:

MichaelOrlov:

PR #1410, opened 2023-07-05
update converter namespace

Aposhian:
The converter development guide says that the converter plugins should be in the `rosbag2` namespace, but they are actually in the `rosbag2_cpp` namespace.


emersonknapp (approved):
Good catch, thanks

PR #1416, opened 2023-07-12
Fix the build with rmw_fastrtps_dynamic.

clalancette:
When building with *only* rmw_fastrtps_dynamic, there is no test named "test_play_services__rmw_fastrtps_cpp" to mark as xfail.  Instead, it is called
"test_play_services__rmw_fastrtps_dynamic_cpp", so make sure to add a different xfail marking for that test.

This should fix the failing build at https://build.ros2.org/view/Rci/job/Rci__nightly-fastrtps-dynamic_ubuntu_jammy_amd64/

### DIFF
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -242,8 +242,11 @@ function(create_tests_for_rmw_implementation)
     ament_add_test_label(test_play_services__rmw_cyclonedds_cpp xfail)
   endif()
 
-  if(${rmw_implementation} MATCHES "rmw_fastrtps(.*)")
+  if(${rmw_implementation} MATCHES "rmw_fastrtps_cpp")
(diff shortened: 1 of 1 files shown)


clalancette:
FYI: the Rpr job here is failing because `test_play_services__rmw_fastrtps_cpp` failed.  Looking at this PR, that should really be ignored since it is marked `xfail`.  However, the Rpr jobs do not currently respect the `xfail` label, so they get run anyway.

All of that is to say is that I don't think that it should prevent this PR from going in.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19145/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13700/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19874/)

MichaelOrlov (approved):
LGTM

Crola1702:

Crola1702:

PR #1424, opened 2023-07-17
Stop inheriting from std::iterator.

clalancette:
In C++17, inheriting from std::iterator has been
deprecated: https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

Here, switch away from inheriting and just define the interface ourselves (which is the current recommended best practice).

This removes some warning when building with gcc 13.1.1

### DIFF
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_statement_wrapper.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_statement_wrapper.hpp
@@ -53,9 +53,15 @@ public:
   {
 public:
     using RowType = std::tuple<Columns ...>;
-    class Iterator : public std::iterator<std::input_iterator_tag, RowType>
+    class Iterator
     {
 public:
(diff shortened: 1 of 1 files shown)


MichaelOrlov (approved):
LGTM.

fujitatomoya (approved):
lgtm

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19196/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13755/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19938/)

MichaelOrlov:
Notes about backporting of this PR.
According to the https://acodersjourney.com/20-abi-breaking-changes/
> 6. Change the class hierarchy  of the class exposed in your dll - for example, 
> - make the exposed class inherit from a different base class
> - make the exposed class  inherit from additional base classes
> - **remove base class(es) from the exposed class**

Therefore this PR is not backportable since it will cause ABI breakage.

PR #1434, opened 2023-08-07
[humble] Reduce message spam when topics to be recorded do not exist

DanMesh:
This ports #1018 to Humble.

The message spamming reported in #1011 was occurring when using `0.15.5` as it seems the change above was never backported.

FYI @bastianhjaeger


MichaelOrlov (commented):
@DanMesh Thank you for trying to backport this PR.  
There is one major issue with this backport.  
While the fix and backport look pretty trivial the reason why it was not backported before is because it introduces ABI breaking changes when adding the new `std::unique_ptr<TopicFilter> topic_filter_;` member variable to the recorder class.   
According to the [ABI BREAKING CHANGES](https://www.acodersjourney.com/20-abi-breaking-changes/)  
> 4. Add, remove, or re-order member variables in a class

 is an ABI breaking change.  

Unfortunately, we can't introduce API/ABI breaking changes to the stable release of the ROS2 core package.  

Although, there is a chance that if we will add this new member variable to the end of the class it will still keep ABI compatibility. Need to try and run the ABI checker.

clalancette:
> Although, there is a chance that if we will add this new member variable to the end of the class it will still keep ABI compatibility. Need to try and run the ABI checker.

My understanding is that while this generally works, it is not guaranteed.  That is, the compiler is free to rearrange the in-memory layout of structures/classes in certain cases (and optimization levels).  So I don't think we can rely on that.

DanMesh:
Thanks for the feedback.
If I were to make the change suggested by @MichaelOrlov, how could that be run through the ABI checker? Is it worth pursuing?

clalancette:
> If I were to make the change suggested by @MichaelOrlov, how could that be run through the ABI checker? Is it worth pursuing?

Personally, I don't think so.  The ABI checker will likely tell you it is fine, but again it is not guaranteed so I don't think we should rely on that.

What we've done in other situations like this is to introduce a sort of global map that holds a mapping of the pointer to the new member we want to add.  Then you can look up the new variable you need when you need it, without changing ABI.  You can see an example of this kind of thing in https://github.com/ros2/rclcpp/pull/2183

DanMesh:
Thanks @clalancette. I'll give that a go and come back to this.

Moma98:
any update on this? this issue is still present on humble
Screenshot from 2023-08-30 12-55-16

DanMesh:
@Moma98 No updates at the moment. At some point, I would like to try the [suggestion above](https://github.com/ros2/rosbag2/pull/1434#issuecomment-1671665534), but have not got round to it.

Jannkar:
Hey! I have been also struggling today with this issue. We had to build rosbag2 from source using @DanMesh's branch, because this spamming issue basically made the latest humble release unusable for us. We would greatly appreciate if this issue was somehow solved for Humble as well.

MichaelOrlov:
@Jannkar Perhaps the fix in the https://github.com/ros2/rosbag2/pull/1466 will be easily portable to the humble and iron branches.

Jannkar:
Perfect! Tested it in Humble and it works as expected. Thanks for the fast solution.

DanMesh:
Closing this since https://github.com/ros2/rosbag2/pull/1469 seems to have solved the issue.
Thanks @MichaelOrlov!

Issue #1437, opened 2023-08-08
:farmer: test_rosbag2_play_end_to_end flaky on Windows


Crola1702:
## Description

Flaky test [test_rosbag2_play_end_to_end](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/testReport/junit/(root)/projectroot/test_rosbag2_play_end_to_end/) in windows CI (debug, release and repeated)

Test regressions:
* [projectroot.test_rosbag2_play_end_to_end](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/testReport/junit/(root)/projectroot/test_rosbag2_play_end_to_end/)
* [rosbag2_tests.test_rosbag2_play_end_to_end.gtest.missing_result](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/testReport/junit/(root)/rosbag2_tests/test_rosbag2_play_end_to_end_gtest_missing_result/)

### Expected Behavior
Test should pass

### Actual Behavior
Test failing because of a timeout

## To Reproduce
1. Run a build in [nightly_win_deb](https://ci.ros2.org/view/nightly/job/nightly_win_deb/), [nightly_win_rel](https://ci.ros2.org/view/nightly/job/nightly_win_rel/) or [nightly_win_rep](https://ci.ros2.org/view/nightly/job/nightly_win_rep/)
2. See projectroot.test_rosbag2_play_end_to_end fail

## System (please complete the following information)
 - OS: Windows
 - ROS 2 Distro: Rolling
 - Install Method: Source
 - Version: Rolling latest

## Additional context

Reference build: https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/

Test regressions:
* [projectroot.test_rosbag2_play_end_to_end](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/testReport/junit/(root)/projectroot/test_rosbag2_play_end_to_end/)
* [rosbag2_tests.test_rosbag2_play_end_to_end.gtest.missing_result](https://ci.ros2.org/view/nightly/job/nightly_win_deb/2813/testReport/junit/(root)/rosbag2_tests/test_rosbag2_play_end_to_end_gtest_missing_result/)

Test is failing because of a timeout:

Log output:
```
[WARN] [1691491268.193064100] [subscription_manager_1691491266065392000]: SubscriptionManager::continue_spinning(..) finished by timeout
[...]

PR #1452, opened 2023-08-23
Mark play_end_to_end test as xfail in Windows

Crola1702:
As mentioned in https://github.com/ros2/rosbag2/issues/1437

### DIFF
--- a/rosbag2_tests/CMakeLists.txt
+++ b/rosbag2_tests/CMakeLists.txt
@@ -69,6 +69,7 @@ if(BUILD_TESTING)
       rosbag2_test_common::rosbag2_test_common
       ${test_msgs_TARGETS}
     )
+    ament_add_test_label(test_rosbag2_play_end_to_end xfail)
   endif()
 
   ament_add_gmock(test_rosbag2_info_end_to_end
--- a/rosbag2_tests/CMakeLists.txt
+++ b/rosbag2_tests/CMakeLists.txt
@@ -69,7 +69,9 @@ if(BUILD_TESTING)
       rosbag2_test_common::rosbag2_test_common
       ${test_msgs_TARGETS}
     )
-    ament_add_test_label(test_rosbag2_play_end_to_end xfail)
+    if(WIN32)
+        ament_add_test_label(test_rosbag2_play_end_to_end xfail)
+    endif()
   endif()
 
   ament_add_gmock(test_rosbag2_info_end_to_end


Crola1702:
CI: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19380/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13904/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20100/)

emersonknapp:
Should we only do this on windows?

rosbag2_tests/CMakeLists.txt
```diff
@@ -69,6 +69,7 @@ if(BUILD_TESTING)
       rosbag2_test_common::rosbag2_test_common
       ${test_msgs_TARGETS}
     )
+    ament_add_test_label(test_rosbag2_play_end_to_end xfail)
```
MichaelOrlov:
```suggestion
if(WIN32)
    ament_add_test_label(test_rosbag2_play_end_to_end xfail)
endif()
```

MichaelOrlov (changes requested):
I think it makes sense to mark the test as xfail only for the Windows platform seems it only regularly fails on Windows.

rosbag2_tests/CMakeLists.txt
Crola1702: Yes you're right. I totally missed that.

Crola1702:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19399/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13928/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20120/)

clalancette (approved):
Looks reasonable to me, but I'd like to wait for final confirmation from @MichaelOrlov before merging.

MichaelOrlov (approved):
:+1:

MichaelOrlov:
@clalancette Sorry, I haven't had a chance yet to make a comprehensive analysis about the reason why this test fails on Windows. But as I said before, I don't mind disabling this test for Windows for a while. Anyway, I will have to reproduce the failure locally.

clalancette:
> @clalancette Sorry, I haven't had a chance yet to make a comprehensive analysis about the reason why this test fails on Windows. But as I said before, I don't mind disabling this test for Windows for a while. Anyway, I will have to reproduce the failure locally.

No worries at all, thanks for any time you spend on it.  I think disabling it just for Windows is a good compromise for now.

PR #1131, opened 2022-10-19
Add pause and resume service calls for rosbag2 recorder

rshanor:
Add service calls to pause recording and resume recording.

- Closes https://github.com/ros2/rosbag2/issues/1130

Signed-off-by: Rick Shanor <rickshanor@gmail.com>

PR #1453, opened 2023-08-24
[humble] Add pause and resume service calls for rosbag2 recorder (Backport #1131)

haudren-woven:
I have found a need to start & stop recording in my Humble installation. I figured that the services already exist in rosbag2_interfaces, so backporting is a very easy job. There were a couple of conflicts to manually resolve, which prevented me from going through Mergify.

Let me know if this is appropriate!

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/recorder.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/recorder.hpp
@@ -30,6 +30,9 @@
 
 #include "rosbag2_cpp/writer.hpp"
 
+#include "rosbag2_interfaces/srv/is_paused.hpp"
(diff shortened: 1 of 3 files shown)


clalancette (commented):
The big problem in general with this backport is that it changes ABI by adding members to the `Recorder` class.  Since our policy is to never break ABI in stable distributions, this is a problem.

One way to get around this is to to add in a global map of Recorder pointers to a structure containing the pieces you need.  You can see an example of this pattern in https://github.com/ros2/rclcpp/pull/2236 (look for the `BackportMembers` class).

MichaelOrlov:
- Closing as wouldn't be merged due to the breaking ABI changes. 
Feel free to reopen this PR if you going to workaround ABI breaking changes.

PR #1455, opened 2023-09-03
ros2 bag convert now excludes messages not in [start_time;end_time]

### LINKED ISSUE #599: Ability to split already existing ros2 bags by time
## Description
In rosbag there is the ability to split already existing bags by time. That functionality would be useful for ros2 as well.

## Related Issues

None

## Completion Criteria

- A bag can be split by time

Check how it was done in rosbag and how it is currently done for bags which get split by time which are recorded live.

Have a bag of 10s, split it after 5s to get 2 bags of 5s each.

pfavr2:
resolves https://github.com/ros2/rosbag2/issues/599

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -177,7 +177,7 @@ protected:
 
   // Checks if the message to be written is within accepted time range
   bool message_within_accepted_time_range(
-    const std::chrono::time_point<std::chrono::high_resolution_clock> & current_time) const;
+    const rcutils_time_point_value_t current_time) const;
 
   // Prepares the metadata by setting initial values.
   virtual void init_metadata();
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -117,7 +117,7 @@ public:
 
   /**
    * Write a message to a bagfile. The topic needs to have been created before writing is possible.
-   * Only writes message if timestamp is within start_time_ns and end_time_ns.
+   * Only writes message if within start_time_ns and end_time_ns (from storage_options).
    *
    * \param message to be written to the bagfile
    * \throws runtime_error if the Writer is not open.
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -117,6 +117,7 @@ public:
 
   /**
    * Write a message to a bagfile. The topic needs to have been created before writing is possible.
+   * Only writes message if timestamp is within start_time_ns and end_time_ns.
    *
    * \param message to be written to the bagfile
    * \throws runtime_error if the Writer is not open.
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -174,6 +174,10 @@ protected:
   bool should_split_bagfile(
     const std::chrono::time_point<std::chrono::high_resolution_clock> & current_time) const;
 
+  // Checks if the message to be written is within accepted time range
+  bool message_within_accepted_time_range(
+    const std::chrono::time_point<std::chrono::high_resolution_clock> & current_time) const;
+
   // Prepares the metadata by setting initial values.
   virtual void init_metadata();
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -436,23 +436,23 @@ bool SequentialWriter::should_split_bagfile(
   return should_split;
 }
 
-bool SequentialWriter::message_within_accepted_time_range(  
-  const rcutils_time_point_value_t current_time) const  
-{  
-  if (storage_options_.start_time_ns >= 0 &&  
-      static_cast<int64_t>(current_time) < storage_options_.start_time_ns)  
-  {  
-    return false;  
-  }  
-  
-  if (storage_options_.end_time_ns >= 0 &&  
-      static_cast<int64_t>(current_time) > storage_options_.end_time_ns)  
-  {  
-    return false;  
-  }  
-  
-  return true;  
-}  
+bool SequentialWriter::message_within_accepted_time_range(
+  const rcutils_time_point_value_t current_time) const
+{
+  if (storage_options_.start_time_ns >= 0 &&
+      static_cast<int64_t>(current_time) < storage_options_.start_time_ns)
+  {
+    return false;
+  }
+
+  if (storage_options_.end_time_ns >= 0 &&
+      static_cast<int64_t>(current_time) > storage_options_.end_time_ns)
+  {
+    return false;
+  }
+
+  return true;
+}
 
 void SequentialWriter::finalize_metadata()
 {
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -354,8 +354,9 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
 
-  if (!message_within_accepted_time_range(message_timestamp))
+  if (!message_within_accepted_time_range(message_timestamp)) {
     return;
+  }
 
   if (is_first_message_) {
     // Update bagfile starting time
@@ -436,17 +437,20 @@ bool SequentialWriter::should_split_bagfile(
 }
 
 bool SequentialWriter::message_within_accepted_time_range(
-  const std::chrono::time_point<std::chrono::high_resolution_clock> &current_time) const
+  const std::chrono::time_point<std::chrono::high_resolution_clock> & current_time) const
 {
   bool retVal = true;
   if ( (storage_options_.start_time > 0) &&
-        current_time < std::chrono::time_point<std::chrono::high_resolution_clock>(
-                       std::chrono::nanoseconds(storage_options_.start_time)) )
-          retVal = false;
-  else if ( ( storage_options_.end_time > 0) &&
-        current_time > std::chrono::time_point<std::chrono::high_resolution_clock>(
-                       std::chrono::nanoseconds(storage_options_.end_time)) )
-          retVal = false;
+    current_time < std::chrono::time_point<std::chrono::high_resolution_clock>(
+      std::chrono::nanoseconds(storage_options_.start_time)) )
+  {
+    retVal = false;
+  } else if ( ( storage_options_.end_time > 0) &&
+    current_time > std::chrono::time_point<std::chrono::high_resolution_clock>(
+      std::chrono::nanoseconds(storage_options_.end_time)) )
+  {
+    retVal = false;
+  }
   return retVal;
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -354,6 +354,8 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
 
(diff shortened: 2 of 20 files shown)


pfavr2:
This is my first PR to ROS2. If it is not good enough I look forward to hearing your comments!

I hope to be able to contribute a lot more in the coming years. Best regards, Peter

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -432,6 +436,26 @@ bool SequentialWriter::should_split_bagfile(
   return should_split;
 }
 
+bool SequentialWriter::message_within_accepted_time_range(
+  const std::chrono::time_point<std::chrono::high_resolution_clock> & current_time) const
+{
+  bool retVal = true;
```
MichaelOrlov: Style. `retValue` => `ret_value`

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -354,6 +354,10 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
   const auto message_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
     std::chrono::nanoseconds(message->time_stamp));
 
+  if (!message_within_accepted_time_range(message_timestamp)) {
```
MichaelOrlov: Please put this check before extracting `topic_information`

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -56,6 +57,10 @@ struct StorageOptions
   // Defaults to disabled.
   bool snapshot_mode = false;
 
+  // Start and end time for cutting
+  rcutils_time_point_value_t start_time = 0;
+  rcutils_time_point_value_t end_time = 0;
+
```
MichaelOrlov, line 62: To avoid bringing extra dependencies I would recommend using a plain `uint64_t` type for datestamps or durations similar to what we already use for the `max_bagfile_duration`

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: Please update the description for the `write(..)` function in the header file according to the changed behavior.

MichaelOrlov (changes requested):
@pfavr2 Thank you for your contribution.
I was thinking about the usability of this feature. 
For the final user probably would be difficult to find what the exact timestamps in nanoseconds should be for cutting to put them in yaml config file for `bag rewrite`.  
It would be much easier to operate with time offset or duration from the beginning of the bag for start and end time when configuring cutting interval.  
Also curious about if the final user would like to have multiple intervals for bag rewrite.
It might make sense to define a vector or array of the start-end intervals.

Another point is that we would like to avoid extra dependencies in the bottom layers of the rosbag2.
To avoid bringing extra dependencies I would recommend using a plain `uint64_t` type for datestamps or durations similar to what we have already used for the `max_bagfile_duration`.

Also need to add e few unit tests to make sure that data stamps parsing properly from and to the yaml files and make sure that sequential_writer correctly respects start and end time.

pfavr2:
Thank you very much for the review. I'll look into the improvements soon!

I agree it would be nice to have multiple time formats for the yaml file and this could be added later.

Specifying the start and end time in nanoseconds are easy when using foxglove for previewing the original file. The time of the marker can be shown in seconds with nanoseconds precision. So copy-paste and delete the decimal point. One possibility would be to rename the start_time and end_time to start_time_ns and end_time_ns for the yaml file? Then later additional time formats can be added as well... would this suffice for now?


The ability to split into multiple output-files with different start_time and end_time is useful for our application of this feature. I think it could be useful for others (you don't have to use it if you only need a single output file - but it won't hurt to offer the possibility).

Thank you again for the review comments.

Best regards,

Peter

MichaelOrlov:
@pfavr2 Using nanoseconds representation and copy past and deleting decimal point from Foxglove studio - sounds like a reasonable approach. 
For instance, if we have a time stamp from Foxglove studio `1666744557.446749846` **seconds** it could be easily transformed to the  `1666744557446749846` **nanoseconds** by removing the dot from the middle.

And yes, please rename variables to the  `start_time_ns` and `end_time_ns` for the yaml file.

When I was talking about the array of the  `start` - `end` timestamps I meant that those chunks would be cut and inserted in one new file, not in the multiple files.  
If you need to cut multiple chunks to each new file you can use a split example
```
$ ros2 bag convert -i bag1 -o out.yaml
# out.yaml
output_bags:
- uri: split1
  topics: [/topic1, /topic2]
- uri: split2
  topics: [/topic1, /topic2]
```

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
pfavr2: In rosbag2_transport/include/rosbag2_transport/play_options.hpp the datatype rcutils_time_point_value_t is used for start_offset (line 87):

```
  // Time to start playback as an offset from the beginning of the bag.
  rcutils_time_point_value_t start_offset = 0;
```

I used this as guideline. Do you think this would be better of using uint64_t also (IMHO this should be same type everywhere)?

MichaelOrlov: @pfavr2 The difference is that `start_offset` lives only on the `rosbag2_transport` and upper layers.
As have said we would like to avoid extra dependencies on the bottom layers and already using `uint64_t` type for time duration. 
For consistency, it would be more reasonable to use the same `uint64_t` type for the `start` and `end`  timestamps. Since they are in nanoseconds it shouldn't cause any problems or difficulties.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -117,6 +117,7 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
 
   /**
    * Write a message to a bagfile. The topic needs to have been created before writing is possible.
+   * Only writes message if timestamp is within start_time_ns and end_time_ns.
```
MichaelOrlov: @pfavr2 Could you please add a note that `start_time_ns` and `end_time_ns` are from `storage_options`?

rosbag2_storage_mcap/CMakeLists.txt
```diff
@@ -43,6 +44,7 @@ target_compile_definitions(${PROJECT_NAME} PRIVATE "ROSBAG2_STORAGE_MCAP_BUILDIN
 target_link_libraries(${PROJECT_NAME}
   mcap_vendor::mcap
   pluginlib::pluginlib
+  rclcpp::rclcpp
```
MichaelOrlov: @pfavr2 Have I missed something?  
Why do we need `rclcpp` package dependencies on the storage plugins layer?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -432,6 +436,26 @@ bool SequentialWriter::should_split_bagfile(
...
+  } else {
+    if ( (storage_options_.end_time_ns > 0) &&
+      current_time > std::chrono::time_point<std::chrono::high_resolution_clock>(
+        std::chrono::nanoseconds(storage_options_.end_time_ns)) )
+    {
+      ret_value = false;
+    }
+  }
+  return ret_value;
+}
```
MichaelOrlov: We can avoid at least 3 type conversions to the `std::chrono::time_point<std::chrono::high_resolution_clock>(timestamp)` if we would be using  `uint32_t` type for the `start_time_ns` and `end_time_ns` parameters and use native comparison for timestamps. It will be better from the performance point of view.

MichaelOrlov (commented):
@pfavr2 We can avoid at least 3 type conversions to the `std::chrono::time_point<std::chrono::high_resolution_clock>(timestamp)` when checking if message in the accepting time range if we would be using  `uint32_t` type for the `start_time_ns` and `end_time_ns` parameters and use native comparison for timestamps. It will be better from the performance point of view.
Because `uint32_t` is exactly half of the `int64_t` range and we wouldn't need a type cast for the `message->time_stamp > start_time_ns` comparison.

Or maybe even using `int64_t` type in this case 0 will be a valid range and we can default to the -1 to determine that start or end time limit is not set.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: Or maybe using `int64_t` type in this case 0 will be a valid range and we can default to the -1 to determine that start or end time limit is not set.

pfavr2: Are we optimising for 64bit or 32bit machines - the 2^32 nanoseconds are only ~4 seconds so 64bit is needed I think?

I'll try to avoid the type conversions.

rosbag2_storage_mcap/CMakeLists.txt
pfavr2: Thank you - I'll remove the superfluous dependency here.

I do not have the architectural diagram in my head - is there an overview somewhere that you can point me to?

(I might be missing other places where rclcpp is not needed?)

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
pfavr2: I have now simplified message_within_accepted_time_range() and now it uses int64_t.

I did look into performance also. Converting a 900MB mcap takes ~25s on my Mac M2.
The simplified code did not change the performance noticeably.
As a matter of fact - the writing of the new .mcap (and the within-check) is negligible. Writing to 2 new files simultaneously takes also ~25s. I guess the grunt work is happening in the reading.

rosbag2_storage_sqlite3/CMakeLists.txt
```diff
@@ -42,6 +43,7 @@ target_link_libraries(${PROJECT_NAME}
   pluginlib::pluginlib
   rosbag2_storage::rosbag2_storage
   rcpputils::rcpputils
+  rclcpp::rclcpp
```
MichaelOrlov:
```suggestion

```

rosbag2_storage_sqlite3/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
 find_package(pluginlib REQUIRED)
 find_package(rcpputils REQUIRED)
+find_package(rclcpp REQUIRED)
```
MichaelOrlov:
```suggestion
```
We don't need `rclcpp` dependencies on the SLite3 storage plugin level.

MichaelOrlov (changes requested):
@pfavr2 Thanks for changing the type to the `int64_t`. 
uint32_ t certainly is not enough, thanks for pointing it out.
Now it looks good! Almost ready to go. 
Please remove `rclcpp` dependencies from the SQLite3 storage plugin.  
We don't have a structural diagram with rules but we are trying to keep lower rosbag2 levels such as storage and storage plugins to be almost ROS agnostic to avoid cyclic dependencies if will be needed to use in other core packages.  
Also usually we catch out on code reviews if someone adds unnecessary dependencies in any packages :)

MichaelOrlov (approved):
LGTM with green CI.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/4af23e6b6151048a6b10b02c16815726/raw/e98144f2a24dd82b42d0c621a2137b2bf908c861/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_storage rosbag2_py
TEST args:  --packages-above rosbag2_cpp rosbag2_storage rosbag2_py
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12716
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19580/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14091/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20305/)

Yuchen966:
Is this feature compatible with humble?

MichaelOrlov:
Unfortunately not, since it changes ABI.

berndpfrommer:
Is there any documentation for this feature, like an example syntax. I can't get it to work on rolling. Option does nothing.

MichaelOrlov:
@berndpfrommer Have you tried to use 
```
start_time_ns
end_time_ns 
```
for the `output_bags:` in yaml config files for the `ros2 bag convert`?

berndpfrommer:
By now I have, and yes it works, and it would have worked even sooner had I not written "stop_time_ns".
I looked at the source code, and still messed it up.
If this had been in the README example I would have just cut-and-paste it over and not made this mistake.

Will you accept a PR from me with doc update for it? I was going to just put it into the example file .yaml file, but
commented out.

MichaelOrlov:
> Will you accept a PR from me with doc update for it? I was going to just put it into the example file .yaml file, but
commented out.

Sure, the PR is welcome!

PR #385, opened 2020-04-21
Run Action CI on Ubuntu 20.04 Focal

emersonknapp:
The master branch of this repository targets ROS 2 Foxy, which targets Ubuntu 20.04

Update CI to reflect this target environment

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

PR #1457, opened 2023-09-04
Add option to set compression threads priority

jmachowinski:
Added option to the the nice value of the compression thread.

### DIFF
--- a/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
@@ -17,6 +17,7 @@
 
 #include <cstdint>
 #include <string>
+#include <optional>
 
 #include "visibility_control.hpp"
 
@@ -61,6 +62,9 @@ struct CompressionOptions
   CompressionMode compression_mode;
   uint64_t compression_queue_size;
   uint64_t compression_threads;
+  /// if set, the compression thread(s) will try to set
+  /// the given priority for itself
+  std::optional<int8_t> thread_priority;
 };
 
 }  // namespace rosbag2_compression
--- a/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
@@ -64,7 +64,7 @@ struct CompressionOptions
   uint64_t compression_threads;
   /// if set, the compression thread(s) will try to set
   /// the given priority for itself
-  std::optional<int8_t> thread_priority;
+  std::optional<int32_t> thread_priority;
 };
 
 }  // namespace rosbag2_compression
--- a/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
@@ -61,9 +61,11 @@ struct CompressionOptions
   std::string compression_format;
   CompressionMode compression_mode;
   uint64_t compression_queue_size;
+  /// \brief // The number of compression threads
   uint64_t compression_threads;
-  /// if set, the compression thread(s) will try to set
-  /// the given priority for itself
+  /// \brief If set, the compression thread(s) will try to set the given priority for itself
+  /// For Windows the valid values are: THREAD_PRIORITY_LOWEST, THREAD_PRIORITY_BELOW_NORMAL and
+  /// THREAD_PRIORITY_NORMAL. For POSIX compatible OSes this is the "nice" value.
   std::optional<int32_t> thread_priority;
 };
 
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cstring>
 #include <functional>
 #include <memory>
 #include <stdexcept>
@@ -33,6 +34,12 @@
 #include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
 
 #include "logging.hpp"
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#include <sys/resource.h>
+#endif
 
 namespace rosbag2_compression
 {
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
 
 void SequentialCompressionWriter::compression_thread_fn()
 {
+  if (compression_options_.thread_priority) {
+
+#ifdef _WIN32
+    ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+      "Changing thread priority is not implemented for windows");
+
+    /**
+     * The implementation should look something like this
+
+    uint8_t nice_value = *compression_options_.thread_nice_value;
+
+    // this must match THREAD_PRIORITY_IDLE, THREAD_PRIORITY_LOWEST...
+    DWORD dwThreadPri = *compression_options_.thread_nice_value;
+
+    if(!SetThreadPriority(GetCurrentThread(), dwThreadPri))
+    {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(GetLastError()));
+    }
+    */
+
+#else
+    int wanted_nice_value = *compression_options_.thread_priority;
+
+    errno = 0;
+    int cur_nice_value = getpriority(PRIO_PROCESS, 0);
+    if (cur_nice_value != -1 && errno == 0) {
+      int new_nice_value = nice(wanted_nice_value - cur_nice_value);
+      if ((new_nice_value == -1 && errno != 0) || new_nice_value != wanted_nice_value) {
+        ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+          "Could not set nice value of compression thread to " << wanted_nice_value << " : " << std::strerror(
+            errno));
+      }
+    } else {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << wanted_nice_value <<
+          " : Could not determine cur nice value");
+    }
+#endif
+  }
+
   // Every thread needs to have its own compression context for thread safety.
   auto compressor = compression_factory_->create_compressor(
     compression_options_.compression_format);
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -71,24 +71,23 @@ void SequentialCompressionWriter::compression_thread_fn()
 {
   if (compression_options_.thread_priority) {
 #ifdef _WIN32
-    ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-      "Changing thread priority is not implemented for windows");
-
-    /**
-     * The implementation should look something like this
-
-    uint8_t nice_value = *compression_options_.thread_nice_value;
-
-    // this must match THREAD_PRIORITY_IDLE, THREAD_PRIORITY_LOWEST...
-    DWORD dwThreadPri = *compression_options_.thread_nice_value;
-
-    if(!SetThreadPriority(GetCurrentThread(), dwThreadPri))
-    {
+    // This must match THREAD_PRIORITY_IDLE, THREAD_PRIORITY_LOWEST...
+    int wanted_thread_priority = *compression_options_.thread_priority;
+    if (!SetThreadPriority(GetCurrentThread(), wanted_thread_priority)) {
       ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(GetLastError()));
+        "Could not set thread priority of compression thread to the: " << wanted_thread_priority <<
+          ". Error code: " << GetLastError());
+    } else {
+      auto detected_thread_priority = GetThreadPriority(GetCurrentThread());
+      if (detected_thread_priority == THREAD_PRIORITY_ERROR_RETURN) {
+        ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+          "Failed to get current thread priority. Error code: " << GetLastError());
+      } else if (wanted_thread_priority != detected_thread_priority) {
+        ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+          "Could not set thread priority of compression thread to the: " <<
+            wanted_thread_priority << ". Detected thread priority: " << detected_thread_priority);
+      }
     }
-    */
-
 #else
     int wanted_nice_value = *compression_options_.thread_priority;
 
@@ -96,14 +95,14 @@ void SequentialCompressionWriter::compression_thread_fn()
     int cur_nice_value = getpriority(PRIO_PROCESS, 0);
     if (cur_nice_value == -1 && errno != 0) {
       ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-        "Could not set nice value of compression thread to " << wanted_nice_value <<
+        "Could not set nice value of compression thread to the: " << wanted_nice_value <<
           " : Could not determine cur nice value");
     } else {
       int new_nice_value = nice(wanted_nice_value - cur_nice_value);
       if ((new_nice_value == -1 && errno != 0)) {
         ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-          "Could not set nice value of compression thread to " << wanted_nice_value <<
-            " : " << std::strerror(errno));
+          "Could not set nice value of compression thread to the: " << wanted_nice_value <<
+            ". Error : " << std::strerror(errno));
       }
     }
 #endif
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -70,7 +70,6 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
 void SequentialCompressionWriter::compression_thread_fn()
 {
   if (compression_options_.thread_priority) {
-
 #ifdef _WIN32
     ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
       "Changing thread priority is not implemented for windows");
@@ -103,8 +102,8 @@ void SequentialCompressionWriter::compression_thread_fn()
       int new_nice_value = nice(wanted_nice_value - cur_nice_value);
       if ((new_nice_value == -1 && errno != 0)) {
         ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-          "Could not set nice value of compression thread to " << wanted_nice_value << " : " << std::strerror(
-            errno));
+          "Could not set nice value of compression thread to " << wanted_nice_value <<
+            " : " << std::strerror(errno));
       }
     }
 #endif
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -75,7 +75,7 @@ void SequentialCompressionWriter::compression_thread_fn()
     int wanted_thread_priority = *compression_options_.thread_priority;
     if (!SetThreadPriority(GetCurrentThread(), wanted_thread_priority)) {
       ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-        "Could not set thread priority of compression thread to the: " << wanted_thread_priority <<
+        "Could not set thread priority of compression thread to: " << wanted_thread_priority <<
           ". Error code: " << GetLastError());
     } else {
       auto detected_thread_priority = GetThreadPriority(GetCurrentThread());
@@ -84,7 +84,7 @@ void SequentialCompressionWriter::compression_thread_fn()
           "Failed to get current thread priority. Error code: " << GetLastError());
       } else if (wanted_thread_priority != detected_thread_priority) {
         ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-          "Could not set thread priority of compression thread to the: " <<
+          "Could not set thread priority of compression thread to: " <<
             wanted_thread_priority << ". Detected thread priority: " << detected_thread_priority);
       }
     }
@@ -95,13 +95,13 @@ void SequentialCompressionWriter::compression_thread_fn()
     int cur_nice_value = getpriority(PRIO_PROCESS, 0);
     if (cur_nice_value == -1 && errno != 0) {
       ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-        "Could not set nice value of compression thread to the: " << wanted_nice_value <<
+        "Could not set nice value of compression thread to: " << wanted_nice_value <<
           " : Could not determine cur nice value");
     } else {
       int new_nice_value = nice(wanted_nice_value - cur_nice_value);
       if ((new_nice_value == -1 && errno != 0)) {
         ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-          "Could not set nice value of compression thread to the: " << wanted_nice_value <<
+          "Could not set nice value of compression thread to: " << wanted_nice_value <<
             ". Error : " << std::strerror(errno));
       }
     }
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
@@ -95,17 +95,17 @@ void SequentialCompressionWriter::compression_thread_fn()
 
     errno = 0;
     int cur_nice_value = getpriority(PRIO_PROCESS, 0);
-    if (cur_nice_value != -1 && errno == 0) {
+    if (cur_nice_value == -1 && errno != 0) {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << wanted_nice_value <<
+          " : Could not determine cur nice value");
+    } else {
       int new_nice_value = nice(wanted_nice_value - cur_nice_value);
-      if ((new_nice_value == -1 && errno != 0) || new_nice_value != wanted_nice_value) {
+      if ((new_nice_value == -1 && errno != 0)) {
         ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
           "Could not set nice value of compression thread to " << wanted_nice_value << " : " << std::strerror(
             errno));
       }
-    } else {
-      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
-        "Could not set nice value of compression thread to " << wanted_nice_value <<
-          " : Could not determine cur nice value");
     }
 #endif
   }
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -36,12 +36,10 @@
 #include "mock_storage_factory.hpp"
 
 #include "mock_compression_factory.hpp"
+#include "fake_compression_factory.hpp"
 
 #ifdef _WIN32
 #include <windows.h>
-#else
-#include <unistd.h>
-#include <sys/resource.h>
 #endif
 
 using namespace testing;  // NOLINT
@@ -401,67 +399,16 @@ TEST_P(SequentialCompressionWriterTest, writer_writes_with_compression_queue_siz
   EXPECT_EQ(fake_storage_size_, kNumMessagesToWrite);
 }
 
-
-TEST_P(SequentialCompressionWriterTest, writer_sets_nice_value)
+TEST_P(SequentialCompressionWriterTest, writer_sets_threads_priority)
 {
   const std::string test_topic_name = "test_topic";
   const std::string test_topic_type = "test_msgs/BasicTypes";
   const uint64_t kCompressionQueueSize = GetParam();
-  const int wanted_nice_value = 10;
-
-  class TestCompressor : public rosbag2_compression::BaseCompressorInterface
-  {
-    int & detected_nice_value;
-
-public:
-    TestCompressor(int & detected_nice_value)
-    : detected_nice_value(detected_nice_value) {}
-    virtual std::string compress_uri(const std::string & uri)
-    {
-      return uri;
-    }
-
-    virtual void compress_serialized_bag_message(
-      const rosbag2_storage::SerializedBagMessage * bag_message,
-      rosbag2_storage::SerializedBagMessage * compressed_message)
-    {
-#ifdef _WIN32
-      int cur_nice_value = getpriority(PRIO_PROCESS, 0);
-      if (cur_nice_value != -1 && errno == 0) {
-
-        detected_nice_value = cur_nice_value;
-      }
+#ifndef _WIN32
+  const int wanted_thread_priority = 10;
 #else
-      //FIXME implement windows version
+  const int wanted_thread_priority = THREAD_MODE_BACKGROUND_BEGIN;
 #endif
-
-      *compressed_message = *bag_message;
-    }
-
-    /**
-     * Get the identifier of the compression algorithm.
-     * This is appended to the extension of the compressed file.
-     */
-    virtual std::string get_compression_identifier() const
-    {
-      return "niceTest";
-    }
-  };
-
-  class FakeFactory : public rosbag2_compression::CompressionFactory
-  {
-    int & detected_nice_value;
-
-public:
-    FakeFactory(int & detected_nice_value)
-    : detected_nice_value(detected_nice_value) {}
-
-    virtual std::shared_ptr<rosbag2_compression::BaseCompressorInterface>
-    create_compressor(const std::string & /*compression_format*/)
-    {
-      return std::make_shared<TestCompressor>(detected_nice_value);
-    }
-  };
 
   // queue size should be 0 or at least the number of remaining messages to prevent message loss
   rosbag2_compression::CompressionOptions compression_options {
@@ -469,14 +416,20 @@ public:
     rosbag2_compression::CompressionMode::MESSAGE,
     kCompressionQueueSize,
     kDefaultCompressionQueueThreads,
-    wanted_nice_value
+    wanted_thread_priority
   };
 
+#ifndef _WIN32
   // nice values are in the range from -20 to +19, so this value will never be read
-  int detected_nice_value = 100;
+  int detected_thread_priority = 100;
+#else
+  int detected_thread_priority = THREAD_PRIORITY_ERROR_RETURN;
+#endif
 
   initializeFakeFileStorage();
-  initializeWriter(compression_options, std::make_unique<FakeFactory>(detected_nice_value));
+  initializeWriter(
+    compression_options,
+    std::make_unique<FakeCompressionFactory>(detected_thread_priority));
 
   writer_->open(tmp_dir_storage_options_);
   writer_->create_topic({test_topic_name, test_topic_type, "", "", ""});
@@ -490,8 +443,7 @@ public:
   }
   writer_.reset();    // reset will call writer destructor
 
-  EXPECT_EQ(detected_nice_value, *compression_options.thread_priority);
-
+  EXPECT_EQ(detected_thread_priority, *compression_options.thread_priority);
   EXPECT_EQ(fake_storage_size_, kNumMessagesToWrite);
 }
 
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -148,7 +148,7 @@ public:
 
   const uint64_t kDefaultCompressionQueueSize = 1;
   const uint64_t kDefaultCompressionQueueThreads = 4;
-  const std::optional<int8_t> kDefaultCompressionQueueThreadsPriority = std::nullopt;
+  const std::optional<int32_t> kDefaultCompressionQueueThreadsPriority = std::nullopt;
 };
 
 TEST_F(SequentialCompressionWriterTest, open_throws_on_empty_storage_options_uri)
@@ -405,9 +405,9 @@ TEST_P(SequentialCompressionWriterTest, writer_sets_threads_priority)
   const std::string test_topic_type = "test_msgs/BasicTypes";
   const uint64_t kCompressionQueueSize = GetParam();
 #ifndef _WIN32
-  const int wanted_thread_priority = 10;
+  const int32_t wanted_thread_priority = 10;
 #else
-  const int wanted_thread_priority = THREAD_MODE_BACKGROUND_BEGIN;
+  const int32_t wanted_thread_priority = THREAD_MODE_BACKGROUND_BEGIN;
 #endif
 
   // queue size should be 0 or at least the number of remaining messages to prevent message loss
@@ -421,9 +421,9 @@ TEST_P(SequentialCompressionWriterTest, writer_sets_threads_priority)
 
 #ifndef _WIN32
   // nice values are in the range from -20 to +19, so this value will never be read
-  int detected_thread_priority = 100;
+  int32_t detected_thread_priority = 100;
 #else
-  int detected_thread_priority = THREAD_PRIORITY_ERROR_RETURN;
+  int32_t detected_thread_priority = THREAD_PRIORITY_ERROR_RETURN;
 #endif
 
   initializeFakeFileStorage();
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -37,6 +37,13 @@
 
 #include "mock_compression_factory.hpp"
 
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#include <sys/resource.h>
+#endif
+
 using namespace testing;  // NOLINT
 
 static constexpr const char * DefaultTestCompressor = "fake_comp";
@@ -390,6 +397,100 @@ TEST_P(SequentialCompressionWriterTest, writer_writes_with_compression_queue_siz
     writer_->write(message);
   }
   writer_.reset();  // reset will call writer destructor
+
+  EXPECT_EQ(fake_storage_size_, kNumMessagesToWrite);
+}
+
+
+TEST_P(SequentialCompressionWriterTest, writer_sets_nice_value)
+{
+  const std::string test_topic_name = "test_topic";
+  const std::string test_topic_type = "test_msgs/BasicTypes";
+  const uint64_t kCompressionQueueSize = GetParam();
+  const int wanted_nice_value = 10;
+
+  class TestCompressor : public rosbag2_compression::BaseCompressorInterface
+  {
+    int & detected_nice_value;
+
+public:
+    TestCompressor(int & detected_nice_value)
+    : detected_nice_value(detected_nice_value) {}
+    virtual std::string compress_uri(const std::string & uri)
+    {
+      return uri;
+    }
+
+    virtual void compress_serialized_bag_message(
+      const rosbag2_storage::SerializedBagMessage * bag_message,
+      rosbag2_storage::SerializedBagMessage * compressed_message)
+    {
+#ifdef _WIN32
+      int cur_nice_value = getpriority(PRIO_PROCESS, 0);
+      if (cur_nice_value != -1 && errno == 0) {
+
+        detected_nice_value = cur_nice_value;
+      }
+#else
+      //FIXME implement windows version
+#endif
+
+      *compressed_message = *bag_message;
+    }
+
+    /**
+     * Get the identifier of the compression algorithm.
+     * This is appended to the extension of the compressed file.
+     */
+    virtual std::string get_compression_identifier() const
+    {
+      return "niceTest";
+    }
+  };
+
+  class FakeFactory : public rosbag2_compression::CompressionFactory
+  {
+    int & detected_nice_value;
+
+public:
+    FakeFactory(int & detected_nice_value)
+    : detected_nice_value(detected_nice_value) {}
+
+    virtual std::shared_ptr<rosbag2_compression::BaseCompressorInterface>
+    create_compressor(const std::string & /*compression_format*/)
+    {
+      return std::make_shared<TestCompressor>(detected_nice_value);
+    }
+  };
+
+  // queue size should be 0 or at least the number of remaining messages to prevent message loss
+  rosbag2_compression::CompressionOptions compression_options {
+    DefaultTestCompressor,
+    rosbag2_compression::CompressionMode::MESSAGE,
+    kCompressionQueueSize,
+    kDefaultCompressionQueueThreads,
+    wanted_nice_value
+  };
+
+  // nice values are in the range from -20 to +19, so this value will never be read
+  int detected_nice_value = 100;
+
+  initializeFakeFileStorage();
+  initializeWriter(compression_options, std::make_unique<FakeFactory>(detected_nice_value));
+
+  writer_->open(tmp_dir_storage_options_);
+  writer_->create_topic({test_topic_name, test_topic_type, "", "", ""});
+
+  auto message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
+  message->topic_name = test_topic_name;
+
+  const size_t kNumMessagesToWrite = 5;
+  for (size_t i = 0; i < kNumMessagesToWrite; i++) {
+    writer_->write(message);
+  }
+  writer_.reset();    // reset will call writer destructor
+
+  EXPECT_EQ(detected_nice_value, *compression_options.thread_priority);
 
   EXPECT_EQ(fake_storage_size_, kNumMessagesToWrite);
 }
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -143,13 +143,15 @@ public:
 
   const uint64_t kDefaultCompressionQueueSize = 1;
   const uint64_t kDefaultCompressionQueueThreads = 4;
+  const std::optional<int8_t> kDefaultCompressionQueueThreadsPriority = std::nullopt;
 };
 
 TEST_F(SequentialCompressionWriterTest, open_throws_on_empty_storage_options_uri)
 {
   rosbag2_compression::CompressionOptions compression_options{
     DefaultTestCompressor, rosbag2_compression::CompressionMode::FILE,
-    kDefaultCompressionQueueSize, kDefaultCompressionQueueThreads};
+    kDefaultCompressionQueueSize, kDefaultCompressionQueueThreads,
+    kDefaultCompressionQueueThreadsPriority};
   initializeWriter(compression_options);
 
   EXPECT_THROW(
@@ -163,7 +165,8 @@ TEST_F(SequentialCompressionWriterTest, open_throws_on_bad_compression_format)
 {
   rosbag2_compression::CompressionOptions compression_options{
     "bad_format", rosbag2_compression::CompressionMode::FILE,
-    kDefaultCompressionQueueSize, kDefaultCompressionQueueThreads};
+    kDefaultCompressionQueueSize, kDefaultCompressionQueueThreads,
+    kDefaultCompressionQueueThreadsPriority};
   initializeWriter(compression_options);
 
   EXPECT_THROW(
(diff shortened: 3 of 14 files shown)


arneboe:
@emersonknapp Could you review this (or find someone to to it)? This is an important feature that enables us to use rosbag2 in production environments without grinding the whole system to a halt when writing large bag files.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +64,21 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
 {
+  if (compression_options_.thread_nice_value) {
+    uint8_t nice_value = *compression_options_.thread_nice_value;
+
+    // according to posix this sets the nice value for the whole process
+    // but in reality this sets the thread priority on linux.
+    // There does not seem to be any other way to set the thread priority on
+    // a non-realtime kernel :-(
+    // https://stackoverflow.com/questions/10876342/equivalent-of-setthreadpriority-on-linux-pthreads
+    if (0 != setpriority(PRIO_PROCESS, 0, nice_value)) {
```
clalancette: Does this work on Windows?

clalancette: Also, I'm not convinced that this is the correct place to do this (even though it is hidden behind an option).  If you compose this node with other nodes, then you'll be setting the niceness of the whole process, not just this node.  That may be your intention, but in that case why not just use the system tools (like `nice`) to do this?  Can you describe your use case in more detail?

jmachowinski: good point, I'll add an ifdef version....

jmachowinski: > Also, I'm not convinced that this is the correct place to do this (even though it is hidden behind an option). If you compose this node with other nodes, then you'll be setting the niceness of the whole process, not just this node. That may be your intention, but in that case why not just use the system tools (like `nice`) to do this? Can you describe your use case in more detail?

Its only setting the nice value of the calling thread, not the whole process, see
man 2 setpriority 
section BUGS

So the use case is, that while we dump our data into the logdata, we want the main thread to continue receiving data with normal scheduling priority. Also we don't want other processes, to get less scheduler time due to the writing of the log file.

jmachowinski: I added an ifdef for windows, and an implementation that should work, but which is commented, as I can not test it due to a lack of a windows...

For the linux implementation I switched to the nice function, which by documentation only modifies the current thread:
`DESCRIPTION
       nice() adds inc to the nice value for the calling thread.  (A higher nice value means a lower priority.)
`
, without using dubious bugs in the posix implementation.

MichaelOrlov: @jmachowinski It is a bit of a hacky and non-deterministic way to manage resources.  
The `nice` and `setpriority` actually doing exactly the same. The reason why you see in Linux documentation that nice changing niceness for the thread rather than a priority is not in a different implementation of the `nice` for Linux. 
It is just because of another issue that Linux threads implementation doesn't follow the POSIX.1 standard.
See details at the https://stackoverflow.com/a/7684821

The more appropriate and deterministic way to manage resources would be:
1. Run rosbag2 recorder in a separate process
2. Run rosbag2 recorder on separate isolated CPU cores to avoid interference with other processes.

jmachowinski: I should elaborate on the context of this commit.
Our use case is an industrial system, that has a node, that records messages from various topics into a **memory limited** ringbuffer. (https://github.com/cellumation/snapshotter) The reception of the messages is done in the main thread, and this should have a normal scheduler priority, so that we don't drop messages. 
For debugging of problems, the customers can trigger a so called snapshot, that dumps the ringbuffer to the disk.
This is done, while the machine is running, therefore we do not want the compression and writing thread to have a high priority, to not interfere with the rest of the system.

Note, that we do not use rosbag2 directly, as the memory overhead of saving the topic name with each message is not acceptable for our use case.

@MichaelOrlov 
> Run rosbag2 recorder in a separate process

In this case we would eventually drop messages as the reception thread would also have a high nice level.

>  Run rosbag2 recorder on separate isolated CPU cores to avoid interference with other processes.

We don't have enough CPUs for that.

Anyhow, I just found out, that the nice level of a new thread is inherited from its parent, and that therefore I could drop it in my code before calling open. Therefore I would be fine with dropping this PR if you don't think this in general a useful feature.

MichaelOrlov: @jmachowinski Thanks for the explanation and details.  
Now it makes sense, I see the point. Let's work together and try to come to some solution for that.

MichaelOrlov: @jmachowinski BTW as regards
> Note, that we do not use rosbag2 directly, as the memory overhead of saving the topic name with each message is not acceptable for our use case.

With the new MCAP storage format, we are not storing the topic name with each message.
See detail in the
https://github.com/ros2/rosbag2/blob/dcd239ceea842752c344f0048572deb5f502feb8/rosbag2_storage_mcap/src/mcap_storage.cpp#L732-L742

jmachowinski: > @jmachowinski BTW as regards
> 
> With the new MCAP storage format, we are not storing the topic name with each message. See detail in the
> 

I meant the MessageCacheInterface, which is bound to the rosbag2_storage::SerializedBagMessage.

I could not see a way of implementing my own buffer, without using rosbag2_storage::SerializedBagMessage.
Note, I was talking about the **in memory** overhead, which is significant, if you have a lot of small messages and per message the topic name is saved.

Here you can find a comparison of memory usage for the ros1 version, were we ran into the same issue but with the ShapeShifterMsg : 
https://github.com/cellumation/snapshotter/tree/ros1

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,42 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+    }
+    */
+
+#else
+    int wanted_nice_value = *compression_options_.thread_nice_value;
+
+    int cur_nice_value = getpriority(PRIO_PROCESS, 0);
+
+    int new_nice_value = nice(wanted_nice_value - cur_nice_value);
+    if(new_nice_value == -1 || new_nice_value != wanted_nice_value)
```
MichaelOrlov: According to the https://linux.die.net/man/2/nice the **-1** could be a valid return value starting from the glibc 2.24.
It was recommended to check `errno` as a trigger for error return value. It seems doesn't make sense to check for -1 at all. Need to set `0` for `errno` before `nice` call and check `if (errno != 0 && new_nice_value != wanted_nice_value)` right after it to report errors.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,42 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(
+          errno));
+    }
+    */
+
+#else
+    int wanted_nice_value = *compression_options_.thread_nice_value;
+
+    int cur_nice_value = getpriority(PRIO_PROCESS, 0);
```
MichaelOrlov, line 95: Please add a check for `errno` as recommended per https://linux.die.net/man/2/getpriority

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,42 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+    if(!SetThreadPriority(GetCurrentThread(), dwThreadPri))
+    {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(
+          errno));
+    }
+    */
+
+#else
+    int wanted_nice_value = *compression_options_.thread_nice_value;
```
MichaelOrlov: Please use either `auto` or native `int8_t` for variable

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,42 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+    uint8_t nice_value = *compression_options_.thread_nice_value;
+
+    // this must match THREAD_PRIORITY_IDLE, THREAD_PRIORITY_LOWEST...
+    DWORD dwThreadPri = *compression_options_.thread_nice_value;
+
+    if(!SetThreadPriority(GetCurrentThread(), dwThreadPri))
+    {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(
+          errno));
```
MichaelOrlov: The `errno` is not applicable here. need to use [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

MichaelOrlov (changes requested):
@jmachowinski Thank you for your contribution.  
I think we can move forward with this PR and the proposed approach but need to address some minor issues and add missing parts in the dependent rosbag2 packages.

In particular, will need to:
1. update `rosbag2_py` package with the new data structure field.
2. update `rosbag2_transport::RecordOptions` package with the new data structure field.
3. update `ros2bag/verb/record.py` with a new command line option for compression threads priority value.
4. Consider if it will be possible to add some unit tests to check if threads priority sets up as expected.

I will doublecheck yet Windows implementation and let you know if changes are needed.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
MichaelOrlov: @jmachowinski Thanks for the input.  
I understand that in theory if we would store some index to the `topic_name` instead of strings it would save us potentially some memory.  
However, performance measurements results from the https://github.com/cellumation/snapshotter/tree/ros1#performance-evaluation-and-comparision a bit confusing.
It looks like `rosbag2_snapshot` implementation consumes 200 MBytes less than `snapsotter`

<h4 tabindex="-1" dir="auto" style="box-sizing: border-box; margin-top: 24px; margin-bottom: 16px; font-size: 16px; font-weight: var(--base-text-weight-semibold, 600); line-height: 1.25; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Scenario 2</h4><p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Run a gazebo simulation with 5 depth cameras and several small topics. Limit memory usage to 10GB.</p>

  | rosbag_snapshot | snapshotter
-- | -- | --
cpu usage avg % | 56% | 52%
Ram usage | 9.8GB | 10.0GB
Resulting Bag Size | 8.4GB | 9.8GB
Resulting Bag Duration | 9.6s | 10.7s

<h4 tabindex="-1" dir="auto" style="box-sizing: border-box; margin-top: 24px; margin-bottom: 16px; font-size: 16px; font-weight: var(--base-text-weight-semibold, 600); line-height: 1.25; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a id="user-content-scenario-3" class="anchor" aria-hidden="true" tabindex="-1" href="https://github.com/cellumation/snapshotter/tree/ros1#scenario-3" style="box-sizing: border-box; background-color: transparent; color: var(--fgColor-accent, var(--color-accent-fg)); text-decoration: none; float: left; padding-right: 4px; margin-left: -20px; line-height: 1; position: absolute;"><svg class="octicon octicon-link" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path d="m7.775 3.275 1.25-1.25a3.5 3.5 0 1 1 4.95 4.95l-2.5 2.5a3.5 3.5 0 0 1-4.95 0 .751.751 0 0 1 .018-1.042.751.751 0 0 1 1.042-.018 1.998 1.998 0 0 0 2.83 0l2.5-2.5a2.002 2.002 0 0 0-2.83-2.83l-1.25 1.25a.751.751 0 0 1-1.042-.018.751.751 0 0 1-.018-1.042Zm-4.69 9.64a1.998 1.998 0 0 0 2.83 0l1.25-1.25a.751.751 0 0 1 1.042.018.751.751 0 0 1 .018 1.042l-1.25 1.25a3.5 3.5 0 1 1-4.95-4.95l2.5-2.5a3.5 3.5 0 0 1 4.95 0 .751.751 0 0 1-.018 1.042.751.751 0 0 1-1.042.018 1.998 1.998 0 0 0-2.83 0l-2.5 2.5a1.998 1.998 0 0 0 0 2.83Z"></path></svg></a></h4>

<h4 tabindex="-1" dir="auto" style="box-sizing: border-box; margin-top: 24px; margin-bottom: 16px; font-size: 16px; font-weight: var(--base-text-weight-semibold, 600); line-height: 1.25; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Scenario 3</h4><p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Run a gazebo simulation with only small topics. Limit memory usage to 10GB.</p>

  | rosbag_snapshot | snapshotter
-- | -- | --
cpu usage avg % | 10% | 8%
Ram usage | 9.8GB | 10.0GB
Resulting Bag Size | 6.8GB | 9.8GB
Resulting Bag Duration | 619s | 954s

<h4 tabindex="-1" dir="auto" style="box-sizing: border-box; margin-top: 24px; margin-bottom: 16px; font-size: 16px; font-weight: var(--base-text-weight-semibold, 600); line-height: 1.25; color: rgb(31, 35, 40); font-family: -apple-system, BlinkMacSystemFont, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a id="user-content-results" class="anchor" aria-hidden="true" tabindex="-1" href="https://github.com/cellumation/snapshotter/tree/ros1#results" style="box-sizing: border-box; background-color: transparent; color: var(--fgColor-accent, var(--color-accent-fg)); text-decoration: none; float: left; padding-right: 4px; margin-left: -20px; line-height: 1; position: absolute;"><svg class="octicon octicon-link" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path d="m7.775 3.275 1.25-1.25a3.5 3.5 0 1 1 4.95 4.95l-2.5 2.5a3.5 3.5 0 0 1-4.95 0 .751.751 0 0 1 .018-1.042.751.751 0 0 1 1.042-.018 1.998 1.998 0 0 0 2.83 0l2.5-2.5a2.002 2.002 0 0 0-2.83-2.83l-1.25 1.25a.751.751 0 0 1-1.042-.018.751.751 0 0 1-.018-1.042Zm-4.69 9.64a1.998 1.998 0 0 0 2.83 0l1.25-1.25a.751.751 0 0 1 1.042.018.751.751 0 0 1 .018 1.042l-1.25 1.25a3.5 3.5 0 1 1-4.95-4.95l2.5-2.5a3.5 3.5 0 0 1 4.95 0 .751.751 0 0 1-.018 1.042.751.751 0 0 1-1.042.018 1.998 1.998 0 0 0-2.83 0l-2.5 2.5a1.998 1.998 0 0 0 0 2.83Z"></path></svg></a></h4>

jmachowinski: > It looks like rosbag2_snapshot implementation consumes 200 MBytes less than snapsotter

@arneboe can you comment on this, you did the tests back in the day...

My guess would be, it is a bug in rosback_snapshot and both programs were given a parameter to run  with a memory limit of 10 GB.

jmachowinski: @MichaelOrlov This is a test between https://github.com/ros/rosbag_snapshot and our snapshotter, not rosbag2_snapshot. Note, the results should be still comparable, to give an idea of the overhead created by the topic name.

arneboe: This is a test that we did back in the day for ros1.
In this test both snapshotters where limited to 10gb of ram.
The interesting thing here is the difference in the resulting bag size. Of the 10gb of ram used in the snapshotter, 9.8gb end up in the bag file while for rosbag_snapshot only 6.8gb end up in the bag file. This shows the memory overhead that occurs because rosbag_snapshot stores a copy of the topic metadata for each message.

I do not know why rosbag_snapshot limits itself to 9.8gb of ram when a 10gb limit is given.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: I went for
`if ((new_nice_value == -1 && errno != 0) || new_nice_value != wanted_nice_value)`
As this would additionally catch other errors, were new_nice_value != wanted_nice_value, even though errno would not be set.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: If I would do that, the error output down below would be broken, as it would interpret wanted_nice_value as char, or I would need to use static cast all over the place.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: done

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: done

jmachowinski:
> In particular, will need to:
> 
>     1. update `rosbag2_py` package with the new data structure field.
> 
>     2. update `rosbag2_transport::RecordOptions` package with the new data structure field.
> 
>     3. update `ros2bag/verb/record.py` with a new command line option for compression threads priority value.
> 
>     4. Consider if it will be possible to add some unit tests to check if threads priority sets up as expected.

The python stuff is nothing, were a commit from ME would meet any quality standards... @kohrt can you have a look at this ?

rosbag2_compression/include/rosbag2_compression/compression_options.hpp
```diff
@@ -61,6 +62,9 @@ struct CompressionOptions
   CompressionMode compression_mode;
   uint64_t compression_queue_size;
   uint64_t compression_threads;
+  /// if set, the compression thread will try to set
```
fujitatomoya:
```suggestion
  /// if set, the compression thread(s) will try to set
```

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
 
 void SequentialCompressionWriter::compression_thread_fn()
 {
+  if (compression_options_.thread_nice_value) {
```
fujitatomoya: Do we really need to do this in here? I would suggest this setting (and setting back) in the `SequentialCompressionWriter::setup_compressor_threads` instead? so that each thread does not need to issue the system calls and printing the same warnings?

note: compression threads can be multiple.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+
+#ifdef _WIN32
+    ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+      "Lowering of process priority is not implemented for windows");
```
fujitatomoya: does this depends on the application for lower or higher?

```suggestion
      "Changing process priority is not implemented for windows");
```

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
 
 void SequentialCompressionWriter::compression_thread_fn()
 {
+  if (compression_options_.thread_nice_value) {
+
+#ifdef _WIN32
```
fujitatomoya, line 73: although i understand the use case and quick fix availability, i would suggest this thread priority and policy control supported with `rcutils` and `rcpputils`, so that dependent packages are agnostic from this platform dependencies.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+    {
+      ROSBAG2_COMPRESSION_LOG_WARN_STREAM(
+        "Could not set nice value of compression thread to " << nice_value << " : " << std::strerror(GetLastError()));
+    }
+    */
+
+#else
+    int wanted_nice_value = *compression_options_.thread_nice_value;
+
+    errno = 0;
```
fujitatomoya, line 94: (comment) I believe `errno` should be multi-thread safe. just checking if i got this wrong. (this is only because current PR accesses `errno` with multi-thread possibly.)

fujitatomoya (changes requested):
besides all comments, and 

```bash
Manual page getpriority(2)
...
BUGS
       According  to POSIX, the nice value is a per-process setting.  However, under the current Linux/NPTL implementation of
       POSIX threads, the nice value is a per-thread attribute: different threads in the same process can have different nice
       values.   Portable  applications should avoid relying on the Linux behavior, which may be made standards conformant in
       the future.

```

IMO, i am not inclined to take this fix.

related REP(ROS Enhancement Proposal): https://github.com/ros-infrastructure/rep/pull/385

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: it must be done here, if we would do this in setup_compressor_threads we would modify the nice value of the main thread, which is not the intention.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: This is out of the scope of this merge request. If functions will be available in rclutils etc, we can still change the code here later on.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: man 3 errno :
`errno is thread-local; setting it in one thread does not affect its value in any other thread.`

jmachowinski:
>     2. update `rosbag2_transport::RecordOptions` package with the new data structure field.
Done
>     4. Consider if it will be possible to add some unit tests to check if threads priority sets up as expected.
Added a test case

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
fujitatomoya: i mean, we can change it back on main thread, so that we can avoid all these system calls in those threads?

jmachowinski: I don't know, if the change back in the main thread would fail, we would be in a pretty bad state...

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
MichaelOrlov: @jmachowinski According to the https://linux.die.net/man/2/nice
> Notes
SUSv2 and POSIX.1-2001 specify that nice() should return the new nice value. However, the Linux syscall and the nice() library function provided in older versions of (g)libc (earlier than glibc 2.2.4) return 0 on success. The new nice value can be found using [getpriority](https://linux.die.net/man/2/getpriority)(2).

Adding `|| new_nice_value != wanted_nice_value` to the condition will not get the advantage of catching any other errors.  

It will affect the portability of such code. One could fork `rosbag2` repo and try to run it with QNX or some other RTOS  and may get in trouble due to this extra check.

I would strongly recommend to not use this check for error handling.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
MichaelOrlov: @fujitatomoya Changing back priority might need `sudo` privileges.  
It woks in such a way that you can lower priority for the thread without `sudo` but if you want to elevate priority need to have a `sudo` permissions.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
MichaelOrlov: I think we can add support for windows. I can contribute.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
fujitatomoya: @MichaelOrlov ah right, that is not what we want to meet here.

MichaelOrlov:
@jmachowinski Could you please rename `compression_threads_nice_value` parameter to something more generic like 
`compression_threads_priority` and update descriptions in the header files?  
Since on Windows we don't have such entities as `nice value` we do really have a thread priority.
I  remember I've added a comment about it but it seems my comment disappeared in Github after you made a git push with --force parameter.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -62,6 +69,47 @@ SequentialCompressionWriter::~SequentialCompressionWriter()
...
+    int cur_nice_value = getpriority(PRIO_PROCESS, 0);
+    if (cur_nice_value != -1 && errno == 0) {
```
MichaelOrlov: @jmachowinski I need to double check yet. But it seems here mistake in the if condition.
The point is that `cur_nice_value == -1` is a valid return value from the `getpriority(..)`
It seems the correct logic shall be `if (!(new_nice_value == -1 && errno != 0))` or its equivalent `if (new_nice_value != -1 || errno != 0)`

jmachowinski: good catch, this is indeed wrong

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -384,6 +401,100 @@ TEST_P(SequentialCompressionWriterTest, writer_writes_with_compression_queue_siz
...
+  for (size_t i = 0; i < kNumMessagesToWrite; i++) {
+    writer_->write(message);
+  }
+  writer_.reset();    // reset will call writer destructor
+
+  EXPECT_EQ(detected_nice_value, *compression_options.thread_nice_value);
+
+  EXPECT_EQ(fake_storage_size_, kNumMessagesToWrite);
+}
+
```
MichaelOrlov: @jmachowinski This approach for testing is not going to work.  
I would propose creating a regular **non-fake** `sequential_compression_writer` and capture std::err output to verify logged error messages.
Something like 
```
    auto writer = ReaderWriterFactory::make_writer(record_options);
    testing::internal::CaptureStderr();
    writer->open(zero_cache_storage_options);
    std::string test_output = testing::internal::GetCapturedStderr();
    EXPECT_TRUE(
      test_output.find(
        "Could not set priority for compression thread to'") == std::string::npos);
```
and make two test cases with valid thread priority and invalid thread priority values.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -384,6 +401,100 @@ TEST_P(SequentialCompressionWriterTest, writer_writes_with_compression_queue_siz
...
+    virtual std::string compress_uri(const std::string & uri)
+    {
+      return uri;
+    }
+
+    virtual void compress_serialized_bag_message(
+      const rosbag2_storage::SerializedBagMessage * bag_message,
+      rosbag2_storage::SerializedBagMessage * compressed_message)
+    {
+#ifdef _WIN32
```
MichaelOrlov: @jmachowinski The test passes on Linux only because there is a mistake with `ifdef _WIN32` statement.  
It shall be **ifndef _WIN32**

MichaelOrlov (changes requested):
@jmachowinski The selected approach for tests is not going to work. The test shall be rewritten.
See my comments in code.
I would propose creating a regular non-fake sequential_compression_writer and capture std::err output to verify logged error messages.
Something like

    auto writer = ReaderWriterFactory::make_writer(record_options);
    testing::internal::CaptureStderr();
    writer->open(zero_cache_storage_options);
    std::string test_output = testing::internal::GetCapturedStderr();
    EXPECT_TRUE(
      test_output.find(
        "Could not set priority for compression thread to'") == std::string::npos);
and make two test cases with valid thread priority and invalid thread priority values.

Also please rename `compression_threads_nice_value` and `thread_nice_value` parameters to the `compression_threads_priority` and `thread_priority` respectively.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
jmachowinski: @MichaelOrlov Why do you think this test is not working ? Its actually working just fine, and failing, as the ifdef mentioned below is switched.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
jmachowinski: The test is actually failing. I added the ifdef **after** running the test locally, which was stupid....

jmachowinski:
> @jmachowinski Could you please rename `compression_threads_nice_value` parameter to something more generic like `compression_threads_priority` and update descriptions in the header files? Since on Windows we don't have such entities as `nice value` we do really have a thread priority. I remember I've added a comment about it but it seems my comment disappeared in Github after you made a git push with --force parameter.

A priority is something special in the linux world, therefore I would not name it that.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
MichaelOrlov: @jmachowinski I said that 
> The selected approach for tests is not going to work.

Because currently test is trying to test against the fake `TestCompressor` instead of the real `SequentialCompressionWriter`. That doesn't make sense.

The unit tests shall call `SequentialCompressionWriter` where the changes are - directly or indirectly.

jmachowinski: @MichaelOrlov have a look at https://github.com/ros2/rosbag2/blob/64e29724cbea3058bfa4c1688867dc9f2f85bdbd/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp#L111
This function is used inside the test, and initializes a SequentialCompressionWriter. I only use the factory, to provide a custom compressor, that is used within the compression thread of the SequentialCompressionWriter. Therefore it is the perfect place to test for the correct priority.

MichaelOrlov: @jmachowinski Sorry, I overlooked it. The test was a bit messy. 
It does really make sense but needs a cleanup.
I've rewritten it. Please consider to merge or cherry-pick my changes from PR [Redesign test for the case when compression writer sets threads priority #1](https://github.com/cellumation/rosbag2/pull/1)

MichaelOrlov:
@jmachowinski As regards
> A priority is something special in the linux world, therefore I would not name it that.

The `thread priority` is a generic term and will be applicable to the cross-platform implementation.
The fact that in the current implementation, the thread priority on Unix/Linux platforms will be settling up via `nice value` is  an implementation details and would be considered as temporary workaround until [[REP-2017] Thread attributes configuration support in rcl #385](https://github.com/ros-infrastructure/rep/pull/385) will not be fully implemented.
Eventually, we will replace part of the code that is setting/getting `nice value` to the API from the `[REP-2017]`. But what is important is that the variables and parameter's names shall be agnostic to this temporary workaround.

jmachowinski:
@MichaelOrlov 
You latest changes trigger this warning : 
Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occurred with plugin factory for class FakeCompressor. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.

I suggest we move the check back into a class inside of the test function.

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: fixed

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
jmachowinski: removed

jmachowinski:
> @jmachowinski Could you please rename `compression_threads_nice_value` parameter to something more generic like `compression_threads_priority` 

Done

MichaelOrlov:
> I suggest we move the check back into a class inside of the test function.

Hold on let me double-check. We didn't change the way how the library was linked.
This warning might have already existed before.

jmachowinski:
> > I suggest we move the check back into a class inside of the test function.
> 
> Hold on let me double-check. We didn't change the way how the library was linked. This warning might have already existed before.

You added  test/rosbag2_compression/fake_compressor.cpp to test_sequential_compression_writer this, triggers the warning

MichaelOrlov:
@jmachowinski You can disregard this warning.  
I made some investigation and found out that the case about what this warning signaling is safe to be used in this particular scenario.  
I would prefer to have a such runtime warning in one test rather than duplicated class for `fake_compressor`.

If this warning is really bothering you and you want to get rid of it you can fix it by specifying `"zstd"` compression format instead of the `DefaultTestCompressor` in the `open_succeeds_on_supported_compression_format` test.
https://github.com/ros2/rosbag2/blob/96c11b073c9bf56fb35586205b941e14f8f15b6a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp#L195-L199
But in this case, will need to add `rosbag2_compression_zstd` as a runtime dependenises for the tests in the `package.xml` file

MichaelOrlov (commented):
@jmachowinski Thanks for the renames and fixes.
We do have currently `uint8_t` for `thread_priority` and it is totally aligned with the range of the nice value on Linux/POSIX since it varies from -19 to 20. 
Although it doesn't align with the possible thread's priority values on Windows. 
According to the https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreadpriority one of the biggest values could be settled up is
```
THREAD_MODE_BACKGROUND_END 0x00020000
```
In decimal, this is **131 072**. On Windows, the type is **int** but this is not good to use `int` type in data structures since not strictly defined length of the integer type. i.e. it could vary from different platforms and compilators.
I think we need to change the type for the thread's priority at least to the `int32_t` or even better  `int64_t` to accommodate all possible values.
We can use a static cast to int when passing it to set priority functions.

MichaelOrlov:
@jmachowinski Could you please fix DCO and cpplint warnings?

jmachowinski:
@MichaelOrlov done

MichaelOrlov (commented):
@jmachowinski Thanks for the fixes. LGTM.
However, will need to uncomment Windows related part and test it. I can do it on the next week.

MichaelOrlov:
@jmachowinski Meanwhile could you please fix the initialization for the newly added field for thread_priority in the `rosbag2_peformance_benchmarking` package?
See compiler errors:
```
023-10-06T17:40:28.2668526Z [ 30%] Building CXX object CMakeFiles/writer_benchmark.dir/src/writer_benchmark.cpp.o
2023-10-06T17:40:28.2669889Z /__w/rosbag2/rosbag2/ros_ws/src/zzkbmflvv2n/rosbag2/rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp: In member function ‘void WriterBenchmark::create_writer()’:
2023-10-06T17:40:28.2755587Z /__w/rosbag2/rosbag2/ros_ws/src/zzkbmflvv2n/rosbag2/rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp:181:74: error: missing initializer for member ‘rosbag2_compression::CompressionOptions::thread_priority’ [-Werror=missing-field-initializers]
2023-10-06T17:40:28.2756610Z   181 |       bag_config_.compression_queue_size, bag_config_.compression_threads};
```

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-10-12/34178/1

jmachowinski:
> @jmachowinski Meanwhile could you please fix the initialization for the newly added field for thread_priority in the `rosbag2_peformance_benchmarking` package? See compiler errors:
> 
> ```
> 023-10-06T17:40:28.2668526Z [ 30%] Building CXX object CMakeFiles/writer_benchmark.dir/src/writer_benchmark.cpp.o
> 2023-10-06T17:40:28.2669889Z /__w/rosbag2/rosbag2/ros_ws/src/zzkbmflvv2n/rosbag2/rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp: In member function ‘void WriterBenchmark::create_writer()’:
> 2023-10-06T17:40:28.2755587Z /__w/rosbag2/rosbag2/ros_ws/src/zzkbmflvv2n/rosbag2/rosbag2_performance/rosbag2_performance_benchmarking/src/writer_benchmark.cpp:181:74: error: missing initializer for member ‘rosbag2_compression::CompressionOptions::thread_priority’ [-Werror=missing-field-initializers]
> 2023-10-06T17:40:28.2756610Z   181 |       bag_config_.compression_queue_size, bag_config_.compression_threads};
> ```

Fixed, sorry for the delay, I was in vacation.
Whats the state one the windows part ?

MichaelOrlov:
@jmachowinski Thanks for the updates.
I am sorry, I also was busy with ROSCon2023 and then with other tasks. 
Going to proceed with this PR this week.

MichaelOrlov:
@jmachowinski I've conditioned and tested Windows implementation. See my changes in the PR https://github.com/cellumation/rosbag2/pull/2 .

Feel free to review and merge to your branch.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-11-16/34757/1

MichaelOrlov:
@jmachowinski Friendly ping for review https://github.com/cellumation/rosbag2/pull/2

jmachowinski:
Is there anything left to do ? I am still not that happy with the rewrite of the one test, as it puts out this big fat warning. As a new developer to this component, this warning would freak me out, if I would see it running the test.

MichaelOrlov:
@jmachowinski As regards:
> I am still not that happy with the rewrite of the one test, as it puts out this big fat warning. As a new developer to this component, this warning would freak me out, if I would see it running the test.

Could you please remind me about what warning you have concern or in what place it is?

MichaelOrlov (approved):
@jmachowinski Thanks for your contribution and initiatives.  
Now it looks good to me. However, will need to rebase on top of the latest rolling and resolve merge conflicts. And as usual make sure that the CI is green.

jmachowinski:
> @jmachowinski As regards:
> 
> > I am still not that happy with the rewrite of the one test, as it puts out this big fat warning. As a new developer to this component, this warning would freak me out, if I would see it running the test.
> 
> Could you please remind me about what warning you have concern or in what place it is?

[ RUN      ] SequentialCompressionWriterTest.open_succeeds_on_supported_compression_format
Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occurred with plugin factory for class FakeCompressor. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.

This warning was introduced by the refactoring of the test in https://github.com/ros2/rosbag2/pull/1457/commits/c8cdcba3e149834e7925867affdccc46a8b8fe04

jmachowinski:
Fixed the warning, and rebased everything.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/d7bfc580442557ee1e29fd6d2fa0c92f/raw/a16b9fbcb901cc2c8db6f45570ee8f15db61a717/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_transport rosbag2_compression rosbag2_transport rosbag2_performance rosbag2_tests
TEST args:  --packages-above rosbag2_transport rosbag2_compression rosbag2_transport rosbag2_performance rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13013
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19933/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20640/)

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/299ef38317bcc7b314fd3569edd4f864/raw/a16b9fbcb901cc2c8db6f45570ee8f15db61a717/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_transport rosbag2_compression rosbag2_transport rosbag2_performance_benchmarking rosbag2_tests
TEST args:  --packages-above rosbag2_transport rosbag2_compression rosbag2_transport rosbag2_performance_benchmarking rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13016
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19936/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20643/)

MichaelOrlov:
The Windows CI job fails with error messages
```
12:11:02   Building Custom Rule C:/ci/ws/src/cellumation/rosbag2/rosbag2_compression/CMakeLists.txt

12:11:02   test_sequential_compression_writer.cpp

12:11:02 LINK : fatal error LNK1181: cannot open input file 'Release\fake_plugin.lib' [C:\ci\ws\build\rosbag2_compression\test_sequential_compression_writer.vcxproj]
```
@jmachowinski It seems it relates to your latest commit with an attempt to fix a warning.

jmachowinski:
I added visibility attributes to the classes in fakeplugin, this should force windows to create a .lib file. Can you rerun the pipeline ?

MichaelOrlov:
The linter is angry a bit:
```
Error Message
Diff with 6 lines
Stacktrace
--- test/rosbag2_compression/fake_compressor.hpp
+++ test/rosbag2_compression/fake_compressor.hpp.uncrustify
@@ -23 +23,2 @@
-class ROSBAG2_COMPRESSION_PUBLIC FakeCompressor : public rosbag2_compression::BaseCompressorInterface
+class ROSBAG2_COMPRESSION_PUBLIC FakeCompressor : public rosbag2_compression::
+  BaseCompressorInterface
```

jmachowinski:
>The linter is angry a bit:

I calmed it down ;-)

MichaelOrlov:
Re-run Windows CI after an attempt to fix linker errors
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20673/)

MichaelOrlov:
@jmachowinski I am sorry, but we just merged another big PR and this branch has to be rebased and conflicts shall be resolved.
Could you please rebase your branch on top of the latest rolling ?

jmachowinski:
rebased

MichaelOrlov:
@jmachowinski since we added recently a feature for composable recorder. This is when `recorder_options` values getting from a node params. We need to add `compression_thread_priority`  to the relevant parser.

Could you please add
```
record_options.compression_threads_priority = param_utils::declare_integer_node_params<int32_t>(
    node, "compression_threads_priority", std::numeric_limits<int32_t>::min(),
    std::numeric_limits<int32_t>::max(), record_options.compression_threads_priority);
```
to the `RecordOptions get_record_options_from_node_params(rclcpp::Node & node)` and modify the relevant test to make sure that the value parsed correctly?

MichaelOrlov:
@jmachowinski I've made the requested change in the node params parser and relevant test for the `compression_thread_priority` by myself.   
Please review and merge or cherry-pick changes to your branch if you don't have any concerns.

jmachowinski:
Looks good to me, another try to merge it ?

MichaelOrlov:
Re-run CI after an attempt to fix Windows build errors and rebase.
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13044
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14441/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20680/)

MichaelOrlov:
@jmachowinski Unfortunately Windows build still fails with the following error messages
```
:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp(26,1): warning C4273: 'FakeCompressor::FakeCompressor': inconsistent dll linkage [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.hpp(29,3): message : see previous definition of '{ctor}' (compiling source file C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp(38,1): warning C4273: 'FakeCompressor::compress_uri': inconsistent dll linkage [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.hpp(31,15): message : see previous definition of 'compress_uri' (compiling source file C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp(44,44): warning C4273: 'FakeCompressor::compress_serialized_bag_message': inconsistent dll linkage [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.hpp(33,8): message : see previous definition of 'compress_serialized_bag_message' (compiling source file C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp(47,1): warning C4273: 'FakeCompressor::get_compression_identifier': inconsistent dll linkage [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.hpp(37,15): message : see previous definition of 'get_compression_identifier' (compiling source file C:\ci\ws\src\cellumation\rosbag2\rosbag2_compression\test\rosbag2_compression\fake_compressor.cpp) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

     Creating library C:/ci/ws/build/rosbag2_compression/Release/fake_plugin.lib and object C:/ci/ws/build/rosbag2_compression/Release/fake_plugin.exp

fake_compressor.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl FakeCompressor::FakeCompressor(void)" (__imp_??0FakeCompressor@@QEAA@XZ) referenced in function "public: virtual class rosbag2_compression::BaseCompressorInterface * __cdecl class_loader::impl::MetaObject<class FakeCompressor,class rosbag2_compression::BaseCompressorInterface>::create(void)const " (?create@?$MetaObject@VFakeCompressor@@VBaseCompressorInterface@rosbag2_compression@@@impl@class_loader@@UEBAPEAVBaseCompressorInterface@rosbag2_compression@@XZ) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

fake_compressor.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl FakeCompressor::~FakeCompressor(void)" (__imp_??1FakeCompressor@@UEAA@XZ) referenced in function "public: virtual void * __cdecl FakeCompressor::`scalar deleting destructor'(unsigned int)" (??_GFakeCompressor@@UEAAPEAXI@Z) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

fake_compressor.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) const FakeCompressor::`vftable'" (__imp_??_7FakeCompressor@@6B@) referenced in function "public: __cdecl FakeCompressor::FakeCompressor(int &)" (??0FakeCompressor@@QEAA@AEAH@Z) [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

C:\ci\ws\build\rosbag2_compression\Release\fake_plugin.dll : fatal error LNK1120: 3 unresolved externals [C:\ci\ws\build\rosbag2_compression\fake_plugin.vcxproj]

```

jmachowinski:
I just set up a windows machine....

jmachowinski:
Compiles and tests now fine under windows.

MichaelOrlov:
Re-run Windows CI job after the attempt to fix build errors.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20688/)

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-for-2023-12-14/35153/1

ros-discourse:

Issue #1459, opened 2023-09-06
Type stubs for rosbag2_py


MichaelFishmanBDAII:
## Description

I'd like type stubs for `rosbag2_py`, so that I can get autocomplete/static analysis.

## Completion Criteria

** What needs to be true before we can call this "Done"? Bullet lists are appropriate. **

- The maximal version: All classes/methods/functions exposed by `rosbag2_py` have full type annotations. A static analysis tool like mypy/pylance can tell parse these type annotations.

## Implementation Notes / Suggestions
Stubs can be generated with `pybind11-stubgen`:

```sh
pip install pip install pybind11-stubgen

pybind11-stubgen rosbag2_py
```
After doing this, making a `setup.py`, and installing the newly created `rosbag2_py-stubs` library, we get some type information, though the function argument and return types are incomplete:

![image](https://github.com/ros2/rosbag2/assets/142259135/fe0e943f-bb25-41e9-9f22-8b1fd13493cd)

The output from the `pybind11-stubgen` command is

```
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialCompressionReader : Invalid expression 'rosbag2_storage::TopicMetadata'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialCompressionReader : Invalid expression 'rosbag2_storage::StorageOptions'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialCompressionReader : Invalid expression 'rosbag2_cpp::ConverterOptions'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialCompressionReader : Invalid expression 'rosbag2_storage::StorageFilter'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialReader : Invalid expression 'rosbag2_storage::TopicMetadata'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialReader : Invalid expression 'rosbag2_storage::StorageOptions'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialReader : Invalid expression 'rosbag2_cpp::ConverterOptions'
pybind11_stubgen - [  ERROR] In rosbag2_py._reader.SequentialReader : Invalid expression 'rosbag2_storage::StorageFilter'
pybind11_stubgen - [WARNING] Raw C++ types/values were found in signatures extracted from docstrings.
[...]

PR #1463, opened 2023-09-08
[wip] More precise type hints for `rosbag2_py`

MichaelFishmanBDAII:
**VERY WIP**

Meant to address #1459

For now, the only improved type hint is for `SequentialReader.read_next` and the type stubs need to be generated manually.

This branch has many other changes I made to make development easier for me. I'm guessing `rosbag2` doesn't actually want them, in which case they'd be removed before this PR is un-drafted.
These changes are:

1. `.devcontainer` and `Dockerfile`. Currently these mention my username. If the devcontainer is kept, I'd remove that username-specific bit. These were made following [this](https://docs.ros.org/en/humble/Installation.html) guide for getting ROS2 in a devcontainer.
2. The tests for `rosbag2_py` now use `unittest.TestCase`, so that I can run them from the VSCode test panel, without running the entire `colcon test`.
3. I removed the `.github` workflows (you _definitely_ don't want this change)


MichaelFishmanBDAII:
Closing this because it's old

PR #1474, opened 2023-09-30
Remove rcpputils::fs dependencies from rosbag2_storages

yoneken:
Because the package will be no longer available. https://github.com/ros2/rcpputils/issues/164

This operation was carried out with the following simple rules.
1. Replace rcpputils::fs::path class to std::filesystem::path class.
2. Replace rcpputils::fs::path::exists function to std::filesystem::exists function.
3. Replace rcpputils::fs::path::string function to std::filesystem::path::generic_string function.

I will create several PRs for the same operation for other rosbag2 packages.

PR #1478, opened 2023-10-06
Remove rcpputils::fs dependency from rosbag2_cpp

yoneken:
Sequential PRs from https://github.com/ros2/rosbag2/pull/1474

This PR and #1474 are independent of each other.
So, whenever they can be merged, and do not need to care about the order.


MichaelOrlov:
- Closing as done in the #1576

Issue #1485, opened 2023-10-20
CPU overhead when discovery is used


doisyg:
## Description
Related to https://github.com/ros2/rosbag2/issues/1170
When recording a list of topics that are not all published, on a system with multiple topics, actions and services (typically a mobile robot with the nav2 stack), and with (default) automatic discovery, there is an important CPU usage from the bag recorder process.Typically 20% CPU usage on a very powerful machine (i7 13th gen)
We don't observe this CPU overhead with the same setup but `--no-discovery` added.

### Expected Behavior
Reasonable CPU usage

### Actual Behavior
Non reasonable CPU usage

## System (please complete the following information)
 - OS: Ubuntu 22.04
 - ROS 2 Distro: Iron latest bin 
 - Install Method: apt
 - Version: 0.22.3-1jammy.20230911.053809

doisyg:
Update: no issue when using rosbag built from source and latest Iron commit: https://github.com/ros2/rosbag2/commit/7adc0f9da7bccb25d7eef56be69b393a9c3220d9
However when using source v0.22.3, the issue is there. So it feels that (good timing), the latest commit fix this issue.

Can you trigger a release ?

fujitatomoya:
@doisyg i am not sure which commit actually fixes your problem here. but for releasing new packages would take some time.

CC: @Yadunund

doisyg:
> i am not sure which commit actually fixes your problem here. but for releasing new packages would take some time.

I am not sure myself but I know that I have the problem with v0.22.3 (even from source), and I don't with the latest version of the Iron branch (so probably these changes: https://github.com/ros2/rosbag2/commit/51a83f4421ff335c2237e6214e9bb7e6ae206a92)

What's blocking for releasing the latest Iron branch in the next sync ?

MichaelOrlov:
@doisyg Yes, https://github.com/ros2/rosbag2/commit/51a83f4421ff335c2237e6214e9bb7e6ae206a92 might have improved the performance burden during discovery since we are deferring more heavy check to the end of the list when filtering out topics.

We will spawn a new release for iron, however as others mentioned it could take some time.
Meanwhile to workaround the issue you can explicitly specify a list of topics for the rosabg2 recorder. In this case, as soon as all topics are subscribed we are exiting from the rosbag2 discovery thread and there will not be extra burden on the performance.

MichaelOrlov:
[...]

PR #1486, opened 2023-10-24
Add cache for topic filter to avoid performance burden on discovery

### LINKED ISSUE #2200: High CPU overhead when discovery is used
## Description
It is known for a long time that our Rosbag2 recorder discovery mechanism is suboptimal and has a significant CPU consumption overhead when up and running.
It was multiple attempts to address this performance overhead, but most of them were either cosmetic or abandoned and not finished. For instance, a https://github.com/ros2/rosbag2/pull/1486.
The reason why we have a  significant CPU overhead in the discovery is that we continiously polling the graph and try to examine available topics with the inner topics filter if there are new topics to subscribe to.  
It would be more efficient as we would wait on a GraphListener change event instead of unconditional polling and use a cache for the topics_filter class, as it was proposed in the https://github.com/ros2/rosbag2/pull/1486 to avoid heavy routines inside for the already evaluated topics.

## Related Issues
- #1485 

## Completion Criteria

* [ ] The CPU overhead of the Rosbag2 recorder's discovery thread shall be marginally improved
* [ ] Finish https://github.com/ros2/rosbag2/pull/1486
* [ ] Implement wait on the GraphListener change event instead of the continuous unconditional polling with interval.

1. Will need to use wait on a GraphListener change event instead of unconditional polling for the `all_topics_and_types = node->get_topic_names_and_types();`
2. Would be good to use a cache for the TopicsFilter class as it was outlined in the https://github.com/ros2/rosbag2/pull/1486

- For the https://github.com/ros2/rosbag2/pull/1486 will likely need to add a few unit tests for the caching mechanism
- For using wait on the GrapListener event, no need for extra test coverage. Since this will be covered by the existing tests.

MichaelOrlov:
### Description
This PR adds cache for the `TopicFilter` class to avoid a performance burden on discovery.
- Closes #2200
- Relates #1485
Rationale:
When rosbag2 discovery is enabled, we are periodically doing a lot of checks in the TopicFilter::take_topic(..), which could be avoided by using a cache.

Yes. Sort of. The user may notice less CPU load during recording when the discovery thread is running.

Yes, I used GitHub Copilot, GPT-4.1 and Claude Code 3.7 Thinking model to assist with writing tests and Doxygen comments.

### Additional Information
This PR **can not** be backported due to the breaking ABI changes in the `TopicFilter.hpp` file


MichaelOrlov:
One test is unexpectedly failing. Movin to the WIP until further investigation.

rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
```diff
@@ -94,21 +94,31 @@ TopicFilter::TopicFilter(
...
+    if (take_topics_cache_it != take_topics_cache_.end()) {
+      if (take_topics_cache_it->second) {
+        filtered_topics.insert(std::make_pair(topic_name, topic_types[0]));
+      }
+    } else {
+      if (take_topic(topic_name, topic_types)) {
+        filtered_topics.insert(std::make_pair(topic_name, topic_types[0]));
+        take_topics_cache_[topic_name] = true;
+      } else {
+        take_topics_cache_[topic_name] = false;
```
fujitatomoya: 1st discovery, lets say if the topic is bound with unknown type, and the register here with `false`. after 2nd discovery, even if the topic is rebound to the appropriate type, it cannot be taken since it does not reconsider the topic type every time. how about using `already_warned_unknown_types_` to check if we already know that is unknown type or not to deny the topic instead?

MichaelOrlov: @fujitatomoya I got back to this PR.
I see your point; it looks like a valid case. I am going to handle it by adding one more flag to the `take_topics_cache_` map and checking findings in the `already_warned_unknown_types_` as you proposed.

MichaelOrlov: Ahh, it doesn't work this way. 
Among the cases described above, i.e.
> 1st discovery, lets say if the topic is bound with unknown type, and the register here with false. after 2nd discovery, even if the topic is rebound to the appropriate type, it cannot be taken since it does not reconsider the topic type every time.

There could be similar cases when using `record_options_.include_unpublished_topics` and `record_options_.ignore_leaf_topics`. We will need to check the up-to-date status from the node graph regardless of the cache.

I have to revert everything that I have done so far and reimplement the caching mechanism by using a separate caches for the topics selected by lists or regex, and for the known topics.

Please review everything one more time. I hope it is a final and clean implementation.

MichaelOrlov:
Pulls: ros2/rosbag2#1486
Gist: https://gist.githubusercontent.com/MichaelOrlov/220210a59195b760fd6da633d8f72fde/raw/0261caf7227d45cb2210dc9268b67cef9b406edf/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17259
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19467/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4913/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25694/)

rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
```diff
@@ -287,6 +288,32 @@ bool TopicFilter::take_topic(
...
+  }
+
+  const std::string & topic_type = topic_types[0];
+
+  bool topic_selected = false;
+  // Check cache first
+  auto lists_or_regex_cache_it =
+    topic_selected_by_lists_or_regex_cache_.find(topic_name + topic_type);
+  if (lists_or_regex_cache_it != topic_selected_by_lists_or_regex_cache_.end()) {
+    topic_selected = lists_or_regex_cache_it->second;
```
christophebedard, line 308: What happens if the topic initially has a type that is excluded, but later that publisher goes away and a different publisher starts publishing the same topic with a type that's not excluded? It would just set `false` in the cache initially and never re-check. Just trying to avoid some sneaky change in behaviour.

MichaelOrlov: Good question. That case is covered because I am caching by `topic_name + topic_type` as a key in the `unordered_map`.  
In case if the same topic will be discovered with another type, it will not be found in the cache as a match due to the `topic_name + topic_type` key i.e. the `topic_type` will be different.

christophebedard: Ah! I missed that. All good then

christophebedard (approved):

PR #1503, opened 2023-11-26
Make `rosbag2_transport::Player::play()` run in a separate thread

### LINKED ISSUE #1500: play service call doesn't return until play finished
## Description
When using the service interface to pull request https://github.com/ros2/rosbag2/pull/1419 (I'm not sure how to set up the same situation for `ros2 bag play`) with the command:

    $ rosservice call /bag_player/play rosbag2_interfaces/srv/Play "{start_offset: {sec: 0, nanosec: 0}, playback_duration: {sec: 0, nanosec: 0}, playback_until_timestamp: {sec: 0, nanosec: 0}}" requester: making request: rosbag2_interfaces.srv.Play_Request(start_offset=builtin_interfaces.msg.Time(sec=0, nanosec=0), playback_duration=builtin_interfaces.msg.Duration(sec=0, nanosec=0), playback_until_timestamp=builtin_interfaces.msg.Time(sec=0, nanosec=0))

### Expected Behavior
According to [this page](https://docs.ros.org/en/iron/How-To-Guides/Topics-Services-Actions.html#services) I would expect it to return quickly after starting the playback:

## Services

* Should be used for remote procedure calls that terminate quickly, e.g. for querying the state of a node or doing a quick calculation such as IK. They should never be used for longer running processes

### Actual Behavior
The service call does not return until play completes, and if looping is on, then it never returns until CTRL+C.

Ubuntu 22.04 + Rolling

roncapat:
- Closes #1500 

This is intended to solve the issue reported in #1500.
CC: @MichaelOrlov @KenYN

This can be also beneficial introductory work for #1419. As it has been noticed, playback shall start automatically when the `Player` component is loaded, but this can't be done via syncronous call by the component constructor.

Current status:
- separate thread implemented
- all tests preliminarly adapted

Anyway, I believe that a round of refactor can be done here. For example, direclty returning an `std::future` from `wait_for_playback_to_end()`. Opinions?

@KenYN may I ask you also to do a test with this branch to confirm that the `play` service now returns immediately?

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -155,7 +155,7 @@ public:
         exec.spin();
       });
     player->play();
-    player->wait_for_playback_to_end();
+    player->wait_for_playback_to_finish();
 
     exec.cancel();
     spin_thread.join();
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -155,6 +155,7 @@ public:
         exec.spin();
       });
     player->play();
+    player->wait_for_playback_to_end();
 
     exec.cancel();
     spin_thread.join();
@@ -175,15 +176,11 @@ public:
       [&exec]() {
         exec.spin();
       });
-    auto play_thread = std::thread(
-      [&player]() {
-        player->play();
-      });
+    player->play();
     player->burst(num_messages);
 
     exec.cancel();
     spin_thread.join();
-    play_thread.join();
   }
 };
 
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -117,9 +117,13 @@ public:
   ROSBAG2_TRANSPORT_PUBLIC
   bool play();
 
-  /// \brief Blocks and waits on the condition variable until the play thread finishes
+  /// \brief Waits on the condition variable until the play thread finishes.
+  /// @param timeout Maximum time in the fraction of seconds to wait for player to finish.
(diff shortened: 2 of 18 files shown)


KenYN:
I did `git checkout patch-2 && git merge roncagliolo/async-play` and it appears to be operating correctly, although by no means is this a thorough test, but thank you very much for this.

MichaelOrlov (changes requested):
@roncapat See my suggestions for review fixes in the https://github.com/roncapat/rosbag2/pull/6 PR

roncapat:
@MichaelOrlov review branch merged :)

MichaelOrlov (approved):
LGTM

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/aa32bdbe242c8c06daca683655e76bb4/raw/4be47e99df1fdad56dd18ce5f2d2c4f04ff13980/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_transport rosbag2_tests rosbag2_py
TEST args:  --packages-above rosbag2_transport rosbag2_tests rosbag2_py
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12985
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19896/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14380/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20603/)

MichaelOrlov:

MichaelOrlov:
Note:
Backporting to the Iron and Humble is not possible because It will cause ABI-breaking changes.
-  it requires adding new member variables to the Player class and we don't have `pimpl` design for the Player class on the Iron and Humble branches.

PR #1507, opened 2023-11-30
Clang fixes for the latest PlayerImpl code.

clalancette:
1. The annotations for threading don't understand the {} initialization syntax.  Just use = false, which is what we do everywhere else.

2. We need to mark the destructor on PlayerImpl as virtual since it has virtual methods.

After this, clang compiles cleanly.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -85,7 +85,7 @@ public:
     const rosbag2_storage::StorageOptions & storage_options,
     const rosbag2_transport::PlayOptions & play_options);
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19910/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14389/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20618/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/43/)

MichaelOrlov (approved):
LGTM

PR #1516, opened 2023-12-17
Improve performance in SqliteStorage::get_bagfile_size()

### LINKED ISSUE #647: Improving performance of should_split_bagfile()
## Description
Bag performance suffers due to should_split_bagfile() checking file size for each incoming message. This is a simple disk operation but adds up very quickly, especially when lots of messages are going through (e. g. high-frequency publishers and/or many publishers).

**Note that one way to circumvent this in the meanwhile is to split by bagfile duration instead.**

### Expected Behavior
Determining whether to split the bag has a negligible cost compared to the cost of writing the message itself.

### Actual Behavior
Checking for the split condition is dominating the main thread in some cases (credit to @eboasson for finding out with a macOS datapoint).

## To Reproduce
Use the rosbag2_performance_benchmarking package. Set max bag size to a non-zero value. Make sure to run benchmarks setup with high message throughput (high frequencies and/or plenty of publishers). Smaller message size with the same MB/s would show a more pronounced effect.

## Additional context
Bag splitting itself could also be improved: https://github.com/ros2/rosbag2/issues/640

MichaelOrlov:
- Closes https://github.com/ros2/rosbag2/issues/647
- According to the https://www.sqlabs.com/blog/2010/12/getting-the-size-of-an-sqlite-database/. 
In SQLite3, the DB file size can be calculated as db `page_count * page_size`.

- Use DB query to get page_count and page_size instead of a filesystem blocking call.
- Added regression unit test to make sure that `SqliteStorage::get_bagfile_size()` returns correct value for both storages SQLite3 and MCAP.


MichaelOrlov:
@emersonknapp If you have experience with SQLite3 storage especially with raw database access I would appreciate your help with this PR.

I've made an implementation that is IMO pretty trivial and simple. However, I am getting one integration test failure with an error message that the database table is locked during the call for the open() method I think. I wouldn't expect this type of error for any scenarios in the current implementation and have no more ideas why DB table became locked.  
Any ideas or thoughts will be helpful. BTW  this error doesn't reproduce on my local setup with regular colcon build/test commands.
Here is the fragment of the log with failure.
```
[ RUN      ] TestRecordEndToEnd/RecordFixture.rosbag2_record_and_play_multiple_topics_with_filter/0
stdin is not a terminal device. Keyboard handling disabled.[INFO] [1702867875.584426960] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1702867875.587127627] [rosbag2_storage]: Opened database '/tmp/tmp_test_dir_SmHn5C/rosbag2_record_and_play_multiple_topics_with_filter_0_sqlite3/rosbag2_record_and_play_multiple_topics_with_filter_0_sqlite3_0.db3' for READ_WRITE.
[WARN] [1702867875.587304449] [rosbag2_cpp]: Environment variable ROS_DISTRO not set, can't store value in bag metadata.
[INFO] [1702867875.591818014] [rosbag2_recorder]: Event publisher thread: Starting
[INFO] [1702867875.591819858] [rosbag2_recorder]: Listening for topics...
[INFO] [1702867875.596126592] [rosbag2_recorder]: Subscribed to topic '/test_topic1'
[INFO] [1702867875.597216102] [rosbag2_recorder]: Subscribed to topic '/test_topic0'
[INFO] [1702867875.598653087] [rosbag2_recorder]: Subscribed to topic '/test_topic'
[INFO] [1702867875.600905696] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1702867875.602679117] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
[INFO] [1702867875.602791071] [rosbag2_recorder]: Recording...
terminate called after throwing an instance of 'rosbag2_storage_plugins::SqliteException'
  what():  Error when processing SQL statement. SQLite error (262): database table is locked
/tmp/ws/install_isolated/rosbag2_test_common/include/rosbag2_test_common/rosbag2_test_common/process_execution_helpers_unix.hpp:154: Failure
Expected equality of these values:
  (((status) & 0x7f) == 0)
    Which is: false
```

MichaelOrlov:
@emersonknapp Never mind, I have figured it out.
According to https://www2.sqlite.org/cvstrac/wiki?p=DatabaseIsLocked 
Only the first variant is possible 
> Trying to CREATE or DROP a table or index while a SELECT statement is still pending.
Sometimes people think they have finished with a SELECT statement because sqlite3_step() has returned SQLITE_DONE. But the SELECT is not really complete until sqlite3_reset() or sqlite3_finalize() have been called.

It turns out that when we are calling `if (should_split_bagfile(message_timestamp))` here 
https://github.com/ros2/rosbag2/blob/cea3fb027e52b9445266adde9d7629b9957a3c35/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L367-L370
the message_cache consumer thread can call `storage_->write(messages)` at the same time.
There is a race condition and possible situation when two threads will try to execute two different `SELECT` commands on the same database at the same time.

I will try to make protection with mutex and condition_variable on the `rosbag2_storage_sqlite3` layer

MichaelOrlov:
Update:
It turns out that the idea of synchronizing two threads with mutexes on the `rosbag2_storage_sqlite3 layer` wasn't a good solution.
Since it would cause locking for producer-consumer threads on the upper `SecuantialWriter` layer and discard all benefits of having a double-buffer scheme to split the transport and storage layers.

I made a workaround to avoid locking mutex and doing actual DB requests for the current number of pages. Instead, I moved it to the end of the DB writing functions. This way `get_bagfile_size()` Will return the already pre-evaluated bag file size value. We will do bag size calculation and evaluation per each DB write operation. 
To avoid extra burden on each message write operation when we are dumping the cache buffer to storage I will do a readout for the current number of pages from DB only after finishing dumping a batch of the messages from the cache buffer. 
Also, intermediate filesize evaluation for each `write(message)` call will be made based on the number of bytes written to the DB per each `write(message)` method call.

**Ready for review**

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
```diff
@@ -125,13 +125,16 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   RCPPUTILS_TSA_REQUIRES(database_write_mutex_);
   int get_last_rowid();
   int read_db_schema_version();
+  uint64_t get_page_size();
```
clalancette: This is logically:

```suggestion
  uint64_t get_page_size() const;
```

Though I'm not sure if that will work given the accesses to the database.  Something to consider, though.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -323,9 +337,14 @@ void SqliteStorage::write(
 
   for (auto & message : messages) {
     write_locked(message);
+    // Update db_file_size_ with the estimated data size written to the DB
+    // + 3 * sizeof(int64_t) This is for storing timestamp, topic_id and index for messages table
+    db_file_size_ += message->serialized_data->buffer_length + 3 * sizeof(int64_t);
```
clalancette, line 342: Is this necessary?  We are going to throw all of this away after the `commit_transaction` below, so can we just skip this?

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -852,6 +878,36 @@ int SqliteStorage::read_db_schema_version()
   return schema_version;
 }
 
+uint64_t SqliteStorage::read_total_page_count_locked()
+{
+  auto scope_exit_reset_statement = rcpputils::make_scope_exit(
+    [this]() {
+      page_count_statement_->reset();
+    });
```
clalancette, line 886: Why are we resetting `page_count_statement` after we use it?  I don't know this code very well, but it seems like this statement should just live for the lifetime of the object.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -852,6 +878,36 @@ int SqliteStorage::read_db_schema_version()
...
+  auto page_count_query_result = page_count_statement_->execute_query<int>();
+  auto page_count_query_result_begin = page_count_query_result.begin();
```
clalancette, line 888: This is just a matter of taste, but it seems unnecessary to assign a variable here.  You are only saving 2 characters while typing :).

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
```diff
@@ -125,13 +125,16 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
...
+  uint64_t read_total_page_count_locked() RCPPUTILS_TSA_GUARDED_BY(database_write_mutex_);
```
clalancette: This is marked as "GUARDED_BY" `database_write_mutex_`, but this isn't a write to the database, is it?  It seems like either the mutex is misnamed, or we need another mutex.

Also, this function is logically "const", though I don't know if it can be that for the same reasons as `get_page_size`.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: Yeah. I also was confused about this reset when was started implementing this feature and was trying to keep the statement state between invocations. Then I've got some intermittent errors with a message that the database is locked and can't perform other operations on it.  
After some investigation, it turns out that this reset is required for each statement invocation and part of the SQLite API workflow.  
From the `sqlite3.h`
> ** The sqlite3_reset() function is called to reset a [prepared statement]
** object back to its initial state, ready to be re-executed.
** ^Any SQL statement variables that had values bound to them using
** the [sqlite3_bind_blob | sqlite3_bind_*() API] retain their values.
** Use [sqlite3_clear_bindings()] to reset the bindings.
**
** ^The [sqlite3_reset(S)] interface resets the [prepared statement] S
** back to the beginning of its program.
**
** ^If the most recent call to [sqlite3_step(S)] for the
** [prepared statement] S returned [SQLITE_ROW] or [SQLITE_DONE],
** or if [sqlite3_step(S)] has never before been called on S,
** then [sqlite3_reset(S)] returns [SQLITE_OK].
**
** ^If the most recent call to [sqlite3_step(S)] for the
** [prepared statement] S indicated an error, then
** [sqlite3_reset(S)] returns an appropriate [error code].

Also please refer to my findings here https://github.com/ros2/rosbag2/pull/1516#issuecomment-1863342775

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: Unfortunately not. And in this case this sort of workaround with temporary `page_count_query_result_begin` is required because when we call `page_count_query_result.begin()` we actually calling `statement_->step()` inside and check if we have already called begin() iterator to not allow to call it more than once.
```
    Iterator begin()
    {
      try_access_data();
      return Iterator(statement_, 0);
    }
    void try_access_data()
    {
      if (is_already_accessed_) {
        throw SqliteException("Only one iterator per query result is supported!");
      }
      is_already_accessed_ = true;
    }
```
The rationale for caching `query_result.begin()` is because we need to call it twice (one more time in the return statement at the end of the function) and API allows to call it only once.
One more "skeleton in the SQLite3 wrapper API cabinet"

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: Yes. It is required because we have two threads in the rosbag2 recorder. The cache consumer and cache producer. There is possible a situation when we are dumping cache i.e. vector of messages to the disk by the cache consumer and at the same time cache producer requesting up-to-date `db_file_size` that will be needed on the upper layer to decide if bag already exceeds the limit and we need to make a bag split.

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
MichaelOrlov: Yes, probably the name of the mutex is not accurate. The situation is that we can't write and read from the database concurrently. Only one SQL statement is allowed per time. I am guessing that the name originally was chosen as `write` mutex because we didn't have a situation before when it was possible to call DB read and write operations from multiple threads.  
However, I can rename it to the `db_read_write_mutex`.

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
MichaelOrlov: It seems ok to make it const. I will add it.

MichaelOrlov:
@clalancette Thank you for the review. I've added `const` qualifiers for the functions and renamed mutex.  
Please take a look one more time.

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
```diff
@@ -152,11 +155,13 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   // This mutex is necessary to protect:
   // a) database access (this could also be done with FULLMUTEX), but see b)
   // b) topics_ collection - since we could be writing and reading it at the same time
-  std::mutex database_write_mutex_;
+  std::mutex db_read_write_mutex_;
```
fujitatomoya, line 158: this does not block the PR, i thought this is originally going to use `std::shared_mutex` for performance improvement especially read operation, but i only can find write operation for now.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -852,6 +878,36 @@ int SqliteStorage::read_db_schema_version()
...
+  return std::get<0>(*page_count_query_result_begin);
+}
+
+uint64_t SqliteStorage::get_page_size() const
+{
+  if (!database_) {
+    return 0;  // Trying to call get_page_size when SqliteStorage::open was not called or db
+    // failed to open
+  }
+  SqliteStatement get_page_size_statement = database_->prepare_statement("PRAGMA page_size;");
```
fujitatomoya, line 901: (making sure i understand correctly) i think that is intention not to have `get_page_size_statement_` member variable, instead using stack memory. because `get_page_size` is now called once during opening the sqlite database file, and not expected to be called many time after wards.

fujitatomoya (approved):
lgtm with green CI

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
MichaelOrlov: @fujitatomoya I've done some research about SQLite3 multithreaded access.  It turns out that SQLite doesn;t really support multithreaded access to the database.
From the https://www.sqlite.org/threadsafe.html
> SQLite supports three different threading modes:

> Single-thread. In this mode, all mutexes are disabled and SQLite is unsafe to use in more than a single thread at once.

> Multi-thread. In this mode, SQLite can be safely used by multiple threads provided that no single [database connection](https://www.sqlite.org/c3ref/sqlite3.html) nor any object derived from database connection, such as a [prepared statement](https://www.sqlite.org/c3ref/stmt.html), is used in two or more threads at the same time.

> Serialized. In serialized mode, API calls to affect or use any SQLite [database connection](https://www.sqlite.org/c3ref/sqlite3.html) or any object derived from such a database connection can be made safely from multiple threads. The effect on an individual object is the same as if the API calls had all been made in the same order from a single thread. The name "serialized" arises from the fact that SQLite uses mutexes to serialize access to each object.

Therefore as far as I understand correctly even when **Multi-thread** mode is enabled it is not allowed to call more than one prepared statement concurrently from multiple threads. In serialized mode, even worse all requests will be performed sequentially anyway. They will be quied in one thread on the DB level.

In addition to the SQLite3 limitations, we opted out of using DB in the unprotected Single-threaded mode in our `sqlite_wrapper` by explicitly specifying the `SQLITE_OPEN_NOMUTEX` flag during DB open operation.
https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_wrapper.cpp#L67-L81
I guess we do this for the sake of performance.

Given everything listed above we have only one choice to use only exclusive mutex.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
MichaelOrlov: @fujitatomoya Yes. your understanding is correct.  
And even if we would have a private member variable for the `get_page_size_statement` we would have to call database_->prepare_statement("PRAGMA page_size;"); and `get_page_size_statement->reset()` each time per `get_page_size` call due to the SQLite3 API limitation. Therefore it doesn't make sense to cache it in the member variable.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/858b61f3724319fdb05b3e8cb28fce6a/raw/2d6648ca88600ddc9ee2e390c25d256f31e2f64b/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_storage rosbag2_storage_sqlite3 rosbag2_tests
TEST args:  --packages-above rosbag2_storage rosbag2_storage_sqlite3 rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13110
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14512/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20779/)

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
fujitatomoya: @MichaelOrlov thanks for sharing this, it gives me very good information.

MichaelOrlov: Just to clarify. I think SQLite3 supports multithreaded access in the `Multi-thread` mode but in this case, need to use multiple DB connections. One per each concurrent thread.

MichaelOrlov:
Re-run Windows CI build after fixing warnings
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20785/)

MichaelOrlov:
@clalancette @emersonknapp The CI is green. However, I need formal approval from those who have write permissions to the rosbag2 repo to be able to merge it.  
Wouldn't you mind to additionally review or approve this PR?

emersonknapp (approved):

ros-discourse:

PR #1518, opened 2023-12-20
Cleanup the rosbag2_transport tests

### LINKED ISSUE #872: :man_farmer:  rosbag2_transport can run out of memory in coverage jobs
## Description

** Clear and concise description of the feature that you'd like to exist, from a user's perspective. Don't get into implementation details, describe what you want it to do. Put any consideration of alternate features that might solve your problem as well. **

The package `rosbag2_transport appears to have run out of memory while building in the Rolling coverage nightly.
https://ci.ros2.org/job/nightly_linux_coverage/1371/ . Building locally, I notice the package required ~15GB peak to build on my desktop, but the peak number probably varies with the number of CPU cores used.

I looked briefly at `rosbag2_transport` with `include-what-you-use` to see if there are excessive includes that could be increasing the memory required to compile. Nothing really sticks out to me except `rclcpp/rclcpp.hpp` can be replaced with one or two more specific headers in a lot of cases. I didn't try it to see what the impact to memory usage is. I attached the output from the tool in case anyone wants to try its suggestions. [iwyu.txt](https://github.com/ros2/rosbag2/files/7206571/iwyu.txt)

One thing that does standout is a lot of files get compiled multiple times when they don't need to be. `rosbag2_transport_add_gmock` calls `ament_add_gmock` with different `RMW_IMPLEMENTATION` environment variables set, but the code shouldn't need to be rebuilt because of that. A way of splitting building the exectuable from configuring the runtime properties of the tests would likely save a bunch of compilation time.

clalancette:
The main goal here is to reduce compile times and memory usage.  The way we do it:

1.  Make as many of the dependencies private as possible.  That way the main executable is exporting the minimum of dependencies to downstream consumers.
2.  Compile the test files once, instead of once per RMW.  We can do this because we change the RMW at runtime with an environment variable.  This requires some changes to ament_cmake, which are in https://github.com/ament/ament_cmake/pull/497

With this PR in place, build times drop by about 3x for me locally (from about 3 minutes to about 1 minute).

A draft until https://github.com/ament/ament_cmake/pull/497 lands.

Fixes #872


clalancette:
All right, https://github.com/ament/ament_cmake/pull/497 is merged now.  So converting this to a real PR and running CI on it.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19999/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14464/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20707/)

clalancette:
After a small fix here, let's try CI again:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20002/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20711/)

mjcarroll (approved):

rosbag2_transport/CMakeLists.txt
```diff
@@ -99,144 +100,153 @@ ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(export_${PROJECT_NAME})
 
 ament_export_dependencies(
-  rclcpp_components
   keyboard_handler
+  rclcpp
   rosbag2_cpp
-  rosbag2_compression
```
MichaelOrlov, line 105: Out of curiosity, why remove `rosbag2_compression` from dependencies? 
Implying that `rosbag2_compression` includes as dependencies for the `rosbag2_storage`?

rosbag2_transport/CMakeLists.txt
```diff
@@ -99,144 +100,153 @@ ament_export_libraries(${PROJECT_NAME})
...
-    ${rclcpp_components_TARGETS})
-
-  if(${rmw_implementation} MATCHES "rmw_cyclonedds(.*)")
-    ament_add_test_label(test_play_services__rmw_cyclonedds_cpp xfail)
-  endif()
+  ament_add_gmock_test(test_play
+    TEST_NAME test_play${target_suffix}
+    ENV ${rmw_implementation_env_var}
+    ${SKIP_TEST}
+  )
```
MichaelOrlov: @clalancette Can we somehow group invocation of the `ament_add_gmock_test(..) ` with the 
```
  ament_add_gmock_executable(test_play_loop test/rosbag2_transport/test_play_loop.cpp)
  target_link_libraries(test_play_loop rosbag2_transport ${test_msgs_TARGETS} rosbag2_test_common::rosbag2_test_common)
```
to land it in one place rather in two different locations in the CMake file?

MichaelOrlov (commented):
@clalancette Thanks for the PR. 
Since the https://github.com/ros2/rosbag2/issues/872 popup again I was considering to do something similar sonner.

You mentioned that build speed increased 3x, that is a good result and could be expectable, if we now compiling for 1 RMW implementation instead of 3 + other optimizations.

But what about memory consumption during compilation? Do you see any improvements?
Sorry, I have not had a chance to experiment with it myself.
I am curious if can do even better.

clalancette:
> But what about memory consumption during compilation? Do you see any improvements?

I did not measure it during build times.  I expect it will be better, just from the fact that we are compiling fewer things.  In terms of improvement, I think this is the best we can do with changes to the build system only; using less memory would probably require changes to the templates, in particular.  But I think that is a follow-up.

rosbag2_transport/CMakeLists.txt
clalancette: To be clear, we are only removing it from the *exported* dependencies.  That's because `rosbag2_compression` is only used in the source files, not the header files.

In other words, you only need to export dependencies that your header files depend on.  Downstream packages don't need to know about the internal dependencies.  In turn, this should make linking of downstream packages a bit faster.

rosbag2_transport/CMakeLists.txt
clalancette: It's possible, but I think it is going to be tricky.  For instance, we could add each test to a list, and then have `create_tests_for_rmw_implementation` iterate over the list, registering each one.  But then we would also need a way to mark certain tests as `SKIP_TEST`, and it isn't clear to me how we would do that in a list.  Ideas?

MichaelOrlov: @clalancette I came across today the following approach
https://github.com/ros2/rclcpp/blob/2bb1dc2c716ae12a24145891724f8436aa4e52bd/rclcpp/test/rclcpp/CMakeLists.txt#L295-L315
```cpp
function(test_generic_pubsub_for_rmw_implementation)
  set(rmw_implementation_env_var RMW_IMPLEMENTATION=${rmw_implementation})
  ament_add_gmock(test_generic_pubsub${target_suffix} test_generic_pubsub.cpp
    ENV ${rmw_implementation_env_var}
  )
  if(TARGET test_generic_pubsub${target_suffix})
    target_link_libraries(test_generic_pubsub${target_suffix} ${PROJECT_NAME} rcl::rcl ${test_msgs_TARGETS})
  endif()
endfunction()
call_for_each_rmw_implementation(test_generic_pubsub_for_rmw_implementation)

function(test_qos_event_for_rmw_implementation)
  set(rmw_implementation_env_var RMW_IMPLEMENTATION=${rmw_implementation})
  ament_add_gmock(test_qos_event${target_suffix} test_qos_event.cpp
    ENV ${rmw_implementation_env_var}
  )
  if(TARGET test_qos_event${target_suffix})
    target_link_libraries(test_qos_event${target_suffix} ${PROJECT_NAME} mimick rcutils::rcutils rmw::rmw ${test_msgs_TARGETS})
  endif()
endfunction()
call_for_each_rmw_implementation(test_qos_event_for_rmw_implementation)
```
It doesn't form a list of tests for running with `call_for_each_rmw_implementation` but rather calls tests one by one.
But it least everything which belongs to the one test in the one place.

clalancette: All right, [1fce988](https://github.com/ros2/rosbag2/pull/1518/commits/1fce988cc6a4baf3bf78715a120e811929140377) does this.  It is a little verbose, but it gets the job done.  @MichaelOrlov please take another look when you get a chance.

MichaelOrlov (approved):
@clalancette Thank you! LGTM.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20014/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14478/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20726/)

Issue #1519, opened 2023-12-21
TimeControllerClock::sleep_until(some_time_in_the_past) sleeps too long


KenYN:
## Description
When processing a high-frequency file (eg 5Gbps split into 1400 byte messages, or one message every 2.2 microseconds) `PlayerImpl::play_messages_from_queue()` starts publishing at slower than real-time

### Expected Behavior
I would expect publishing to happen in realtime.

### Actual Behavior
Extensive debugging pinned down the culprit to this function call:

https://github.com/ros2/rosbag2/blob/cea3fb027e52b9445266adde9d7629b9957a3c35/rosbag2_transport/src/rosbag2_transport/player.cpp#L852

If the message header time is behind the current Sim time, rather than returning immediately as there is no waiting to be done, it instead calls this line, which takes 2.2 microseconds to execute:

https://github.com/ros2/rosbag2/blob/cea3fb027e52b9445266adde9d7629b9957a3c35/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp#L171

According to [Stack Overflow](https://stackoverflow.com/q/70375285/1270789):

> Thus, wait_until() unconditionally releases the lock, and even if it unblocks right away on account of timeout, it must then reacquire the lock. This leaves open the possibility that another thread acquires the lock in between, in which case that other thread can hold the lock for an arbitrary amount of time, delaying the return of the wait_until().

I wonder if something other thread is grabbing the lock, or is a context switch stalling the thread?

## To Reproduce
Try playing a high-frequency file - I'd attach my test file, but it's over 3GB in size.

## To Fix

Before the line:

https://github.com/ros2/rosbag2/blob/cea3fb027e52b9445266adde9d7629b9957a3c35/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp#L171

I added code that skips that call if we have already passed `steady_until`. This hack works for me, but I'm unable due to company policies, etc, to work it into a PR, and I've not shown any code just in case my obvious one-liner causes any licensing issues.

## System (please complete the following information)
 - OS:Ubuntu 22.04
 - ROS 2 Distro: Rolling
 - Install Method: Source
 - Version: With Player composable node patches

MichaelOrlov:
@KenYN The first questions which arise reading the description of the issue, - Do you use RTOS of Linux with RTOS patch?
And what is the maximum context switch time in the OS?

[...]

PR #1521, opened 2023-12-22
call cv.wait_until only if necessary.

fujitatomoya:
address https://github.com/ros2/rosbag2/issues/1519

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
@@ -168,7 +168,9 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
       impl_->cv.wait_for(lock, impl_->sleep_time_while_paused);
     } else {
       const auto steady_until = impl_->ros_to_steady(until);
-      impl_->cv.wait_until(lock, steady_until);
+      if (steady_until > impl_->now_fn()) {
+        impl_->cv.wait_until(lock, steady_until);
+      }
     }
     if (impl_->paused) {
       // Don't allow publishing any messages while paused
--- a/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
@@ -168,6 +168,7 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
       impl_->cv.wait_for(lock, impl_->sleep_time_while_paused);
     } else {
       const auto steady_until = impl_->ros_to_steady(until);
+      // wait only if necessary for performance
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@KenYN what do you think? i would add your name under Co-Author, if that is fine for you?

CC: @MichaelOrlov

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -168,7 +168,9 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
       impl_->cv.wait_for(lock, impl_->sleep_time_while_paused);
     } else {
       const auto steady_until = impl_->ros_to_steady(until);
-      impl_->cv.wait_until(lock, steady_until);
+      if (steady_until > impl_->now_fn()) {
```
fujitatomoya, line 172: i do not think we need `else` statement here, instead it should fall down to the current sequence.

MichaelOrlov: @fujitatomoya Sorry I didn't get. What `else` statement did you mean?
Also it would be nice to add some comment above why we do this trick. To avoid removal it by someone in the future.

MichaelOrlov (commented):
@fujitatomoya LGTM if add a comment in the code with brief explanation why we need this check. i.g. to avoid possible extra delays in the `cv.wait_until()` when the timer is already due.

KenYN:
@fujitatomoya Thank you, that's just the code I had, and a name check would be nice too!

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
fujitatomoya: ah, that was supposed to mean, in `else` statement it does not return `true`, instead it honors `pause` state. (that was originally a note for myself, but thanks for checking.)

fujitatomoya:
@MichaelOrlov added the comment line for the reference in the future.

MichaelOrlov (approved):
LGTM.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/57b38338c7899819b223914f8e22e502/raw/ee222938d77b6edc2e97da2206bd7fce1314cec2/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13081
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20015/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20727/)

MichaelOrlov:

ros-discourse:

Issue #1529, opened 2024-01-05
`Recording stopped` message is duplicated when stop recording


fujitatomoya:
## Description

`Recording stopped` is printed twice when we stop recording with `ros2 bag record...`

### Expected Behavior

it should print only once.

### Actual Behavior

```
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 bag record --all-topics
[INFO] [1704478296.853012766] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1704478296.865478656] [rosbag2_recorder]: Listening for topics...
[INFO] [1704478296.865513875] [rosbag2_recorder]: Event publisher thread: Starting
[INFO] [1704478296.870645864] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1704478296.872183975] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
[INFO] [1704478296.872236186] [rosbag2_recorder]: Recording...
[INFO] [1704478296.995752165] [rosbag2_recorder]: Subscribed to topic '/chatter'
[INFO] [1704478297.006186393] [rosbag2_recorder]: Subscribed to topic '/parameter_events'
[INFO] [1704478305.969940135] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1704478305.975847475] [rosbag2_recorder]: Event publisher thread: Exiting
[INFO] [1704478305.975882003] [rosbag2_recorder]: Recording stopped
[INFO] [1704478305.977988895] [rosbag2_recorder]: Recording stopped
```

## To Reproduce

`ros2 bag record ...`

## System (please complete the following information)

 - OS: Ubuntu 22.04
 - ROS 2 Distro: Rolling
 - Install Method: Source build
 - Version: https://github.com/ros2/ros2/commit/ed4995259068d8e8a3c6eb586f4bc9a848dfbc89

fujitatomoya:
### Implementation Consideration

I think we can add the state control mutex lock (or rwlock, probably not needed since atomic operation can be enough to read the data such as `RecorderImpl::is_paused()`) aligned with `RecorderImpl::paused_` and `RecorderImpl::in_recording_`. i guess that we need to add some state checks at the beginning of methods.
this should avoid this original problem since if the recorder is already stopped, it does not have to do the stop procedure again.

MichaelOrlov:
- Closing a done in the #1530

PR #1530, opened 2024-01-05
`Recording stopped` prints only once.

fujitatomoya:
part of https://github.com/ros2/rosbag2/issues/1529.

the follow up PR is required for https://github.com/ros2/rosbag2/issues/1529#issuecomment-1884132718

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -209,7 +209,9 @@ RecorderImpl::RecorderImpl(
 RecorderImpl::~RecorderImpl()
 {
   keyboard_handler_->delete_key_press_callback(toggle_paused_key_callback_handle_);
-  stop();
+  if (in_recording_) {
+    stop();
+  }
 }
 
 void RecorderImpl::stop()


fujitatomoya (commented):
@MichaelOrlov minor fix, could you review?

MichaelOrlov (approved):
@fujitatomoya I was considering to put this check inside `RecorderImpl::stop()` method to cover other possible cases.
It is ok solution as intermitient fix.
However, will need to add protection with mutex for the `in_recording_` in the future to avoid race conditions.

fujitatomoya:
> However, will need to add protection with mutex for the in_recording_ in the future to avoid race conditions.

https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L139

it is atomic operation, so it should be protected?

MichaelOrlov:
@fujitatomoya Changing a flag is atomic however, transitioning from one state to another (recording/not recording) is not.
There are possible situations when we are in the process of starting recording and calling the `stop()` method from another thread.
Let's do protection with mutex in a separate PR to be able to backport the current PR.

fujitatomoya:
Ah, i see that resolves the problem too. i think we can close this PR.

fujitatomoya:
@MichaelOrlov why did you reopen this one? do you want to merge this 1st and follow-up PR with mutex locks?

MichaelOrlov:
@fujitatomoya 
> do you want to merge this 1st and follow-up PR with mutex locks?

Yes, exactly. Becasue if we will add a new mutex we will not be able to backport it to the Humble and Iron.

fujitatomoya:
one thing i want to to confirm is, i believe https://github.com/ros2/rosbag2/issues/1529#issuecomment-1884132718 is planned to be in `RecorderImpl`, which does not break the userspace. is that not acceptable?

MichaelOrlov:
@fujitatomoya The problem is that we introduced `RecorderImpl` only on the Iron distro. Humble doesn't have it
https://github.com/ros2/rosbag2/blob/a296fb57577fc4f121a9f5ecebb459a385e5f13c/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L113-L131

fujitatomoya:
> The problem is that we introduced RecorderImpl only on the Iron distro. Humble doesn't have it

Ah 😅 I see. can you run CI for this? i am gonna follow up for https://github.com/ros2/rosbag2/issues/1529#issuecomment-1884132718 with another PR.

MichaelOrlov:
@fujitatomoya Yes, please go ahead and run CI. I 've approved this PR.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20061/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14514/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20781/)

fujitatomoya:
@MichaelOrlov looks good to go!

MichaelOrlov:

PR #1531, opened 2024-01-08
Use middleware send and receive timestamps from message_info during recording

jmachowinski:
- depends from https://github.com/ros2/rclcpp/pull/1928

This should create a better timestamp in cases of high load.

Note, I am unsure, if we should use the received_timestamp or if the source_timestamp would be the better option.

Disclaimer: The `send_timestamp` will be available only for the `MCAP` storage format for a while, since we already have a special field for it in `MCAP` format specification.


jmachowinski:
Note, this was made possible by the merge of https://github.com/ros2/rclcpp/pull/1928

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -518,9 +518,12 @@ RecorderImpl::create_subscription(
...
     topic_type,
     qos,
-    [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message) {
+    [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
+    const rclcpp::MessageInfo & mi) {
       if (!paused_.load()) {
-        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
+        writer_->write(
+          message, topic_name, topic_type,
+          rclcpp::Time(mi.get_rmw_message_info().received_timestamp));
```
fujitatomoya: this can be better system time stamp in the rosbag2, but one question.

what if `--use-sim-time` option is enabled with `ros2 bag record`? i guess it does not work as expected? rmw is not aware of ros simulation time clock, so this will always store the system time from rmw implementation.

fujitatomoya:
CC: @MichaelOrlov

rosbag2_transport/src/rosbag2_transport/recorder.cpp
jmachowinski: fair point, i will rework the patch for this case

MichaelOrlov (changes requested):
@jmachowinski @fujitatomoya Not this way, and unfortunately not so easy.

We have reserved special fields `publishTime` and `sequence` for `mcap` storage format specifically for this use case. However, currently they are populated with dummy data.
https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_storage_mcap/src/mcap_storage.cpp#L752-L757

From mcap vendor `Message` struct
```
  /**
   * @brief An optional sequence number. If non-zero, sequence numbers should be
   * unique per channel and increasing over time.
   */
  uint32_t sequence;
  /**
   * @brief Nanosecond timestamp when this message was recorded or received for
   * recording.
   */
  Timestamp logTime;
  /**
   * @brief Nanosecond timestamp when this message was initially published. If
   * not available, this should be set to `logTime`.
   */
  Timestamp publishTime;
```

The proper implementation will be to add those two fields `publish_time` and `sequence_number` to the `rosbag2_storage::SerializedBagMessage` struct
https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_storage/include/rosbag2_storage/serialized_bag_message.hpp#L27-L32
and change or add a new rosbag2_cpp::Writer::write(..) API
https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L185-L200
to be able to pass those two new fields from the rosbag2 transport layer. i.g. recorder.

jmachowinski:
> The proper implementation will be to add those two fields publish_time and sequence_number to the rosbag2_storage::SerializedBagMessage struct

Do we really need the sequence_number ? I don't see any practical value in it, though I might be missing something...

jmachowinski:
Updated the patch.

@MichaelOrlov  I also started to implement your suggestion, of passing the send timestamp along.
While doing this, I was wondering, if we should add a recoder option for using the send timestamp as recv timestamp.

The reason for this suggestion would be, that none of the tooling and the API seems to be set up to use the send timestamp, but by switching the timestamps, we could 'hack' our way around this shortcoming.

MichaelOrlov:
@jmachowinski As reagards:
> Do we really need the sequence_number ? I don't see any practical value in it, though I might be missing something...

The sequence number needed to be able to detect on what step messages were lost and how many of them were lost. Or the other way around to make sure that it wasn't data losses during recording. i.e. check for data consistency.

As regards:
> While doing this, I was wondering, if we should add a recoder option for using the send timestamp as recv timestamp.

> The reason for this suggestion would be, that none of the tooling and the API seems to be set up to use the send timestamp, but by switching the timestamps, we could 'hack' our way around this shortcoming.

No hacking solutions will be accepted for the rosbag2. Sorry, I have a strong opinion on that.   
Instead, we would add an extra parameter for the rosbag2 player to be able to playback based on the published timestamp.

fujitatomoya:
@MichaelOrlov 

> The sequence number needed to be able to detect on what step messages were lost and how many of them were lost. Or the other way around to make sure that it wasn't data losses during recording. i.e. check for data consistency.

i think use case of `sequence` is clear but, i guess @jmachowinski was asking what can be filled in it from `rmw_message_info_t`? if i am not mistaken... @jmachowinski correct me if i got wrong.

i guess, we can generate incremental number for `sequence` during recording, but what's in it for us?
or are you asking the we need to formulate the `sequence` number based on `publisher_gid` and `publication_sequence_number` to detect the message gap?

jmachowinski:
> No hacking solutions will be accepted for the rosbag2. Sorry, I have a strong opinion on that.
Instead, we would add an extra parameter for the rosbag2 player to be able to playback based on the published timestamp.

I also got a strong opinion against workarounds. But in this case I don't see a good way around it. The problem is, that its not only this library. We got rqt_bag and foxglove studio down the line, that would need patches as well. 

For our use case the receive timestamp is mostly useless and we are way more interested in analyzing the data in send order. And I guess we are not the only one... 

This leads to a more general question, is the receive timestamp actually relevant to anyone ?

> The sequence number needed to be able to detect on what step messages were lost and how many of them were lost. Or the other way around to make sure that it wasn't data losses during recording. i.e. check for data consistency.

Okay, that is a valid application. 
Which leads to followup questions:
 - Is there anyone really evaluating this field ?
 - Is it worth the memory overhead per message, or would print a warning on recording also be sufficient ?

@fujitatomoya The technical part on how to fill in the sequence number is clear to me, its more about is it worth the memory overhead. Adding a uint32_t might not sound like much, but for comparison, we have a own implementation, of the snapshot feature, which has a memory food print that is about ~1/3 smaller compared to rosbag2, as we don't save the topic name but a 16 bit topic id in every message.

MichaelOrlov:
@jmachowinski As regards to the:
> I also got a strong opinion against workarounds. But in this case I don't see a good way around it. The problem is, that its not only this library. We got rqt_bag and foxglove studio down the line, that would need patches as well.

It only means that if other tools like `Foxglove studio` and `rqt_bag` need to be extended for the new functionality they will get it eventually. AFAIK Foxglove studio doesn't have this feature only because we are not yet writing valid data to the relevant fields from the rosbag2.

As regards 
> This leads to a more general question, is the receive timestamp actually relevant to anyone ?

Well, at least currently everybody else using the receive timestamp and changing it by default to the published timestamp may cause breakage in workflow and other algorithms. The safest way will be to add an option with the new fields and new functionality.

As regards to the:
> Adding a uint32_t might not sound like much, but for comparison, we have a own implementation, of the snapshot feature, which has a memory food print that is about ~1/3 smaller compared to rosbag2, as we don't save the topic name but a 16 bit topic id in every message.

This is not true with MCAP file format anymore because:
1. We don't save the topic name as a string per each message in MCAP format. See
https://github.com/ros2/rosbag2/blob/195e406194a0bd1901261b8fef180245b70f1a9d/rosbag2_storage_mcap/src/mcap_storage.cpp#L738-L751
2. We are not adding new fields to the storage. Those fields are already present in MCAP file format and currently written with dummy data.

jmachowinski:
> This is not true with MCAP file format anymore because

I am talking about RAM usage, not SSD/HDD usage.

The problem is within SerializedBagMessage that is stored in CircularMessageCache while using the snapshot feature.
If we add additional members to this structure it will have an impact on the RAM usage.


>>  This leads to a more general question, is the receive timestamp actually relevant to anyone ?

> Well, at least currently everybody else using the receive timestamp and changing it by default to the published timestamp may cause breakage in workflow and other algorithms. The safest way will be to add an option with the new fields and new functionality.

I agree that it is the safest route, that is out of the question.

I am more wondering, if it was a deliberate decision to use the receive timestamp in the first place, or was it just used because it was available. 
And as a followup question, would it make sense to do a transition away from the receive timestamp in the long run ?

fujitatomoya:
> if it was a deliberate decision to use the receive timestamp in the first place, or was it just used because it was available.
> And as a followup question, would it make sense to do a transition away from the receive timestamp in the long run ?

good question.
in the application perspective especially dealing with sensing data, i do need `published time`(close to data generation time) more than `received time` to see if we can use this data. but that is just my opinion, maybe there could be more history for this decision.

MichaelOrlov:
@jmachowinski @fujitatomoya 
As regards to the 
> if it was a deliberate decision to use the receive timestamp in the first place, or was it just used because it was available.
And as a followup question, would it make sense to do a transition away from the receive timestamp in the long run ?

I also don't know the answer to this question. It was before me.  
I don't think that it make sense to do a complete transition away from the receive timestamp in the long run. 
I envision it to be useful when will need to determine message delay on the transport layer at least.

Another valuable point that was likely driving the decision is clock synchronization on different ECUs and nodes.  
On a heterogenous system with multiple HW units, clock synchronization is not an easy task, especially during the R&D phase or quick prototyping. 
Without precise clock synchronization on such a configuration, it will be impossible to playback messages using the sent timestamp since they likely will be out of order in the time domain after recording.

@jmachowinski  As regards:
>   am talking about RAM usage, not SSD/HDD usage.

> The problem is within SerializedBagMessage that is stored in CircularMessageCache while using the snapshot feature.
If we add additional members to this structure it will have an impact on the RAM usage.

Noted. As mitigation for such a problem, I only can suggest to start using `topic_id` instead of the `topic_name` in the `SerializedBagMessage` as you mentioned above. Perhaps `topic_id` should be returned by the underlying storage plugin e.g MCAP or SQLIte3 when adding a new topic to the storage. And will need yet to figure out how to match `topic_id` to the `topic_name` during the playback.
However, anyway, we need two new additional fields for the `published_timestamp` and `sequence_number` in the `rosbag2_storage::SerializedBagMessage` struct.

ros-discourse:

jmachowinski:
I added the publisher timestamp. And renamed the existing one to receive timestamp.

The sequence number is a different beast, and I would put it into a separate pull request.  
The problem with the sequence number is, you have either the receive sequence number, which is continuous, but does not tell you, if you lost a msg in transmission, as it is incremented every time a message is received. So there is no real point in recording it.

The second available sequence number is the one of the sender, which would tell you that you lost messages in transmission, but it will jump all over the place if you got multiple senders. Don't have an idea how to solve this...

rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -360,7 +360,7 @@ SequentialCompressionWriter::compress_message(
   std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message)
 {
   auto compressed_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
-  compressed_message->time_stamp = message->time_stamp;
+  compressed_message->receive_time_stamp = message->receive_time_stamp;
```
MichaelOrlov: Need to update `publish_time_stamp` as well here.
```
  compressed_message->publish_time_stamp = message->send_time_stamp;
```

rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -199,6 +199,26 @@ class ROSBAG2_CPP_PUBLIC Writer
...
+   * \param send_time The time stamp when the message was send
+   * \throws runtime_error if the Writer is not open.
+   */
+  void write(
+    std::shared_ptr<const rclcpp::SerializedMessage> message,
+    const std::string & topic_name,
+    const std::string & type_name,
+    const rclcpp::Time & recv_time,
+    const rclcpp::Time & send_time);
+
```
MichaelOrlov:
```suggestion
  void write(
    std::shared_ptr<const rclcpp::SerializedMessage> message,
    const std::string & topic_name,
    const std::string & type_name,
    const rcutils_time_point_value_t & recv_timestamp,
    const rcutils_time_point_value_t & send_timestamp);
```
It doesn't make sense to use `rclcpp::Time` for the newly created API since we will have to convert timestamp twice. It will be more efficient to use `rcutils_time_point_value_t` the same type as in the destination structure. Please note that the source timestamp from message info is also derived from rcutils time point
```
typedef rcutils_time_point_value_t rmw_time_point_value_t;
```

rosbag2_cpp/src/rosbag2_cpp/converter.cpp
```diff
@@ -75,13 +75,13 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> Converter::convert(
...
     allocated_ros_message.get(), message->topic_name.c_str());
-  allocated_ros_message->time_stamp = message->time_stamp;
+  allocated_ros_message->time_stamp = message->receive_time_stamp;
   input_converter_->deserialize(message, introspection_ts, allocated_ros_message);
 
   // re-serialize
   output_message->serialized_data = rosbag2_storage::make_empty_serialized_message(0);
   output_message->topic_name = std::string(allocated_ros_message->topic_name);
-  output_message->time_stamp = allocated_ros_message->time_stamp;
+  output_message->receive_time_stamp = allocated_ros_message->time_stamp;
```
MichaelOrlov:
```suggestion
  output_message->receive_time_stamp = message->receive_time_stamp;
  output_message->publish_time_stamp = message->send_time_stamp;
```

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -182,10 +183,21 @@ void Writer::write(
...
+{
+  write(message, topic_name, type_name, time, time);
+}
+
+void Writer::write(
+  std::shared_ptr<const rclcpp::SerializedMessage> message,
+  const std::string & topic_name,
+  const std::string & type_name,
+  const rclcpp::Time & recv_time,
+  const rclcpp::Time & send_time)
```
MichaelOrlov:
```suggestion
void Writer::write(
  std::shared_ptr<const rclcpp::SerializedMessage> message,
  const std::string & topic_name,
  const std::string & type_name,
  const rcutils_time_point_value_t & recv_timestamp,
  const rcutils_time_point_value_t & send_timestamp)
```

rosbag2_cpp/src/rosbag2_cpp/writer.cpp
```diff
@@ -182,10 +183,21 @@ void Writer::write(
...
 {
   auto serialized_bag_message = std::make_shared<rosbag2_storage::SerializedBagMessage>();
   serialized_bag_message->topic_name = topic_name;
-  serialized_bag_message->time_stamp = time.nanoseconds();
+  serialized_bag_message->receive_time_stamp = recv_time.nanoseconds();
+  serialized_bag_message->publish_time_stamp = send_time.nanoseconds();
```
MichaelOrlov: If it will be `rcutils_time_point_value_t` we will not need to use `nanoseconds()` getter
```suggestion
  serialized_bag_message->receive_time_stamp = recv_timestamp;
  serialized_bag_message->publish_time_stamp = send_timestamp;
```

rosbag2_py/src/rosbag2_py/_writer.cpp
```diff
@@ -54,7 +54,7 @@ class Writer : public rosbag2_cpp::Writer
     bag_message->topic_name = topic_name;
     bag_message->serialized_data =
       rosbag2_storage::make_serialized_message(message.c_str(), message.length());
-    bag_message->time_stamp = time_stamp;
+    bag_message->receive_time_stamp = time_stamp;
```
MichaelOrlov: Need to update ` bag_message->publish_time_stamp = time_stamp;` as well here.

rosbag2_storage/include/rosbag2_storage/serialized_bag_message.hpp
```diff
@@ -27,7 +27,15 @@ namespace rosbag2_storage
...
-  rcutils_time_point_value_t time_stamp;
+  /**
+   * @brief Nanosecond timestamp when this message was received.
+   */
+  rcutils_time_point_value_t receive_time_stamp;
+  /**
+   * @brief Nanosecond timestamp when this message was initially published. If
+   * not available, this will be set to time_stamp.
+   */
+  rcutils_time_point_value_t publish_time_stamp;
```
MichaelOrlov: @jmachowinski Could you please rename variables with timestamps?
```suggestion
  /**
   * @brief Nanosecond timestamp when this message was received.
   */
  rcutils_time_point_value_t recv_timestamp;
  /**
   * @brief Nanosecond timestamp when this message was initially published. If
   * not available, this will be set to recv_timestamp.
   */
  rcutils_time_point_value_t send_timestamp;
```
The motivation is following:
1. Don't use underscore between `time_stamp` 
  - a) for consistency with rclcpp and other core packages. it will be shorter in writing and it looks and reads like one entity `timestamp`. 
  - b) We usually divide with an underscore when more than one logical entity is combined. I know in some places in rosbag2 we are already using `time_stamp` but let's not repeat that style and make it better.
2. The `recv` and `send` is shorter than `receive` and `publish`.
3. The `publish` might be confusing. Since when one will read its name it will be not clear whether this timestamp for the player with what timestamp this message will be published during playback or some other sophisticated logic or truly timestamp when the message was published. It would be more clear if it would be `received_timestamp` and `published_timestamp`. However,  `recv_timestamp` and `send_timestamp` are shorter and also not introducing confusion about `publish`.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,16 +523,35 @@ std::shared_ptr<rclcpp::GenericSubscription>
...
       if (!paused_.load()) {
-        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
+        writer_->write(
+          message, topic_name, topic_type,
+          rclcpp::Time(mi.get_rmw_message_info().received_timestamp),
+          rclcpp::Time(mi.get_rmw_message_info().source_timestamp));
       }
     });
-  return subscription;
 }
```
MichaelOrlov: @jmachowinski I think we can simplify the implementation of this function and reduce delta by doing it like this
```cpp
std::shared_ptr<rclcpp::GenericSubscription>
RecorderImpl::create_subscription(
{
  return node->create_generic_subscription(
    topic_name,
    topic_type,
    qos,
    [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
    const rclcpp::MessageInfo & mi) {
      if (!paused_.load()) {
        writer_->write(
          message, topic_name, topic_type,
          record_options_.use_sim_time ?
          node->get_clock()->now().nanoseconds() : mi.get_rmw_message_info().received_timestamp),
          mi.get_rmw_message_info().source_timestamp);
      }
    });
}
```

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
```diff
@@ -135,5 +135,5 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
     }
   }
   // check that the timestamp is same as the clock message
-  EXPECT_THAT(string_messages[0]->time_stamp, time_value);
+  EXPECT_THAT(string_messages[0]->receive_time_stamp, time_value);
```
MichaelOrlov: Please add extra check
```
  // Check that the publish_time_stamp is not the same as the clock message
  EXPECT_NE(string_messages[0]->publish_time_stamp, time_value);
  EXPECT_NE(string_messages[0]->publish_time_stamp, 0);
```

MichaelOrlov (changes requested):
@jmachowinski Thanks for your contribution.

I think it does make sense to use `rcutils_timestamp_value_t` instead of the `rclcpp::Time` in the newly added `write(..)` API for efficiency. See explanations and details in the dedicated comment.
Also, I would like to ask a renames for the time stamp variables. See motivation and explanation in the dedicated comment.

MichaelOrlov:
@jmachowinski As regards
> I added the publisher timestamp. And renamed the existing one to receive timestamp.
> 
> The sequence number is a different beast, and I would put it into a separate pull request. The problem with the sequence number is, you have either the receive sequence number, which is continuous, but does not tell you, if you lost a msg in transmission, as it is incremented every time a message is received. So there is no real point in recording it.
> 
> The second available sequence number is the one of the sender, which would tell you that you lost messages in transmission, but it will jump all over the place if you got multiple senders. Don't have an idea how to solve this...

Yeah, that is not easy. We tried to brainstorm a situation around the sequence number during the last `Rosbag2 and Tooling Working Group` meeting last Friday. 
The straightforward solution would be to record or encode in the sequence number the publisher id along with each message to distinguish and rule out the situation when multiple publishers exist on the same topic.
However, it turns out that `publisher_gid` type defined as 
```cpp
typedef struct RMW_PUBLIC_TYPE rmw_gid_s
{
  /// Name of the rmw implementation
  const char * implementation_identifier;

  /// Byte data GID value
  uint8_t data[RMW_GID_STORAGE_SIZE];
} rmw_gid_t;
```
where `#define RMW_GID_STORAGE_SIZE 16u` 
i.e. We would need an extra **16** bytes for each message to store the publisher id along with it. 
This sounds like a prohibitive overhead which we can't afford.

After some more consideration, it seems I found an elegant and I think proper solution.
If we start differentiating topics by `topic_id` assigned by the storage plugin instead of the `topic_name` i.e. finish with https://github.com/ros2/rosbag2/issues/1553. We can add `publisher_id` to the `TopicMetadata` and create unique `topic_id` aka `channel_id` for each publisher. Then we will split the recording per each publisher as a separate topic. And the problem will go away since the sequence number will be unique per channel. We can keep unorderd_map on the `rosbag2_transport` layer to find a match between `publisher_id` and generated `topic_id` and will not need to save publisher id with each message.
However will need to figure out yet how to fuse those topic channels during the playback. Likely there will not be extra effort and they will be organically aligned with the current design since the messages will be saved and playback in the same order since they are arriving in the same order and timestamp. We usually just read the next message from storage. If we will need some sorting before playback then we will need to account that topic could be split on multiple channels. But this is another story...

As regards the current implementation.
I am ok if we will add a sequence number field in the separate follow-up PR.
However, it would be ~~better~~ preferable to add it to the scope of this PR at least to the newly added `write(..)` API with timestamps as a placeholder. It could be assigned to 0 for now. The motivation is that each time when we have to change API we can't really add some parameters to the method or function we have to create a new one and deprecate the old one. To avoid breakage in someone's downstream code and dependent packages.

cc: @fujitatomoya @clalancette

rosbag2_transport/src/rosbag2_transport/recorder.cpp
jmachowinski: This would add an additional if on every message receive.

jmachowinski: Hm, do we want to set the rmw send timestamp in the sim case ? It will be in a different time domain

jmachowinski:
@MichaelOrlov  adopted code to the request changes

rosbag2_py/src/rosbag2_py/_writer.cpp
```diff
@@ -54,7 +54,7 @@ class Writer : public rosbag2_cpp::Writer
     bag_message->topic_name = topic_name;
     bag_message->serialized_data =
       rosbag2_storage::make_serialized_message(message.c_str(), message.length());
-    bag_message->receive_time_stamp = time_stamp;
+    bag_message->recv_timestamp = time_stamp;
```
MichaelOrlov: @jmachowinski Please add forgotten initialization for the `send_timestamp`
```
bag_message->send_timestamp = time_stamp;;
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Hi @jmachowinski Sorry for a bit late response
As regards:
> Hm, do we want to set the rmw send timestamp in the sim case ? It will be in a different time domain

I was thinking why not? May be for someone would be useful to have a published timestamp from the original domain not simulated. Better to have this extra information rather than lose it by rewriting it with the received timestamp. 
I also noticed that you changed `node->get_clock()->now()` to the `node->now()` as far as I remember they are not the same and we had some issue before with it and deliberately using `node->get_clock()->now()` for simulation time.

As regards:
> This would add an additional if on every message receive.

This is a very good point and I overlooked it. I agree that it would better to avoid extra check for the sake of the performance.
In this case, I would suggest refactoring this method as:
```cpp
std::shared_ptr<rclcpp::GenericSubscription>
RecorderImpl::create_subscription(
{
  if (record_options_.use_sim_time) {
    return node->create_generic_subscription(
      topic_name,
      topic_type,
      qos,
      [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
      const rclcpp::MessageInfo & msg_info) {
        if (!paused_.load()) {
          writer_->write(
            std::move(message), topic_name, topic_type,
            node->get_clock()->now().nanoseconds(),
            msg_info.get_rmw_message_info().source_timestamp);
        }
      });
  } else {
    return node->create_generic_subscription(
      topic_name,
      topic_type,
      qos,
      [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
      const rclcpp::MessageInfo & msg_info) {
        if (!paused_.load()) {
          writer_->write(
            std::move(message), topic_name, topic_type,
            msg_info.get_rmw_message_info().received_timestamp,
            msg_info.get_rmw_message_info().source_timestamp);
        }
      });
  }
}
```

jmachowinski: > I also noticed that you changed `node->get_clock()->now()` to the `node->now()` as far as I remember they are not the same and we had some issue before with it and deliberately using `node->get_clock()->now()` for simulation time.

There should be literally not difference
```c++
rclcpp::Time
Node::now() const
{
  return node_clock_->get_clock()->now();
}
``` 
```c++
rclcpp::Clock::ConstSharedPtr
Node::get_clock() const
{
  return node_clock_->get_clock();
}
```

jmachowinski:
About the channel_id idea, I thing this could work out nicely. The only problem I see is, that rosbag info would show topics multiple times, and some work might be needed there.

fujitatomoya (commented):
lgtm with green CI.

i think we can discuss on `sequence number` feature with another PR. (https://github.com/ros2/rosbag2/issues/1553)

fujitatomoya (approved):

ros-discourse:

https://discourse.ros.org/t/ros-2-meeting-minutes-2024-02-15/36221/1

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,16 +523,37 @@ std::shared_ptr<rclcpp::GenericSubscription>
...
+  if (record_options_.use_sim_time) {
+    return node->create_generic_subscription(
+      topic_name,
+      topic_type,
+      qos,
+      [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
+      const rclcpp::MessageInfo & mi) {
+        if (!paused_.load()) {
+          writer_->write(
+            message, topic_name, topic_type, node->now().nanoseconds(),
```
MichaelOrlov: Neatpick.
We can use `std::move(message)` when passing the message to the `writer::write(..)` method.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,16 +523,37 @@ std::shared_ptr<rclcpp::GenericSubscription>
...
+  } else {
+    return node->create_generic_subscription(
+      topic_name,
+      topic_type,
+      qos,
+      [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
+      const rclcpp::MessageInfo & mi) {
+        if (!paused_.load()) {
+          writer_->write(
+            message, topic_name, topic_type,
```
MichaelOrlov: Neatpick.
We can use `std::move(message)` when passing the message to the `writer::write(..)` method.

MichaelOrlov (approved):
@jmachowinski Thanks for addressing the review comments.
 LGTM with one neat pick.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @jmachowinski Thanks for elaborating and digging deeper into the rclcpp node's clock routines.
I agree there are no differences. It seems I messed up with something else.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
jmachowinski: good catch, updated

rosbag2_transport/src/rosbag2_transport/recorder.cpp
jmachowinski: updated

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,16 +523,37 @@ std::shared_ptr<rclcpp::GenericSubscription>
...
-    topic_type,
-    qos,
-    [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message) {
-      if (!paused_.load()) {
-        writer_->write(message, topic_name, topic_type, node->get_clock()->now());
-      }
-    });
-  return subscription;
+  std::function<void(std::shared_ptr<const rclcpp::SerializedMessage>,
+    const rclcpp::MessageInfo &)> recvFun;
```
MichaelOrlov: @jmachowinski Sorry I overlooked it.  
Could you please remove this unused `recvFun` declaration?
```suggestion
```

MichaelOrlov:
@clalancette Have you seen such errors `"failed to initialize rosout publisher"` on node creation from rclpy before?
```
                                                                 namespace = namespace or ''
2024-02-21T08:38:46.4532922Z             if not self._context.ok():
2024-02-21T08:38:46.4533682Z                 raise NotInitializedException('cannot create node')
2024-02-21T08:38:46.4534385Z             with self._context.handle:
2024-02-21T08:38:46.4534889Z                 try:
2024-02-21T08:38:46.4535324Z     >               self.__node = _rclpy.Node(
2024-02-21T08:38:46.4535879Z                         node_name,
2024-02-21T08:38:46.4536356Z                         namespace,
2024-02-21T08:38:46.4536868Z                         self._context.handle,
2024-02-21T08:38:46.4537414Z                         cli_args,
2024-02-21T08:38:46.4537918Z                         use_global_arguments,
2024-02-21T08:38:46.4538470Z                         enable_rosout
2024-02-21T08:38:46.4538946Z                     )
2024-02-21T08:38:46.4541163Z     E               rclpy._rclpy_pybind11.RCLError: failed to initialize rosout publisher: create_publisher() could not create data writer, at /__w/rosbag2/rosbag2/ros_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/publisher.cpp:275, at /__w/rosbag2/rosbag2/ros_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:117
2024-02-21T08:38:46.4543482Z     
2024-02-21T08:38:46.4544174Z     ../../install/rclpy/lib/python3.10/site-packages/rclpy/node.py:187: RCLError
2024-02-21T08:38:46.4545387Z     ----------------------------- Captured stdout call -----------------------------
2024-02-21T08:38:46.4548498Z     [NON-XML-CHAR-0x1B][37;1m2024-02-21 08:38:03.292 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_QOS_CHECK[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mHISTORY DEPTH must be lower or equal to the max_samples_per_instance value.[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcheck_qos[NON-XML-CHAR-0x1B][m
2024-02-21T08:38:46.4554892Z     [NON-XML-CHAR-0x1B][37;1m2024-02-21 08:38:03.324 [NON-XML-CHAR-0x1B][31;1m[[NON-XML-CHAR-0x1B][37;1mRTPS_QOS_CHECK[NON-XML-CHAR-0x1B][31;1m Error] [NON-XML-CHAR-0x1B][37mHISTORY DEPTH must be lower or equal to the max_samples_per_instance value.[NON-XML-CHAR-0x1B][34;1m -> Function [NON-XML-CHAR-0x1B][36mcheck_qos[NON-XML-CHAR-0x1B][m
2024-02-21T08:38:46.4557740Z     ----------------------------- Captured stderr call -----------------------------
2024-02-21T08:38:46.4558602Z     stdin is not a terminal device. Keyboard handling disabled.
```
I've got it on failing [build_and_test](https://github.com/ros2/rosbag2/actions/runs/7985746952/job/21804776556?pr=1531) job from this PR.
However, the changes in this PR seem unrelated to the failure.  
I can disable the "rosout" for the rosbag2 recorder node at all in the foreseeable future, but IMO it should work as is. Otherwise seems something is broken on the lower layers.

MichaelOrlov:
I see the detailed failure info from the FastRTPS
```
HISTORY DEPTH must be lower or equal to the max_samples_per_instance value
```
But what is the `max_samples_per_instance` ?

clalancette:
> @clalancette Have you seen such errors `"failed to initialize rosout publisher"` on node creation from rclpy before?

No, but all of the nightlies failed last night because of a bug in Fast-DDS.  We have now pinned to a known working version: https://github.com/ros2/ros2/pull/1527 .  That could be the same problem here.

MichaelOrlov:
@clalancette Thanks for information. Now `build_and_test` job passes green.
I will run CI then.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/aedd0afd1f58c3779a18cf5bdd0c379c/raw/201805c10b344285609d0317dd7366a06cc74cf6/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_performance_benchmarking rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_performance_benchmarking rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13341
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20338/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21087/)

MichaelOrlov:
@jmachowinski The `TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)` 
https://github.com/cellumation/rosbag2/blob/aa9fed89abe82fe25d45518a2315ee9d65c8e5f6/rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp#L140-L142
failed on Windows CI job
https://ci.ros2.org/job/ci_windows/21087/testReport/junit/rosbag2_transport/RecordIntegrationTestFixture/record_all_with_sim_time/

It turns out that `send_timestamp` is equal to **0** on Windows.

cc: @clalancette Do you know if we have a different implementation for `message_info::source_timestamp` on Windows?

clalancette:
> cc: @clalancette Do you know if we have a different implementation for `message_info::source_timestamp` on Windows?

I don't really know.  It looks like it only happens on Connext, so maybe @asorbini knows.

MichaelOrlov:
@jmachowinski @clalancette  I've found the place where it is settled up to zero.
https://github.com/ros2/rmw_connextdds/blob/d2bd90826e19097a429ba150b12e67d3822c1ccc/rmw_connextdds_common/src/common/rmw_impl.cpp#L1970-L1983
```cpp
rmw_connextdds_message_info_from_dds(
  rmw_message_info_t * const to,
  const DDS_SampleInfo * const from)
{
  rmw_connextdds_ih_to_gid(from->publication_handle, to->publisher_gid);
// Message timestamps are disabled on Windows because RTI Connext DDS
// does not support a high enough clock resolution by default (see: _ftime()).
#if !RTI_WIN32
  to->source_timestamp = dds_time_to_u64(&from->source_timestamp);
  to->received_timestamp = dds_time_to_u64(&from->reception_timestamp);
#else
  to->source_timestamp = 0;
  to->received_timestamp = 0;
#endif /* !RTI_WIN32 */
```
We need either a workaround for it or a fix in the `rmw_connextdds_message_info_from_dds(..)`.
@asorbini Where is `RTI_WIN32` defined? I can't find it. I am curious if we could use it in the rosbag2 or maybe if would need to create the same definition in the rosbag2 to what should we map?

jmachowinski:
@MichaelOrlov Should I add a workaround ? I would go with an ifdef in the recorder.cpp that would disable this feature if RTI_WIN32 is set.

MichaelOrlov:
@jmachowinski Yes, please. However, I am not sure that `RTI_WIN32 ` will be visble/(ever defined) on rosbag2 level.
Anyway, let's give it a try.

Please note that we need to fallback to the old behavior in recorded if `RTI_WIN32 ` is defined not only in test.

jmachowinski:
@MichaelOrlov added a workaround, that disables this feature on windows.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,6 +523,20 @@ std::shared_ptr<rclcpp::GenericSubscription>
...
+  return node->create_generic_subscription(
+    topic_name,
+    topic_type,
+    qos,
+    [this, topic_name, topic_type](std::shared_ptr<const rclcpp::SerializedMessage> message,
+    const rclcpp::MessageInfo & mi) {
+      if (!paused_.load()) {
+        writer_->write(
+          std::move(message), topic_name, topic_type, node->now().nanoseconds(),
+          node->now().nanoseconds());
```
MichaelOrlov: According to the MCAP spec, the source timestamp shall be assigned to **0** if not available.

MichaelOrlov (changes requested):
@jmachowinski It turns out that in this workaround we are falling back to the old behavior for all RMWs under the Windows.  
However, we need to make an exclusion only for the RTI Connext DDS.
Perhaps we can use following if construct to detect it
```
 if (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") != 0) {
```

Also, will need to adjust a newly added test for this workaround as well.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,16 +523,49 @@ std::shared_ptr<rclcpp::GenericSubscription>
 RecorderImpl::create_subscription(
   const std::string & topic_name, const std::string & topic_type, const rclcpp::QoS & qos)
 {
-  auto subscription = node->create_generic_subscription(
+#if defined _WIN32 || defined __CYGWIN__
```
MichaelOrlov: @jmachowinski Why `|| defined __CYGWIN__` is needed?

jmachowinski:
I was under the impression, that we don't have any rwm timestamp support on windows, regardless of the DDS implementation. I updated the patch, to only blacklist connextdds. I'll have a look at the test later.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -523,6 +533,23 @@ std::shared_ptr<rclcpp::GenericSubscription>
 RecorderImpl::create_subscription(
   const std::string & topic_name, const std::string & topic_type, const rclcpp::QoS & qos)
 {
+#ifdef _WIN32
+  if (!rmw_has_time_stamping_support_) {
```
MichaelOrlov: @jmachowinski We are going to use this variable only once and only in this place.  
Could you please replace it with the explicit condition? 
```
 if (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") != std::string::npos) {
```

ros-discourse:

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @jmachowinski Could you please address CI failures and my comments?
We need to merge this PR ASAP this week if we want it to be landed on the LTS Jazzy Jalisco release.

jmachowinski: ohh sorry, totally forgot about this PR, I was busy fixing race conditions in the executor for jazzy. I will try to address this tomorrow.

jmachowinski: Rebased, and changed as requested.

MichaelOrlov (approved):
@jmachowinski Thanks! LGTM.

MichaelOrlov (changes requested):
@jmachowinski Sorry. Completely forgot.
Can you please regenerate Python stub files (.pyi)?  
The instruction is in the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/7fafb68b299a4679142305c2a3cf6c99/raw/02674aa34517082806b39b18fdf48b8cbb37791d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_performance_benchmarking rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_performance_benchmarking rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13673
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21510/)

MichaelOrlov (approved):
LGTM with green CI

PR #1536, opened 2024-01-11
resolve conflict.

fujitatomoya:


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -723,7 +113,9 @@ Recorder::Recorder(
 Recorder::~Recorder()
 {
   keyboard_handler_->delete_key_press_callback(toggle_paused_key_callback_handle_);
-  stop();
+  if (!subscriptions_.empty()) {
```
fujitatomoya, line 116: to keep the ABI compatibility, we cannot add member variables for `Recorder` class. instead, we could use `subscriptions_` map to see if that is NOT empty? or `event_publisher_thread_should_exit_` (probably not this one, since this will never be back to `false` once it becomes `true`?) ? any thoughts?

fujitatomoya (commented):
CC: @MichaelOrlov

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @fujitatomoya I think it would be more appropriate and safe to use `event_publisher_thread_should_exit_` for this purpose. I have pushed a fix to the original backporting PR https://github.com/ros2/rosbag2/pull/1535.
You can close this PR.

fujitatomoya: @MichaelOrlov thanks!

PR #1538, opened 2024-01-14
Add topic_id returned by storage to the TopicMetadata

MichaelOrlov:
- Relates #1553
- Rationale. To be able to distinguish topics by unique topic ID rather than by topic name in the future.
Motivation is the same as in #1553 
> 1. Users complain multiple times for instance here https://github.com/ros2/rosbag2/pull/1531#issuecomment-1884435993 that rosbag2 stores messages in internal buffers in a very inefficient way and that memory footprint for the same data payload could be reduced down to the factor of the **2** or **3** by using `topic_id` represented as integer type instead of full `topic_name` represented as a string for each message.
This is very meaningful when using the snapshot feature and recording on the host HW with limited resources.
> 2. Another motivating factor for using `topic_id` instead of the `topic_name` in internal data structures and buffers is the ability to differentiate different publishers with the same topic name. It might be a situation for instance when multiple publishers exist on the same topic but have different QoS settings. Another use case could be when multiple publishers exist on the same topic but with different versions of the type definitions. 
Currently, we don't support these use cases in the rosbag2. However, if we will differentiate topics by topic id in inner rosbag2 representation on all layers we would be able to add support for the aforementioned use cases in the future.

About the size of the type for the newly added `topic_id` field.
In MCAP specification we reserved 16 bit for the topic_id. This is the `ChannelId`, the MCAP inner type that corresponds to the `uint16_t`. When we were creating the MCAP format we decided that the ability to differentiate a maximum `65535` topics would be enough in the vast majority of cases. However, in the `SQLite3` storage plugin, we are using `topic_id` as an index to map records from the `messages` DB table to the corresponding `topics` table. According to the SQLite3 specification, the index is the `int64_t` data type and can't be changed.
~~Therefore the type of the newly added `topic_id` filed shall be no less than `int64_t` and shall be with the sign.~~
Update:
Decided to keep `uint16_t` for `topic_id` in the `TopicMetada` and have a hash map to `int64_t` indexes inside the sql storage plugin. The performance will not be affected during recording because currently we already have an internal hash map lookup to find match for `topic_name` to the `topic_id` index. It will be similar hash map lookup, however we will be using external `topic_id` instead of `topic_name`.

### DIFF
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -441,6 +441,7 @@ void MCAPStorage::read_metadata()
 
     // Create a TopicInformation for this topic
     rosbag2_storage::TopicInformation topic_info{};
+    topic_info.topic_metadata.id = channel.id;
     topic_info.topic_metadata.name = channel.topic;
     topic_info.topic_metadata.serialization_format = channel.messageEncoding;
     topic_info.topic_metadata.type = schema_ptr->name;
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
@@ -115,6 +115,10 @@ public:
 
 private:
   void initialize();
+  void add_topic_to_metadata(
+    int64_t topic_id, std::string topic_name, std::string topic_type, std::string ser_format,
+    int64_t msg_count, const std::string & offered_qos_profiles_str,
+    const std::string & type_hash);
   void read_metadata();
   void prepare_for_writing();
   void prepare_for_reading();
@@ -138,8 +142,8 @@ private:
   ReadQueryResult message_result_ {nullptr};
   ReadQueryResult::Iterator current_message_row_ {
     nullptr, SqliteStatementWrapper::QueryResult<>::Iterator::POSITION_END};
-  std::unordered_map<std::string, int> topics_ RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
-  std::unordered_map<std::string, int> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
+  std::unordered_map<std::string, int64_t> topics_ RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
+  std::unordered_map<std::string, int64_t> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
     db_read_write_mutex_);
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
@@ -148,7 +148,7 @@ private:
   std::unordered_map<std::string, int64_t> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
     db_read_write_mutex_);
   std::unordered_map<int64_t, uint16_t> inner_to_extern_topic_id_map_;
-  uint16_t last_extern_topic_id_ = 0;  // 0 corresponds to the invalid topic_id
+  std::atomic<uint16_t> last_extern_topic_id_{0};  // 0 corresponds to the invalid topic_id
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
@@ -116,7 +116,7 @@ public:
 private:
   void initialize();
   void add_topic_to_metadata(
-    int64_t topic_id, std::string topic_name, std::string topic_type, std::string ser_format,
+    int64_t inner_topic_id, std::string topic_name, std::string topic_type, std::string ser_format,
     int64_t msg_count, const std::string & offered_qos_profiles_str,
     const std::string & type_hash);
   void read_metadata();
@@ -131,6 +131,8 @@ private:
   int read_db_schema_version();
   uint64_t get_page_size() const;
   uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_);
+  uint16_t get_extern_topic_id(int64_t inner_topic_id) const;
+  uint16_t get_or_generate_extern_topic_id(int64_t inner_topic_id);
 
   using ReadQueryResult = SqliteStatementWrapper::QueryResult<
     std::shared_ptr<rcutils_uint8_array_t>, rcutils_time_point_value_t, std::string, int>;
@@ -145,6 +147,8 @@ private:
   std::unordered_map<std::string, int64_t> topics_ RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
   std::unordered_map<std::string, int64_t> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
     db_read_write_mutex_);
+  std::unordered_map<int64_t, uint16_t> inner_to_extern_topic_id_map_;
+  uint16_t last_extern_topic_id_ = 0;  // 0 corresponds to the invalid topic_id
   std::vector<rosbag2_storage::TopicMetadata> all_topics_and_types_;
   std::string relative_path_;
   std::atomic_bool active_transaction_ {false};
--- a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
+++ b/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
@@ -611,48 +611,47 @@ void SqliteStorage::fill_topics_and_types()
   if (database_->field_exists("topics", "offered_qos_profiles")) {
     if (database_->field_exists("topics", "type_description_hash")) {
       auto statement = database_->prepare_statement(
-        "SELECT name, type, serialization_format, offered_qos_profiles, type_description_hash"
+        "SELECT id, name, type, serialization_format, offered_qos_profiles, type_description_hash"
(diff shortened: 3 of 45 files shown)


MichaelOrlov:
Moving it to the draft again.
Going to implement proposal form the https://github.com/ros2/rosbag2/issues/1553#issuecomment-1914547903
Will keep **uint16_t** for `topic_id` in the `TopicMetada` and have a hash map to **int64_t** indexes inside the sql storage plugin.

Motivation:
Well, on one hand, it will require keeping a hash map with two types of IDs in the SQLite3 storage plugin side and having a hash lookup operation for each message read/write. The hash map is not very memory efficient but still could take less memory footprint rather than keeping an extra **6** bytes per message. Especially when we are having a lot of high-frequency messages like IMU. Another drawback of this is a little bit of performance hit for lookup for inner ID in the hash map for each message write.
However, on the other hand, we probably can sacrifice a bit of performance penalty on the SQLite3 plugin side for the sake of a better memory footprint.
Please note that MCAP storage plugin will not be affected and users will have benefits using rosbag2 with it.

MichaelOrlov:
Ready for review

fujitatomoya:
@MichaelOrlov thanks for the PR. this one looks big change, i will try review in this week. you can go ahead to assign this to me.

MichaelOrlov:
@fujitatomoya Thanks for helping with review.
While it looks like a big PR with a lot of changes although, it is because it affects a lot of tests where we need to add `0` or `1` for initializing the newly added field to the TopicMetadata. The most of the changes where need to look at and review are in the `sqlite3_storage` package (there a few files)

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -434,6 +434,7 @@ void MCAPStorage::read_metadata()
 
     // Create a TopicInformation for this topic
     rosbag2_storage::TopicInformation topic_info{};
+    topic_info.topic_metadata.id = channel.id;
```
fujitatomoya, line 444: So this [channel.id](https://mcap.dev/spec#channel-op0x04) is generated by MCAP, right? Since i am not familiar with mcap library, i would like to check if i understand correctly.

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
```diff
@@ -138,9 +144,11 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   ReadQueryResult message_result_ {nullptr};
   ReadQueryResult::Iterator current_message_row_ {
     nullptr, SqliteStatementWrapper::QueryResult<>::Iterator::POSITION_END};
-  std::unordered_map<std::string, int> topics_ RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
-  std::unordered_map<std::string, int> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
+  std::unordered_map<std::string, int64_t> topics_ RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
+  std::unordered_map<std::string, int64_t> msg_definitions_ RCPPUTILS_TSA_GUARDED_BY(
     db_read_write_mutex_);
+  std::unordered_map<int64_t, uint16_t> inner_to_extern_topic_id_map_;
+  uint16_t last_extern_topic_id_ = 0;  // 0 corresponds to the invalid topic_id
```
fujitatomoya: how about the following?

```suggestion
  std::atomic<uint16_t> last_extern_topic_id_(0);  // 0 corresponds to the invalid topic_id
```

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -908,6 +915,32 @@ uint64_t SqliteStorage::get_page_size() const
...
+{
+  uint16_t extern_topic_id = get_extern_topic_id(inner_topic_id);
+  if (extern_topic_id == 0) {
+    if (last_extern_topic_id_ == std::numeric_limits<uint16_t>::max()) {
+      ROSBAG2_STORAGE_DEFAULT_PLUGINS_LOG_ERROR_STREAM(
+        "External topic_id reached maximum allowed value" <<
+          std::to_string(std::numeric_limits<uint16_t>::max()));
+      throw std::range_error("External topic_id reached maximum allowed value");
+    }
+    extern_topic_id = ++last_extern_topic_id_;
```
fujitatomoya:
```suggestion
    last_extern_topic_id_.fetch_add(1, std::memory_order_relaxed);
    extern_topic_id = last_extern_topic_id_;
```

fujitatomoya (commented):
overall lgtm, a few comments.

rosbag2_storage_mcap/src/mcap_storage.cpp
MichaelOrlov: Yes. The channel in the mcap is sort of more generic entity decribing data stream as topic in ROS.
We creating a new channel in `create_topic(..) API`
https://github.com/ros2/rosbag2/blob/90d1da835d23965891f7d3e6bbdf24e23fe0602b/rosbag2_storage_mcap/src/mcap_storage.cpp#L830-L843
And `mcap_writer_->addChannel(channel);` actually assigning a new `channel.id`
https://github.com/foxglove/mcap/blob/4d62967abe95e755b91f16cce390f6560d18e07e/cpp/mcap/include/mcap/writer.hpp#L363-L370
```
   * @brief Add a new channel to the MCAP file and set `channel.id` to a
   * generated channel id. The channel id is used when adding messages to the
   * file.
   *
   * @param channel Description of the channel to register. The `id` value is
   *   ignored and will be set to a generated channel id.
   */
  void addChannel(Channel& channel);
```
The implementation is very simple
https://github.com/foxglove/mcap/blob/4d62967abe95e755b91f16cce390f6560d18e07e/cpp/mcap/include/mcap/writer.inl#L493-L496
```
void McapWriter::addChannel(Channel& channel) {
  channel.id = uint16_t(channels_.size() + 1);
  channels_.push_back(channel);
}
```

fujitatomoya (approved):

MichaelOrlov:
@emersonknapp @clalancette Could some of you please give formal approval for this PR to be able to merge it? 
@fujitatomoya already reviewed it.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/9d2827dcef7e219879574e49386f1596/raw/16f993bb8bcfd59af7ac24dbeaa4bec51ead6950/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_examples rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_examples rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13237
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20225/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14642/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20943/)

emersonknapp (approved):
lgtm

MichaelOrlov:
Re-run CI with the correctly listed `rosbag2_examples_cpp` and, `rosbag2_examples_py` instead `rosbag2_examples` package.
Gist: https://gist.githubusercontent.com/MichaelOrlov/7b5b5d317f83b855c612ecd3a833a50e/raw/16f993bb8bcfd59af7ac24dbeaa4bec51ead6950/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_examples_cpp rosbag2_examples_py rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_examples_cpp rosbag2_examples_py rosbag2_py rosbag2_storage rosbag2_storage_mcap  rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13240
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20228/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14645/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20946/)

DLu:
There's not much information in the [migration notes](https://docs.ros.org/en/jazzy/Releases/Release-Jazzy-Jalisco.html#added-topic-id-returned-by-storage-to-the-topicmetadata), and all the examples seem to just use the value 0. So if I'm writing a bag file from scratch ([using Python say](https://docs.ros.org/en/rolling/Tutorials/Advanced/Recording-A-Bag-From-Your-Own-Node-Py.html#record-synthetic-data-from-a-node)), should each `TopicMetaData` i construct use `id=0`? 

https://github.com/MetroRobots/classic_bags/issues/16

MichaelOrlov:
The `topic_id` shall be unique for each topic. The numbering is implementation details for storage plugins. 
In the following example
```python
        topic_info = rosbag2_py._storage.TopicMetadata(
            id=0,
            name='synthetic',
            type='example_interfaces/msg/Int32',
            serialization_format='cdr')
        self.writer.create_topic(topic_info)
```
it is assigned to 0 because it will be overwritten by the storage plugin after a call to the `writer.create_topic(topic_info)`
You can get assigned value by checking `topic_info.id` after `writer.create_topic(topic_info)`.

Therefore - no special steps for migration are required.

DLu:
I guess what I'm trying to figure out is why zero is not the default value, if it is just going to be overwritten. Then the TopicMetadata constructor could have stayed the same.

MichaelOrlov:
Well, it is zero-initialized by default. The `TopicMetadata` is a struct and doesn't have any explicit constructors. 
However, the constructor could be added.
As always, PRs with improvements are welcome!

DLu:
In the Python bindings introduced by @r7vme in #1569, there **is** an [explicit constructor](https://github.com/ros2/rosbag2/blob/055935d33dd2fed2772657c9dc1f2173eaa7f752/rosbag2_py/rosbag2_py/_storage.pyi#L128), and `id` is first argument. However, I will admit I'm not fully familiar with "stubs" or the `...` notation.

PR #1539, opened 2024-01-15
Add default initialization for CompressionOptions

arneboe:
This PR adds sane defaults for the `CompressionOptions` struct.
Sane defaults are important because the struct does not have a constructor and it is possible (and very easy) to use it without completely initializing it.

Another option would be to add a constructor and delete the default constructor, but that would not be backward compatible.

### DIFF
--- a/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
@@ -62,7 +62,7 @@ struct CompressionOptions
   CompressionMode compression_mode{CompressionMode::NONE};
   uint64_t compression_queue_size{0};
   /// \brief // The number of compression threads
-  uint64_t compression_threads{5};
+  uint64_t compression_threads{0};
   /// \brief If set, the compression thread(s) will try to set the given priority for itself
   /// For Windows the valid values are: THREAD_PRIORITY_LOWEST, THREAD_PRIORITY_BELOW_NORMAL and
   /// THREAD_PRIORITY_NORMAL. For POSIX compatible OSes this is the "nice" value.
--- a/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
+++ b/rosbag2_compression/include/rosbag2_compression/compression_options.hpp
@@ -58,11 +58,11 @@ ROSBAG2_COMPRESSION_PUBLIC std::string compression_mode_to_string(CompressionMod
  */
 struct CompressionOptions
 {
-  std::string compression_format;
-  CompressionMode compression_mode;
-  uint64_t compression_queue_size;
+  std::string compression_format{};
+  CompressionMode compression_mode{CompressionMode::NONE};
+  uint64_t compression_queue_size{0};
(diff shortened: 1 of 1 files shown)


rosbag2_compression/include/rosbag2_compression/compression_options.hpp
```diff
@@ -58,11 +58,11 @@ ROSBAG2_COMPRESSION_PUBLIC std::string compression_mode_to_string(CompressionMod
...
 {
-  std::string compression_format;
-  CompressionMode compression_mode;
-  uint64_t compression_queue_size;
+  std::string compression_format{};
+  CompressionMode compression_mode{CompressionMode::NONE};
+  uint64_t compression_queue_size{0};
   /// \brief // The number of compression threads
-  uint64_t compression_threads;
+  uint64_t compression_threads{5};
```
fujitatomoya: how about putting `0` instead? that will set up with `std::thread::hardware_concurrency()`?

```suggestion
  uint64_t compression_threads{0};
```

fujitatomoya (changes requested):
minor comment, but needs to be resolved.

CC: @MichaelOrlov

rosbag2_compression/include/rosbag2_compression/compression_options.hpp
MichaelOrlov: I agree with @fujitatomoya better way to put **0** and we will override it with HW threads available here
https://github.com/ros2/rosbag2/blob/317286c98472205593d5b0a791d06fea42613ba3/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp#L176-L183

MichaelOrlov (changes requested):
@arneboe  I have a minor concern about initial value for number of threads. 
Also please fix DCO. See details in the https://github.com/ros2/rosbag2/pull/1539/checks?check_run_id=20499622275

arneboe:
> @arneboe I have a minor concern about initial value for number of threads. Also please fix DCO. See details in the https://github.com/ros2/rosbag2/pull/1539/checks?check_run_id=20499622275

Done ;)

MichaelOrlov:

MichaelOrlov (approved):
LGTM

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/d2aeb87a86a3fc667fbe88950d5f1ea8/raw/fd15696f0f6d091943f8f91903f39eb3682df54b/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression
TEST args:  --packages-above rosbag2_compression
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13144
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20104/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14546/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20819/)

MichaelOrlov:

PR #1547, opened 2024-01-22
Add transactional state mutex for RecorderImpl class.

fujitatomoya:
address https://github.com/ros2/rosbag2/issues/1529#issuecomment-1884132718

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -137,7 +137,7 @@ private:
   rclcpp::Service<rosbag2_interfaces::srv::Snapshot>::SharedPtr srv_snapshot_;
   rclcpp::Service<rosbag2_interfaces::srv::SplitBagfile>::SharedPtr srv_split_bagfile_;
 
-  std::recursive_mutex state_transaction_mutex_;
+  std::recursive_mutex state_transition_mutex_;
   std::atomic<bool> stop_discovery_ = false;
   std::atomic<bool> paused_ = false;
   std::atomic<bool> in_recording_ = false;
@@ -220,32 +220,32 @@ RecorderImpl::~RecorderImpl()
 
 void RecorderImpl::stop()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
-  stop_discovery();
-
-  if (in_recording_ == false) {
-    RCLCPP_DEBUG(node->get_logger(), "Recording has already stopped.");
-  } else {
-    pause();
-    subscriptions_.clear();
-    writer_->close();  // Call writer->close() to finalize current bag file and write metadata
-
-    {
-      std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
-      event_publisher_thread_should_exit_ = true;
-    }
-    event_publisher_thread_wake_cv_.notify_all();
-    if (event_publisher_thread_.joinable()) {
-      event_publisher_thread_.join();
-    }
-    in_recording_ = false;
-    RCLCPP_INFO(node->get_logger(), "Recording stopped");
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
+  if (!in_recording_) {
+    RCLCPP_DEBUG(node->get_logger(), "Recording has already been stopped or not running.");
+    return;
   }
+
+  stop_discovery();
+  pause();
+  subscriptions_.clear();
+  writer_->close();  // Call writer->close() to finalize current bag file and write metadata
+
+  {
+    std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
+    event_publisher_thread_should_exit_ = true;
+  }
+  event_publisher_thread_wake_cv_.notify_all();
+  if (event_publisher_thread_.joinable()) {
+    event_publisher_thread_.join();
+  }
+  in_recording_ = false;
+  RCLCPP_INFO(node->get_logger(), "Recording stopped");
 }
 
 void RecorderImpl::record()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
   if (in_recording_.exchange(true)) {
     RCLCPP_WARN_STREAM(
       node->get_logger(),
@@ -395,9 +395,9 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
 
 void RecorderImpl::pause()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
-  if (paused_ == true) {
-    RCLCPP_DEBUG(node->get_logger(), "Recording has already paused.");
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
+  if (paused_) {
+    RCLCPP_DEBUG(node->get_logger(), "Recorder is alreadyin pause state.");
   } else {
     paused_.store(true);
     RCLCPP_INFO_STREAM(node->get_logger(), "Pausing recording.");
@@ -406,9 +406,9 @@ void RecorderImpl::pause()
 
 void RecorderImpl::resume()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
-  if (paused_ == false) {
-    RCLCPP_DEBUG(node->get_logger(), "Recording has already resumed.");
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
+  if (!paused_) {
+    RCLCPP_DEBUG(node->get_logger(), "Already in the recording.");
   } else {
     paused_.store(false);
     RCLCPP_INFO_STREAM(node->get_logger(), "Resuming recording.");
@@ -417,7 +417,7 @@ void RecorderImpl::resume()
 
 void RecorderImpl::toggle_paused()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
   if (paused_.load()) {
     this->resume();
   } else {
@@ -432,9 +432,11 @@ bool RecorderImpl::is_paused()
 
 void RecorderImpl::stop_discovery()
 {
-  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
-  if (stop_discovery_ == true) {
-    RCLCPP_DEBUG(node->get_logger(), "Recorder Topic Discovery has already stopped.");
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
+  if (stop_discovery_) {
+    RCLCPP_DEBUG(
+      node->get_logger(),
+      "Recorder topic discovery has already been stopped or not running.");
   } else {
     stop_discovery_ = true;
     if (discovery_future_.valid()) {
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -79,6 +79,9 @@ public:
   /// Return the current paused state.
   bool is_paused();
 
+  /// Stop discovery
+  void stop_discovery();
+
   std::unordered_map<std::string, std::string> get_requested_or_available_topics();
 
   /// Public members for access by wrapper
@@ -86,7 +89,6 @@ public:
   std::shared_ptr<rosbag2_cpp::Writer> writer_;
   rosbag2_storage::StorageOptions storage_options_;
   rosbag2_transport::RecordOptions record_options_;
-  std::atomic<bool> stop_discovery_ = false;
   std::unordered_map<std::string, std::shared_ptr<rclcpp::SubscriptionBase>> subscriptions_;
 
 private:
@@ -135,6 +137,8 @@ private:
   rclcpp::Service<rosbag2_interfaces::srv::Snapshot>::SharedPtr srv_snapshot_;
   rclcpp::Service<rosbag2_interfaces::srv::SplitBagfile>::SharedPtr srv_split_bagfile_;
 
+  std::recursive_mutex state_transaction_mutex_;
+  std::atomic<bool> stop_discovery_ = false;
   std::atomic<bool> paused_ = false;
   std::atomic<bool> in_recording_ = false;
   std::shared_ptr<KeyboardHandler> keyboard_handler_;
@@ -160,8 +164,8 @@ RecorderImpl::RecorderImpl(
 : writer_(std::move(writer)),
   storage_options_(storage_options),
   record_options_(record_options),
-  stop_discovery_(record_options_.is_discovery_disabled),
   node(owner),
+  stop_discovery_(record_options_.is_discovery_disabled),
   paused_(record_options.start_paused),
   keyboard_handler_(std::move(keyboard_handler))
 {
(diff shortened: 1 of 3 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -160,8 +164,8 @@ RecorderImpl::RecorderImpl(
 : writer_(std::move(writer)),
   storage_options_(storage_options),
   record_options_(record_options),
-  stop_discovery_(record_options_.is_discovery_disabled),
   node(owner),
+  stop_discovery_(record_options_.is_discovery_disabled),
```
fujitatomoya: this is required to avoid `recorder` warning since construction order has been changed.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +395,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
...
+  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  if (paused_ == true) {
+    RCLCPP_DEBUG(node->get_logger(), "Recording has already paused.");
+  } else {
+    paused_.store(true);
+    RCLCPP_INFO_STREAM(node->get_logger(), "Pausing recording.");
+  }
 }
 
 void RecorderImpl::resume()
```
fujitatomoya, line 406: this can be implemented as `pause(false)` to avoid the implementation redundancy. any thoughts?

fujitatomoya:
@MichaelOrlov minor fix, could you take a look when you have time?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @fujitatomoya Well. From one hand yes it could be simplified with `pause(false)` but we deliberately choose to use `pause()` and `resume()` methods to have a clean and more obvious API for managing rosbag2 recorder state from remote via service calls and during distributed recording.  
This is to facilitate `Pause`, `Resume` and `IsPaused` service calls. 
In addition IMO `pause(false)` is not very clear what does it really mean those `false`.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -135,6 +137,8 @@ class RecorderImpl
   rclcpp::Service<rosbag2_interfaces::srv::Snapshot>::SharedPtr srv_snapshot_;
   rclcpp::Service<rosbag2_interfaces::srv::SplitBagfile>::SharedPtr srv_split_bagfile_;
 
+  std::recursive_mutex state_transaction_mutex_;
```
MichaelOrlov:
```suggestion
  std::mutex state_transition_mutex_;
```
IMO `state_transition_mutex` name better fit for its purpose

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -211,41 +215,37 @@ RecorderImpl::RecorderImpl(
...
+    {
+      std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
+      event_publisher_thread_should_exit_ = true;
+    }
+    event_publisher_thread_wake_cv_.notify_all();
+    if (event_publisher_thread_.joinable()) {
+      event_publisher_thread_.join();
+    }
+    in_recording_ = false;
+    RCLCPP_INFO(node->get_logger(), "Recording stopped");
```
MichaelOrlov: Suggesting following simplification in logic and for the purpose of reducing delta
```cpp
void RecorderImpl::stop()
{
  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
  if (!in_recording_) {
    RCLCPP_DEBUG(node->get_logger(), "Recording has already been stopped or not running.");
    return;
  }

  stop_discovery();
  pause();
  subscriptions_.clear();
  writer_->close();  // Call writer->close() to finalize current bag file and write metadata

  {
    std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
    event_publisher_thread_should_exit_ = true;
  }
  event_publisher_thread_wake_cv_.notify_all();
  if (event_publisher_thread_.joinable()) {
    event_publisher_thread_.join();
  }
  in_recording_ = false;
  RCLCPP_INFO(node->get_logger(), "Recording stopped");
}
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +395,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
 
 void RecorderImpl::pause()
 {
-  paused_.store(true);
-  RCLCPP_INFO_STREAM(node->get_logger(), "Pausing recording.");
+  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  if (paused_ == true) {
+    RCLCPP_DEBUG(node->get_logger(), "Recording has already paused.");
+  } else {
```
MichaelOrlov:
```suggestion
  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
  if (paused_) {
    RCLCPP_DEBUG(node->get_logger(), "Recording already in pause state.");
  } else {
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +395,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
...
 }
 
 void RecorderImpl::resume()
 {
-  paused_.store(false);
-  RCLCPP_INFO_STREAM(node->get_logger(), "Resuming recording.");
+  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  if (paused_ == false) {
+    RCLCPP_DEBUG(node->get_logger(), "Recording has already resumed.");
+  } else {
```
MichaelOrlov:
```suggestion
  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
  if (!paused_) {
    RCLCPP_DEBUG(node->get_logger(), "Already in the recording.");
  } else {
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -419,6 +430,26 @@ bool RecorderImpl::is_paused()
   return paused_.load();
 }
 
+void RecorderImpl::stop_discovery()
+{
+  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  if (stop_discovery_ == true) {
+    RCLCPP_DEBUG(node->get_logger(), "Recorder Topic Discovery has already stopped.");
+  } else {
+    stop_discovery_ = true;
```
MichaelOrlov:
```suggestion
 if (stop_discovery_) {
    RCLCPP_DEBUG(
      node->get_logger(),
      "Recorder topic discovery has already been stopped or not running.");
  } else {
    stop_discovery_ = true;
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -419,6 +430,26 @@ bool RecorderImpl::is_paused()
   return paused_.load();
 }
 
+void RecorderImpl::stop_discovery()
+{
+  std::lock_guard<std::recursive_mutex> lock(state_transaction_mutex_);
+  if (stop_discovery_ == true) {
```
MichaelOrlov: @fujitatomoya I don't see a reason or rationale for protecting discovery start/stop operations with mutex. 
After some analysis, I came to the conclusion it would be enough to have an atomic `stop_discovery_` variable. 
I don't see possibilities for race conditions or undefined behavior in discovery start/stop operations.

If you agree please remove mutex lock here and use regular **non**-recursive mutex for the `state_transition_mutex_`.

MichaelOrlov (changes requested):
@fujitatomoya Thanks for your contribution.
I have a few nitpick findings and suggestions and also have a concern about using recursive_mutex for protecting state transition.

I don't see a reason or rationale for protecting discovery start/stop operations with mutex at all. After some analysis, I came to the conclusion it would be enough to have an atomic `stop_discovery_` variable.
I don't see possibilities for race conditions or undefined behavior in discovery start/stop operations even if it would be a public API and could be called from different threads.

If you agree please remove mutex lock in `stop_discovery()` method and use regular **non**-recursive mutex for the `state_transition_mutex_`.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
fujitatomoya: okay, that makes sense too. resolving.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
fujitatomoya: i was thinking that if multiple threads calling on `RecorderImpl::stop_discovery`, that would be inconsistent behavior since this does `read &  write` operation on `stop_discovery_`. that is the main reason to move `stop_discovery_` in the private scope to protect this variable with mutex lock.

e.g) 10 threads possibly call on `discovery_future_.wait_for` that leads to the timeout except one.

if that is acceptable, i am okay with having `std::mutex`. what do you think?
i understand your concern is recursive lock generates more complication here, so i am okay with either way.

fujitatomoya: @MichaelOrlov only this comment is alive, could you kindly check?

fujitatomoya (commented):
@MichaelOrlov sorry, i did not send this to you. cached for like a week in my browser... 😓

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Hi @fujitatomoya 
For the case you described 
> e.g) 10 threads possibly call on discovery_future_.wait_for that leads to the timeout except one.

I think that protection with ` if (stop_discovery_.exchange(true)) {` would be enough?
i.e. It would look like that without locking the mutex
```cpp
void RecorderImpl::stop_discovery()
{
  if (stop_discovery_.exchange(true)) {
    RCLCPP_DEBUG(
      node->get_logger(),
      "Recorder topic discovery has already been stopped or not running.");
  } else {
    if (discovery_future_.valid()) {
      auto status = discovery_future_.wait_for(2 * record_options_.topic_polling_interval);
      if (status != std::future_status::ready) {
        RCLCPP_ERROR_STREAM(
          node->get_logger(),
          "discovery_future_.wait_for(" << record_options_.topic_polling_interval.count() <<
            ") return status: " <<
            (status == std::future_status::timeout ? "timeout" : "deferred"));
      }
    }
  }
}
```
For other cases when we are calling `stop_discovery()` from inner class members (RecorderImpl::stop()) it is already protected with mutex.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -211,24 +215,19 @@ RecorderImpl::RecorderImpl(
...
-  stop_discovery_ = true;
-  if (discovery_future_.valid()) {
-    auto status = discovery_future_.wait_for(2 * record_options_.topic_polling_interval);
-    if (status != std::future_status::ready) {
-      RCLCPP_ERROR_STREAM(
-        node->get_logger(),
-        "discovery_future_.wait_for(" << record_options_.topic_polling_interval.count() <<
-          ") return status: " << (status == std::future_status::timeout ? "timeout" : "deferred"));
-    }
+  std::lock_guard<std::recursive_mutex> lock(state_transition_mutex_);
```
MichaelOrlov: @fujitatomoya, The Clang-Tidy in CLion, giving a warning on this line with `lock.`
```
Previous declaration is here recorder.cpp235:33:
warning: declaration shadows a local variable
```
It would be better to rename `lock` to something like `state_transition_lock`

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: If considering other cases when we can potentially change `stop_discovery_` atomic variable.
We are changing it only in `RecorderImpl::record()` and `RecorderImpl::stop()` or from destructor when calling `RecorderImpl::stop()` in all that scenarious we aleady locking the transitional mutex on the upper layer.

MichaelOrlov: Only possible situation for race condition could be if we would call concurrently from different threads `RecorderImpl::record()` and `RecorderImpl::stop_discovery()`.
In this case, it will be better to create a separate mutex for discovery and wrap start discovery routines to a separate method e.g. `Recorder::start_discovery()`. 
I think it might be useful in the future to link those API to the service calls.

fujitatomoya: @MichaelOrlov thanks!

`stop_discovery_.exchange(true)` should do work!

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +395,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
...
+    RCLCPP_DEBUG(node->get_logger(), "Already in the recording.");
+  } else {
+    paused_.store(false);
+    RCLCPP_INFO_STREAM(node->get_logger(), "Resuming recording.");
+  }
 }
 
 void RecorderImpl::toggle_paused()
 {
+  std::lock_guard<std::recursive_mutex> state_lock(state_transition_mutex_);
```
fujitatomoya: > a concern about using recursive_mutex for protecting state transition.

`resume` or `pause` would be called via another thread, so without having the mutex lock in them recursively, the behavior is not gonna be transactional. for example, this thread comes in here with `state_transition_mutex_` and check it is paused, then another thread call `resume` without `state_transition_mutex_`, then this `resume` does print `Already in the recording.`.

if this is acceptable, we can remove recursive mutex lock, i think. what do you think? @MichaelOrlov

MichaelOrlov: @fujitatomoya I don't think that we need to remove the mutex lock at all in resume and toggle. But rather chnage it to the non-recursive mutex.

I will do a prototype.

MichaelOrlov: @fujitatomoya See my new commit https://github.com/ros2/rosbag2/pull/1547/commits/4fba7a49ebe3993a6c0d2e86deb0f41ee8957166 on your branch

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +396,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
 
 void RecorderImpl::pause()
 {
-  paused_.store(true);
-  RCLCPP_INFO_STREAM(node->get_logger(), "Pausing recording.");
+  std::lock_guard<std::mutex> state_lock(state_transition_mutex_);
```
fujitatomoya: deadlock here? if this is called from `RecorderImpl::stop` or `RecorderImpl::toggle`?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -395,18 +396,29 @@ const rosbag2_cpp::Writer & RecorderImpl::get_writer_handle()
...
+    paused_.store(true);
+    RCLCPP_INFO_STREAM(node->get_logger(), "Pausing recording.");
+  }
 }
 
 void RecorderImpl::resume()
 {
-  paused_.store(false);
-  RCLCPP_INFO_STREAM(node->get_logger(), "Resuming recording.");
+  std::lock_guard<std::mutex> state_lock(state_transition_mutex_);
```
fujitatomoya: deadlock here if this is called from `RecorderImpl::toggle_paused`?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @fujitatomoya Sorry about the deadlocks It was premature and I was a bit in a rush.  
From the last round of review, I was under the impression that we don't really need a mutex lock in such operations as `pause`, `resume` and ~~`toggle_paused`~~ since there is only one atomic variable change inside.  
It seems I have forgotten to rework it.

fujitatomoya: @MichaelOrlov yeah, that is also fine to me if we have agreement on that.

something that i was concerned about that is, https://github.com/ros2/rosbag2/pull/1547#discussion_r1525624065. as you think, probably this is not a problem, so maybe we can remove mutex from `pause` and `resume`?

MichaelOrlov: @fujitatomoya I think that ultimately currently we shall protect with mutex only transition between start and stop operation since these operations are not atomic and could cause undefined behavior. All other operations as pause, resume and toggle_paused are atomic operations that change internally only one atomic variable.

I've fix deadlocks in pause(), resume() and toggle_paused() methods see my last commit https://github.com/ros2/rosbag2/pull/1547/commits/d7880740458fa0d96ffb9f8b0185130fdae286e0
- Removed mutex lock in pause(), resume() and toggle_paused() methods
and use only atomic value change for `paused_` variable.
- Renamed `state_transition_mutex_` to the`start_stop_transition_mutex_` because it really intends to protect only those two operations
- Add unit test for `toggle_paused()`.

fujitatomoya (commented):
@MichaelOrlov thanks for the fix, lgtm with green CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20493/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14895/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21242/)

MichaelOrlov (approved):

MichaelOrlov:
@fujitatomoya There is a warning for my "hacky" cast part in the `toggle_paused()`
```
/home/morlov/ros2_rolling/src/ros2/rosbag2/rosbag2_transport/src/rosbag2_transport/recorder.cpp:417:24: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  417 |   if (atomic_fetch_xor(reinterpret_cast<std::atomic_uchar *>(&paused_), 1)) {
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Unfortunately, I was not able to find a way how to fix this warning in a good way without changing the type of the `paused_` variable to the `std::atomic_uchar` or without rolling back to the simple `if (paused_.load()) {..} else {..}` statement.  

I've tried to change the type to the `std::atomic_uchar` and seems backward implicit type conversion from `bool` to the `std::atomic_uchar` works fine without warnings and we can continue using `paused_` as a boolean variable. See my latest commit https://github.com/ros2/rosbag2/pull/1547/commits/9241cc598fe2e0fcad9e7e58b39203d3608123d2

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/5f018e1125c75305282285cfd52d9352/raw/6c7bfb3b37e6ad32d4e0d2a521a41a02f5364390/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13483
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20495/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14897/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21244/)

fujitatomoya:
@MichaelOrlov green light, we are good to go.

fujitatomoya:
@MichaelOrlov thanks for the help!

MichaelOrlov:
@fujitatomoya Likewise! :wink:

PR #1552, opened 2024-01-24
Remove an incorrect TSA annotation.

clalancette:
read_total_page_count_locked is not protected by the mutex in question, so remove the annotation.

This fixes a warning when building under clang: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1776/clang/new/source.5173c086-9b9f-4319-96c5-67a99529b3cd/#129 .

This is followup from #1516 .  @MichaelOrlov FYI

### DIFF
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
@@ -126,7 +126,7 @@ private:
   int get_last_rowid();
   int read_db_schema_version();
   uint64_t get_page_size() const;
-  uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
+  uint64_t read_total_page_count_locked() const;
 
   using ReadQueryResult = SqliteStatementWrapper::QueryResult<
     std::shared_ptr<rcutils_uint8_array_t>, rcutils_time_point_value_t, std::string, int>;
--- a/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
+++ b/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
@@ -126,7 +126,7 @@ private:
   int get_last_rowid();
   int read_db_schema_version();
   uint64_t get_page_size() const;
-  uint64_t read_total_page_count_locked() const;
+  uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_);
 
   using ReadQueryResult = SqliteStatementWrapper::QueryResult<
     std::shared_ptr<rcutils_uint8_array_t>, rcutils_time_point_value_t, std::string, int>;


rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
```diff
@@ -126,7 +126,7 @@ class ROSBAG2_STORAGE_DEFAULT_PLUGINS_PUBLIC SqliteStorage
   int get_last_rowid();
   int read_db_schema_version();
   uint64_t get_page_size() const;
-  uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
+  uint64_t read_total_page_count_locked() const;
```
fujitatomoya: I think `read_total_page_count_locked` actually requires `db_read_write_mutex_` lock since it queries SQL?

```suggestion
  uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_);
```

clalancette: But it never takes the lock: https://github.com/ros2/rosbag2/blob/931bf541a3547dd7d1bcec119d68cb95d7409a72/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp#L881-L893

Oh, but you are saying maybe it *should* take the lock?  Yeah, that's possible.  @MichaelOrlov what do you think?

MichaelOrlov: @fujitatomoya The function itself shouldn't take a lock, however, it should rather be called only when the lock has already been acquired.
This is why I added `_locked` prefix to the name of the function and TSA annotation ` RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_);` to make sure that this function will ever be called when the relevant mutex is already locked.

BTW. We have another function a few lines above in the header file with exactly the same TSA guard lock and behavior and no warnings.
```
void write_locked(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> message)
  RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_);
```

MichaelOrlov (commented):
@clalancette Giving the fact that the similar function with the same TSA annotation exists a few line above and there are no warnings https://github.com/ros2/rosbag2/blob/931bf541a3547dd7d1bcec119d68cb95d7409a72/rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp#L124-L125 and also considering a warning message which you are trying to address **'"guarded_by' attribute only applies to non-static data members and global variables"**  I have a suspicious that the warning appears because we have **const** annotation on the 
```
uint64_t read_total_page_count_locked() const RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_);
``` 
As I've already mentioned in the @fujitatomoya comment the function itself shouldn't take a lock, however, it should rather be called only when the lock has already been acquired.
This is why I added `_locked` postfix to the name of the function and TSA annotation `RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_)` to make sure that this function will ever be called when the relevant mutex is already locked.

MichaelOrlov (commented):
@clalancette @fujitatomoya After reread-  @fujitatomoya suggestions about replacing `RCPPUTILS_TSA_GUARDED_BY(db_read_write_mutex_)` to the `RCPPUTILS_TSA_REQUIRES(db_read_write_mutex_)`  I would agree with it. It should fix the warning and facilitate the purpose of it.

rosbag2_storage_sqlite3/include/rosbag2_storage_sqlite3/sqlite_storage.hpp
clalancette: I see, thanks for the explanation.  I've now changed it to `RCPPUTILS_TSA_REQUIRED`; please take another look.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20132/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14570/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20843/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/57/)

MichaelOrlov (approved):
@clalancette LGTM.
BTW how to compile with clang locally to see warnings?
Using `--mixin clang-libcxx ` parameter for `colcon build`?

clalancette:
> Using `--mixin clang-libcxx ` parameter for `colcon build`?

I use the following command-line, which is derived from what we do in https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/ :

```
colcon build --base-paths src --build-base build --install-base install --event-handlers console_direct+ --cmake-args -DBUILD_TESTING=ON --no-warn-unused-cli -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_CXX_COMPLIER=/usr/bin/clang++ -DINSTALL_EXAMPLES=OFF -DSECURITY=ON -DAPPEND_PROJECT_NAME_TO_INCLUDEDIR=ON -DFORCE_BUILD_VENDOR_PKG=ON --mixin clang-libcxx --packages-up-to rosbag2_storage_sqlite3
```

(all of that may not be required, especially the `DCMAKE_*_COMPILER` flags)

MichaelOrlov:
Thanks for sharing

PR #1557, opened 2024-01-31
Gracefully handle SIGINT and SIGTERM signals for play and burst CLI

MichaelOrlov:
- Depends from https://github.com/ros2/rosbag2/pull/1592
- Intercept signals to call Player::stop() instead of relying on the rclcpp::shutdown() in default signal handlers.
- Also added static Player::Cancel() method.
- Added test coverage for Cancel during playback. Note: adding unit test coverage for Cancel during `burst` is not feasible because it runs very fast without delays between messages and without a repetition option.


MichaelOrlov:
@fujitatomoya @r7vme @sangteak601 I would use a review for this PR.

sangteak601:
Looks good to me.

ros-discourse:

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/6988fe3cba817e9ea4882a6957d37d28/raw/73ff5df4c6165b9d480abd7caeff0fbb350484b4/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_py rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13536
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20598/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14954/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21319/)

fujitatomoya:
@MichaelOrlov i can review this tomorrow.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -140,50 +140,109 @@ class Player
...
+    size_t burst_num_messages = 0)
+  {
+    auto old_sigterm_handler = std::signal(
+      SIGTERM, [](int /* signal */) {
+        rosbag2_py::Player::cancel();
+      });
+    auto old_sigint_handler = std::signal(
+      SIGINT, [](int /* signal */) {
+        rosbag2_py::Player::cancel();
       });
```
fujitatomoya: I do not see those original handlers be called on anywhere, is that correct behavior? i was expecting that on signal SIGINT and SIGTERM, we can cancel the player and then original signal handlers should be called accordingly based on signal number? but as far as i can see, it reset the original handler during exit, never called?

fujitatomoya (commented):
@MichaelOrlov one thing i want to confirm, probably that can be intention but i really do not see the reason, why it does not call the original signal handler from user application.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: @fujitatomoya That is not easy to implement since I am not saving signal number and handling signals asynchronously.  
There is basically no information at the end of the Play or Burst method if the signal was triggered or not.

fujitatomoya: understood the current implementation. probably this can be followed up but not in this PR. thank you for the explanation.

fujitatomoya (approved):
lgtm

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: @fujitatomoya Apparently, we do have a signal handler class on the rclcpp layer https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/signal_handler.hpp#L43-L55 but it is an inner class just for the rclcpp.  
I will try to see if I will be able to create something similar but lightweight for rosbag2 with low effort to be able to handle signals more systematically.

MichaelOrlov: @fujitatomoya I've added the call of the original signal handler for the `Player` and `Recorder` classes in the https://github.com/ros2/rosbag2/pull/1557/commits/290933247fbc8a75140573c62a9edda73b1bb8a1. Please review.

MichaelOrlov:
@clalancette I will need your formal approval or additional review for this PR.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -140,53 +142,144 @@ class Player
...
+    }
+    process_deferred_signal();
+    uninstall_signal_handlers();
   }
+
+  static std::atomic_bool exit_;
+  static std::condition_variable wait_for_exit_cv_;
+  static SignalHandlerType old_sigint_handler_;
+  static SignalHandlerType old_sigterm_handler_;
+  static int deferred_sig_number_;
```
fujitatomoya, line 269: really unlikely but we could have dedicated `deferred_sig_number_` for sigint and sigterm, so that we can guarantee that both original signal handers can be called later?

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -140,53 +142,144 @@ class Player
...
-    spin_thread.join();
+    play_impl(storage_options, play_options, true, num_messages);
+  }
+
+protected:
+  static void signal_handler(int sig_num)
+  {
+    if (sig_num == SIGINT || sig_num == SIGTERM) {
+      deferred_sig_number_ = sig_num;
+      rosbag2_py::Player::cancel();
```
fujitatomoya, line 171: canceling `Player` right away, and wait for `play` exits, and then call original signal handlers. i think this sounds reasonable to me.

fujitatomoya (commented):
@MichaelOrlov thanks for the fix, i had a couple of minor comments.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: I've tried to consider this option but it became very tricky how to maintain the order in which signals were called.  
The tricky part is that it needs to be a static variable. We can't use containers like vectors or queues in static variables since they allocate memory.  
At the end this is all done on `pybind_11` level and unlikely will be important at all if we will call `sigint` than `sigterm`  rather than just one `sigterm`.
I decided to impose a limitation that it will be called only the latest one. Anyway usually both of them will do the same actions just terminate the program.

MichaelOrlov:
@clalancette @emersonknapp I need a formal review/approval for this PR. 
@fujitatomoya Doesn't have write permissions to the rosbag2 repository and I can't merge with only his approaval.
If you will approve this PR I will re-run CI and if it is green will merge it.

emersonknapp (approved):

clalancette (approved):

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/82537b0fd13a1c970b1dc96f41168b65/raw/35835c4132ebe7f1094443d2366e3a27aabaaf48/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_py rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13670
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20820/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15124/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21505/)

MichaelOrlov:
Did rebase before merge to see if RPR and build_test jobs will pass.

MichaelOrlov:
RPR job failed with flake8 errors in unrelated packages. It seems a known issue on the CI and unrelated to this PR. Merging then.

MichaelOrlov:

PR #1558, opened 2024-01-31
Remove rcpputils::fs dependencies from rosbag2_storages

r7vme:
Original author: Kenta Yonekura <yoneken@ieee.org>

Copy of stalled PR https://github.com/ros2/rosbag2/pull/1474

### DIFF
--- a/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
+++ b/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
@@ -55,7 +55,7 @@ BagMetadata MetadataIo::read_metadata(const std::string & uri)
 
 std::string MetadataIo::get_metadata_file_name(const std::string & uri)
 {
-  std::string metadata_file = (std::filesystem::path(uri) / metadata_filename).string();
+  std::string metadata_file = (std::filesystem::path(uri) / metadata_filename).generic_string();
 
   return metadata_file;
 }
--- a/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
+++ b/rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
@@ -14,11 +14,10 @@
 
 #include "rosbag2_storage/metadata_io.hpp"
 
+#include <filesystem>
 #include <fstream>
 #include <string>
 #include <vector>
-
-#include "rcpputils/filesystem_helper.hpp"
 
 #include "rcutils/filesystem.h"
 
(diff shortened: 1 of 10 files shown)


r7vme:
@MichaelOrlov could you please take a look?

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
```diff
@@ -56,14 +55,14 @@ BagMetadata MetadataIo::read_metadata(const std::string & uri)
 
 std::string MetadataIo::get_metadata_file_name(const std::string & uri)
 {
-  std::string metadata_file = (rcpputils::fs::path(uri) / metadata_filename).string();
+  std::string metadata_file = (std::filesystem::path(uri) / metadata_filename).generic_string();
```
r7vme, line 58: @MichaelOrlov I'm not 100% sure that `generic_string` is necessary. This only affects behaviour on Windows

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20197/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14617/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20908/)

rosbag2_storage/src/rosbag2_storage/metadata_io.cpp
clalancette: > @MichaelOrlov I'm not 100% sure that `generic_string` is necessary. This only affects behaviour on Windows

I believe that's why we added it earlier, but we'll see what Windows CI has to say about this code.

clalancette:
@r7vme There are quite a few linter failures here.  Please run the linters locally and fix the problems, then we can try again.

r7vme:
@clalancette fixed tests, PTAL

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14628/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20927/)

clalancette (approved):
This looks good to me, but I'd like another opinion from @MichaelOrlov before we merge.

MichaelOrlov (approved):
@r7vme Thanks for the contribution. The changes looks good to me.
However, there is one open question for the future.
Do we really need to use the ".generic_string()" or simple `.string()` would be enough?

clalancette:
> Do we really need to use the ".generic_string()" or simple `.string()` would be enough?

Someone is going to have to test it out on Windows to tell for sure.

MichaelOrlov:
As far as I understand from documentation the one should use  `.generic_string()` if needed to make sure that the Windows-specific quirks with backslashes `\` in the path will be converted to the regular slashes `/` when the path needs to be represented as string.

MichaelOrlov:

Issue #1542, opened 2024-01-17
Early (incorrect) exit in certain situations with mcap logging


anrp-tri:
## Description
We run the ros2 bagger in CI to collect logs of a (CI) simulation run. In this instance, we segregate parallel simulations with different ROS2 domain IDs. Each individual simulation has on the order of ~50 topics. Due to the parallel-ness of this, the machine is significantly loaded i.e. scheduling delays are expected. About 2% of the time, the logger exits during system startup complaining about topic information.

### Expected Behavior
The logger does not exit or crash.

### Actual Behavior
We get one of three exits, occasionally, during startup while topics are appearing (actual topic in question is not consistent), some logs:
```
2024-01-15 13:49:52,859 ERROR:[INFO] [1705326592.534510178] [rosbag2_recorder]: Subscribed to topic '/carla/objects'
2024-01-15 13:49:52,859 ERROR:terminate called after throwing an instance of 'std::runtime_error'
2024-01-15 13:49:52,859 ERROR:what():  Channel reference not found for topic: "/ic/speedometer"
```

```
2024-01-12 13:29:40,012 ERROR:[INFO] [1705066179.755091747] [rosbag2_recorder]: Subscribed to topic '/ic/warnings/fuel'
2024-01-12 13:29:40,012 ERROR:terminate called after throwing an instance of 'std::runtime_error'
2024-01-12 13:29:40,012 ERROR:what():  Unknown message topic "/rosout"
```

These are both uncaught `std::runtime_error` exceptions, so the logger exits.

Very occasionally, we get an error (presumably) from STL which smells like memory corruption (combined with the stack traces):
```
2024-01-11 13:51:23,861 ERROR:[INFO] [1704981083.671525050] [rosbag2_recorder]: Subscribed to topic '/telemetry/roadwheel_angle'
2024-01-11 13:51:23,861 ERROR:terminate called after throwing an instance of 'std::bad_alloc'
2024-01-11 13:51:23,861 ERROR:what():  std::bad_alloc
```

## To Reproduce
I was unable to reproduce this with a test program that spammed new topics. We see this quite reliably in our CI (about 1/50 runs of a simulation). We can instrument the built version of rosbag2_storage_mcap if useful.

## System (please complete the following information)
 - OS: Ubuntu Jammy (also focal)
 - ROS 2 Distro:  humble (also foxy)
 - Install Method: We build rosbag2_storage_mcap from source
[...]

PR #1561, opened 2024-02-01
Use rw_lock to protect mcap metadata lists.

fujitatomoya:
address https://github.com/ros2/rosbag2/issues/1542

### DIFF
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -42,8 +42,8 @@
 #include <algorithm>
 #include <filesystem>
 #include <memory>
+#include <mutex>
 #include <optional>
-#include <shared_mutex>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -231,6 +231,7 @@ public:
 
 private:
   void read_metadata();
+  void write_lock_free(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg);
   void open_impl(const std::string & uri, const std::string & preset_profile,
                  rosbag2_storage::storage_interfaces::IOFlag io_flag,
                  const std::string & storage_config_uri);
@@ -247,13 +248,13 @@ private:
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> next_;
 
   rosbag2_storage::BagMetadata metadata_{};
-  std::shared_mutex mcap_rw_shared_mutex_;
+  std::mutex mcap_storage_mutex_;
   std::unordered_map<std::string, rosbag2_storage::TopicInformation> topics_
-    RCPPUTILS_TSA_GUARDED_BY(mcap_rw_shared_mutex_);
+    RCPPUTILS_TSA_GUARDED_BY(mcap_storage_mutex_);
   std::unordered_map<std::string, mcap::SchemaId> schema_ids_
-    RCPPUTILS_TSA_GUARDED_BY(mcap_rw_shared_mutex_);  // datatype -> schema_id
+    RCPPUTILS_TSA_GUARDED_BY(mcap_storage_mutex_);  // datatype -> schema_id
   std::unordered_map<std::string, mcap::ChannelId> channel_ids_
-    RCPPUTILS_TSA_GUARDED_BY(mcap_rw_shared_mutex_);  // topic -> channel_id
+    RCPPUTILS_TSA_GUARDED_BY(mcap_storage_mutex_);  // topic -> channel_id
   rosbag2_storage::StorageFilter storage_filter_{};
   mcap::ReadMessageOptions::ReadOrder read_order_ = mcap::ReadMessageOptions::ReadOrder::FileOrder;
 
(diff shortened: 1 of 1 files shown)


rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -735,6 +741,7 @@ uint64_t MCAPStorage::get_minimum_split_file_size() const
 /** BaseWriteInterface **/
 void MCAPStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg)
 {
+  std::shared_lock read_lock(mcap_rw_shared_mutex_);
```
fujitatomoya: this is protected shared_mutex, this is intentionally done to keep the performance to write the data into mcap file as much as possible.

MichaelOrlov: @fujitatomoya Actually the `shared_mutex` in this particular case has the opposite effect on performance and will be on average **4.5 times slower** than regular `std::mutex` with `unique_lock` because we don't have cases in rosbag2 when `MCAPStorage::write(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg)` could be called from multiple threads.

fujitatomoya: @MichaelOrlov what if we have MultiThreadedExecutor? i was expecting that will call backend storage `write` call concurrently? i may be mistaken...

MichaelOrlov (changes requested):
@fujitatomoya Thanks for trying to fix this issue with race condition.
At first glance, the fix is trivial. However, we can do way better.
- We don't need to use `shared_mutex` because there are no situations in rosbag2 when write() method could be called from multiple threads and unlikely would be because 
   1. it is protected with mutex on upper `rosbag2_cpp::writer::write(..)` method for the cases when we are writing without cache. 
   2. We are using `SingleThreadedExecutor` which is processing all subscription callbacks one by one sequentially in the one thread. 
   3. When double-buffer cache is enabled we are using another method https://github.com/ros2/rosbag2/blob/b8a9b06c983898da17898bd9248e38bd2cd6a893/rosbag2_storage_mcap/src/mcap_storage.cpp#L784-L790
- And here we can do optimization. It will be more efficient to lock mutex once before we iterate through the vector of messages and call the write method for each message. Other API calls as `create_topic(..)` or `remove_topic(..)` have lower priority and could wait until we will finish dumping the vector of messages from our double-buffer cache to the storage. It will be more elegant to create a new private method for writing one message without lock. We can name it `write_lock_free(msg)` or `write_locked(msg)` and lock mutex in public write(..) method right before calling newly created `write_lock_free(msg)` and in the 
https://github.com/ros2/rosbag2/blob/b8a9b06c983898da17898bd9248e38bd2cd6a893/rosbag2_storage_mcap/src/mcap_storage.cpp#L784-L790
right before for loop.

fujitatomoya:
@MichaelOrlov thanks!

> because there are no situations in rosbag2 when write() method could be called from multiple threads and unlikely would be because

I see, `rosbag2_cpp::writer::write` protects the methods with mutex lock.

whole point is `rosbag2_cpp::cache::CacheConsumer::exec_consuming()`, this one is staying outside of the lock mechanism.

> We are using SingleThreadedExecutor which is processing all subscription callbacks one by one sequentially in the one thread.

currently true. minor concern was if we support `MultiThreadedExecutor` in the future, we will need to come back to this discussion. probably that is not concern, i guess `rosbag2` takes the different path to take advantage of `cache` with dedicated single thread for the performance?

> It will be more elegant to create a new private method for writing one message without lock. We can name it write_lock_free(msg) or write_locked(msg) and lock mutex in public write(..) method right before calling newly created write_lock_free(msg)

okay, can do that.

- will use `std::mutex`, instead of `shared_mutex`.
- conceal lock free implementation in private scope, then control lock in public scope.

fujitatomoya:
@MichaelOrlov requesting another look! thanks!

MichaelOrlov:
@fujitatomoya Thanks for the followup. As reagrds.
> currently true. minor concern was if we support MultiThreadedExecutor in the future, we will need to come back to this discussion. probably that is not concern, i guess rosbag2 takes the different path to take advantage of cache with dedicated single thread for the performance?

Yes. Currently using double-buffer cache gives the same performance gain as a multithreaded executor but in two threads and decouples writing to the storage and transport layer. Which is very important due to the possible delays with filesystem or DB operations. 
Even if we would start using MutithreadedExecutor for whatever reasons in the future we still bounded to the locking on one `std::mutex` on the rosbag2_cpp writer(message) layer here
https://github.com/ros2/rosbag2/blob/edda3769ffe0986d75cabe4464597a516f7aa969/rosbag2_cpp/src/rosbag2_cpp/writer.cpp#L104-L108
And would need to reconsider and make changes on that level as well.  
Switching to `shared_mutex` on the `rosbag2_storage_mcap` layer if would needed in the future will not cause difficulties even with backporting since the whole changes in the `.cpp` file only and will be API/ABI compatible.

MichaelOrlov (approved):
@fujitatomoya Thanks. LGTM with green CI.

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/a06b14d20610f8f8cc8007db7fc61efe/raw/06daf70cbe74c9b1eb6715c512d655cd101f648f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13233
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20218/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14638/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20937/)

MichaelOrlov:

ros-discourse:

https://discourse.ros.org/t/ros-2-meeting-minutes-2024-02-15/36221/1

PR #1578, opened 2024-02-28
Make some changes for newer versions of uncrustify.

clalancette:
Uncrustify 0.78 is the new version in Ubuntu 24.04. It has some small differences in indentation compared to Uncrustify 0.72, which is what is in Ubuntu 22.04.

In this PR, we make the changes to match what Uncrustify 0.78 wants.  In most cases, I think this is a "better" indentation.

We also add in strategic use of "*INDENT-OFF*" comments, which instructs Uncrustify to not indent this particular section of code.  That is so that this will still pass the linters on Uncrustify 0.72 while Rolling is still on that version.

With this in place, we should get green CI on both Ubuntu 22.04 and 24.04.

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/logging.hpp
+++ b/rosbag2_compression/src/rosbag2_compression/logging.hpp
@@ -22,40 +22,44 @@
 
 #define ROSBAG2_COMPRESSION_PACKAGE_NAME "rosbag2_compression"
 
+// *INDENT-OFF*
+
 #define ROSBAG2_COMPRESSION_LOG_INFO(...) \
   RCUTILS_LOG_INFO_NAMED(ROSBAG2_COMPRESSION_PACKAGE_NAME, __VA_ARGS__)
 
 #define ROSBAG2_COMPRESSION_LOG_INFO_STREAM(args) do { \
-    std::stringstream __ss; \
-    __ss << args; \
-    RCUTILS_LOG_INFO_NAMED(ROSBAG2_COMPRESSION_PACKAGE_NAME, "%s", __ss.str().c_str()); \
+  std::stringstream __ss; \
+  __ss << args; \
+  RCUTILS_LOG_INFO_NAMED(ROSBAG2_COMPRESSION_PACKAGE_NAME, "%s", __ss.str().c_str()); \
 } while (0)
 
 #define ROSBAG2_COMPRESSION_LOG_ERROR(...) \
   RCUTILS_LOG_ERROR_NAMED(ROSBAG2_COMPRESSION_PACKAGE_NAME, __VA_ARGS__)
 
 #define ROSBAG2_COMPRESSION_LOG_ERROR_STREAM(args) do { \
-    std::stringstream __ss; \
-    __ss << args; \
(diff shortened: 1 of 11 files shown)


MichaelOrlov (approved):
@clalancette LGTM.
However, as more I see such PRs as more I want to reformat everything with clang-format (at least in rosbag2) as we have done it for the `rosbag2_storage_mcap`.  
Maybe it is a good time? @clalancette What do you think?

clalancette:
> However, as more I see such PRs as more I want to reformat everything with clang-format (at least in rosbag2) as we have done it for the `rosbag2_storage_mcap`.
> Maybe it is a good time? @clalancette What do you think?

Yeah, I see your point here.  But I'd really prefer that rosbag2 not go its own way here, as it is a part of the core.

At this point, I could definitely be convinced to switch away from uncrustify to clang-format for the whole core.  However, the reason I did not do this here is that it is going to be significantly more work.  While it *is* annoying that we have to do this between uncrustify versions, the fact is that this is going to take ~10 PRs to fix all of the core.  If we switch to clang-format, we are going to have to open a PR to pretty much every repository that has C/C++ code, so probably around 75 PRs.

If we do want to take this on, I'm going to suggest that we do it *after* we have transitioned to Ubuntu 24.04.  That way at least we'll be on a stable base and won't be doing too many things at once.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20388/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14798/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21138/)

MichaelOrlov:
@clalancette 
> If we do want to take this on, I'm going to suggest that we do it after we have transitioned to Ubuntu 24.04. That way at least we'll be on a stable base and won't be doing too many things at once.

Please let me know when it will be a good time and I will be glad to reformat all rosbag2 code with clang-format linter.

I personally dislike how uncrustify formatting code from day 1 when I looked at the ROS 2 code.  
I've tried to investigate whether may be possible to tweak uncrustify config to make it look more pretty. Unfortunately not feasible to get closer to what is easy to do with clang-format.

clalancette:
> Please let me know when it will be a good time and I will be glad to reformat all rosbag2 code with clang-format linter.

Just to be clear, I don't think we should do this.

That is, I think rosbag2 code should be formatted the same as the rest of the ROS 2 core.  If we decide to reformat *everything* with clang-format, then we can reformat rosbag2 as well.  But just reformatting rosbag2 defeats the entire purpose of having a consistent style.

MichaelOrlov:
Sorry for the misunderstanding I was under the impression that we are almost ready for decision `"to reformat everything with clang-format"`

clalancette:
> Sorry for the misunderstanding I was under the impression that we are almost ready for decision `"to reformat everything with clang-format"`

No problem.  I put something on the agenda for the weekly meeting so we can discuss it next week.

clalancette:
Given that this PR is just a change in formatting, and none of the failing tests above has to do with formatting, I'm going to go ahead and merge this in.  Thanks for the review.

PR #1580, opened 2024-03-06
Add simple bag player examples

### LINKED ISSUE #1449: Add additional examples for reading ROSBag data and serialized messages
## Description

Currently, the ROSBag2 examples contain three different examples for writing data to a ROSBag via C++  or Python. However, there are no specific examples of reading data from ROSBags (neither for the standard reader or the sequential reader). In addition, the writing examples do not show how to serialize various messages before writing them to a bag file.
Inside the file under _rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_api.cpp_, there are test cases for reading data and also for the message serialization. These helped me when using the ROSBag2 API, but in my opinion, a better idea is to create dedicated examples, because the unit tests should be meant solely for testing.

## Completion Criteria

- Example for reading ROSBag data using a simple reader (`rosbag2_cpp::Reader`)
- Example for reading ROSBag data using a sequential reader (`rosbag2_cpp::SequentialReader`)
- Example for writing ROSBag data using serialized messages (`rclcpp::SerializedMessage` and `rclcpp::Serialization<>`)

sangteak601:
Fixes #1449.

Added simple bag player examples in cpp and Python. I also tried to read or write bag files without specifying topic types. However, it became too complicated and was no longer a "simple" example, so I decided not to push the changes.

### DIFF
--- a/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
+++ b/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
@@ -34,7 +34,7 @@ public:
 
     timer_ = this->create_wall_timer(
       100ms,
-      [this](){return this->timer_callback();}
+      [this]() {return this->timer_callback();}
     );
 
     reader_.open(bag_filename);
--- a/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
+++ b/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
@@ -7,7 +7,7 @@
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/serialization.hpp"
 #include "rosbag2_cpp/reader.hpp"
-#include "example_interfaces/msg/string.hpp"
+#include "std_msgs/msg/string.hpp"
 
 using namespace std::chrono_literals;
 
@@ -17,7 +17,7 @@ class SimpleBagPlayer : public rclcpp::Node
     SimpleBagPlayer(const std::string & bag_filename)
     : Node("playback_node")
     {
-      publisher_ = this->create_publisher<example_interfaces::msg::String>("chatter", 10);
+      publisher_ = this->create_publisher<std_msgs::msg::String>("chatter", 10);
 
       timer_ = this->create_wall_timer(100ms,
           [this](){return this->timer_callback();}
@@ -37,7 +37,7 @@ class SimpleBagPlayer : public rclcpp::Node
         }
 
         rclcpp::SerializedMessage serialized_msg(*msg->serialized_data);
-        example_interfaces::msg::String::SharedPtr ros_msg = std::make_shared<example_interfaces::msg::String>();
+        std_msgs::msg::String::SharedPtr ros_msg = std::make_shared<std_msgs::msg::String>();
 
         serialization_.deserialize_message(&serialized_msg, ros_msg.get());
 
@@ -49,9 +49,9 @@ class SimpleBagPlayer : public rclcpp::Node
     }
 
     rclcpp::TimerBase::SharedPtr timer_;
-    rclcpp::Publisher<example_interfaces::msg::String>::SharedPtr publisher_;
+    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_;
 
-    rclcpp::Serialization<example_interfaces::msg::String> serialization_;
+    rclcpp::Serialization<std_msgs::msg::String> serialization_;
     rosbag2_cpp::Reader reader_;
 };
 
--- a/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
+++ b/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
@@ -30,12 +30,15 @@ public:
   explicit SimpleBagPlayer(const std::string & bag_filename)
   : Node("simple_bag_player")
   {
+    declare_parameter("edit", false);
+    message_needs_to_be_edit_before_send_ = get_parameter("edit").as_bool();
+
     publisher_ = this->create_publisher<std_msgs::msg::String>("chatter", 10);
 
     // ignore timestamp and publish at a fixed rate (10 Hz).
     timer_ = this->create_wall_timer(
       100ms,
-      [this]() {return this->timer_callback();}
+      [this]() {this->timer_callback();}
     );
 
     reader_.open(bag_filename);
@@ -73,12 +76,12 @@ private:
   rclcpp::Serialization<std_msgs::msg::String> serialization_;
   rosbag2_cpp::Reader reader_;
 
-  bool message_needs_to_be_edit_before_send_ {true};
+  bool message_needs_to_be_edit_before_send_;
 };
 
 int main(int argc, char ** argv)
 {
-  if (argc != 2) {
+  if (argc < 2) {
     std::cerr << "Usage: " << argv[0] << " <bag>" << std::endl;
     return 1;
   }
--- a/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
+++ b/rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
@@ -1,5 +1,18 @@
+// Copyright 2024 Open Source Robotics Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include <chrono>
-#include <functional>
 #include <iostream>
 #include <memory>
 #include <string>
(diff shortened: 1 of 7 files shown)


sangteak601:
Hi @MichaelOrlov, I think I screwed up with reviewers. Do you mind fixing it and reviewing this PR ?

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "rosbag2_cpp/reader.hpp"
+#include "std_msgs/msg/string.hpp"
+
+using namespace std::chrono_literals;
+
+class SimpleBagPlayer : public rclcpp::Node
+{
+public:
+  explicit SimpleBagPlayer(const std::string & bag_filename)
+  : Node("playback_node")
```
fujitatomoya: nit: against `simple_bag_recorder`.

```suggestion
  : Node("simple_bag_player")
```

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,84 @@
...
+      rosbag2_storage::SerializedBagMessageSharedPtr msg = reader_.read_next();
+
+      if (msg->topic_name != "chatter") {
+        continue;
+      }
+
+      rclcpp::SerializedMessage serialized_msg(*msg->serialized_data);
+      std_msgs::msg::String::SharedPtr ros_msg = std::make_shared<std_msgs::msg::String>();
+
+      serialization_.deserialize_message(&serialized_msg, ros_msg.get());
```
fujitatomoya: Do we need to deserialize the message to be published? why can we not do `publish(rclcpp::SerializedMessage)`?

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+public:
+  explicit SimpleBagPlayer(const std::string & bag_filename)
+  : Node("playback_node")
+  {
+    publisher_ = this->create_publisher<std_msgs::msg::String>("chatter", 10);
+
+    timer_ = this->create_wall_timer(
+      100ms,
+      [this]() {return this->timer_callback();}
```
fujitatomoya: So this simple bag player does not care the timestamp in the bag file at all.

if user publishes `chatter` topic with 100Hz (10 msec) and record it with `simple_bag_recorder`, this will actually replay the messages with 1/10 frequency against the bag file timestamps.

i think this is no problem at all as example, but i think user needs to be aware of this situation. how about the adding comment somewhere here to explain this?

rosbag2_examples/rosbag2_examples_py/rosbag2_examples_py/simple_bag_player.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2023 Open Source Robotics Foundation, Inc.
```
fujitatomoya:
```suggestion
# Copyright 2024 Open Source Robotics Foundation, Inc.
```

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
sangteak601: Thanks for the review.
Resolved.

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
sangteak601: Resolved

rosbag2_examples/rosbag2_examples_py/rosbag2_examples_py/simple_bag_player.py
sangteak601: Resolved

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
sangteak601: One of the requests from #1449 was to show how to serialize/deserialize messages. I also think it would be nice to include (de)serialization in the examples.

How about adding comments to inform that it is also possible to publish serialized messages?

MichaelOrlov:
@sangteak601 @fujitatomoya I am curious how this PR overlaps with the existing tutorial and examples from https://docs.ros.org/en/rolling/Tutorials/Advanced/Reading-From-A-Bag-File-CPP.html ?

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
fujitatomoya: I see, as an example, that makes more sense. adding comment would be good to, i can leave this to you. thanks!

fujitatomoya:
@MichaelOrlov 

> I am curious how this PR overlaps with the existing tutorial and examples from https://docs.ros.org/en/rolling/Tutorials/Advanced/Reading-From-A-Bag-File-CPP.html ?

that is what i found too.

i was thinking that https://docs.ros.org/en/rolling/Tutorials/Advanced/Reading-From-A-Bag-File-CPP.html can also be updated to point the source code from this example.

MichaelOrlov:
> i was thinking that https://docs.ros.org/en/rolling/Tutorials/Advanced/Reading-From-A-Bag-File-CPP.html can also be updated to point the source code from this example.

I would prefer to make them aligned as much as possible but in the tutorial preferable to have the full text with source code instead of just a link on it.  

Also, I think that in the examples still unclear point that if the message always shall be deserialized or not before publishing.
To make it clear and show that it is optional I would suggest adding some class member variable or global variable with name like `message_needs_to_be_edit_before_send` and have an `if-else` statement showing two options for publishing. With deserialized and altering value before publishing and non-serialized publishing.

cc: @sangteak601

sangteak601:
@MichaelOrlov Thanks for the review. I will make a modification.

Based on my knowledge(correct me if I'm wrong), we can also receive serialized messages directly. Therefore, perhaps we can modify the recorder example in a similar manner?

sangteak601:
@fujitatomoya @MichaelOrlov I have addressed issues. Do you mind taking another look?

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,91 @@
...
+public:
+  explicit SimpleBagPlayer(const std::string & bag_filename)
+  : Node("simple_bag_player")
+  {
+    publisher_ = this->create_publisher<std_msgs::msg::String>("chatter", 10);
+
+    // ignore timestamp and publish at a fixed rate (10 Hz).
+    timer_ = this->create_wall_timer(
+      100ms,
+      [this]() {return this->timer_callback();}
```
fujitatomoya: nothing to return?

```suggestion
      [this]() {this->timer_callback();}
```

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  rclcpp::TimerBase::SharedPtr timer_;
+  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_;
+
+  rclcpp::Serialization<std_msgs::msg::String> serialization_;
+  rosbag2_cpp::Reader reader_;
+
+  bool message_needs_to_be_edit_before_send_ {true};
```
fujitatomoya: i think it would be better for user to configure this flag via cli option to play around with this example package? unless, user needs to modify the source code and build it by them self from source, that requires more work for user?

rosbag2_examples/rosbag2_examples_py/rosbag2_examples_py/simple_bag_player.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class SimpleBagPlayer(Node):
+
+    def __init__(self, bag_filename):
+        super().__init__('simple_bag_player')
+        self.reader = rosbag2_py.SequentialReader()
+        storage_options = rosbag2_py._storage.StorageOptions(
+            uri=bag_filename,
+            storage_id='sqlite3')
```
fujitatomoya, line 30: if i am not mistaken, this is intentionally enabled as `sqlite3`? but this bag file can be also red from `simple_bag_player.cpp`, because it reads the meta data from the bag file and use the appropriate storage plugin if available? just checking this because user does not meet the failure during the tutorial.

fujitatomoya (commented):
overall lgtm, but i have a few comments.

besides,

> I would prefer to make them aligned as much as possible but in the tutorial preferable to have the full text with source code instead of just a link on it.

i do agree with this. we need to update https://github.com/ros2/ros2_documentation accordingly to align with these samples.

rosbag2_examples/rosbag2_examples_py/rosbag2_examples_py/simple_bag_player.py
sangteak601: It was set as `sqlite3` in `simple_bag_recorder.py`. I don't know the reason exactly.
It is possible to read `sqlite3` files from `simple_bag_player.cpp`, but the opposite(reading `mcap` files from `simple_bag_player.py`) isn't.

Maybe we can delete `storage_id`?

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
sangteak601: Done

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
sangteak601: Add a ROS parameter `edit` to enable configuration.

sangteak601:
> > I would prefer to make them aligned as much as possible but in the tutorial preferable to have the full text with source code instead of just a link on it.
> 
> i do agree with this. we need to update https://github.com/ros2/ros2_documentation accordingly to align with these samples.

I will update the documentation once this is merged.

rosbag2_examples/rosbag2_examples_py/rosbag2_examples_py/simple_bag_player.py
fujitatomoya: No, i think that is okay with current code. this also includes an example how to specify the storage id, i think that is fine.

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,94 @@
...
+using namespace std::chrono_literals;
+
+class SimpleBagPlayer : public rclcpp::Node
+{
+public:
+  explicit SimpleBagPlayer(const std::string & bag_filename)
+  : Node("simple_bag_player")
+  {
+    declare_parameter("edit", false);
+    message_needs_to_be_edit_before_send_ = get_parameter("edit").as_bool();
```
fujitatomoya, line 34: comment for the documentation. this needs to be described how to enable with command line parameter argument. so that user can understand we can publish either publishing serialized data as it is or editing data before publishing in the example.

fujitatomoya (approved):

rosbag2_examples/rosbag2_examples_cpp/src/simple_bag_player.cpp
```diff
@@ -0,0 +1,94 @@
...
+      if (message_needs_to_be_edit_before_send_) {
+        rclcpp::SerializedMessage serialized_msg(*msg->serialized_data);
+        std_msgs::msg::String::SharedPtr ros_msg = std::make_shared<std_msgs::msg::String>();
+
+        serialization_.deserialize_message(&serialized_msg, ros_msg.get());
+        ros_msg->data += "[edited]";
+        publisher_->publish(*ros_msg);
+        std::cout << ros_msg->data << "\n";
+      } else {
+        publisher_->publish(*msg->serialized_data);
```
MichaelOrlov, line 66: @sangteak601 As far as I recall we need another type of publisher aka `GenericPublisher` to be able to publish serialized data.

MichaelOrlov:
- Closing as stale PR in favor of the #1683

PR #1583, opened 2024-03-12
Switch to using ament_vendor_package for lz4.

clalancette:
This is more explicit, and more like how we vendor the rest of the packages in our ecosystem.

Note that this is the first of two PRs to switch `mcap_vendor` over to using `ament_vendor_package`.  We want to do that to make it more like the other vendor packages in the core, and to fix building of binary `mcap_vendor` packages on Noble (we currently have a hack that allows us to do this, but we would like to get rid of it).


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20458/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14860/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21201/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/597/)

lz4_vendor/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.10)
+project(lz4_vendor)
+
+find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_vendor_package REQUIRED)
+
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
+find_package(lz4 QUIET)
```
cottsay: Should this have a minimum version?

lz4_vendor/README.md
```diff
@@ -0,0 +1,3 @@
+# lz4_vendor
```
cottsay: Do we want/need a quality declaration?

lz4_vendor/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+ament_vendor(lz4_vendor
+  SATISFIED ${lz4_FOUND}
+  VCS_URL https://github.com/lz4/lz4.git
+  VCS_VERSION v1.9.4
+  SOURCE_SUBDIR build/cmake
+  CMAKE_ARGS
+    -DBUILD_STATIC_LIBS:BOOL=OFF
+    -DBUILD_SHARED_LIBS:BOOL=ON
```
cottsay: Consider adding a call to `ament_add_default_options` to handle this, and possibly other things in the future.

lz4_vendor/cmake/Modules/Findlz4.cmake
```diff
@@ -0,0 +1,54 @@
...
+
+if(lz4_FOUND)
+  set(lz4_INCLUDE_DIRS ${lz4_INCLUDE_DIR})
+  set(lz4_LIBRARIES ${lz4_LIBRARY})
+
+  if(NOT TARGET lz4::lz4)
+    add_library(lz4::lz4 UNKNOWN IMPORTED)
+    set_property(TARGET lz4::lz4 PROPERTY IMPORTED_LOCATION ${lz4_LIBRARY})
+    set_property(TARGET lz4::lz4 PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${lz4_INCLUDE_DIR})
+  endif()
```
cottsay: The recently introduced official CMake config appears to use the `LZ4::` namespace: https://github.com/lz4/lz4/blob/5ff839680134437dbf4678f3d0c7b371d84f4964/build/cmake/CMakeLists.txt#L223-L225

lz4_vendor/CMakeLists.txt
clalancette: It's not clear to me.  I guess we could set this to 1.9.3 (which is what is in Ubuntu 22.04), but I don't have a strong opinion either way.

lz4_vendor/README.md
clalancette: It shouldn't be necessary; none of rosbag2 has quality declarations.

lz4_vendor/cmake/Modules/Findlz4.cmake
clalancette: Yeah, good call.  I'll fix this in an update.

clalancette: Should be fixed in 3fe38c4.

lz4_vendor/CMakeLists.txt
clalancette: Changes in 3fe38c4.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20462/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14863/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21204/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/598/)

cottsay (approved):

MichaelOrlov:
@clalancette Any idea why `ros2bag` tests fails with error message:
```
 ==================================== ERRORS ====================================
2024-03-12T22:32:52.5615828Z ____________________________ ERROR collecting test _____________________________
2024-03-12T22:32:52.5620514Z /usr/local/lib/python3.10/dist-packages/pluggy/_hooks.py:501: in __call__
2024-03-12T22:32:52.5623335Z     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
2024-03-12T22:32:52.5625134Z /usr/local/lib/python3.10/dist-packages/pluggy/_manager.py:119: in _hookexec
2024-03-12T22:32:52.5626189Z     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
2024-03-12T22:32:52.5627625Z /usr/local/lib/python3.10/dist-packages/_pytest/python.py:225: in pytest_collect_file
2024-03-12T22:32:52.5628637Z     module: Module = ihook.pytest_pycollect_makemodule(
2024-03-12T22:32:52.5630579Z /usr/local/lib/python3.10/dist-packages/_pytest/config/compat.py:79: in fixed_hook
2024-03-12T22:32:52.5631429Z     return hook(**kw)
2024-03-12T22:32:52.5646994Z /usr/local/lib/python3.10/dist-packages/pluggy/_hooks.py:501: in __call__
2024-03-12T22:32:52.5648220Z     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
2024-03-12T22:32:52.5649653Z /usr/local/lib/python3.10/dist-packages/pluggy/_manager.py:119: in _hookexec
2024-03-12T22:32:52.5650835Z     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
2024-03-12T22:32:52.5652056Z ../../../../build/launch_testing/launch_testing/pytest/hooks.py:188: in pytest_pycollect_makemodule
2024-03-12T22:32:52.5653140Z     entrypoint = find_launch_test_entrypoint(path)
2024-03-12T22:32:52.5654253Z ../../../../build/launch_testing/launch_testing/pytest/hooks.py:178: in find_launch_test_entrypoint
2024-03-12T22:32:52.5655327Z     module = import_path(path, root=None)
2024-03-12T22:32:52.5656607Z E   TypeError: import_path() missing 1 required keyword-only argument: 'consider_namespace_packages'
2024-03-12T22:32:52.5658146Z --- generated xml file: /__w/rosbag2/rosbag2/ros_ws/build/ros2bag/pytest.xml ---
2024-03-12T22:32:52.5659198Z =========================== short test summary info ============================
2024-03-12T22:32:52.5660652Z ERROR test - TypeError: import_path() missing 1 required keyword-only argument: 'consider_namespace_packages'
2024-03-12T22:32:52.5661641Z !!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
2024-03-12T22:32:52.5662128Z =============================== 1 error in 0.16s ===============================
2024-03-12T22:32:52.5948706Z Failed   <<< ros2bag [1.54s, exited with code 2]
```
I am pretty sure it is unrelated to the changes from this PR, but still unclear for me what is wrong with it.

lz4_vendor/package.xml
```diff
@@ -0,0 +1,24 @@
...
+  <version>0.24.0</version>
+  <description>LZ4 compression vendor package, providing a dependency for LZ4.</description>
+  <maintainer email="michael.orlov@apex.ai">Michael Orlov</maintainer>
+  <maintainer email="me@emersonknapp.com">Emerson Knapp</maintainer>
+  <maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>
+  <license>Apache License 2.0</license>  <!-- the contents of this package are Apache 2.0 -->
+  <license>BSD</license>  <!-- LZ4 is BSD -->
+  <license>GPLv2</license>  <!-- LZ4 is GPLv2 -->
+
+  <url type="website">https://facebook.github.io/zstd/</url>
```
MichaelOrlov:
```suggestion
  <url type="website">https://github.com/lz4/lz4/</url>
```

MichaelOrlov (changes requested):
@clalancette @cottsay If I am not mistaken and recall correctly, the reason why we didn't "vendorize" `lz4` at the very beginning and use it directly during the compilation of the  `mcap` library from sources is because:
1. `lz4` is released under the dual license. From the https://github.com/lz4/lz4?tab=License-1-ov-file
   ```
     This repository uses 2 different licenses :
      - all files in the `lib` directory use a BSD 2-Clause license
      - all other files use a GPLv2 license, unless explicitly stated otherwise
   ```
   It turns out, if do it the way how it is done in the current PR. i.e. mix and match by compiling everything from `lz4` sources into the vendor package library and then link against it in the `mcap` library. The `mcap` library will become **poisoned** by the `GPLv2 license` and everything else downstream linking mcap library will become poisoned by the `GPLv2 license`!!!
It might be "ok" for open source projects but certainly **NO-NO** for any commercial products. The consequences will be that all closed-source "poisoned" by the `GPLv2` code or library will automatically become under the `GPLv2` license and could be enforced to be open sourced.
Originally we workaround this problem by using explicitly only those part of the `lz4` which is under the `BSD` license via 
```
  file(GLOB _lz4_srcs
    ${lz4_SOURCE_DIR}/lib/*.c)

  add_library(mcap SHARED
    src/main.cpp
    ${_lz4_srcs}
  )
```
2. No one else uses `lz4` anywhere else in the other ROS 2 core packages.

MichaelOrlov:
@clalancette I recall one more argument as to why we didn't "vendorize" `lz4` from the very beginning.
3. The `lz4` is a high-performant compression library and compiling `mcap` library with `lz4` raw sources gives better results in performance rather than wrapping `lz4` into the library and linking against it.
Folks from Foxglove even did some measurements before as far as I recall.

cc: @james-rms @defunctzombie

clalancette:
> 1. `lz4` is released under the dual license. From the https://github.com/lz4/lz4?tab=License-1-ov-file

I'll have to look at this one in more detail, though clearly I am not a lawyer so there's only so much I can do.  All I'll say is that it seems very dicey to rely on that from a legal point-of-view.

>     2. No one else uses `lz4` anywhere else in the other ROS 2 core packages.

This part doesn't really matter; that's the case for e.g. `zstd_vendor` as well.

> 3. The `lz4` is a high-performant compression library and compiling `mcap` library with `lz4` raw sources gives better results in performance rather than wrapping `lz4` into the library and linking against it.

This would be extremely surprising.  The only way I can think that would happen would be if the compiler inlined everything.

MichaelOrlov:
@clalancette My major concern is a viral "infection" of other libraries and executables by the `GPLv2` license from the `lz4` library.
We at Apex.AI have a strict requirement not to link against GPL-licensed libraries and code, and this strict requirement comes to us from big customers with well-known names.  
Please note that at the same time it is ok to use standalone GPL-licensed tools or libraries which is not part of the product.

Curious, if we can include in the `lz4_vendor` package only BSD licensed part i.e. `${lz4_SOURCE_DIR}/lib/*.c`?
It will solve the licensing issue.

clalancette:
> @clalancette My major concern is a viral "infection" of other libraries and executables by the `GPLv2` license from the `lz4` library.

I honestly don't think this is going to be much of a difference.  In particular, we are still only building those same BSD-licensed files; the only difference is that we are now using a bit more of the build infrastructure.  But I am not a lawyer and I cannot give legal advice.

In all honesty, if this is a huge concern we should not be using lz4 at all, and should switch to something where the license is clearer.

cottsay:
@ros-pull-request-builder retest this

lz4_vendor/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+ament_vendor(lz4_vendor
+  SATISFIED ${lz4_FOUND}
+  VCS_URL https://github.com/lz4/lz4.git
+  VCS_VERSION v1.9.4
+  SOURCE_SUBDIR build/cmake
+  CMAKE_ARGS
+    -DBUILD_STATIC_LIBS:BOOL=OFF
+    -DBUILD_SHARED_LIBS:BOOL=${BUILD_SHARED_LIBS}
+    -DLZ4_BUILD_LEGACY_LZ4C:BOOL=OFF
+    -DLZ4_BUILD_CLI:BOOL=OFF
```
MichaelOrlov: @clalancette According to the https://github.com/lz4/lz4/blob/v1.9.4/build/cmake/CMakeLists.txt 
If using these two flags settled to `OFF`
```
    -DLZ4_BUILD_LEGACY_LZ4C:BOOL=OFF
    -DLZ4_BUILD_CLI:BOOL=OFF
```
CMake will use only files from the `\lib` subfolder to compile liblz4 library.
Therefore the library will contain only those part which is licensed under the `BSD license`. Perhaps we can remove mentions about GPLv2 license in this case. Thoughts?
Maybe add a comment for these two flags that we settling them OFF to ensure to use only the BSD licensed part?

clalancette: > Therefore the library will contain only those part which is licensed under the `BSD license`. Perhaps we can remove mentions about GPLv2 license in this case. Thoughts?

Ah, that's good to know, thanks.  Let's talk about it at the meeting tomorrow.

> Maybe add a comment for these two flags that we settling them OFF to ensure to use only the BSD licensed part?

Yes, I think this is a good idea regardless.

MichaelOrlov: @clalancette There is one caveat with it. 
As far as I understand if `liblz4` library will be found from the Ubuntu distro it will not be compiled and these flags will not be taken into account.  And Since `LZ4_BUILD_LEGACY_LZ4C` and `LZ4_BUILD_CLI` settled to `ON` by default the `liblz4` library from Ubuntu distros contain `GPLv2`  code.

clalancette: > As far as I understand if `liblz4` library will be found from the Ubuntu distro it will not be compiled and these flags will not be taken into account. And Since `LZ4_BUILD_LEGACY_LZ4C` and `LZ4_BUILD_CLI` settled to `ON` by default the `liblz4` library from Ubuntu distros contain `GPLv2` code.

I don't think this is going to be a problem.

I pulled down the orig.tar.gz and debian.tar.xz from https://packages.ubuntu.com/noble/liblz4-1, and then built them locally with `debuild -uc -us`.  When building, I see the following:

```
make[3]: Entering directory '/home/ubuntu/lz4_compile/lz4-1.9.4/lib'
compiling static library
x86_64-linux-gnu-gcc -g -O2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/ubuntu/lz4_compile/lz4-1.9.4=. -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fdebug-prefix-map=/home/ubuntu/lz4_compile/lz4-1.9.4=/usr/src/lz4-1.9.4-1 -Wdate-time -D_FORTIFY_SOURCE=3 -DLZ4F_PUBLISH_STATIC_FUNCTIONS  -Wdate-time -D_FORTIFY_SOURCE=3 -DLZ4F_PUBLISH_STATIC_FUNCTIONS -DXXH_NAMESPACE=LZ4_   -c lz4.c lz4file.c lz4frame.c lz4hc.c xxhash.c
compiling dynamic library 1.9.4
x86_64-linux-gnu-gcc -g -O2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/home/ubuntu/lz4_compile/lz4-1.9.4=. -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fdebug-prefix-map=/home/ubuntu/lz4_compile/lz4-1.9.4=/usr/src/lz4-1.9.4-1 -Wdate-time -D_FORTIFY_SOURCE=3 -DLZ4F_PUBLISH_STATIC_FUNCTIONS  -Wdate-time -D_FORTIFY_SOURCE=3 -DLZ4F_PUBLISH_STATIC_FUNCTIONS -DXXH_NAMESPACE=LZ4_  -Wl,-Bsymbolic-functions -flto=auto -ffat-lto-objects -Wl,-z,relro -shared lz4.c lz4file.c lz4frame.c lz4hc.c xxhash.c -fPIC -fvisibility=hidden -Wl,-soname=liblz4.so.1 -o liblz4.so.1.9.4
```

As far as I can tell, both the static and the shared library only use files from the lib directory.  It is true that the rest of the build goes on to build the programs and so on, but those will never be directly linked against by `rosbag2` or `mcap`, so we should be fine.

clalancette:
All right.  As we discussed, my latest push here renames the package to `liblz4_vendor` to emphasize we are only vendoring the library itself, none of the surrounding bits.  I also added some explanation to the package.xml and the README.md about this.

Note that this now requires https://github.com/ros/rosdistro/pull/40263 to go in first, since that makes an explicit distinction between liblz4 (the runtime library) and liblz4-devel (the development libraries).

@cottsay @MichaelOrlov Please take another look when you get a chance.

MichaelOrlov (approved):
@clalancette :+1:  LGTM.

clalancette:
Here is CI for it.  Note that we can't merge this until we merge https://github.com/ros/rosdistro/pull/40263, but at least we can be ready to go:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20499/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14908/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21250/)

mcap_vendor/CMakeLists.txt
```diff
@@ -18,7 +20,7 @@ if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_EXTENSIONS OFF)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic -Werror)
```
defunctzombie, line 21: Why the removal of `-Werror`? Was something failing?

defunctzombie:
@clalancette I know you wrote this as the justification - but I'm wondering about the justification that led to this being the approach?

> This is more explicit, and more like how we vendor the rest of the packages in our ecosystem.

Does this open the possibility that other packages can depend on this new _vendor_ package? What if you don't want that to be the case and prefer to have this as an internal dependency?

What you have here seems fine but I was curious at the reasoning for making something that was previously internal now external.

mcap_vendor/CMakeLists.txt
clalancette: As a general policy, I think using `-Werror` is a mistake, and we don't use it anywhere else in the ROS 2 core.

The problem is that while we can, and do, make things compile warning-free on our Tier-1 platforms, ROS 2 is used across a wide variety of platforms, with a wide variety of compilers.  Turning warnings into errors just makes it annoying for those people not using our Tier-1 platforms.

clalancette:
> Does this open the possibility that other packages can depend on this new _vendor_ package? What if you don't want that to be the case and prefer to have this as an internal dependency?

It does open this possibility, yes.

However, I will point out a few things:
1. On Linux (both Ubuntu and RHEL), this will actually not compile lz4 at all, and instead use the system version.  That's our typical policy, as it lets ROS 2 use the same system libraries as everything else.
2. On Windows, you are right that this will make lz4 available for other packages to depend upon.
3. It is possible to make a library more "private"; we do that with [rviz_ogre_vendor](https://github.com/ros2/rviz/blob/2273179562a451fe5382839d50a46afa5e1af076/rviz_ogre_vendor/CMakeLists.txt), for instance.  I don't think it is necessary in this case, particularly because we are going to be using the system version whenever possible here.

defunctzombie:
> 1. On Linux (both Ubuntu and RHEL), this will actually not compile lz4 at all, and instead use the system version.  That's our typical policy, as it lets ROS 2 use the same system libraries as everything else

FWIW I've found this a hinderance in the past. System versions can sometimes be out of date and also drift which makes reasoning about which actual version of the library you have built against difficult. At a previous place of work we had a large effort to never depend on any system library for the robotics stack. Might be something to consider for the future.

clalancette:
> FWIW I've found this a hinderance in the past. System versions can sometimes be out of date and also drift which makes reasoning about which actual version of the library you have built against difficult. At a previous place of work we had a large effort to never depend on any system library for the robotics stack. Might be something to consider for the future.

Heh.  That is the exact opposite of our policy, which is to use system versions wherever possible.  If you don't do this, then you make it dangerous to use the system version of *anything*, because then you run the risk of multiple libraries presenting the same API but with different ABIs.  It is part of being in the ecosystem.

cottsay:
Worth calling out is that you can absolutely specify the `FORCE_BUILD_VENDOR_PKG:BOOL=ON` CMake argument to ignore the system packages and always build vendor payloads from source.

I think that for a general user, the risk of ABI collision, additional build time, and additional file size, are not worth the cost, but there has always been a mechanism to opt-out of using system packages for vendor payloads.

clalancette:
Usual CI is happy.  I'm going to run a couple of additional pieces of CI just to make sure:

* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/606/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21267/)

clalancette:
All right, all of the CI failures are known flakes.  We still have to merge in https://github.com/ros/rosdistro/pull/40263 first, so I'll wait for that to happen before merging this.

clalancette:
All right, I've merged in all of the necessary changes elsewhere.  I'm going to run one more set of CI on this after rebasing on the latest, then merge this in.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20723/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15032/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21407/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21412/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/629/)

clalancette:
I kind of lost this one in the shuffle, but it looks like CI was all green, and there have been no changes to this repository since then.  It is also approved, so going ahead and merging this one.

Issue #1099, opened 2022-09-25
Snapshot should write to a different file every time it is triggered


bijoua29:
## Description
When you trigger multiple snapshots through the service call, each snapshot data is written to the same file. This is not very useful for multiple snapshots.

The typical use-case for the snapshot feature is to save a snapshot when the user see something "interesting". This might happen multiple times. Usually triggering a snapshot is not always a one-time event.

## Related Issues

N/A

## Completion Criteria

- Save to a different file when a snapshot is triggered
- Add ability to specify filename for each snapshot or allow option for timestamped file

## Implementation Notes / Suggestions
N/A

## Testing Notes / Suggestions
N/A

swiz23:
I second this feature request. This seems helpful.

wvergouwenavular:
+1

emersonknapp:
Note - the recorder does provide the service `split_bagfile` - you could call that immediately after the snapshot to split. Though, I think it's not unreasonable to provide a `split` option in `Snapshot.srv` so that it can happen atomically. PRs welcome!

Timple:
Shall we extend the `Snapshot.srv` from this:
```
---
bool success
```
To this?
```
string filename   # Name of bag file to save snapshot to.
---
bool success
```
In line with this: https://github.com/ros/rosbag_snapshot/blob/main/rosbag_snapshot_msgs/srv/TriggerSnapshot.srv

We probably need to come up with a way to append the timestamp as well. Perhaps if we enter something like `my_file_{stamp}` that this would result in `my_file_2024_01_22-14_18_28/`?
Suggestions welcome

MichaelOrlov:
@Timple Extend `Snapshot.srv ` with a few new parameters is not a big deal or a problem.  
However, it turns out that it is not trivial to implement this feature on the lower rosbag2 layers.  
There is also another request to extend `snapshot` feature to be able to write a buffer after the event as well. 
These two extensions are sort of overlaps in implementation and even alone are not trivial to implement.

bijoua29:
I guess I'm surprised that just adding the ability to name the file seems not trivial to implement. But I have no knowledge of the structure of the code.

Timple:
Would it make sense to align on the service call syntax already? 

ROS2 Jade is coming up, and changing a service after release won't be possible I guess. Implementation can then be worked on after the interface is agreed.

bijoua29:
Yes, it would make sense on the service call syntax. Was there a proposal for an updated syntax?

[...]

PR #1584, opened 2024-03-14
Improve logic of snapshot feature

dkorenkovnl:
For enabling snapshot feature you need to run bag node with the following parameters
ros2 bag record --max-cache-size 100000 --snapshot-mode /topic_you_want_to_record

The key --max-cache-size 100000 makes SequentialWriter to use rosbag2_cpp::cache::CircularMessageCache
and rosbag2_cpp::cache::CacheConsumer to process messages.
The key --snapshot-mode will enable rosbag2_interfaces::srv::Snapshot service.

The disadvantages of current implementation is that we are still writing data to the same file.
This PR will provide functionality of saving snapshot to separate file each time when we are calling
rosbag2_interfaces::srv::Snapshot service

For enabling new feature you need to provide the following parameters
storage_options_.max_bagfile_size = 1024*100;
storage_options_.storage_id = "mcap_snapshot";

### DIFF
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -841,11 +841,244 @@ void MCAPStorage::ensure_rosdistro_metadata_added()
   has_added_ros_distro_metadata_ = true;
 }
 
+/**
+ * A storage implementation for the MCAP file format of snapshots.
+ */
+class ROSBAG2_STORAGE_MCAP_PUBLIC MCAPStorageSnapshot
+: public rosbag2_storage::storage_interfaces::ReadWriteInterface
+{
+public:
+  MCAPStorageSnapshot();
+  ~MCAPStorageSnapshot() override;
+  /** BaseIOInterface **/
+  #ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
+    void open(const rosbag2_storage::StorageOptions & storage_options,
+              rosbag2_storage::storage_interfaces::IOFlag io_flag =
+                rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE) override;
+    void open(const std::string & uri, rosbag2_storage::storage_interfaces::IOFlag io_flag =
+                                         rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE);
+  #else
+    void open(const std::string & uri,
+            rosbag2_storage::storage_interfaces::IOFlag io_flag =
+            rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE) override;
+  #endif
+
+
+  /** BaseInfoInterface **/
+  rosbag2_storage::BagMetadata get_metadata() override;
+  std::string get_relative_file_path() const override;
+  uint64_t get_bagfile_size() const override;
+  std::string get_storage_identifier() const override;
+
+  /** BaseReadInterface **/
+  #ifdef ROSBAG2_STORAGE_MCAP_HAS_SET_READ_ORDER
+  bool set_read_order(const rosbag2_storage::ReadOrder &) override;
+  #endif
+  bool has_next() override;
+  std::shared_ptr<rosbag2_storage::SerializedBagMessage> read_next() override;
+  std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() override;
+  void get_all_message_definitions(
+          std::vector<rosbag2_storage::MessageDefinition> & definitions) override;
+
+  /** ReadOnlyInterface **/
+  void set_filter(const rosbag2_storage::StorageFilter & storage_filter) override;
+  void reset_filter() override;
+  #ifdef ROSBAG2_STORAGE_MCAP_OVERRIDE_SEEK_METHOD
+  void seek(const rcutils_time_point_value_t & time_stamp) override;
+  #else
+  void seek(const rcutils_time_point_value_t & time_stamp);
+  #endif
+
+  /** ReadWriteInterface **/
+  uint64_t get_minimum_split_file_size() const override;
+
+  /** BaseWriteInterface **/
+  void write(std::shared_ptr<const rosbag2_storage::SerializedBagMessage> msg) override;
+  void write(
+          const std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> & msg) override;
+  void create_topic(const rosbag2_storage::TopicMetadata & topic,
+                    const rosbag2_storage::MessageDefinition & message_definition) override;
+  void remove_topic(const rosbag2_storage::TopicMetadata & topic) override;
+  #ifdef ROSBAG2_STORAGE_MCAP_HAS_UPDATE_METADATA
+  void update_metadata(const rosbag2_storage::BagMetadata &) override;
+  #endif
+
+private:
+   std::shared_ptr<MCAPStorage> storage_;
+   std::string uri_;
+   rosbag2_storage::storage_interfaces::IOFlag io_flag_ = rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE;
+   rosbag2_storage::BagMetadata metadata_;
+   std::vector<std::pair<rosbag2_storage::TopicMetadata,rosbag2_storage::MessageDefinition>> topics_;
+   rosbag2_storage::StorageOptions storage_options_;
+   uint64_t current_storage_size_;
+};
+
+MCAPStorageSnapshot::MCAPStorageSnapshot():storage_(nullptr),current_storage_size_(0)
+{
+}
+
+MCAPStorageSnapshot::~MCAPStorageSnapshot()
+{
+}
+
+/** BaseIOInterface **/
+#ifdef ROSBAG2_STORAGE_MCAP_HAS_STORAGE_OPTIONS
+void MCAPStorageSnapshot::open(const rosbag2_storage::StorageOptions & storage_options,
+                       rosbag2_storage::storage_interfaces::IOFlag io_flag)
+{
+  (void)io_flag;
+  storage_options_ = storage_options;
+  uri_ = storage_options.uri;
+}
+#endif
+
+void MCAPStorageSnapshot::open(const std::string & uri, rosbag2_storage::storage_interfaces::IOFlag io_flag)
+{
+  uri_ = uri;
+  //this class is used only for READ_WRITE
+  (void)io_flag;
+}
+
+/** BaseInfoInterface **/
+rosbag2_storage::BagMetadata MCAPStorageSnapshot::get_metadata()
+{
+  return metadata_;
+}
+
+std::string MCAPStorageSnapshot::get_relative_file_path() const
+{
+  return uri_;
+}
+
+uint64_t MCAPStorageSnapshot::get_bagfile_size() const
+{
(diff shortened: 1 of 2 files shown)


rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -841,12 +841,245 @@ void MCAPStorage::ensure_rosdistro_metadata_added()
...
+  #else
+    void open(const std::string & uri,
+            rosbag2_storage::storage_interfaces::IOFlag io_flag =
+            rosbag2_storage::storage_interfaces::IOFlag::READ_WRITE) override;
+  #endif
+
+
+  /** BaseInfoInterface **/
+  rosbag2_storage::BagMetadata get_metadata() override;
+  std::string get_relative_file_path() const override;
```
Timple, line 869: I assume the original function did the same thing? So why override it? Just leave it out I think.

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -841,12 +841,245 @@ void MCAPStorage::ensure_rosdistro_metadata_added()
...
+std::string MCAPStorageSnapshot::get_storage_identifier() const
+{
+  return "mcap_snapshot";
+}
+
+/** BaseReadInterface **/
+#ifdef ROSBAG2_STORAGE_MCAP_HAS_SET_READ_ORDER
+bool MCAPStorageSnapshot::set_read_order(const rosbag2_storage::ReadOrder & read_order)
+{
+  (void)read_order;
```
Timple: Don't `(void)` variables. Declare them as comments in the function definition:

```c++
bool MCAPStorageSnapshot::set_read_order(const rosbag2_storage::ReadOrder & /*read_order*/)
```

MichaelOrlov:
@dkorenkovnl link to the relevant issue and some explanation of what this PR is about would very helpfull.

dkorenkovnl:
I will close this PR because it is not ready to be public yet)

dkorenkovnl:
PR is ready for review

rosbag2_storage_mcap/src/mcap_storage.cpp
dkorenkovnl: fixed

MichaelOrlov:
@dkorenkovnl As regards
> The disadvantages of current implementation is that we are still writing data to the same file.
This PR will provide functionality of saving snapshot to separate file each time when we are calling
rosbag2_interfaces::srv::Snapshot service

Better to start a discussion in an issue rather than doing implementation right away.

The fact that snapshot writing to the same file is a well known "feature". And currently the workaround exists for it.
i.e. Need to send `split_bag` service call each time right after calling the snapshot service call. 

If you want to make a really it as a new feature to be able to save snapshots in a new file each time. The following shall be considered:
1. It shall be done on the upper layer in the `rosbag2_cpp` package to be available for all storage plugins and storage formats.
2. Need to add an optional parameter to the `rosbag2_interfaces::srv::Snapshot` service to trigger this behavior.

dkorenkovnl:
Hi @MichaelOrlov !
We have such discussion [here](https://github.com/ros2/rosbag2/issues/1099).

>The fact that snapshot writing to the same file is a well known "feature". And currently the workaround exists for it. i.e. Need to send split_bag service call each time right after calling the snapshot service call.

A couple thought about it current implementation:
- the procedure of creation of snapshot is async nevertheless that service return us true
and we do not know right now when snapshot is truly be ready
- it would be good to know the name of the file

>If you want to make a really it as a new feature to be able to save snapshots in a new file each time. The following shall be considered:

>It shall be done on the upper layer in the rosbag2_cpp package to be available for all storage plugins and storage formats.
>Need to add an optional parameter to the rosbag2_interfaces::srv::Snapshot service to trigger this behavior.

It was my first try but when we did so that broke down some unit tests and then I remembered SOLID design principle,
specially Open/Closed Principle(software entities should be open for extension, but closed for modification)

I would like to remark that Class SequentialWriter created in a good way. It contains basic algorithm that operate of abstractions(converter_,storage_) but the details are closed inside of implementation(MCAPStorage,SqliteStorage).

And finally for up and run current changes you need to change only configuration.

Timple:
This would indeed open the door to implementing a preferred filename as suggested here: https://github.com/ros2/rosbag2/issues/1099#issuecomment-1903990853

MichaelOrlov:
@dkorenkovnl A couple questions to clarify:
> - the procedure of creation of snapshot is async nevertheless that service return us true
and we do not know right now when snapshot is truly be ready

It was made async and returned true immediately on purporse because writing a snapshot could take a significant time.
The question is, let's assume that snapshot service call will be synchronous and will not return confirmation until the file will be written - how it will help? What is a real-world scenario?

> - it would be good to know the name of the file

I think we can know the filename in advance and can return it before finishing writing.

dkorenkovnl:
@MichaelOrlov 
>It was made async and returned true immediately on purporse because writing a snapshot could take a significant time.
The question is, let's assume that snapshot service call will be synchronous and will not return confirmation until the file will be written - how it will help? What is a real-world scenario?

The suggestion from my side is to use Action Server.(but it not scope of this PR)
From user point of view is when he calls "/snapshot" and return back responce with bool success = true
he is thinking that snapshot is ready but it is not 
and without diving in src code he could not understand that.

>I think we can know the filename in advance and can return it before finishing writing.

Yes I agree that bunch "/snapshot" and "/split_bagfile" services doing current job well.
And we could use "/split_bagfile" to return name of the created bag file. @Timple  what do you think about that?

Timple:
> bunch "/snapshot" and "/split_bagfile" services doing current job well.

I'm not quite sure about this. 

1. This means I have to make two service calls (minor)
2. Files are created even if the `/snapshot` was never called
3. The filename is completely unrelated to the file content
The filename says: 12:45:00 but the snapshot contains 30 second data from 16:15:10 to 16:15:40.
4. We still don't control the filename. 
I would like a file for example: `bagfile_{stamp}_emergency_stop/` so that these can easily be spotted between other bagfiles (main reason for https://github.com/ros2/rosbag2/issues/1099#issuecomment-1903990853)
5. Even if the filename was returned, we cannot rename it in hindsight. As we don't know when writing is finished. And the filename is embedded in the metadata.

So that's why I though this PR was a good start on getting filename as a request variable instead of a result variable.

_I numbered the items be able to address them on a per-item basis, not to sound provocative_ :slightly_smiling_face:

MichaelOrlov:
@Timple @dkorenkovnl I understand your concerns about the proposed workaround by calling the "/snapshot" and "/split_bagfile"  services one by one in sequence. But please recognize that this is a short-term remedy and still a workaround rather than a work-as-designed approach. You are not the first one who has concerns about the snapshot feature. However, it turns out that it is not so straightforward to implement writing a snapshot in a separate file in a proper way.
At the same time, we can't move forward with this PR because the way you implemented this feature contradicts the general rosbag2 design. The proper way to implement this feature will be to make changes on the `rosbag2_cpp` and `rosbag2_transport` layers rather than trying to create an ad hoc solution by creating another storage plugin inside an already existent mcap storage plugin.

I will try to answer most of your concerns regarding the proposed workaround and listed above.
> 1. This means I have to make two service calls (minor) 

[mo]: Yes. But this is a workaround.

> 2. Files are created even if the /snapshot was never called

[mo]: IMO. This is not true since if a snapshot is never called, then "/split_bagfile" is also never called - no files will be created. The only caveat could be that we might have an empty last file. I don't remember exactly, but we might already have a mechanism to prevent it; need to double-check. Anyway, it shouldn't be a real problem.

> 3. The filename is completely unrelated to the file content
The filename says: 12:45:00 but the snapshot contains 30 second data from 16:15:10 to 16:15:40.

[mo]: That sounds like a `bug` or at least unexpected behavior. Could you please create a separate issue for this case with a clear and concise explanation of how to reproduce it or with minimum valuable reproducer on some branch?

> 4. We still don't control the filename.
I would like a file for example: bagfile_{stamp}_emergency_stop/ so that these can easily be spotted between other bagfiles (main reason for https://github.com/ros2/rosbag2/issues/1099#issuecomment-1903990853)

[mo]: Yes, this is currently a limitation. Sorry about this. Please see the answer to question 5 for a potential workaround.

> 5. Even if the filename was returned, we cannot rename it in hindsight. As we don't know when writing is finished. And the filename is embedded in the metadata.

[mo]: To be able to know a moment when the current file writing finished after the "~/split_bagfile" event, you can subscribe to the topic "events/write_split". Rosbag2 will publish a special message on this topic to notify remote nodes that the newly written file is ready and a new one has been created. The old and new filenames are part of this message in those notification. Yes, we have such a feature in the rosbag2, and sorry that we don't have documentation for it. Please refer to the following part of the code to better understand how to use it.
https://github.com/ros2/rosbag2/blob/30328006f872de4e1c82fef4ecca8ca815b2d4dc/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L321-L337
and 
https://github.com/ros2/rosbag2/blob/30328006f872de4e1c82fef4ecca8ca815b2d4dc/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L356-L376
The key info there is 
```
       auto message = rosbag2_interfaces::msg::WriteSplitEvent(); 
       message.closed_file = bag_split_info_.closed_file; 
       message.opened_file = bag_split_info_.opened_file; 
```

dkorenkovnl:
> [mo]: IMO. This is not true since if a snapshot is never called, then "/split_bagfile" is also never called - no files will be created. >The only caveat could be that we might have an empty last file. I don't remember exactly, but we might already have a >mechanism to prevent it; need to double-check. Anyway, it shouldn't be a real problem.

@MichaelOrlov just checked one more time and it is true :
-we create file right after ros2 bag record was run
-we create file after we called split_bagfile service

I got that result by testing a branch iron.
Steps:
ros2 topic pub -r 10 /greetings example_interfaces/msg/String "{data: 'Hello from terminal'}"
ros2 bag record --max-cache-size 100000 --snapshot-mode /greetings

ros2 service call /rosbag2_recorder/snapshot rosbag2_interfaces/Snapshot "{}"
ros2 service call  /rosbag2_recorder/split_bagfile   rosbag2_interfaces/SplitBagfile

MichaelOrlov:
@dkorenkovnl As regards
> @MichaelOrlov just checked one more time and it is true :
-we create file right after ros2 bag record was run
-we create file after we called split_bagfile service

That looks like expected and correct behavior.
After the Steps:
> ros2 topic pub -r 10 /greetings example_interfaces/msg/String "{data: 'Hello from terminal'}"
> ros2 bag record --max-cache-size 100000 --snapshot-mode /greetings

> ros2 service call /rosbag2_recorder/snapshot rosbag2_interfaces/Snapshot "{}"
> ros2 service call /rosbag2_recorder/split_bagfile rosbag2_interfaces/SplitBagfile

How many `bag` files do you have?

MichaelOrlov:
I've tried to reproduce and got a 2 mcap files the last one is without any messages only metadata inside - you can disregard the last file in this case.

MichaelOrlov:
- Closing this PR as will not be merged.
Because the way this feature is implemented in this PR contradicts the general rosbag2 design. The proper way to implement this feature will be to make changes on the rosbag2_cpp and rosbag2_transport layers rather than trying to create an ad hoc solution by creating another storage plugin inside an already existing mcap storage plugin.

PR #1589, opened 2024-03-16
Add Doxyfile

### LINKED ISSUE #1412: rosdoc2 Documentation (Doxyfile)
## Description

Create Doxyfiles for the projects in this repo, making at least https://build.ros2.org/view/Rdoc/job/Rdoc__rosbag2__ubuntu_jammy_amd64/ pass for all packages present in this repo, therefore having an entry for all of them in https://docs.ros.org/en/rolling/p/

This is a very basic start, but would get us on the road to better documentation, and a place to put documentation so it's discoverable. Rosbag2 is a very important tool for ROS users, but it's pretty much always been the case that only the contributors really know all the features it has available.

sangteak601:
Fixes #1412. Added basic doxygen configuration files.
Done with reference to [rclcpp](https://github.com/ros2/rclcpp).

### DIFF
--- a/rosbag2_compression/Doxyfile
+++ b/rosbag2_compression/Doxyfile
@@ -1,0 +1,20 @@
+# All settings not listed here will use the Doxygen default values.
+
+PROJECT_NAME           = "rosbag2_compression"
+PROJECT_NUMBER         = rolling
+PROJECT_BRIEF          = "Compression implementations for rosbag2 bags and messages"
+
+INPUT                  = ./include
+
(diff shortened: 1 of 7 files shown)


sangteak601:
I'm not very sure how I can do with `rosbag2_py` package. Any help would be appreciated.

MichaelOrlov:
@clalancette @fujitatomoya I  am not familiar with Dixygenb config files and the way how they should be generated.
If you are, please help with review for this PR.

PR #1577, opened 2024-02-28
Filter topic by type 

ahcorde:
It will be great to be able to filter topic by type.

Example usage:

```bash
ros2 bag record --topic_types sensor_msgs/msg/Image sensor_msgs/msg/CameraInfo
```

PR #1593, opened 2024-03-20
Update rosbag2_py stubs

r7vme:
Related to a recent change that added stub files https://github.com/ros2/rosbag2/pull/1569

PR above was not rebased to the latest rolling, which had [pybind11 changes](https://github.com/ros2/rosbag2/pull/1577/files#diff-c85fd024e7d68c762d60d25e2ba155bcc527f04c937ed09adc893ea241fdc61c), before merging. This caused a "race condition" - CI was green, but after rebasing it would fail. Seems, when making changes to rosbag2_py (pybind11), we need to make sure to rebase to latest rolling before merge.

### DIFF
--- a/rosbag2_py/rosbag2_py/_transport.pyi
+++ b/rosbag2_py/rosbag2_py/_transport.pyi
@@ -52,6 +52,7 @@ class RecordOptions:
     start_paused: bool
     topic_polling_interval: datetime.timedelta
     topic_qos_profile_overrides: dict
+    topic_types: List[str]
     topics: List[str]
     use_sim_time: bool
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
@r7vme thank you for catching this!

MichaelOrlov (approved):
@r7vme Thanks. LGTM.
Good point about rebase.

MichaelOrlov:
Since changes only in auto-generated `pyi` file which is not participating in the build and test on the CI build farm I am merging this PR without running the full CI pipeline.

PR #1594, opened 2024-03-21
Check existence of a file before passing it to the mcap reader

cwecht:
…mcap::FileStreamReader in order to avoid this assertion: failing https://github.com/foxglove/mcap/blob/9d73125e50121c1eeb2ceb9d13db51ff923b0420/cpp/mcap/include/mcap/reader.inl#L90

This is an issue if you happen to run rosbag build in debug mode (in which assertions are enabled). As it is the behavior of debug and release mode is inconsistent as in debug mode the assertion is raised and in release mode an exception is raised.

### DIFF
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -14,6 +14,7 @@
 
 #include "rcpputils/thread_safety_annotations.hpp"
 #include "rcutils/logging_macros.h"
+#include "rcutils/strerror.h"
 #include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/ros_helper.hpp"
 #include "rosbag2_storage/storage_interfaces/read_write_interface.hpp"
@@ -336,7 +337,9 @@ void MCAPStorage::open_impl(const std::string & uri, const std::string & preset_
       relative_path_ = uri;
       input_ = std::make_unique<std::ifstream>(relative_path_, std::ios::binary);
       if (!input_->is_open()) {
-        throw std::runtime_error(std::strerror(errno));
+        char error_string[1024] = {};
+        rcutils_strerror(error_string, sizeof(error_string));
+        throw std::runtime_error(error_string);
       }
       data_source_ = std::make_unique<mcap::FileStreamReader>(*input_);
       mcap_reader_ = std::make_unique<mcap::McapReader>();
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -40,6 +40,7 @@
 #include <mcap/mcap.hpp>
 
 #include <algorithm>
+#include <cstring>
 #include <memory>
 #include <mutex>
 #include <optional>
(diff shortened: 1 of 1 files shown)


rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -334,6 +335,9 @@ void MCAPStorage::open_impl(const std::string & uri, const std::string & preset_
     case rosbag2_storage::storage_interfaces::IOFlag::READ_ONLY: {
       relative_path_ = uri;
       input_ = std::make_unique<std::ifstream>(relative_path_, std::ios::binary);
+      if (!input_->is_open()) {
+        throw std::runtime_error(std::strerror(errno));
```
fujitatomoya: if `input_` failed to open the file during construction, the errno is set accordingly? i was expecting we need to check the failbit instead, but it would be better that we can just use errno!

How about adding more information for runtime error exception?

```suggestion
        throw std::runtime_error("Failed to open storage " + relative_path_ + std::strerror(errno));
```

cwecht: 1. std::ifstream and errno is a bit complicated: cppreference (hence the C++ standard) don't mention errno at all. However, std::ifstream is usually implemented in terms of fopen/open which in turn is specified to set errno. I think, it is reasonable to realy on this behavior. An alternative could be to set the [execption mask ](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions) of `input_ `. IMO this has the disadvantage that we would need to disable it after that probably. Otherwise there might occour other exceptions in `FileStreamReader`, which might not be expected or not dealt with properly.
2. Regarding the exception message: this exception will be catched here: https://github.com/ros2/rosbag2/blob/rolling/rosbag2_storage/src/rosbag2_storage/impl/storage_factory_impl.hpp#L135 Here, the path is logged already, so there is no need to duplicate this infromation in the exception message.

fujitatomoya: thanks for the explanation! lgtm!

fujitatomoya (approved):

MichaelOrlov (approved):
LGTM

MichaelOrlov:
Gist: https://gist.githubusercontent.com/MichaelOrlov/6dee8d20cb96a4d63b417bf6df46932e/raw/624b11e64730f3df462ee2425c8fcd8c6ca7fa78/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13570
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20661/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14992/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21364/)

MichaelOrlov:
@cwecht There is one new warning on Windows CI build
> 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.

Could you please address it by using `strerror_s ` instead of the `strerror`?

cwecht:
> @cwecht There is one new warning on Windows CI build
> 
> > 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
> 
> Could you please address it by using `strerror_s ` instead of the `strerror`?

Turns out there is already `rcutils_strerror` so I went with that.

fujitatomoya:
@cwecht change looks good, CI just starts. 🤞 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20703/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15017/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21392/)

cwecht:
https://build.ros2.org/job/Rpr__rosbag2__ubuntu_noble_amd64/62/console failing seems to be caused by some kind of hicup/issue on the toolchains/buildservers side, right? In case we need that build to successes, I'd suggest to rerun it.

cwecht:
@fujitatomoya @MichaelOrlov how do we proceed?

MichaelOrlov:
@cwecht The buildfarm CI is green we are good to go. Never mind.

PR #1510, opened 2023-12-12
Update documentation about recorder and player creation via composition

roncapat:
Closes #1509.

Feedbacks welcomed :)

PR #1603, opened 2024-04-09
[WIP] Make Player a `rosbag2` component (instead of `rosbag2_transport`)

### LINKED ISSUE #1736: Strange location of recorder and player executables
## Description
We can run the recorder as a node by:
```
ros2 run rosbag2_transport recorder
```
This cannot be guessed but has to be reverse engineered from code.

### Expected Behavior
```
ros2 run rosbag2 recorder
```
or 
```
ros2 run ros2bag recorder
```

### Actual Behavior
Type `ros2 run rosbag2` and try tab-complete, this fails

## To Reproduce
These templates are getting rediculous.

## System
Jazzy

## Additional context
Yes, there the commanline cli of course. But in launchfiles this:

```XML
<node name="rosbag_operational" pkg="rosbag2" exec="recorder" />
```
Looks much better than
```XML
<executable cmd="ros2 bag record --node-name rosbag_operational"/>
```
As it's consistent with other stuff, better passes ros arguments etc. Also node renaming and remapping is much better.

roncapat:
Closes #1736

This is a proposal to address https://github.com/ros2/rosbag2/pull/1510/files#r1484600093.

CC @MichaelOrlov @fujitatomoya

NOTE: I compiled it and the component gets correclty exposed via `ros2 component types` but I haven't tested it yet.

If it works, I extend it also to `Recorder` and then it's ready for merge.


roncapat:
@MichaelOrlov may I ask if you are still interested in this, and if the proposed method is ok for you, so I can apply the same strategy for Recorder??

MichaelOrlov (commented):
Hi @roncapat, Yes, I am still interested in this PR.
However, it is too long to describe what needs to be changed.
Instead, I did prototyping and made a couple of direct commits to your branch. Please see https://github.com/ros2/rosbag2/pull/1603/commits/c00bf42186e5576a73a9a12befba125389adce7f and https://github.com/ros2/rosbag2/pull/1603/commits/4f6414e625ceac7a57275730196f62c357531a0b

Please review the changes and feel free to add a `composable_recorder.{hpp}cpp` file in a similar way.

I've tested loading a new composable rosbag2::Plyer in my local setup and can confirm that it is compiles and runs as expected. However, it would be good to add some unit tests.

To test it I ran in one terminal
```bash
ros2 run rclcpp_components component_container
```
in another terminal 
```bash
ros2 component load /ComponentManager rosbag2 rosbag2::Player -p storage.uri:=/recordings/talker/talker.mcap -p play.loop:=true
```
After that, you can try to run `ros2 component list` and `ros2 topic echo /topic` to make sure that `rosbag2::Player` composable node is up and running.

I took `talker.mcap` from the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/test/resources/mcap/talker/talker.mcap

rosbag2_transport/CMakeLists.txt
```diff
@@ -72,9 +72,11 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
   yaml-cpp::yaml-cpp
 )
 
+# rosbag2_transport::Player composable node is deprecated. Please use rosbag2::Player instead
```
MichaelOrlov, line 75: @clalancette Do you have any idea how to deprecate early added `"rosbag2_transport::Player"` registered rclcpp component node from further usage?

MichaelOrlov: Ahh never mind. I found a way how to put a deprecation notice. See https://github.com/ros2/rosbag2/pull/1603/commits/8751ebf56a3167e3542d5abd3e781d6eb42f3feb for details.

cc: @roncapat

roncapat:
@MichaelOrlov looks nice to me! As soon as I have a bit of time, I will do the same for Recorder class.

MichaelOrlov:
cc: @emersonknapp

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -86,11 +86,14 @@ class Player : public rclcpp::Node
...
   /// \param node_options Node options which will be used during construction of the underlying
   /// node.
+  [[deprecated("The rosbag2_transport::Player constructor for the composable player is deprecated."
+               "Please use rosbag2::Player constructor instead")]]
   ROSBAG2_TRANSPORT_PUBLIC
   explicit Player(const rclcpp::NodeOptions & node_options);
 
   /// \brief Default constructor and entry point for the composable player.
-  /// Will construct Player class and initialize play_options, storage_options from node
+  /// \note Uses from the rosbag2::Player constructor which is registered with ament resource index.
```
roncapat:
```suggestion
  /// \note Used from the rosbag2::Player constructor which is registered with ament resource index.
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -741,7 +741,11 @@ void RecorderImpl::warn_if_new_qos_for_subscribed_topic(const std::string & topi
 // Recorder public interface
 
 Recorder::Recorder(const rclcpp::NodeOptions & node_options)
-: Recorder("rosbag2_recorder", node_options) {}
+: Recorder("rosbag2_recorder", node_options) {
+  RCLCPP_WARN(this->get_logger(),
+  "The rosbag2_transport::Recorder composable node is deprecated."
+  " Please use rosbag2::Recorder instead");
+}
```
roncapat:
```suggestion
: Recorder("rosbag2_recorder", node_options)
{
  RCLCPP_WARN(this->get_logger(),
    "The rosbag2_transport::Recorder composable node is deprecated."
    " Please use rosbag2::Recorder instead");
}
```

roncapat:
@MichaelOrlov done, sorry for the delay.

MichaelOrlov (changes requested):
@roncapat Thank you for the follow-up.
The latest changes look mostly good to me. I pushed one commit with fixes for the CPPLint warnings.

What we are missing now is a basic integration unit tests to make sure that invoking composable player can be constructed and play messages and composable recorder can be constructed from CLI command and record messages.
In Rosbag2 we have a special package `rosbag2_tests`  for such type of integration tests where almost all required instrumentation already defined.

Could you please write a few integration tests? 
You can use [test_rosbag2_record_end_to_end.cpp](https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp) and [test_rosbag2_play_end_to_end.cpp](https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp) as a good reference and example.
Also feel free to use [cdr_test](https://github.com/ros2/rosbag2/tree/rolling/rosbag2_tests/resources/mcap/cdr_test) bag files from resource folder for playback test. 
For recording tests feel free to use [RecordFixture](https://github.com/ros2/rosbag2/blob/b48354198038fd4e8966035cafc46a7c6e6cb4c9/rosbag2_tests/test/rosbag2_tests/record_fixture.hpp#L46) class. There are already present instrumentation to handle recorded files from Rosbag2 recorder in the system `temp` folder.

Also could you please sign your https://github.com/ros2/rosbag2/pull/1603/commits/51d303f6d6acc6156eb55b034473190ce66ded44 commit to pass DCO check on CI?

PR #1605, opened 2024-04-10
Update to use yaml-cpp version 0.8.0.

clalancette:
In yaml-cpp 0.8.0, they updated to use a proper CMake target, i.e. yaml-cpp::yaml-cpp.  Update that here as well since we are updating our vendor to 0.8.0.

While I was in here, I did a bunch of cleanup to make fewer symbols externally visible.

Let me know if you want me to split the cleanups out into a separate PR; I'm happy to do that.

Depends on https://github.com/ros2/yaml_cpp_vendor/pull/48 , so should not be merged before that one.

### DIFF
--- a/rosbag2_storage_mcap/CMakeLists.txt
+++ b/rosbag2_storage_mcap/CMakeLists.txt
@@ -28,6 +28,8 @@ find_package(mcap_vendor REQUIRED)
 find_package(pluginlib REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(rosbag2_storage REQUIRED)
+find_package(yaml_cpp_vendor REQUIRED)
+find_package(yaml-cpp REQUIRED)
 
 ament_python_install_package(ros2bag_mcap_cli)
 
@@ -40,11 +42,12 @@ target_include_directories(${PROJECT_NAME} PUBLIC
 )
 target_compile_features(${PROJECT_NAME} PUBLIC c_std_99 cxx_std_17)
 target_compile_definitions(${PROJECT_NAME} PRIVATE "ROSBAG2_STORAGE_MCAP_BUILDING_DLL")
-target_link_libraries(${PROJECT_NAME}
+target_link_libraries(${PROJECT_NAME} PRIVATE
   mcap_vendor::mcap
   pluginlib::pluginlib
   rcutils::rcutils
   rosbag2_storage::rosbag2_storage
+  yaml-cpp::yaml-cpp
 )
 
 set(MCAP_COMPILE_DEFS)
@@ -109,6 +112,7 @@ if(BUILD_TESTING)
     rosbag2_storage::rosbag2_storage
     rosbag2_test_common::rosbag2_test_common
     ${std_msgs_TARGETS}
+    yaml-cpp::yaml-cpp
   )
   target_compile_definitions(test_mcap_storage PRIVATE ${MCAP_COMPILE_DEFS})
 
@@ -117,6 +121,5 @@ endif()
 
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(export_${PROJECT_NAME})
-ament_export_dependencies(mcap_vendor pluginlib rosbag2_storage rcutils)
 
 ament_package()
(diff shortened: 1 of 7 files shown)


ahcorde (approved):

rosbag2_storage_mcap/CMakeLists.txt
```diff
@@ -117,6 +121,5 @@ endif()
 
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(export_${PROJECT_NAME})
-ament_export_dependencies(mcap_vendor pluginlib rosbag2_storage rcutils)
```
MichaelOrlov, line 120: Out of curiosity, why remove export dependencies from the `pluginlib`?
The `rosbag2_storage_mcap` is a storage plugin.

MichaelOrlov (approved):
@clalancette Thanks for cleanups. LGTM with green CI.

rosbag2_storage_mcap/CMakeLists.txt
clalancette: > Out of curiosity, why remove export dependencies from the `pluginlib`?
> The `rosbag2_storage_mcap` is a storage plugin.

Good question.

The whole reason to put something in `ament_export_dependencies` is so that this dependency will be propagated to downstream libraries.  That is only needed in the case where the downstream libraries can "see" symbols from the dependency in the public headers.  As an example, if library `foo` depends on `bar`, and has `#include <bar/bar.h>` in its public `foo/foo.h` header file, then we need to export the dependency since libraries downstream to `foo` also need to have a dependency on `bar`.  If `foo` depends on `bar`, but only in private headers or implementation files, then it does not need to export the dependency; the dependency is an implementation detail.

Thus, when looking at the `target_link_libraries` lines, I only put things as PUBLIC if they are used as part of the public header files.  Everything else gets put as PRIVATE.  Following from that, only things that are PUBLIC get put into `ament_export_dependencies`; all PRIVATE dependencies do not need to be exported.

In this particular case, the fact that `rosbag2_storage_mcap` uses `pluginlib` is an implementation detail.  It matters to this library, and to the code that is going to load it, but downstream libraries that depend on `rosbag2_storage_mcap` don't need to care about this detail.

Does that make sense?

MichaelOrlov: @clalancette Thanks for the detailed explanations. Honestly, I thought we had dependencies from the pluginlib in the publicly exposed header file.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20814/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21498/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21499/)
* RHEL-9 [Build Status](http://ci.ros2.org/job/ci_linux-rhel/634/)

jyjblrd:
Hi, this is causing my builds to fail with the error:
```
CMake Error at /opt/ros/rolling/share/ament_cmake_gmock/cmake/ament_add_gmock_executable.cmake:50 (add_executable):
  Target "test_qos" links to target "yaml-cpp::yaml-cpp" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
Call Stack (most recent call first):
  /opt/ros/rolling/share/ament_cmake_gmock/cmake/ament_add_gmock_executable.cmake:37 (_ament_add_gmock_executable)
  /opt/ros/rolling/share/ament_cmake_gmock/cmake/ament_add_gmock.cmake:65 (ament_add_gmock_executable)
  CMakeLists.txt:129 (ament_add_gmock)
```

Pulling from the commit before this one resolves this issue. I noticed that the automated builds are also failing with this issue?

clalancette:
> Pulling from the commit before this one resolves this issue. I noticed that the automated builds are also failing with this issue?

Yeah, that will be the case until we do releases of all 4 of the connected PRs.  Given that we are just 2.5 working days from the freeze, we'll probably wait until Tuesday, April 16 to do that.  Until then, you'll have to build this along with https://github.com/ros2/yaml_cpp_vendor/pull/48

jyjblrd:
Ah, that makes sense. Sorry, was not aware of how ROS handles releases!

PR #1606, opened 2024-04-10
Switch rclpy to be an exec_depend in rosbag2_py.

clalancette:
That's because there is code in _storage.cpp
that imports rclpy.duration at runtime.

In particular, see https://github.com/ros2/rosbag2/blob/1bfe305cf3018fb59bc6571df303a919e7b4bd89/rosbag2_py/src/rosbag2_py/_storage.cpp#L37 and https://github.com/ros2/rosbag2/blob/1bfe305cf3018fb59bc6571df303a919e7b4bd89/rosbag2_py/src/rosbag2_py/_storage.cpp#L50

### DIFF
--- a/rosbag2_py/package.xml
+++ b/rosbag2_py/package.xml
@@ -24,13 +24,13 @@
   <depend>rosbag2_storage</depend>
   <depend>rosbag2_transport</depend>
 
+  <exec_depend>rclpy</exec_depend>
   <exec_depend>rpyutils</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>python3-pytest</test_depend>
   <test_depend>rcl_interfaces</test_depend>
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20818/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15122/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21507/)

MichaelOrlov (approved):
LGTM

MichaelOrlov:
@clalancette It seems that the rpr job is broken. Any thoughts or estimates on when it will get back to the normal state?

clalancette:
> @clalancette It seems that the rpr job is broken. Any thoughts or estimates on when it will get back to the normal state?

Unfortunately no.  We are still working through issues with Ubuntu 24.04, and I don't currently have an estimate when those will be fixed.

PR #1617, opened 2024-04-18
Updated zstd to 1.5.5

ahcorde:
Removed this warning
```bash
/root/ros2_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/compress/zstd_compress_superblock.c:412:12: warning: variable 'litLengthSum' set but not used [-Wunused-but-set-variable]
  412 |     size_t litLengthSum = 0;
```

PR #1627, opened 2024-04-25
Fixed warnings on RHEL

ahcorde:
version 1.5.5 creates new warnings on RHEL https://ci.ros2.org/job/ci_linux-rhel/638/clang-tidy/new/

This PR should fix them


ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20936/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15227/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21610/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/651/)

zstd_vendor/patches/0002.patch
```diff
@@ -0,0 +1,44 @@
+diff --git a/lib/legacy/zstd_v01.c b/lib/legacy/zstd_v01.c
+index 1a3aad07..10071134 100644
+--- a/lib/legacy/zstd_v01.c
++++ b/lib/legacy/zstd_v01.c
+@@ -1487,6 +1487,7 @@ static size_t ZSTDv01_decodeLiteralsBlock(void* ctx,
+     BYTE* const oend = ostart + maxDstSize;
+     blockProperties_t litbp;
+
++    memset(&litbp, 0, sizeof(litbp));
```
clalancette: Hm.  While I think this is mostly harmless, it also seems unnecessary.  That is, `ZSTDv01_getcBlockSize` below either returns an error, or full initializes the `blockProperties_t` structure.  If it returns an error, then the line `if (ZSTDv01_isError(litcSize)) return litcSize;` should get us out of here.  Thus, I don't see how the situations pointed out in https://ci.ros2.org/job/ci_linux-rhel/638/clang-tidy/new/source.d14c05bd-1262-4304-8320-b9a9520fe791/#1495 can happen.

@ahcorde Do you understand this better?  Why is `clang` complaining here?

ahcorde: These two warnings:

```bash
/root/rhel_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/legacy/zstd_v01.c: In function ‘ZSTD_decompressBlock’:
/root/rhel_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/legacy/zstd_v01.c:1504:20: warning: ‘litbp.origSize’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 1504 |             size_t rleSize = litbp.origSize;
      |                    ^~~~~~~
/root/rhel_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/legacy/zstd_v01.c:1488:23: note: ‘litbp.origSize’ was declared here
 1488 |     blockProperties_t litbp;
      |                       ^~~~~
/root/rhel_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/legacy/zstd_v01.c:1495:5: warning: ‘*(unsigned int *)((char *)&litbp + offsetof(blockProperties_t, blockType))’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 1495 |     switch(litbp.blockType)
      |     ^~~~~~
/root/rhel_ws/build/zstd_vendor/zstd_vendor-prefix/src/zstd_vendor/lib/legacy/zstd_v01.c:1488:23: note: ‘*(unsigned int *)((char *)&litbp + offsetof(blockProperties_t, blockType))’ was declared here
 1488 |     blockProperties_t litbp;
      |
```

Other solution is replace `memset` with 

```
+    litbp.blockType = 0;
+    litbp.origSize = 0;
```

MichaelOrlov: @clalancette As regards
> If it returns an error, then the line if (ZSTDv01_isError(litcSize)) return litcSize; should get us out of here.

I think that clang-tidy is not smart enough to figure out that `(ZSTDv01_isError(litcSize)` will be true for settled `ERROR(srcSize_wrong)` from the ` ZSTDv01_getcBlockSize(..)` at least it seems it doesn't connect this flow.

This is certainly a false positive warning in this case.

BTW if go with the second option
```
 + litbp.blockType = 0;
 + litbp.origSize = 0;
```
It would be more appropriate to init `litbp.blockType = bt_end`. In this case, we will return an error from the switch statement at least.

MichaelOrlov (commented):
I am curious about why these warnings appeared only on the RHEL9 ?
Do we have different compiler flags or linter configs?

clalancette (changes requested):
Actually, looking at this more, I think we should go in an entirely different direction.

Up until we merged in #1617, we actually weren't building zstd on RHEL *at all*, and we were just depending on the system version (which is 1.5.1 on RHEL).  I think we should go back to doing that; in particular, I think we should change https://github.com/ros2/rosbag2/blob/cb90408dcc7970057897f3476a4437b3fc40a2c5/zstd_vendor/CMakeLists.txt#L9 to say:

```
# The platform with the oldest supported version is RHEL-9,
# which has zstd 1.5.1.  Make sure we have at least that version.
find_package(zstd 1.5.1 QUIET)
```

clalancette (approved):
Looks good to me with green CI (including RHEL).

MichaelOrlov (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20957/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15256/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21633/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/675/)

clalancette:
The failures on aarch64 are known, so going ahead and merging.

ahcorde:

PR #1632, opened 2024-05-02
Add optional  '--topics' CLI argument for 'ros2 bag record'

MichaelOrlov:
#### Motivation:
- This is a follow-up from the https://github.com/ros2/rosbag2/pull/1629#issuecomment-2084683977
- TLDR:  if one wants to specify a list of services and list of topics for recording, the CLI command should be
```
ros2 bag record --services service1 service2 -- topic1 topic2 --output path/to/bag
```
It works currently, if put a trailing `--` after the list of services. However, that workaround with trailing `--` is not obvious and could cause confusion and be a source of mistakes because the topics parameter can't be specified explicitly. i.e., `--topics`.

#### List of changes:
- Add optional "`--topics`" parameter for the `ros2 bag record` CLI interface.
- Deprecate the positional "`topics`" parameter to be consistent with the"`--services`" parameter.
- Add test coverage for non-trivial cases in the recorder CLI parser.
- Use the optional "`--topics`" parameter instead of the deprecated positional parameter in integration tests.

Note: There are no API/ABI breaking changes in this PR

Issue #1620, opened 2024-04-22
`--help` mentions that `--regex` is applied on top of topic list, but doing so gives "Must specify only one option out of topics, --regex or --all"


Raphtor:
## Description
Doing `ros2 bag record --help` says that `-e` is applied on top of topics list:
```
>> ros2 bag record --help
usage: ros2 bag record [-h] [-a] [-e REGEX] [-x EXCLUDE] [--include-unpublished-topics] [--include-hidden-topics] [-o OUTPUT]
 ...
  -e REGEX, --regex REGEX
         Record only topics containing provided regular expression. Overrides --all, applies on top of topics list.
```

### Expected Behavior
The topics recorded include both the topic list AND all topics that match the regex.
### Actual Behavior
Doing that gives an error.
```
ros2 bag record -e '\/namespace\/.*\/topic' /topic1 /topic2
[ERROR] [ros2bag]: Must specify only one option out of topics, --regex or --all
```
Flipping the arguments also gives an error:

## To Reproduce
Run `ros2 bag record -e '\/namespace\/.*\/topic' /topic1 /topic2`

## System (please complete the following information)
 - OS: Ubuntu 20.04
 - ROS 2 Distro: Humble
 - Install Method: APT
 - Version: `ros-humble-ros2bag/jammy,now 0.15.8-1jammy.20231205.205704`

Raphtor:
A workaround is just to include the topic list in the regex, e.g
`ros2 bag record -e (\/namespace\/.*\/topic)|(\/topic1)|(\/topic2)`
But the documentation shouldn't give false hope.

Additionally, since subscribers are only created _after_ messages get published, its possible to miss messages with from volatile QOS durability publishers _if there are no subscribers_. However, if they are specified normally in the topic list with `--include-unpublished-topics`, they are created at the beginning and capture all messages.

For example, consider the three cases below:
Case 1: regex with no subscriber
1. Bag starts recording with `-e` to specify topics: `ros2 bag record -e --include-unpublished-topics (\/namespace\/.*\/topic)|(\/topic1)|(\/topic2)`
2. Node starts up.
3. Node publishes a single message with volatile durability to `/topic`.
4. Bag makes a subscriber to the message
5. Message is lost.

Case 2: listing topics manually
1. Bag starts recording, listing topics explicitly: `ros2 bag record --include-unpublished-topics /topic1 /topic2`
2. Node starts up.
3. Node publishes a single message with volatile durability to `/topic`.
[...]

PR #1633, opened 2024-05-03
Sweep cleanup in rosbag2 recorder CLI args verification code

MichaelOrlov:
- Relates (fixes on Rolling and Jazzy) issue https://github.com/ros2/rosbag2/issues/1620
- Depends from https://github.com/ros2/rosbag2/pull/1632
#### Motivation:
We have reworked `rosbag2_transport ::topic_filter` in the scope of the Services recording/Replay feature and removed many constraints and limitations which was applied to the topic filter. However, the recorder CLI arguments verification code and help section hasn't been updated so far.

#### List of changes:
- Remove outdated checks that impose limitations that only one option out of others can be used. Such as:
   - "Only one option out of --all, --all-services --services or --regex can be used"
   - "Only one option out of --all, --all-topics, --topics, --topic-types or --regex can be used"
   - "--exclude-regex argument requires either --all, --all-topics, --all-services or --regex"
   - and so on..
- Move arguments checks to a separate method 'validate_parsed_arguments(self, args, uri)'
- Fix the wrong limitation on arguments in the recorder CLI help section.
  - The `--regex` now can be used together with `--topics` and `--services` lists. Also now --all, --all-topics or --all-services will override `--regex`. Before, it was the opposite.
  - `--exclude-topics` can be used together with `--topics`. Sort of nonsense but why not..
  - `--exclude--services` can be used together with `--services`.

Note: We already have test coverage for most (almost all) of the newly defined behavior in the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_transport/test/rosbag2_transport/test_topic_filter.cpp

Note: There are no API/ABI breaking changes in this PR. However, it is not backportable to the Iron or Humble since we have different logic in the `rosbag2_transport ::topic_filter`.

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -201,61 +201,83 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
              'Has no effect if no compression mode is chosen. Default: %(default)s.')
 
 
+def check_necessary_argument(args):
+    # At least one options out of --all, --all-topics, --all-services, --services, --topics,
+    # --topic-types or --regex must be used
+    if not (args.all or args.all_topics or args.all_services or
+            (args.services and len(args.services) > 0) or
+            (args.topics and len(args.topics) > 0) or
+            (args.topic_types and len(args.topic_types) > 0) or args.regex):
+        return False
+    return True
+
+
+def validate_parsed_arguments(args, uri) -> str:
+    if args.topics_positional:
+        print(print_warn('Positional "topics" argument deprecated. '
+                         'Please use optional "--topics" argument instead.'))
+        args.topics = args.topics_positional
+
+    if not check_necessary_argument(args):
+        return print_error('Need to specify at least one option out of --all, --all-topics, '
+                           '--all-services, --services, --topics, --topic-types or --regex')
+
+    if args.exclude_regex and not \
+            (args.all or args.all_topics or args.topic_types or args.all_services or
+             args.regex):
+        return print_error('--exclude-regex argument requires either --all, '
+                           '--all-topics, --topic-types, --all-services or --regex')
+
+    if args.exclude_topics and not \
+            (args.all or args.all_topics or args.topic_types or args.regex):
+        return print_error('--exclude-topics argument requires either --all, --all-topics, '
+                           '--topic-types or --regex')
+
+    if args.exclude_topic_types and not \
+            (args.all or args.all_topics or args.topic_types or args.regex):
+        return print_error('--exclude-topic-types argument requires either --all, '
+                           '--all-topics or --regex')
+
+    if args.exclude_services and not (args.all or args.all_services or args.regex):
+        return print_error('--exclude-services argument requires either --all, --all-services '
+                           'or --regex')
+
+    if (args.all or args.all_services) and args.services:
+        print(print_warn('--all or --all-services will override --services'))
+
+    if (args.all or args.all_topics) and args.topics:
+        print(print_warn('--all or --all-topics will override --topics'))
+
+    if (args.all or args.all_topics or args.all_services) and args.regex:
+        print(print_warn('--all, --all-topics or --all-services will override --regex'))
+
+    if os.path.isdir(uri):
+        return print_error("Output folder '{}' already exists.".format(uri))
+
+    if args.use_sim_time and args.no_discovery:
+        return print_error(
+            '--use-sim-time and --no-discovery both set, but are incompatible settings. '
+            'The /clock topic needs to be discovered to record with sim time.')
+
+    if args.compression_format and args.compression_mode == 'none':
+        return print_error('Invalid choice: Cannot specify compression format '
+                           'without a compression mode.')
+
+    if args.compression_queue_size < 0:
+        return print_error('Compression queue size must be at least 0.')
+
+
 class RecordVerb(VerbExtension):
     """Record ROS data to a bag."""
 
     def add_arguments(self, parser, cli_name):  # noqa: D102
         add_recorder_arguments(parser)
 
-    def _check_necessary_argument(self, args):
-        # At least one options out of --all, --all-topics, --all-services, --services, --topics,
-        # --topic-types or --regex must be used
-        if not (args.all or args.all_topics or args.all_services or
-                (args.services and len(args.services) > 0) or
-                (args.topics and len(args.topics) > 0) or
-                (args.topic_types and len(args.topic_types) > 0) or args.regex):
-            return False
-        return True
-
-    def validate_parsed_arguments(self, args, uri) -> str:
-        if args.topics_positional:
-            print(print_warn('Positional "topics" argument deprecated. '
-                             'Please use optional "--topics" argument instead.'))
-            args.topics = args.topics_positional
-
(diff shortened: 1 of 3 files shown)


ros2bag/ros2bag/verb/record.py
```diff
@@ -211,83 +211,63 @@ def _check_necessary_argument(self, args):
         # At least one options out of --all, --all-topics, --all-services, --services, --topics,
         # --topic-types or --regex must be used
         if not (args.all or args.all_topics or args.all_services or
-           args.services or (args.topics and len(args.topics) > 0) or
-           (args.topic_types and len(args.topic_types) > 0) or args.regex):
+                args.services or (args.topics and len(args.topics) > 0) or
```
Barry-Xu-2018: It is better to change the check of args.services like args.topics.   

```suggestion
                (args.services and len(args.services)>0) or (args.topics and len(args.topics) > 0) or
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -211,83 +211,63 @@ def _check_necessary_argument(self, args):
...
-                               '--topics, --topic-types or --regex')
-
-        if (args.exclude_regex and
-           not (args.regex or args.all or args.all_topics or args.all_services)):
-            return print_error('--exclude-regex argument requires either --all, '
-                               '--all-topics, --all-services or --regex')
-
-        if args.exclude_topics and not (args.regex or args.all or args.all_topics):
-            return print_error('--exclude-topics argument requires either --all, --all-topics '
-                               'or --regex')
```
Barry-Xu-2018, line 253: Is this check still necessary ?   
Users have set `exclude_topics`, but the lack of necessary settings leads that no topics will be recorded. It would be best to provide a prompt for users to ensure they know there is an error in command argument.

ros2bag/ros2bag/verb/record.py
```diff
@@ -211,83 +211,63 @@ def _check_necessary_argument(self, args):
...
-
-        if args.exclude_topic_types and not (args.regex or args.all or args.all_topics):
-            return print_error('--exclude-topic-types argument requires either --all, '
-                               '--all-topics or --regex')
```
Barry-Xu-2018, line 257: The comments are the same as above.  
Besides, `args.topics` should be added.

ros2bag/ros2bag/verb/record.py
```diff
@@ -211,83 +211,63 @@ def _check_necessary_argument(self, args):
...
-
-        if args.exclude_services and not (args.regex or args.all or args.all_services):
-            return print_error('--exclude-services argument requires either --all, --all-services '
-                               'or --regex')
```
Barry-Xu-2018, line 261: The comments are the same as above.

ros2bag/ros2bag/verb/record.py
MichaelOrlov: No. this check is not necessary we already do check for must-have inclusion arguments beforehand in the 
https://github.com/ros2/rosbag2/pull/1633/files#diff-69df89597d099d6eeeb32ecc480c22116dc6c97f32556ef9636eee901f62452cR210-R215

ros2bag/ros2bag/verb/record.py
MichaelOrlov: No. this check is not necessary we already do check for must-have inclusion arguments beforehand in the
https://github.com/ros2/rosbag2/pull/1633/files#diff-69df89597d099d6eeeb32ecc480c22116dc6c97f32556ef9636eee901f62452cR210-R215

ros2bag/ros2bag/verb/record.py
MichaelOrlov: No. this check is not necessary we already do check for must-have inclusion arguments beforehand in the
https://github.com/ros2/rosbag2/pull/1633/files#diff-69df89597d099d6eeeb32ecc480c22116dc6c97f32556ef9636eee901f62452cR210-R215

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Fixed.

ros2bag/ros2bag/verb/record.py
Barry-Xu-2018: I think the check for must-have inclusion arguments isn't enough.  

For example,  
ros2 bag record --all-services --exclude-topics topicA

User forget to set any option for recording topics except `--exclude-topics`. If there is no above check code, user will not get any error prompts.

MichaelOrlov: @Barry-Xu-2018 Thanks for the clarification, and sorry that I misunderstood you from the very beginning.
It does make sense to me. I will add those checks.

MichaelOrlov: Addressed in the https://github.com/ros2/rosbag2/pull/1633/commits/9eb6641595aa55ea8c28c88b7f2a777bb21d86dd

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Fixed in the https://github.com/ros2/rosbag2/pull/1633/commits/9eb6641595aa55ea8c28c88b7f2a777bb21d86dd

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Fixed in the https://github.com/ros2/rosbag2/pull/1633/commits/9eb6641595aa55ea8c28c88b7f2a777bb21d86dd

Barry-Xu-2018:
LGTM

MichaelOrlov:
@clalancette @emersonknapp This PR was reviewed and approved by @Barry-Xu-2018.
However, I need your formal review/approval to be able to merge this PR.

MichaelOrlov:
@clalancette @emersonknapp Friendly ping here for romal review/approval. 
This PR blocks another PR https://github.com/ros2/rosbag2/pull/1646

fujitatomoya (approved):
@MichaelOrlov lgtm with green CI!

MichaelOrlov:
Pulls: ros2/rosbag2#1633
Gist: https://gist.githubusercontent.com/MichaelOrlov/5f0723928c4a6536b269e7fc767de58a/raw/83fc88eb59f337cf880618eeb9a8a56a73dc4b95/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13938
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21126/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15426/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21844/)

MichaelOrlov:
@clalancette @emersonknapp Since the Jazzy release is done, perhaps you will find time to formally approve this PR.
It was already reviewed by @fujitatomoya and @Barry-Xu-2018.

ahcorde (approved):

MichaelOrlov:

PR #1641, opened 2024-05-07
Add unit tests to cover message's send and received timestamps during recording

### LINKED ISSUE #1638: Rosbag2 does not work at all with CycloneDDS on Jazzy
## Description
During the Jazzy tutorial testing party, @mikaelarguedas found a severe issue where rosbag2 recording does not work at all with CycloneDDS.  The original report is in https://github.com/osrf/ros2_test_cases/issues/1279 , but in short when recording with `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp`, the timestamps are set to January 1, 1970.  The fallout from this is that playback never works.

### Expected Behavior
It should be possible to record and play rosbag2 data with `rmw_cyclonedds_cpp`

### Actual Behavior
No data recorded with `rmw_cyclonedds_cpp` can be played back.

## To Reproduce
1. Terminal 1: `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp ros2 launch composition composition_demo_launch.py`
2. Terminal 2: `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp ros2 bag record -a`
3. Ctrl-C both Terminal 1 and Terminal 2
4. `ros2 bag info <rosbagfile>` - Notice that the start and end timestamps are January 1, 1970.
5. `RMW_IMPLEMENTATION ros2 bag play <rosbagfile>` - Nothing plays back

 - OS: Ubuntu 24.04
 - ROS 2 Distro: Jazzy and Rolling
 - Install Method: apt
 - Version: 0.26.2-1noble.20240502

MichaelOrlov:
- Closes https://github.com/ros2/rosbag2/issues/1638
- CycloneDDS doesn't support `received_timestamps` and settling up zeros in `message_info` for that field.
- Fallback to the timestamp from the node clock in case of the CycloneDDS.
- Full RCA can be found in the https://github.com/ros2/rosbag2/issues/1638#issuecomment-2097426872

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -586,10 +586,10 @@ RecorderImpl::create_subscription(
   }
 #endif
 
-  static bool rwm_has_received_timestamp_support =
+  static bool rmw_has_received_timestamp_support =
     std::string(rmw_get_implementation_identifier()).find("rmw_cyclonedds") == std::string::npos;
 
-  if (record_options_.use_sim_time || !rwm_has_received_timestamp_support) {
+  if (record_options_.use_sim_time || !rmw_has_received_timestamp_support) {
     return node->create_generic_subscription(
       topic_name,
       topic_type,
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -586,10 +586,7 @@ RecorderImpl::create_subscription(
   }
 #endif
 
-  static bool rmw_has_received_timestamp_support =
-    std::string(rmw_get_implementation_identifier()).find("rmw_cyclonedds") == std::string::npos;
-
-  if (record_options_.use_sim_time || !rmw_has_received_timestamp_support) {
+  if (record_options_.use_sim_time) {
     return node->create_generic_subscription(
       topic_name,
       topic_type,
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -586,7 +586,10 @@ RecorderImpl::create_subscription(
   }
 #endif
 
-  if (record_options_.use_sim_time) {
+  static bool rwm_has_received_timestamp_support =
+    std::string(rmw_get_implementation_identifier()).find("rmw_cyclonedds") == std::string::npos;
+
+  if (record_options_.use_sim_time || !rwm_has_received_timestamp_support) {
     return node->create_generic_subscription(
       topic_name,
       topic_type,
--- a/rosbag2_transport/test/rosbag2_transport/test_record.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record.cpp
@@ -91,19 +91,20 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
   EXPECT_THAT(array_messages[0]->float32_values, Eq(array_message->float32_values));
 
   // Check for send and received timestamps
-  bool rwm_has_send_timestamp_support = true;
+  bool rmw_has_send_timestamp_support = true;
 #ifdef _WIN32
   if (std::string(rmw_get_implementation_identifier()).find("rmw_connextdds") !=
     std::string::npos)
   {
-    rwm_has_send_timestamp_support = false;
+    rmw_has_send_timestamp_support = false;
   }
 #endif
   for (const auto & message : recorded_messages) {
     EXPECT_NE(message->recv_timestamp, 0) << "topic : " << message->topic_name;
-    if (rwm_has_send_timestamp_support) {
+    if (rmw_has_send_timestamp_support) {
       // Check that the send_timestamp is not the same as the clock message
       EXPECT_NE(message->send_timestamp, 0);
+      EXPECT_THAT(message->recv_timestamp, Ge(message->send_timestamp));
     } else {
       // if rwm has not sent timestamp support, send_timestamp must be zero
       EXPECT_EQ(message->send_timestamp, 0);
--- a/rosbag2_transport/test/rosbag2_transport/test_record.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_record.cpp
@@ -70,8 +70,10 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
       return mock_writer.get_messages().size() >= expected_messages;
     });
   auto recorded_messages = mock_writer.get_messages();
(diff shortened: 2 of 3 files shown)


fujitatomoya (approved):
lgtm with green CI. Good eye 🚀

ahcorde (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21005/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15309/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21686/)

marcoag (approved):

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -586,7 +586,10 @@ RecorderImpl::create_subscription(
   }
 #endif
 
-  if (record_options_.use_sim_time) {
+  static bool rwm_has_received_timestamp_support =
+    std::string(rmw_get_implementation_identifier()).find("rmw_cyclonedds") == std::string::npos;
```
clalancette: I have two thoughts about this.

1.  Does rmw_connextdds properly set the receive_timestamp?
2.  This fix seems awfully specific to `rmw_cyclonedds`.  What about other RMWs that don't happen to set the receive_timestamp properly?  Would it be better to instead detect whether `receive_timestamp` is zero, and then do a fallback behavior if it is?

MichaelOrlov: @clalancette 
- As regards 1. It seems yes since CI is green, and I've added test coverage.
- As regards 2. The changes in this PR is a workaround. The proper fix would be to set received_timestamp at least on RMW layer i.e.,
https://github.com/ros2/rmw_cyclonedds/blob/c6dbe24b2f2be87cf8e4750d89501657ab83566f/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3280-L3282
```
  message_info->source_timestamp = info.source_timestamp;
  // TODO(iluetkeb) add received timestamp, when implemented by Cyclone
  message_info->received_timestamp = 0;
```
- As regards:
> What about other RMWs that don't happen to set the receive_timestamp properly? Would it be better to instead detect whether receive_timestamp is zero, and then do a fallback behavior if it is?

Well we have API where RMW or underlying transport implementation shall settle up value for received timestamp in message info. If they do not do this - this is should considered as bug on their side. I would understand if `send_timestamp` could be not set because it might not be available but with receive - no problem to set it.
- As regards the idea of checking for a timestamp for each message, and if it is zero, overwriting it with our timestamp - IMO, it looks like a hack, and we shouldn't do this.

MichaelOrlov: I only was thinking of adding a warning message on the subscription that RMW doesn't support `received_timestamp`

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -586,7 +586,10 @@ RecorderImpl::create_subscription(
   }
 #endif
 
-  if (record_options_.use_sim_time) {
+  static bool rwm_has_received_timestamp_support =
```
emersonknapp:
```suggestion
  static bool rmw_has_received_timestamp_support =
```

emersonknapp (approved):
LGTM except for typo

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: @clalancette `rmw_connextdds` properly set received and send timestamps among a case with Windows which we have already workaounded in the same function a few lines above (570-587)
https://github.com/ros2/rmw_connextdds/blob/d2bd90826e19097a429ba150b12e67d3822c1ccc/rmw_connextdds_common/src/common/rmw_impl.cpp#L1970-L1983
```cpp
rmw_connextdds_message_info_from_dds(
  rmw_message_info_t * const to,
  const DDS_SampleInfo * const from)
{
  rmw_connextdds_ih_to_gid(from->publication_handle, to->publisher_gid);
// Message timestamps are disabled on Windows because RTI Connext DDS
// does not support a high enough clock resolution by default (see: _ftime()).
#if !RTI_WIN32
  to->source_timestamp = dds_time_to_u64(&from->source_timestamp);
  to->received_timestamp = dds_time_to_u64(&from->reception_timestamp);
#else
  to->source_timestamp = 0;
  to->received_timestamp = 0;
#endif /* !RTI_WIN32 */
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Good catch. Thanks. Fixed.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
clalancette: Doesn't this workaround on Connext Windows mean that rosbag2 won't be able to playback data on that combination?  In other words, should we do a fix to `rmw_connextdds` as well with a similar fix to https://github.com/ros2/rmw_cyclonedds/pull/491 ?

MichaelOrlov: As regards:
> Doesn't this workaround on Connext Windows mean that rosbag2 won't be able to playback data on that combination?

No. the workaround is correct and rosbag2 will playback with `connext_dds` it will take received_timestamp from node clock on Windows and in all other cases it will be settled up by `rmw_connextdds` directly.

From one side, yes we can implement the same fix as we currently do for `rmw_cyclonedds`. However, from another side, it is not an issue currently and a workaround is already in place. I wouldn't bother for it now. At least I have more important things to do.

clalancette: All right, that's good enough for now.  Let's just concentrate on fixing rmw_cyclonedds then.

clalancette:
So with https://github.com/ros2/rmw_cyclonedds/pull/491 now merged, this one isn't strictly required.  Totally up to you if you want to add in the additional tests here, but I'm going to suggest that if we do that, we do *not* backport to Jazzy at this time.

MichaelOrlov:
Pulls: ros2/rosbag2#1641
Gist: https://gist.githubusercontent.com/MichaelOrlov/b31de407050b0f1ce7f4267923e2042d/raw/782365b24d4f679263c4bb60019c89c25a2e7a43/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/13855
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21037/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15339/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21732/)

MichaelOrlov:

PR #1644, opened 2024-05-08
[humble] Resolve recording option problem

### LINKED ISSUE #1620: `--help` mentions that `--regex` is applied on top of topic list, but doing so gives "Must specify only one option out of topics, --regex or --all"

Barry-Xu-2018:
Address #1620

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -55,7 +55,7 @@ class RecordVerb(VerbExtension):
         parser.add_argument(
             '-e', '--regex', default='',
             help='Record only topics containing provided regular expression. '
-            'Overridden by --all, applies on top of topics list.')
+            'Applies on top of topics list.')
         parser.add_argument(
             '-x', '--exclude', default='',
             help='Exclude topics containing provided regular expression. '
@@ -192,6 +192,9 @@ class RecordVerb(VerbExtension):
 
         if args.compression_queue_size < 1:
             return print_error('Compression queue size must be at least 1.')
+
+        if args.all and args.regex:
+            print('[WARN] [ros2bag]: --all will override --regex.')
 
         args.compression_mode = args.compression_mode.upper()
 
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -55,7 +55,7 @@ class RecordVerb(VerbExtension):
         parser.add_argument(
             '-e', '--regex', default='',
             help='Record only topics containing provided regular expression. '
-            'Overrides --all, applies on top of topics list.')
+            'Overridden by --all, applies on top of topics list.')
         parser.add_argument(
             '-x', '--exclude', default='',
             help='Exclude topics containing provided regular expression. '
@@ -168,8 +168,8 @@ class RecordVerb(VerbExtension):
 
     def main(self, *, args):  # noqa: D102
         # both all and topics cannot be true
-        if (args.all and (args.topics or args.regex)) or (args.topics and args.regex):
-            return print_error('Must specify only one option out of topics, --regex or --all')
+        if (args.all and args.topics):
+            return print_error('Specify either --all or topics, but not both simultaneously.')
         # one out of "all", "topics" and "regex" must be true
         if not(args.all or (args.topics and len(args.topics) > 0) or (args.regex)):
             return print_error('Invalid choice: Must specify topic(s), --regex or --all')


ros2bag/ros2bag/verb/record.py
```diff
@@ -168,8 +168,8 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
 
     def main(self, *, args):  # noqa: D102
         # both all and topics cannot be true
-        if (args.all and (args.topics or args.regex)) or (args.topics and args.regex):
-            return print_error('Must specify only one option out of topics, --regex or --all')
+        if (args.all and args.topics):
+            return print_error('Specify either --all or topics, but not both simultaneously.')
```
Barry-Xu-2018, line 172: There is an existing test case to check this

https://github.com/ros2/rosbag2/blob/ed8d98449320fd45ed3b7b78dbc584acfaae8d00/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L665-L673

Barry-Xu-2018:
Test rosbag2 / build_and_test (pull_request) failed.   
After checking, I found the below error log. It seems that the cause is from CI process.
```
Error: The process '/usr/bin/bash' failed with exit code 2
```

Barry-Xu-2018:
BTW, Iron has the same problem.   
After this PR is merged, I will apply this PR to Iron branch.

fujitatomoya (commented):
@Barry-Xu-2018 please take a look at this comment, https://github.com/ros2/rosbag2/issues/1620#issuecomment-2101498389

Barry-Xu-2018:
@fujitatomoya 

> @Barry-Xu-2018 please take a look at this comment, [#1620 (comment)](https://github.com/ros2/rosbag2/issues/1620#issuecomment-2101498389)

I added comments https://github.com/ros2/rosbag2/issues/1620#issuecomment-2101776142.

fujitatomoya (approved):
@Barry-Xu-2018 thanks for the explanation.

fix looks good, btw is this supposed to target to iron 1st. then we can backport it to humble. (i think we can do vice-versa, so just a question)

CC: @MichaelOrlov

ros2bag/ros2bag/verb/record.py
```diff
@@ -55,7 +55,7 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '-e', '--regex', default='',
             help='Record only topics containing provided regular expression. '
-            'Overrides --all, applies on top of topics list.')
+            'Overridden by --all, applies on top of topics list.')
```
MichaelOrlov: @Barry-Xu-2018 IMO `Overridden by --all` sounds a bit confusing. 
I think it would be better to remove it from the help section at all. Just keep `applies on top of topics list.`
And add a check in the main function with a warning message like
```
if args.all and args.regex:
  print_warn('--all will override --regex')
```

Barry-Xu-2018: Yes. This way is clearer than writing it in the parameter description.   
I will change it.

Barry-Xu-2018:
@fujitatomoya 

> fix looks good, btw is this supposed to target to iron 1st. then we can backport it to humble. (i think we can do vice-versa, so just a question)

Yes. I will create a new PR for Iron and then backport it to Humble.

ros2bag/ros2bag/verb/record.py
Barry-Xu-2018: Fixed in https://github.com/ros2/rosbag2/pull/1644/commits/c84824fc92e0a4b43b0d0a0c4f0c443c9cbf3279 

Next, I will create a new PR for Iron based on this PR and close this PR. After PR for Iron is merged, backport it to Humble.

Barry-Xu-2018:
#1649 for Iron has been created.  After it is merged, it will backport to Humble.
So I will close this PR.

PR #1646, opened 2024-05-09
Exclude recorded /clock topic when --clock option is specified

### LINKED ISSUE #1645: Fix double output when playing with --clock in rosbag containing clock
## Description

If we play rosbag containing `/clock` with the -`-clock` option, we will get double `/clock` output.

```
ros2 topic echo /clock

---
clock:
  sec: 1715138101
  nanosec: 880998015
---
clock:
  sec: 1666578700
  nanosec: 606046430
---
clock:
  sec: 1666578700
  nanosec: 608539499
---
clock:
  sec: 1715138101
  nanosec: 891000787
---
clock:
  sec: 1666578700
  nanosec: 611041316
```

## Completion Criteria

only one `/clock` is published

If `--clock` is specified, how about excluding `/clock?` (https://github.com/ros2/rosbag2/pull/1646)
```
        if play_options.clock_publish_frequency > 0:
            play_options.exclude_topics_to_filter.append('/clock')
```

play rosbag containing `/clock` with `--clock`

kosuke55:
As mentioned in https://github.com/ros2/rosbag2/issues/1645, if you play a rosbag containing `/clock` with `--clock`, it will be double-published.
This PR excludes `/clock` when --clock is specified.

```
ros2 bag play sample_bag --clock 100
```

before
```
---
clock:
  sec: 1715138101
  nanosec: 880998015
---
clock:
  sec: 1666578700
  nanosec: 606046430
---
clock:
  sec: 1666578700
  nanosec: 608539499
---
clock:
  sec: 1715138101
  nanosec: 891000787
---
clock:
  sec: 1666578700
  nanosec: 611041316
```

after
```
clock:
  sec: 1715138117
  nanosec: 244567726
---
clock:
  sec: 1715138117
  nanosec: 254594666
---
clock:
  sec: 1715138117
  nanosec: 264606514
---
clock:
  sec: 1715138117
  nanosec: 274554989
---
clock:
  sec: 1715138117
  nanosec: 284571707
```

### DIFF
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -87,7 +87,8 @@ class PlayVerb(VerbExtension):
         clock_args_group.add_argument(
             '--clock', type=positive_float, nargs='?', const=40, default=0,
             help='Publish to /clock at a specific frequency in Hz, to act as a ROS Time Source. '
-                 'Value must be positive. Defaults to not publishing.')
+                 'Value must be positive. Defaults to not publishing.'
+                 'If specified, /clock topic in the bag file is excluded to publish.')
         clock_args_group.add_argument(
             '--clock-topics', type=str, default=[], nargs='+',
             help='List of topics separated by spaces that will trigger a /clock update '
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -208,8 +208,6 @@ class PlayVerb(VerbExtension):
 
         play_options.exclude_regex_to_filter = args.exclude_regex
 
-        play_options.exclude_topics_to_filter = args.exclude_topics if args.exclude_topics else []
-
         play_options.exclude_service_events_to_filter = \
             convert_service_to_service_event_topic(args.exclude_services)
 
@@ -217,6 +215,10 @@ class PlayVerb(VerbExtension):
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
+        exclude_topics = args.exclude_topics if args.exclude_topics else []
+        if play_options.clock_publish_frequency > 0:
+            exclude_topics.append('/clock')
+        play_options.exclude_topics_to_filter = exclude_topics
         if args.clock_topics_all or len(args.clock_topics) > 0:
             play_options.clock_publish_on_topic_publish = True
         play_options.clock_topics = args.clock_topics


MichaelOrlov:
@kosuke55 I am not convinced yet that we need this fix. 
I'm curious as to why not use `ros2 bag play sample_bag --clock 100 --exclude-topics /clock` explicitly from CLI arguments in this case?

kosuke55:
@MichaelOrlov 
Thanks, that is correct and can be solved by explicitly excluding it in the command.
This may be more like a personal situation, but when automating some process, some rosbags include `/clock` and some do not.
When processing such mixed rosbags, it is necessary to check if `/clock` is included as a preprocessing step and separate the commands (or always add `--exclude-topics` option).  Although it requires a bit of additional work, adding this process solves the problem.
However, I can't think of a situation where I would want to publish the duplicate `/clock`, and usually that would cause a bug, so I thought it would be better to automatically exclude it.

fujitatomoya:
@kosuke55 i think automatic exclusion is user-friendly. as @MichaelOrlov mentioned, we `can` do this using options and checking topic information in rosbag file. but imo this provides less-burden for ros2bag user. after all, i think this is nice to have.

ros2bag/ros2bag/verb/play.py
```diff
@@ -208,15 +208,17 @@ def main(self, *, args):  # noqa: D102
...
         play_options.exclude_service_events_to_filter = \
             convert_service_to_service_event_topic(args.exclude_services)
 
         play_options.topic_qos_profile_overrides = qos_profile_overrides
         play_options.loop = args.loop
         play_options.topic_remapping_options = topic_remapping
         play_options.clock_publish_frequency = args.clock
+        exclude_topics = args.exclude_topics if args.exclude_topics else []
+        if play_options.clock_publish_frequency > 0:
+            exclude_topics.append('/clock')
```
fujitatomoya, line 221: i think we should explain this behavior in `--clock` option help field for sure, e.g `If specified, /clock topic in the bag file is excluded to publish.`.

ros2bag/ros2bag/verb/play.py
```diff
@@ -208,15 +208,17 @@ def main(self, *, args):  # noqa: D102
...
+        play_options.exclude_topics_to_filter = exclude_topics
```
fujitatomoya, line 222: Can we just do the following without any other changes above?

```suggestion
        if play_options.clock_publish_frequency > 0:
            play_options.exclude_topics_to_filter.append('/clock')
```

fujitatomoya (changes requested):

ros2bag/ros2bag/verb/play.py
kosuke55: I haven't been able to trace the cause in detail, but I couldn't add it directly :thinking: 

```
print(f"play_options.clock_publish_frequency: {play_options.clock_publish_frequency}")
if play_options.clock_publish_frequency > 0:
    print("Before appending:", play_options.exclude_topics_to_filter)
    play_options.exclude_topics_to_filter.append('/clock')
    print("After appending:", play_options.exclude_topics_to_filter)
print(f"Final state of exclude_topics_to_filter: {play_options.exclude_topics_to_filter}")
```

```
play_options.clock_publish_frequency: 100.0
Before appending: []
After appending: []
Final state of exclude_topics_to_filter: []
```

MichaelOrlov (commented):
@kosuke55 Thanks for the detailed explanation about the reasoning for these changes. 
Now, it does make sense to me.

I agree with @fujitatomoya that we need to put some notes in the CLI help section about this behavior to not surprise someone and not to be considered as a bug.

Also, I think we will need to wait until changes per https://github.com/ros2/rosbag2/pull/1633 will be merged and then rebase on top of it.

ros2bag/ros2bag/verb/play.py
kosuke55: thanks!! I updated option help in https://github.com/ros2/rosbag2/pull/1646/commits/2961ccc8f377877bc4bfa081a9abb4a782fb237e cc @MichaelOrlov

MichaelOrlov (approved):
@kosuke55 Thanks for the PR. LGTM.

MichaelOrlov:
Pulls: ros2/rosbag2#1646
Gist: https://gist.githubusercontent.com/MichaelOrlov/d95ab4465eca8b23ad6a5333051fac6b/raw/1b36ae598a72185cff646cb0be10a77df46c3e6b/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_tests ros2bag
TEST args:  --packages-above rosbag2_tests ros2bag
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14061
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21251/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15571/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/741/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21974/)

MichaelOrlov:
https://github.com/Mergifyio backport jazzy iron humble

PR #1648, opened 2024-05-09
Added method to introspect QoS in Python

### LINKED ISSUE #1770: [Feature Request] Add getters to the python binding of rclcpp::QoS
## Description

In ROS2 iron and lower the TopicMetadata structure offered the available QoS profiles as string, thus, allowing the user to check e.g. for certain settings. In my case I am mostly interested in identifying topics with 'transient local' durability as these have to be treated differently in the post processing of rosbags.

With Jazzy the TopicMetadata structure now returns the python binded equivalent of `std::vector<rclcpp::QoS>` (introduced in abdc408).  An overview of the python QoS structure can be found [here](https://github.com/ros2/rosbag2/blob/990475c7e252c38ee62ab45b9139b46e92fd3360/rosbag2_py/rosbag2_py/_storage.pyi#L51) while the pybind implementation is located [here](https://github.com/ros2/rosbag2/blob/990475c7e252c38ee62ab45b9139b46e92fd3360/rosbag2_py/src/rosbag2_py/_storage.cpp#L191).
As far as I can tell this binding only offers setters to create QoS structures with certain settings. Hence, the request to extend the binding with the corresponding getters in order to allow users a better interaction with the QoS structure.

Alternate ways I considered:
TopicMetadata offers a comparison method but after digging into the implementation it turns out that this comparison considers everything except the offered QoS profiles.

## Related Issues

No related issues (as far as I can tell).

## Completion Criteria

Addition of some getter functions for accessing the QoS settings. As a concrete example I would like to be able to check whether the durability setting of the offered QoS profile is set to 'transient local'. However, accessing the remaining settings is of course desired as well.

The [rclpy equivalent of the QoS structure](https://github.com/ros2/rclpy/blob/464a357a7d9e39a12aed966fa3ccb44d50bb6816/rclpy/rclpy/qos.py#L69) may be helpful as reference.

ahcorde:
Required to fix the QoS of the topic in rqt. Related PR https://github.com/ros-visualization/rqt_bag/pull/156


MichaelOrlov (commented):
@ahcorde Looks good to me. However, we need to regenerate Python stubs to pass CI. Please see https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md

rosbag2_py/rosbag2_py/_storage.pyi
```diff
@@ -53,14 +54,24 @@ class QoS:
     def avoid_ros_namespace_conventions(self, arg0: bool) -> QoS: ...
     def best_effort(self) -> QoS: ...
     def deadline(self, arg0: Duration) -> QoS: ...
+    def depth(self) -> int: ...
+    @overload
+    def durability(self) -> rmw_qos_durability_policy_t: ...
+    @overload
```
clalancette: So the more we add to this, the more I wonder why we have this structure at all; it is essentially duplicating what is in https://github.com/ros2/rclpy/blob/rolling/rclpy/rclpy/qos.py .

@MichaelOrlov Can you explain more about this structure, and why we are using this rather than `rclpy.QoSProfile`?

ahcorde: My guess here, it's that QoS on rclpy is a Python class not a pybind11 wrapper class. Make impossible to return this type, that's why it's duplicated. is this assumption right ?

MichaelOrlov: @clalancette Yes. basically, because this is a pybind11 wrapper class https://github.com/ros2/rosbag2/pull/1648/files#diff-6fb0986aa5f23a37be72468378f8959ba9702304ec4504d318a8c7602211c625R191-R206 sort of converter to the `rclcpp::qos` class.  
I don't know for sure, but I think it is because the all other functions in the rosbag2_py are pybind11 wrappers and need to convert returning or incoming values to/from the `rclcpp::qos` class.

MichaelOrlov (approved):

rosbag2_py/rosbag2_py/_storage.pyi
devrite: @MichaelOrlov Just stumbled upon the issue when trying to adapt for the new API changes in our ros2bag_tools.

Two questions that I have is it not possible to just use a typecaster or otherwise directly map rclcpp QoS or make rclpy QoSProfile somehow a direct bind?

As a side note I still think storage QoS is not publically imported.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21059/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15362/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21775/)

ahcorde:
Let's see if we can find a better solution instead of duplicate the class in `rclcy` and `rosbag2_py`

MichaelOrlov:
@ahcorde Am I understand correctly that this PR is not a blocker for Jazzy release?
cc: @clalancette @marcoag

ahcorde:
@MichaelOrlov, no, it's not a blocker

rosbag2_py/rosbag2_py/_storage.pyi
MichaelOrlov: @devrite I am sorry, I don't have a better solution on top of my head. However, constructive suggestions and contributions are welcome.

As regards
> As a side note I still think storage QoS is not publically imported.

Could you please elaborate on that?

devrite: Have not worked with pybind enough to be sure it works but either way, if QoSProfiles can not be replaced by the QoS bind of the current PR on the long run a converter function on the python side or pybind side (typecaster) may be useful.

Regarding my last comment. In the import list above I do not see the QoS type listed. Meaning if somebody wants to use it, when creating a TopicInfo object, you would need to import it currently from _storage and not the public module.

ahcorde:
@clalancette @MichaelOrlov waht about this solution ? convert the Rosbag rclcpp::QoS warpper into rclcpy QoSProfile Python class with a new function 

Code in `rqt_bag` is much more cleaner https://github.com/ros-visualization/rqt_bag/pull/164/files

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -391,6 +391,41 @@ PYBIND11_MODULE(_storage, m) {
...
+        "seconds"_a = qos_input.deadline().to_rmw_time().sec,
+        "nanoseconds"_a = qos_input.deadline().to_rmw_time().nsec);
+      pybind11::object duration_liveliness_lease_duration = rclcpy_qos.attr("Duration")(
+        "seconds"_a = qos_input.liveliness_lease_duration().to_rmw_time().sec,
+        "nanoseconds"_a = qos_input.liveliness_lease_duration().to_rmw_time().nsec);
+
+      auto history = rclcpp::HistoryPolicy::KeepLast;
+      if (qos_input.history() != rclcpp::HistoryPolicy::Unknown) {
+        history = qos_input.history();
+      }
```
MichaelOrlov: @ahcorde Out of curiosity why this "hack" with history is needed?

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -391,6 +391,41 @@ PYBIND11_MODULE(_storage, m) {
...
+        "durability"_a = static_cast<int>(qos_input.durability()),
+        "lifespan"_a = duration_lifespan,
+        "deadline"_a = duration_deadline,
+        "liveliness"_a = static_cast<int>(qos_input.liveliness()),
+        "liveliness_lease_duration"_a = duration_liveliness_lease_duration,
+        "avoid_ros_namespace_conventions"_a = qos_input.avoid_ros_namespace_conventions());
+
+      return qos_profile;
+    },
+    "Convert QoS in YAML to std::vector<QoS>");
```
MichaelOrlov:
```suggestion
    "Converts rclcpp::QoS to rclpy.qos");
```

rosbag2_py/src/rosbag2_py/_storage.cpp
```diff
@@ -391,6 +391,41 @@ PYBIND11_MODULE(_storage, m) {
     &rosbag2_storage::to_rclcpp_qos_vector,
     "Convert QoS in YAML to std::vector<QoS>");
 
+  m.def(
+    "convert_rosbag_qos_to_rlcpy_qos",
```
MichaelOrlov: What this function really do is converting the `rclcpp::QoS` to the `rclpy.qos`.
There is no specific `rosbag2::QoS` class. The name is misleading.
Please rename it to something more appropriate, like `rclcpp_qos_to_rclpy_qos` or `convert_rclcpp_qos_to_rclpy_qos.`

MichaelOrlov (changes requested):
@ahcorde In general,  I don't mind having a helper conversion function.
However, the name of the `convert_rosbag_qos_to_rlcpy_qos ` function is misleading.
There is no specific `rosbag2::QoS` class. We have only `rclcpp::QoS`.

rosbag2_py/src/rosbag2_py/_storage.cpp
ahcorde: For some reason the history is always returning unknown, I dediced to include this hack.

Maybe I should add a TODO here, what do you think?

ros-discourse:

https://discourse.ros.org/t/ros-2-jazzy-jalisco-released/37862/1

rosbag2_py/src/rosbag2_py/_storage.cpp
MichaelOrlov: @ahcorde As regards
> For some reason the history is always returning unknown

Could you please elaborate on that? In what cases? How to reproduce?

MichaelOrlov: @ahcorde Friendly ping here. I am not convinced that we need this "hack" here.

ahcorde: @MichaelOrlov removed hack

MichaelOrlov (approved):
LGTM with green CI.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21685/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15960/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1186/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22390/)

ahcorde:

PR #1659, opened 2024-05-17
Upgrade to version of MCAP used in rolling

nauticus-stu:
This upgrades the MCAP vendor library to be the same one used in Rolling (0.8.0 -> 1.3.0).

There is at least one bug fix in the CPP library that makes the newer version helpful to those directly using the MCAP library through this package. For us, the 0.8.0 version does not accept channels with no schema even though it is explicitly allowed in the spec. The logic is fixed in the new version.

### DIFF
--- a/mcap_vendor/CMakeLists.txt
+++ b/mcap_vendor/CMakeLists.txt
@@ -32,8 +32,8 @@ endif()
 macro(build_mcap_vendor)
   include(FetchContent)
   fetchcontent_declare(mcap
-    URL https://github.com/foxglove/mcap/archive/refs/tags/releases/cpp/v0.8.0.tar.gz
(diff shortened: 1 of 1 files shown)


MichaelOrlov:
We usually do not update vendor packages in stable ROS 2 releases. We need to make sure no API/ABI breaking changes.

nauticus-stu:
That makes sense to me. Is there a different process I can follow for backporting bug fixes?

MichaelOrlov:
> Is there a different process I can follow for backporting bug fixes?

If it is a small bugfix you can try to make a patch via CMake file. See example in the https://github.com/ros2/rosbag2/tree/rolling/zstd_vendor

MichaelOrlov:
- Closing as wouldn't be merged due to the ABI/API compatibility issue.

Issue #1660, opened 2024-05-17
`ros2 bag convert` not writing any messages on jazzy


ottojo:
## Description
`ros2 bag convert` seems to not write any messages to the output bag on jazzy. I originally experienced this when trying to use the new parameters introduced in #1455, but the issue seems to occur with the "compress" example from the readme as well.

The output bag seems to not contain any data at all when running on jazzy.

### Expected Behavior
This is what happens on iron: The file size is reduced, but most importantly the resulting bag contains all topics and messages:

out.yaml (taken from readme):

```yaml
output_bags:
- uri: bag_compressed_iron
  all: true
  compression_mode: file
  compression_format: zstd
```

```console
jonas@ade ➜  ~ ros2 bag info bag_full_iron 

Files:             bag_full_iron_0.mcap
Bag size:          5.5 KiB
Storage id:        mcap
Duration:          33.999s
Start:             May 17 2024 19:14:06.45 (1715966046.45)
End:               May 17 2024 19:14:40.45 (1715966080.45)
Messages:          35
Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 35 | Serialization Format: cdr

jonas@ade ➜  ~ ros2 bag convert -i bag_full_iron -o out.yaml 
[INFO] [1715966134.703149096] [rosbag2_compression]: Compressing file: bag_compressed_iron/bag_compressed_iron_0.mcap

jonas@ade ➜  ~ ros2 bag info  bag_compressed_iron/

Files:             bag_compressed_iron_0.mcap.zstd
Bag size:          2.3 KiB
Storage id:        mcap
Duration:          33.999s
Start:             May 17 2024 19:14:06.45 (1715966046.45)
End:               May 17 2024 19:14:40.45 (1715966080.45)
Messages:          35
Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 35 | Serialization Format: cdr
```

### Actual Behavior
On jazzy, with out.yaml:
```yaml
output_bags:
- uri: compressed
  all: true
  compression_mode: file
  compression_format: zstd
```

The resulting file contains no messages at all:
```console
jonas@ade ➜  ~ ros2 bag info bag_jazzy_full 

Files:             bag_jazzy_full_0.mcap
Bag size:          6.8 KiB
Storage id:        mcap
ROS Distro:        jazzy
Duration:          24.999s
Start:             May 17 2024 19:16:57.716 (1715966217.716)
[...]

PR #1480, opened 2023-10-13
Implement service recording and display info about recorded services

Barry-Xu-2018:
Split #1414 to 2 PRs.  
This is first PR. This PR implement service record and recorded service info display.  

:warning:  This PR depended on ros2/rclcpp#2209

- Relates https://github.com/ros2/rmw/issues/357

PR #1661, opened 2024-05-19
doc: readme: fix example conversion configs: all -> all_topics

ottojo:
this was changed in #1480, see issue #1660

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,9 +228,15 @@ output_bags:
   storage_config_uri: ""
   all_topics: false
   topics: []
+  topic_types: []
+  all_services: false
+  services: []
   rmw_serialization_format: ""  # defaults to using the format of the input topic
   regex: ""
-  exclude: ""
+  exclude_regex: ""
+  exclude_topics: []
+  exclude_topic_types: []
+  exclude_services: []
   compression_mode: ""
   compression_format: ""
   compression_queue_size: 1
--- a/README.md
+++ b/README.md
@@ -254,6 +254,7 @@ $ ros2 bag convert -i bag1 -i bag2 -o out.yaml
 output_bags:
 - uri: merged_bag
   all_topics: true
+  all_services: true
 ```
 
 Example split:
--- a/README.md
+++ b/README.md
@@ -279,6 +279,7 @@ $ ros2 bag convert -i bag1 -o out.yaml
 output_bags:
 - uri: compressed
   all_topics: true
+  all_services: true
   compression_mode: file
   compression_format: zstd
 ```
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ output_bags:
   max_bagfile_duration: 0
   storage_preset_profile: ""
   storage_config_uri: ""
-  all: false
+  all_topics: false
   topics: []
   rmw_serialization_format: ""  # defaults to using the format of the input topic
   regex: ""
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
good eye! lgtm.

@MichaelOrlov @Barry-Xu-2018 this actually breaks the user configuration as reported. i will go through the documentation again.

Barry-Xu-2018:
The `RecordOptions` has already been changed quite a bit.  
Refer to the code below.

https://github.com/ros2/rosbag2/blob/5a06430cac7e823e5b5522a67ce7bd3b325e516f/rosbag2_transport/src/rosbag2_transport/record_options.cpp#L58-L103

The current README.md only mentions part of it. Is it necessary to point to the relevant code, so users know what can be configured ? @fujitatomoya @MichaelOrlov

https://github.com/ros2/rosbag2/blob/5a06430cac7e823e5b5522a67ce7bd3b325e516f/README.md?plain=1#L221-L240

Barry-Xu-2018:
@ottojo 

Additionally.  ` exclude` should be changed to `exclude_regex`.  
Please refer to https://github.com/ros2/rosbag2/blob/5a06430cac7e823e5b5522a67ce7bd3b325e516f/rosbag2_transport/src/rosbag2_transport/record_options.cpp#L58-L103

ottojo:
@Barry-Xu-2018 thanks for pointing me to the corresponding parts in the source. I updated the `exclude` key and added the other keys relevant for topic and service selection. I did not add all keys found in the source, i think some may already be excluded on purpose, because they are not relevant for bag conversion (such as for example `disable_keyboard_controls`, `compression_threads_priority` and `topic_polling_interval`)

ottojo:
oh, i also just notice that it's not only the `RecordingOptions` but also `StorageOptions`, where the actual options are that i wanted to use prior to hitting this issue:
https://github.com/ros2/rosbag2/blob/5a06430cac7e823e5b5522a67ce7bd3b325e516f/rosbag2_storage/src/rosbag2_storage/storage_options.cpp#L44-L56

fujitatomoya (approved):
lgtm.

i do not think we need to introduce all options here, but all options need to be aligned with the implementation.

README.md
```diff
@@ -247,7 +253,7 @@ $ ros2 bag convert -i bag1 -i bag2 -o out.yaml
 # out.yaml
 output_bags:
 - uri: merged_bag
-  all: true
+  all_topics: true
```
MichaelOrlov, line 256:
```suggestion
  all_topics: true
  all_services: true
```

README.md
```diff
@@ -271,7 +277,7 @@ $ ros2 bag convert -i bag1 -o out.yaml
 # out.yaml
 output_bags:
 - uri: compressed
-  all: true
+  all_topics: true
```
MichaelOrlov, line 281:
```suggestion
  all_topics: true
  all_services: true
```

MichaelOrlov (commented):
I think it will be valuable to add `all_services: true` for examples with bag compression and bag split. 
To avoid other potential questions in the future, like "Why services were lost during conversion?"

MichaelOrlov (approved):
LGTM. Thanks for the PR!

clalancette (approved):
Looks good to me as well.

Given that this is a change to documentation, we can merge this without CI, but @MichaelOrlov I'll wait for you to do that.  I think we should also backport this to the `jazzy` branch.

MichaelOrlov:
Merging without CI run since changes only in the `readme.md` file.

MichaelOrlov:

PR #1665, opened 2024-05-21
Link `rcpputils::filesystem` to solve error on RHEL

audrow:
It seems #1653 created a [regression](https://build.ros2.org/view/Hbin_rhel_el864/job/Hbin_rhel_el864__rosbag2_compression__rhel_8_x86_64__binary/65/) on Humble on RHEL.

Here is an attempt to fix it. I'll run RHEL CI before merging.

### DIFF
--- a/rosbag2_compression/CMakeLists.txt
+++ b/rosbag2_compression/CMakeLists.txt
@@ -45,6 +45,12 @@ ament_target_dependencies(${PROJECT_NAME}
   rosbag2_storage)
 target_compile_definitions(${PROJECT_NAME} PRIVATE ROSBAG2_COMPRESSION_BUILDING_DLL)
 
+if(CMAKE_COMPILER_IS_GNUCXX)
+  target_link_libraries(${PROJECT_NAME}
+    rcpputils::filesystem
+  )
+endif()
+
 install(
   DIRECTORY include/
   DESTINATION include/${PROJECT_NAME})
--- a/rosbag2_compression/CMakeLists.txt
+++ b/rosbag2_compression/CMakeLists.txt
@@ -45,12 +45,6 @@ ament_target_dependencies(${PROJECT_NAME}
   rosbag2_storage)
 target_compile_definitions(${PROJECT_NAME} PRIVATE ROSBAG2_COMPRESSION_BUILDING_DLL)
 
-if(CMAKE_COMPILER_IS_GNUCXX)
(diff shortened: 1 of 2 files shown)


MichaelOrlov (approved):
LGTM.

MichaelOrlov (changes requested):
@audrow Ups not so simple 
```
CMake Error at CMakeLists.txt:30 (add_library):
08:00:04   Target "rosbag2_compression" links to target "rcpputils::filesystem" but
08:00:04   the target was not found.  Perhaps a find_package() call is missing for an
08:00:04   IMPORTED target, or an ALIAS target is missing?
```

Basically need to add `find_package(rcpputils)` above.

rosbag2_compression/CMakeLists.txt
```diff
@@ -45,6 +45,12 @@ ament_target_dependencies(${PROJECT_NAME}
   rosbag2_storage)
 target_compile_definitions(${PROJECT_NAME} PRIVATE ROSBAG2_COMPRESSION_BUILDING_DLL)
 
+if(CMAKE_COMPILER_IS_GNUCXX)
+  target_link_libraries(${PROJECT_NAME}
+    rcpputils::filesystem
+  )
+endif()
+
```
MichaelOrlov: @audrow @clalancette I think we need 
```
if(CMAKE_COMPILER_IS_GNUCXX)
  target_link_libraries(${PROJECT_NAME} stdc++fs)
#endif
```
IMO ` target_link_libraries(${PROJECT_NAME} rcpputils::filesystem)` is not going to help here.

clalancette: This actually isn't needed, as a similar thing is done above with `ament_target_dependencies`.

Instead, we need a change to the code itself to switch back to using `rcpputils::fs`.  I'll push some changes shortly that make it compile for me, though @MichaelOrlov suggests that it may not work in practice.

MichaelOrlov: @clalancette Yes, it is not going to work out of the box. I will do rework shortly. Give me a couple of hours.

MichaelOrlov:
@audrow @clalancette I created a new PR https://github.com/ros2/rosbag2/pull/1666 with fix in replacement to the PR.

Please review https://github.com/ros2/rosbag2/pull/1666

clalancette:
Sounds good, closing this one.

Issue #1597, opened 2024-03-27
Mark bags that are still being recorded


tonynajjar:
## Description

In ROS1, a bag that was still being recorded was suffixed with `.active`, it would be nice if in rosbag2 there was something similar. An example use case is a scripts that uploads bags to a server automatically - that script needs to know not to upload a bag if it's still recording.

defunctzombie:
@MichaelOrlov @emersonknapp any recollection on why this was omitting from rosbag2 or if there's some gotcha with adding support for this?

MichaelOrlov:
@defunctzombie @tonynajjar I don't know about any prior discussions about having a different suffix for currently writing file.  
I guess it might be some corner cases with consistency in filenames in case if recording is abnormally interrupted and we have put current filename to the metadata and haven't renamed it. We are saving metadata at the beginning as well. Also not sure if file renaming could be easily done from C++ std::filesystem.

Also, I think that one of the reasons we didn't follow ROS 1 behavior is because we handling the matter of concern in the question
> that script needs to know not to upload a bag if it's still recording

a better way than in ROS 1.

The one can subscribe to the topic `"events/write_split"` to get a notification when the current file is finished and ready to be processed by third-party script or program. 
Rosbag2 will publish a special message on this topic to notify remote nodes that the newly written file is ready and a new one has been created. The old and new filenames are part of this message in those notification. 
Yes, we have such a feature in the rosbag2, and sorry that we don't have documentation for it. Please refer to the following part of the code to better understand how to use it.
https://github.com/ros2/rosbag2/blob/30328006f872de4e1c82fef4ecca8ca815b2d4dc/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L321-L337
and 
https://github.com/ros2/rosbag2/blob/30328006f872de4e1c82fef4ecca8ca815b2d4dc/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L356-L376
The key info there is 
```
       auto message = rosbag2_interfaces::msg::WriteSplitEvent(); 
       message.closed_file = bag_split_info_.closed_file; 
       message.opened_file = bag_split_info_.opened_file; 
```

tonynajjar:
The script in question is the foxglove-agent which is closed-source so unfortunately I don't have the option to adapt it. Either the Foxglove guys would need to do so (@defunctzombie ), or rosbag2 has to introduce a differentiator in the file name as this is the only thing foxglove-agent can ignore.

[...]

PR #1671, opened 2024-05-22
fix issue #1597: mark files being written as "active"

facontidavide:
**Suggested changes:**

1. Add the suffix ".active" to ` SequentialWriter::format_storage_uri`
2. When `storage_` is closed, rename the file to remove the ".active" suffix.
3. Save the metadata when splitting. In this way all the parts but the last one are still usable in case of crash.

To address #1597 point 3 is not necessary and **maybe** not even desirable, so let me know what you think.

**Note:** 

Since` rcpputils::fs` is not wrapping "rename", I had to use `std::filesystem` instead and upgrade to C++ 17.


rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -269,11 +287,22 @@ void SequentialWriter::switch_to_next_storage()
...
+  // Write the metadata before splitting. This will be replaced by an updated file later.
+  // In this way, if the process crashes, the metadata for the previous parts will still be available.
+  finalize_metadata();
+  metadata_io_->write_metadata(base_folder_, metadata_);
 
   storage_options_.uri = format_storage_uri(
     base_folder_,
     metadata_.relative_file_paths.size());
+  
+  // when the storage_ is closed, rename the file to remove ".inactive" suffix
```
facontidavide:
```suggestion
  // when the storage_ is closed, rename the file to remove ".active" suffix
```

fujitatomoya:
@facontidavide thanks for the PR.

@MichaelOrlov are we supposed to create PR target to rolling 1st? and then backport?

clalancette:
> @MichaelOrlov are we supposed to create PR target to rolling 1st? and then backport?

Yes please.  Also, we are going to run into other problems with the switch to `std::filesystem` on Humble (because of older GCC compilers).  But let's target this to `rolling` first, then deal with that in the backport.

facontidavide:
I am temporarily closing this in favor of #1672

PR #1682, opened 2024-05-31
Allow to specify start offset from CLI arguments equal to 0.0 for the rosbag2 player

reinzor:
- This PR will allow to specify start offset from CLI arguments equal to `0.0` for the rosbag2 player.

- The default is specified as `0.0` but this is not allowed by the user.

### DIFF
--- a/ros2bag/ros2bag/api/__init__.py
+++ b/ros2bag/ros2bag/api/__init__.py
@@ -125,6 +125,17 @@ def check_positive_float(value: Any) -> float:
         raise ArgumentTypeError('{} is not the valid type (float)'.format(value))
 
 
+def check_not_negative_float(value: Any) -> float:
+    """Argparse validator to verify that a value is a float and that not negative."""
+    try:
+        fvalue = float(value)
+        if fvalue < 0.0:
+            raise ArgumentTypeError(f'Value {value} is less than zero.')
+        return fvalue
+    except ValueError:
+        raise ArgumentTypeError('{} is not the valid type (float)'.format(value))
+
+
 def check_path_exists(value: Any) -> str:
     """Argparse validator to verify a path exists."""
     try:
--- a/ros2bag/ros2bag/verb/burst.py
+++ b/ros2bag/ros2bag/verb/burst.py
@@ -16,8 +16,8 @@ from argparse import FileType
 
 from rclpy.qos import InvalidQoSProfileException
 from ros2bag.api import add_standard_reader_args
+from ros2bag.api import check_not_negative_float
 from ros2bag.api import check_not_negative_int
-from ros2bag.api import check_positive_float
 from ros2bag.api import convert_service_to_service_event_topic
 from ros2bag.api import convert_yaml_to_qos_profile
 from ros2bag.api import print_error
@@ -60,7 +60,7 @@ class BurstVerb(VerbExtension):
             help='Path to a yaml file defining storage specific configurations. '
                  'See storage plugin documentation for the format of this file.')
         parser.add_argument(
-            '--start-offset', type=check_positive_float, default=0.0,
+            '--start-offset', type=check_not_negative_float, default=0.0,
             help='Start the playback player this many seconds into the bag file.')
         parser.add_argument(
             '-n', '--num-messages', type=check_not_negative_int, default=0,
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -16,6 +16,7 @@ from argparse import FileType
 
 from rclpy.qos import InvalidQoSProfileException
 from ros2bag.api import add_standard_reader_args
+from ros2bag.api import check_not_negative_float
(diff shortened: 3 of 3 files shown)


MichaelOrlov (commented):
Instead of creating a new function `check_not_negative_float(..)` IMO it seems make sense to change a logic in the already existent `check_positive_float(value: Any)` to include `0.0` in range.

ros2bag/ros2bag/api/__init__.py
```diff
@@ -125,6 +125,17 @@ def check_positive_float(value: Any) -> float:
         raise ArgumentTypeError('{} is not the valid type (float)'.format(value))
 
 
+def check_not_negative_float(value: Any) -> float:
```
fujitatomoya, line 128: the reason for introducing this is, check_positive_float need to stay with `ros2 bag play --rate`, right? in that case 0 should not be taken as an argument.

ros2bag/ros2bag/verb/burst.py
```diff
@@ -60,7 +60,7 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             help='Path to a yaml file defining storage specific configurations. '
                  'See storage plugin documentation for the format of this file.')
         parser.add_argument(
-            '--start-offset', type=check_positive_float, default=0.0,
+            '--start-offset', type=check_not_negative_float, default=0.0,
```
fujitatomoya, line 63: Do we actually needs this to include `0`? that case we just do not need to specify.

Timple: It makes sense for launchfiles which also have a default argument for example.

fujitatomoya: @Timple thanks for the info. can you share the actual example?

ros2bag/ros2bag/api/__init__.py
reinzor: That is indeed the [second usage of the `check_positive_float`](https://github.com/ros2/rosbag2/blob/e01e1efe57c4f226eed21e2eefb0b459fd3bd0b8/ros2bag/ros2bag/verb/play.py#L51), in this case, zero does not make sense.

ros2bag/ros2bag/verb/burst.py
reinzor: Example:

```xml
<launch>
    <arg name="bag_file"/>

    <arg name="clock" default="100" />
    <arg name="rate" default="1.0" />
    <arg name="start_offset" default="0.0" />

    <set_parameter name="use_sim_time" value="true" />

    <executable name="rosbag" cmd="ros2 bag play --clock $(var clock) --rate $(var rate) --start-offset $(var start_offset) $(var bag_file)" />
</launch>
```

For now, we work around this by setting the default of the `start_offset` to a small value. But since the default (not specified) is zero, I think it should also accept this value as an input.

fujitatomoya: thanks! i am not sure the constraints of launch description, so you saying it must be set, right?

> But since the default (not specified) is zero, I think it should also accept this value as an input.

yes, this is true. default should be accepted if specified...

reinzor: A default is not necessary but this would enforce the user to always pass the `start_offset` to the launch file which is not desired.

MichaelOrlov (changes requested):
@reinzor Overall looks good to me.
However, need to fix DCO.
```
To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~1 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin fix/start-offset-may-be-zero
```

reinzor:
Thanks, done

MichaelOrlov (approved):
LGTM with green CI.

MichaelOrlov:
Pulls: ros2/rosbag2#1682
Gist: https://gist.githubusercontent.com/MichaelOrlov/4dfc487841f39d11b672182363081c30/raw/9e4f15ca480d6f610796ec35661ff4159eb199e9/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_tests ros2bag
TEST args:  --packages-above rosbag2_tests ros2bag
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14066
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21257/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15576/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21980/)

MichaelOrlov:
https://github.com/Mergifyio backport jazzy iron humble

PR #1700, opened 2024-06-07
Propagate "custom_data" and "ros_distro" in to the metadata.yaml file during re-indexing

### LINKED ISSUE #1699: Reindexing fails to read custom_data and ros_distro from storage
## Description
Newer distributions support encoding metadata directly into the storage system. if one of these bags is re-indexed, the fields for `custom_data` and `ros_distro` are not properly updated in the newly produced `metadata.yaml`.

### Expected Behavior
The `custom_metadata` and `ros_distro` fields should be propagated from the underlying storage into the new `metadata.yaml` file.

Example data:
```
custom_data:
  something: "hi"
ros_distro: "jazzy"
```

### Actual Behavior
The `custom_metadata` and `ros_distro` fields are set to their default values in the rosbag2_storage::BagMetadata struct.

Example result after re-indexing:
```
custom_data: ~
ros_distro: ""
```

## To Reproduce

1. Record a ros bag with `ros2 bag record -a --custom-data something=hi`
2. Interrupt the `ros2 bag record` after any topic data has been captured
3. View the recorded `metadata.yaml` file and observe `custom_data` containing `something:hi` and `ros_distro` containing `jazzy`.
4. Re-index the bag with `ros2 bag reindex <path/to/bag/dir>`
5. View the updated `metadata.yaml` file and observe `custom_data` and `ros_distro` being set to their default values.

 - OS: Ubuntu Noble
 - ROS 2 Distro: Jazzy
 - Install Method: APT & Source
 - Version: jazzy, rolling

coalman321:
Propagates the `custom_data` and `ros_distro` fields from the underlying storage into the `metadata.yaml` file during re-indexing. 
- Previously this data was not extracted from the storage and the subsequent fields in `metadata.yaml` were left as default. 
- The implementation in this PR uses the `custom_data` and `ros_distro` metadata from the newest storage file contained within the log as custom data can be modified in the storage API between successive split files. 

This fixes #1699

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
@@ -162,8 +162,8 @@ void Reindexer::aggregate_metadata(
   const rosbag2_storage::StorageOptions & storage_options)
 {
   std::map<std::string, rosbag2_storage::TopicInformation> temp_topic_info;
-  std::chrono::system_clock::time_point latest_log_start =
-    std::chrono::system_clock::time_point::min();
+  std::chrono::time_point<std::chrono::high_resolution_clock> latest_log_start =
+    std::chrono::time_point<std::chrono::high_resolution_clock>::min();
 
   // In order to most accurately reconstruct the metadata, we need to
   // visit each of the contained relative files files in the bag,
--- a/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
@@ -191,10 +191,14 @@ void Reindexer::aggregate_metadata(
     metadata_.storage_identifier = temp_metadata.storage_identifier;
 
     // try to find the last log for the most complete custom data section
-    if (latest_log_start < temp_metadata.starting_time) {
+    if (latest_log_start < temp_metadata.starting_time ) {
       latest_log_start = temp_metadata.starting_time;
-      metadata_.custom_data = temp_metadata.custom_data;
-      metadata_.ros_distro = temp_metadata.ros_distro;
+      if (temp_metadata.custom_data.size() > 0) {
+        metadata_.custom_data = temp_metadata.custom_data;
+      }
+      if (temp_metadata.ros_distro.size() > 0) {
+        metadata_.ros_distro = temp_metadata.ros_distro;
+      }
     }
 
     if (temp_metadata.starting_time < metadata_.starting_time) {
--- a/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
@@ -191,12 +191,12 @@ void Reindexer::aggregate_metadata(
     metadata_.storage_identifier = temp_metadata.storage_identifier;
 
     // try to find the last log for the most complete custom data section
-    if (latest_log_start < temp_metadata.starting_time ) {
+    if (latest_log_start < temp_metadata.starting_time) {
       latest_log_start = temp_metadata.starting_time;
-      if (temp_metadata.custom_data.size() > 0) {
+      if (!temp_metadata.custom_data.empty()) {
         metadata_.custom_data = temp_metadata.custom_data;
       }
-      if (temp_metadata.ros_distro.size() > 0) {
+      if (!temp_metadata.ros_distro.empty()) {
         metadata_.ros_distro = temp_metadata.ros_distro;
       }
     }
--- a/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
@@ -162,6 +162,8 @@ void Reindexer::aggregate_metadata(
   const rosbag2_storage::StorageOptions & storage_options)
 {
   std::map<std::string, rosbag2_storage::TopicInformation> temp_topic_info;
+  std::chrono::system_clock::time_point latest_log_start =
+    std::chrono::system_clock::time_point::min();
 
   // In order to most accurately reconstruct the metadata, we need to
   // visit each of the contained relative files files in the bag,
@@ -187,6 +189,13 @@ void Reindexer::aggregate_metadata(
     bag_reader->open(temp_so, blank_converter_options);
     auto temp_metadata = bag_reader->get_metadata();
     metadata_.storage_identifier = temp_metadata.storage_identifier;
+
+    // try to find the last log for the most complete custom data section
+    if(latest_log_start < temp_metadata.starting_time){
+      latest_log_start = temp_metadata.starting_time;
+      metadata_.custom_data = temp_metadata.custom_data;
+      metadata_.ros_distro = temp_metadata.ros_distro;
+    }
 
     if (temp_metadata.starting_time < metadata_.starting_time) {
       metadata_.starting_time = temp_metadata.starting_time;
--- a/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
@@ -191,7 +191,7 @@ void Reindexer::aggregate_metadata(
     metadata_.storage_identifier = temp_metadata.storage_identifier;
 
     // try to find the last log for the most complete custom data section
-    if(latest_log_start < temp_metadata.starting_time){
+    if (latest_log_start < temp_metadata.starting_time) {
       latest_log_start = temp_metadata.starting_time;
       metadata_.custom_data = temp_metadata.custom_data;
       metadata_.ros_distro = temp_metadata.ros_distro;
--- a/rosbag2_tests/test/rosbag2_tests/test_reindexer.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_reindexer.cpp
@@ -81,6 +81,7 @@ public:
       rosbag2_storage::StorageOptions storage_options;
       storage_options.storage_id = GetParam();
       storage_options.uri = root_bag_path_.string();
+      storage_options.custom_data["name"] = "value";
       writer.open(storage_options);
       rosbag2_storage::TopicMetadata topic;
       topic.name = "/test_topic";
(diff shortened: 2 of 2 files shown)


rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -188,6 +190,13 @@ void Reindexer::aggregate_metadata(
     auto temp_metadata = bag_reader->get_metadata();
     metadata_.storage_identifier = temp_metadata.storage_identifier;
 
+    // try to find the last log for the most complete custom data section
+    if (latest_log_start < temp_metadata.starting_time) {
+      latest_log_start = temp_metadata.starting_time;
+      metadata_.custom_data = temp_metadata.custom_data;
+      metadata_.ros_distro = temp_metadata.ros_distro;
+    }
+
```
fujitatomoya: a quick question. are we expecting that `custom_data` and `ros_distro` are different from file to file? if that is not time sensitive information and static, we can store once for each element?

coalman321: If update_metadata is used on the underlying storage, both of these can change between successive split files. 

Personally, I would argue that `ros_distro` is probably far less likely to be subject to change. However, I currently have a custom writer implementation that supports updating the `custom_data` field during recording, meaning that the data is changed progressively between split files. I use this for indicating events that the log contains as well as other custom data I want to store with the log while the system is being recorded.

The only reason for `ros_distro` field being updated to the value from the youngest split file is simply because I had already added the logic to find the youngest file for `custom_data`.

fujitatomoya (approved):
lgtm with green CI

MichaelOrlov (approved):
@coalman321 Thanks for the PR. Overall looks good to me.
However, I have a proposal to check if `temp_metadata.custom_data` is not empty before assigning it to the `metadata_.custom_data` and the same for the `ros_distro`.

coalman321:
> @coalman321 Thanks for the PR. Overall looks good to me. However, I have a proposal to check if `temp_metadata.custom_data` is not empty before assigning it to the `metadata_.custom_data` and the same for the `ros_distro`.

So I changed my thoughts on implementation a few times during this PR. My guiding thought was that the youngest split file (most recent) should be the most up to date version of `custom_data`. I was concerned about the case where the user had edited the info between the split files and having any sort of filtering could cause the re-indexed data to not reflect this updated state. The metadata in the final split file written should already be an aggregation of this information by nature of the storage implementation and thus be the most "accurate" representation of the information the user intends to store. In my thought this extends to being empty as well.

If you think this thought process doesn't match up with the designed intent, then I would be happy to modify it.

MichaelOrlov:
@coalman321 Your assumption that the latest metadata should have the most up to date version of the `custom_data` is correct. 
However, this is true for generic normal workflow. Although it might not be true if recording finishes abnormally, the metadata record could get corrupted or be incomplete.
IMO, it wouldn't hurt to add a sanity check that `temp_metadata.custom_data` is not empty before copying from it for a worst-case scenario.

coalman321:
I had not considered the abnormal case where the recording is not closed properly. Ill adjust the logic to work only if the fields are not empty.

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
```diff
@@ -188,6 +190,17 @@ void Reindexer::aggregate_metadata(
...
+    // try to find the last log for the most complete custom data section
+    if (latest_log_start < temp_metadata.starting_time ) {
+      latest_log_start = temp_metadata.starting_time;
+      if (temp_metadata.custom_data.size() > 0) {
+        metadata_.custom_data = temp_metadata.custom_data;
+      }
+      if (temp_metadata.ros_distro.size() > 0) {
+        metadata_.ros_distro = temp_metadata.ros_distro;
+      }
+    }
```
MichaelOrlov: Please use the `.empty()` property of the containers instead of the `size()` for checking if it is empty or not.
Also, please address `uncrustify` warning.

MichaelOrlov (changes requested):

rosbag2_cpp/src/rosbag2_cpp/reindexer.cpp
coalman321: Switched to use empty and fixed the `uncrustify` warning as requested.

MichaelOrlov (approved):
@coalman321 Thank you for your contribution.
 LGTM with green CI.

coalman321:
> @coalman321 Thank you for your contribution. LGTM with green CI.

Happy to help. Also looks like `ament_cppcheck` failed, but Github cant show status for it. It might just need to be re-run.

MichaelOrlov (commented):
@coalman321 May I ask you to modify https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_reindexer.cpp tests to check `custom_data` and `ROS_DISTRO` in the reindexer test?

coalman321:
> @coalman321 May I ask you to modify https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_reindexer.cpp tests to check `custom_data` and `ROS_DISTRO` in the reindexer test?

Added verification for both in the reindexer test as requested.

MichaelOrlov (approved):
@coalman321 Thanks for updating the tests.
LGTM with green CI.

MichaelOrlov:
Pulls: ros2/rosbag2#1700
Gist: https://gist.githubusercontent.com/MichaelOrlov/23b7ed7f99ad70b70f6d33a8bc52b264/raw/ffe2b74684675f2ba3d508aeae04d238473dce94/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_tests
TEST args:  --packages-above rosbag2_cpp rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14060
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21250/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15570/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/740/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21973/)

MichaelOrlov:
@coalman321 Unfortunately the CI failed on Windows with the following errors:
```
C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(194,55): error C2678: binary '<': no operator found which takes a left-hand operand of type 'std::chrono::system_clock::time_point' (or there is no acceptable conversion) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]

C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\include\system_error(191,28): message : could be 'bool std::operator <(const std::error_code &,const std::error_code &) noexcept' [found using argument-dependent lookup] (compiling source file C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]

C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\include\system_error(277,28): message : or       'bool std::operator <(const std::error_condition &,const std::error_condition &) noexcept' [found using argument-dependent lookup] (compiling source file C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]

C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(194,55): message : while trying to match the argument list '(std::chrono::system_clock::time_point, std::chrono::time_point<std::chrono::steady_clock,std::chrono::nanoseconds>)' [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]

C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(195,53): error C2679: binary '=': no operator found which takes a right-hand operand of type 'std::chrono::time_point<std::chrono::steady_clock,std::chrono::nanoseconds>' (or there is no acceptable conversion) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
```

coalman321:
> @coalman321 Unfortunately the CI failed on Windows with the following errors:
> 
> ```
> C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(194,55): error C2678: binary '<': no operator found which takes a left-hand operand of type 'std::chrono::system_clock::time_point' (or there is no acceptable conversion) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
> 
> C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\include\system_error(191,28): message : could be 'bool std::operator <(const std::error_code &,const std::error_code &) noexcept' [found using argument-dependent lookup] (compiling source file C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
> 
> C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.29.30133\include\system_error(277,28): message : or       'bool std::operator <(const std::error_condition &,const std::error_condition &) noexcept' [found using argument-dependent lookup] (compiling source file C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
> 
> C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(194,55): message : while trying to match the argument list '(std::chrono::system_clock::time_point, std::chrono::time_point<std::chrono::steady_clock,std::chrono::nanoseconds>)' [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
> 
> C:\ci\ws\src\coalman321\rosbag2\rosbag2_cpp\src\rosbag2_cpp\reindexer.cpp(195,53): error C2679: binary '=': no operator found which takes a right-hand operand of type 'std::chrono::time_point<std::chrono::steady_clock,std::chrono::nanoseconds>' (or there is no acceptable conversion) [C:\ci\ws\build\rosbag2_cpp\rosbag2_cpp.vcxproj]
> ```

Looks like it was an issue with the clock type that was being used. GCC must alias the system clock and the high-res clock in chrono. The change uses the exact same time point spec from the BagMetadata structure now.

MichaelOrlov:
- Re-run Windows CI job
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21977/)

MichaelOrlov:
https://github.com/Mergifyio backport jazzy iron

MichaelOrlov:

PR #1705, opened 2024-06-10
Throw an error: storage "mcap" is not compatible with compression_mode "message"

facontidavide:
The title.

Note that this problem has been fixed already in rosbag2_storage version 0.18, with the introduction of `MCAPStorage::update_metadata`

This PR is targetting specifically Humble.

Note that the change in **record.py** is not enough, since we need to also cover the case when rosbag recordinf is done using the C++ API.


rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp
```diff
@@ -187,6 +187,15 @@ void SequentialCompressionWriter::open(
   if (storage_) {
     return;  // The writer already opened.
   }
+  // Note: a better solution was implemente in rosbag2-storage version 0.18+
```
facontidavide:
```suggestion
  // Note: a better solution was implemented in rosbag2-storage version 0.18+
```

MichaelOrlov (changes requested):
@facontidavide Thanks for the PR.
I have a concern regarding mentioning that `compression_mode::message` is only supported for the  SQLite plugin.
>  'Invalid choice: compression_mode \'message\' is only supported '
>                               'with the sqlite3 storage plugin

It is actually more like only `mcap` is not supporting this mode because there are other thirdparty custom storage plugins exists and we don't empose any restriction on them. In general case they would support per-message compression the same way as SQLite storage plugin.

Also need to run uncrustify for the `sequential_compression_writer.cpp`

ros2bag/ros2bag/verb/record.py
```diff
@@ -197,6 +197,11 @@ def main(self, *, args):  # noqa: D102
             return print_error('Invalid choice: Cannot specify compression format '
                                'without a compression mode.')
 
+        if args.compression_mode == 'message' and args.storage == 'mcap':
+            return print_error('Invalid choice: compression_mode \'message\' is only supported '
+                               'with the sqlite3 storage plugin. When using the MCAP storage plugin, you can '
+                               'enable chunk compression by setting `compression: \'Zstd\'` in storage config')
+
```
fujitatomoya: I agree with @MichaelOrlov, why does generic CLI part need to be aware of the implementation details what is supported or not? instead, can we just call API and then catch the error or exception that response should tell why it fails?

@facontidavide if i may, what is the background for this PR? is there any specific issue to avoid this condition?

facontidavide: The background of PR is a request from Foxglove.
I am helping them with the WHAT and HOW, without questioning too much the WHY :wink: 

If you want to remove the code in the `record.py`, let me know, and I will

facontidavide:
@MichaelOrlov you are correct, the message should be changed.

Any other suggestion (I need to uncrustify, I know)?

MichaelOrlov:
@facontidavide As regards,
> Any other suggestion (I need to uncrustify, I know)?

No. In general, I am okay with this PR; we have talked about it with folks from Foxglove before.

ros2bag/ros2bag/verb/record.py
fujitatomoya: I do not think you need to remove this in necessary, thanks for sharing the information!

defunctzombie:
👋 happy to provide any additional details if needed but it seems like the questions were answered to everyone's satisfaction?

The gist is that `mcap` does not have `per-message` compression capabilities and we want to guide users towards success by recording mcap files that they can use inside and out of a ROS workspace (as many often want to do for analysis, viz, etc).

Here's an example where a user ran into a problem and it was not obvious to them they had miss-configured recordings to use per-message compression: https://github.com/foxglove/studio/issues/7315. This report is the genesis of this backport request.

ros2bag/ros2bag/verb/record.py
facontidavide:
```suggestion
        if args.compression_mode == 'message' and args.storage == 'mcap':
            return print_error('Invalid choice: compression_mode \'message\' is not supported '
                               'by the MCAP storage plugin. You can enable chunk compression by '
                               'setting `compression: \'Zstd\'` in storage config')

```

defunctzombie:
@facontidavide CI is failing on a few of the items - is this expected or something changed in this PR?

defunctzombie:
@facontidavide Any update here?

defunctzombie:
@facontidavide This is still failing CI

defunctzombie:
@MichaelOrlov Are there any other changes you want to see to this or is the last blocker passing CI?

MichaelOrlov:
@defunctzombie Yes, rather than CI all good and pretty trivial.

defunctzombie:
@facontidavide Will you be able to resolve the lint issue in this pr (per the above recommendation to run uncrustify) or are you no longer working on this PR?

MichaelOrlov:
- Closing this PR as stale. It will be superseded by https://github.com/ros2/rosbag2/pull/1782

Issue #1719, opened 2024-06-17
Ros bag reindex failed to generate topic inside metadata.yaml if the message_count is 0


KoolElmout:
## Description
On re-indexing a sqllite3 rosbag which originally contains a topic with 0 data inside (message_count:0), the metadata.yaml is not generated properly. The full topic is absent. 

This bug was discovered using the python Teranis rosbags reader's library.  (Confer https://gitlab.com/ternaris/rosbags/-/issues/91). 

### Expected Behavior
The reindex rosbag 's metadata.yaml file shall contains the topic with message_count:0
eg:

```
    - topic_metadata:
        name: /tract/control_signals
        type:msgs/msg/ControlSignal
        serialization_format: cdr
        offered_qos_profiles: "- history: 3\n  depth: 0\n  reliability: 1\n  durability: 2\n  deadline:\n    sec: 9223372036\n    nsec: 854775807\n  lifespan:\n    sec: 9223372036\n    nsec: 854775807\n  liveliness: 1\n  liveliness_lease_duration:\n    sec: 9223372036\n    nsec: 854775807\n  avoid_ros_namespace_conventions: false"
      message_count: 0  

```

### Actual Behavior
Actually, the full topic section is not present after re-index.

## To Reproduce
1) Take a sqlite bag with at least a topic which has a message_count at 0. 
2) Remove the metadata.yaml 
3) Launch the command
`ros2 bag reindex -s sqlite3  mybag
`
4) Check the generated metadata.yaml, all topics with 0 message count have disappeared. 

PS: This permit to re-generate the problem. In the true world, it shall permit to generate the metadata.yaml in case it has not been generated (which happens several time a day in my case.) 

## System (please complete the following information)
 - OS: [e.g. Ubuntu 22.04]
 - ROS 2 Distro: [Humble]
 - Install Method: [APT, release ]
 - Version: ros-humble-ros2bag                                0.15.9-1jammy.20240217.104352           amd6
 
## Additional context
If we place back from the original data to the re-indexed one the topic section, the python Teranis rosbags reader's library read the file properly. If not it failed. We conclude that the information shall be present inside the db3 but not reported in metadata.yaml.

MichaelOrlov:
Preliminary analysis. 
The problem is not really in the reindexer but rather in the `SqliteStorage::get_metadata()` function.
This could be fixed in the 
[...]

PR #1720, opened 2024-06-20
On ros humble, the ros2 bag reindex was generating metadata.yaml without topics which has 0 message count.

KoolElmout:
Add all topics present inside db3 to metadata.yaml  (even the 0 message count):

This patch solve the hereunder issue:
https://github.com/ros2/rosbag2/issues/1719

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ venv/
 **/.pytest_cache/
 __pycache__/
 *.db3-*
-
+log/
+install/
--- a/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
@@ -51,6 +51,10 @@ public:
 
   virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
 
+  virtual void setMetadataAllow0MessageCount(__attribute__((unused)) bool enable)
+  {
+  }
+
   virtual std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() const = 0;
 
   virtual void set_filter(const rosbag2_storage::StorageFilter & storage_filter) = 0;
--- a/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
@@ -72,6 +72,8 @@ public:
 
   const rosbag2_storage::BagMetadata & get_metadata() const override;
 
+  void setMetadataAllow0MessageCount(bool enable) override;
+
   std::vector<rosbag2_storage::TopicMetadata> get_all_topics_and_types() const override;
 
   void set_filter(const rosbag2_storage::StorageFilter & storage_filter) override;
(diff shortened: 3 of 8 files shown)


KoolElmout:
@dco, i am not well verse with git. So i do not want to destroy some work by doing mistake at this stage. Could you do the sign-off for me ? Here is my address "elmout Kool <elmout@rocketmail.com>" . 
 Regards

.gitignore
```diff
@@ -7,4 +7,5 @@ venv/
 **/.pytest_cache/
 __pycache__/
 *.db3-*
-
+log/
+install/
```
fujitatomoya, line 11: I would highly recommend to avoid unnecessary or not related changes to solve the issue. besides, why do we have these directories in `.gitignore` in this repository?

rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
```diff
@@ -51,6 +51,10 @@ class ROSBAG2_CPP_PUBLIC BaseReaderInterface
 
   virtual const rosbag2_storage::BagMetadata & get_metadata() const = 0;
 
+  virtual void setMetadataAllow0MessageCount(__attribute__((unused)) bool enable)
+  {
+  }
+
```
fujitatomoya, line 57: i am not sure why we need this into this class... instead we can address the issue in sqlite3 plugin implementation?

fujitatomoya (commented):
I think we need to target to `rolling` branch since this problem can be reproducible with `rolling` branch when we use `sqlite3` as backend database of rosbag2.

CC: @MichaelOrlov

.gitignore
KoolElmout: I am working with vscode, and if i do not add it inside it it keep proposing to commit it. But i agree it is not part of the patch

rosbag2_cpp/include/rosbag2_cpp/reader_interfaces/base_reader_interface.hpp
KoolElmout: The problem is that the sequential reader uses the base interface and not directly the sqlite implementation of it. So it is necessary to complete the interface by this function, then override it inside sqlite implementation. But maybe there is others solutions.

.gitignore
fujitatomoya: in that case, your global `.gitignore` should be set with this? and it works better for you. besides, not everyone is using vscode, i think that is user dependent setting, not repository global setting.

fujitatomoya:
@KoolElmout btw, https://github.com/ros2/rosbag2/pull/1722 @MichaelOrlov is trying to replace this? i guess.

MichaelOrlov (changes requested):
@KoolElmout Thanks for attempting to try to fix an issue. 
I am sorry, but not. At least not this way.
First of all, we can't introduce API/ABI breaking changes on already released ROS 2 distros. i.e. Humble.
The second one is that it is some mess with tests and overengineered SQL requests.
The implementation could be much more simple and without extra SQL requests, see my PR https://github.com/ros2/rosbag2/pull/1722 for Humble.

MichaelOrlov:
- Closing as wouldn't be merged in favor of the https://github.com/ros2/rosbag2/pull/1722

PR #1722, opened 2024-06-22
[Humble] Add topics with zero message counts to the SQLiteStorage::get_metadata().

MichaelOrlov:
- Closes https://github.com/ros2/rosbag2/issues/1719
- Analog and manual backport of the #1725. 
Have to do manual backporting because the code base is very different on humble branch in the affected places.

PR #1725, opened 2024-06-22
Add topics with zero message counts to the SQLiteStorage::get_metadata().

fujitatomoya:
- Address https://github.com/ros2/rosbag2/issues/1719 for `rolling`.
I guess we can backport this to `jazzy`.

### DIFF
--- a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
+++ b/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
@@ -771,7 +771,7 @@ void SqliteStorage::add_topic_to_metadata(
     // metadata_.version will be lower than 9
     offered_qos_profiles_str, (db_schema_version_ >= 3) ? metadata_.version : 8);
   const rosbag2_storage::TopicMetadata topic_metadata {
-    static_cast<uint16_t>(inner_topic_id), topic_name, topic_type, ser_format,
+    get_or_generate_extern_topic_id(inner_topic_id), topic_name, topic_type, ser_format,
     offered_qos_profiles, type_hash};
   auto & topics_list = metadata_.topics_with_message_count;
   auto it = std::find_if(
--- a/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
+++ b/rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
@@ -815,10 +815,10 @@ void SqliteStorage::read_metadata()
   if (database_->field_exists("topics", "offered_qos_profiles")) {
     if (database_->field_exists("topics", "type_description_hash")) {
       std::string query =
-        "SELECT topics.id, name, type, serialization_format, COUNT(messages.id), "
+        "SELECT messages.topic_id, name, type, serialization_format, COUNT(messages.id), "
         "MIN(messages.timestamp), MAX(messages.timestamp), offered_qos_profiles, "
-        "type_description_hash FROM topics LEFT OUTER JOIN messages on "
-        "topics.id = messages.topic_id GROUP BY topics.name;";
+        "type_description_hash FROM messages JOIN topics on topics.id = messages.topic_id "
+        "GROUP BY topics.name;";
 
       auto statement = database_->prepare_statement(query);
       auto query_results = statement->execute_query<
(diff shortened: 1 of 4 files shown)


rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -815,10 +815,10 @@ void SqliteStorage::read_metadata()
   if (database_->field_exists("topics", "offered_qos_profiles")) {
     if (database_->field_exists("topics", "type_description_hash")) {
       std::string query =
-        "SELECT messages.topic_id, name, type, serialization_format, COUNT(messages.id), "
+        "SELECT topics.id, name, type, serialization_format, COUNT(messages.id), "
         "MIN(messages.timestamp), MAX(messages.timestamp), offered_qos_profiles, "
-        "type_description_hash FROM messages JOIN topics on topics.id = messages.topic_id "
-        "GROUP BY topics.name;";
+        "type_description_hash FROM topics LEFT OUTER JOIN messages on "
```
fujitatomoya: this is the important part to join the table, previously excludes the contents from `topics` table if id does not match.

with this query, we can have the contents including message count is 0 as following.

```bash
sqlite> SELECT topics.id, name, type, serialization_format, COUNT(messages.id), MIN(messages.timestamp), MAX(messages.timestamp) FROM topics LEFT OUTER JOIN messages on topics.id = messages.topic_id GROUP BY topics.name;
5|/events/write_split|rosbag2_interfaces/msg/WriteSplitEvent|cdr|0||
4|/greetings|example_interfaces/msg/String|cdr|0||
1|/parameter_events|rcl_interfaces/msg/ParameterEvent|cdr|0||
3|/rosout|rcl_interfaces/msg/Log|cdr|9|1719079568093052120|1719079568099102171
2|/some_topic|std_msgs/msg/Bool|cdr|1|1719079568100694158|1719079568100694158
```

i have confirmed that this actually works with reproducible procedure explained in https://github.com/ros2/rosbag2/issues/1719

MichaelOrlov (changes requested):
@fujitatomoya Honestly, I am not a fan of overcomplicated SQL requests.
I found a more simple and elegant solution. See my implementation for Humble https://github.com/ros2/rosbag2/pull/1722

fujitatomoya:
> @fujitatomoya Honestly, I am not a fan of overcomplicated SQL requests. I found a more simple and elegant solution. See my implementation for Humble #1722

@MichaelOrlov hmmm, i prefer to let the SQL do the job since it is meant to be queried those index data... calling SQL query multiple times and then constructing the data in the program would be more complicated IMO... i guess current SQL query is not complicated at all, seems to me really straight forward.

but if you want to take your fix, that is fine too.

besides, i think there are slightly different from humble and rolling, right? so if you want to take this PR, i can manage rolling and jazzy on my side. what do you think?

MichaelOrlov:
@fujitatomoya  As regards the extra SQL request, we actually have to do this extra SQL request, `SqliteStorage::fill_topics_and_types()`, anyway at the very beginning, on construction or during storage opening.

Yes. the implementation is slightly different on Rolling and Jazzy, this is why I started from Humble. 
However, I would prefer my version of the fix. I would appreciate it if you would take it for rolling and jazzy.

fujitatomoya:
@MichaelOrlov i do not have strong opinion on this, so okay to take your fix. do you want me to work on rolling and jazzy? or we should wait https://github.com/ros2/rosbag2/pull/1722 and backport it to rolling and jazzy?

MichaelOrlov:
@fujitatomoya I would appreciate it if you could prepare PR for Rolling, and then we can backport it to Jazzy.

fujitatomoya:
@MichaelOrlov okay, i will borrow your code and adjust it to rolling, then backport to jazzy!

fujitatomoya:
@MichaelOrlov it now aligns with https://github.com/ros2/rosbag2/pull/1722, CI is passing on my local env. can you take a look? just a slightly structure is different from humble one but complicated. (note, added you as co-author since i borrowed your code.)

MichaelOrlov (commented):
@fujitatomoya Thanks, overall looks good. 
Only one concern about failure on the `Build_and_Test` job. It seems we missed dependencies from the `std_msgs` package for test.

fujitatomoya:
@MichaelOrlov https://build.ros2.org/job/Rpr__rosbag2__ubuntu_noble_amd64/235/ seems unrelated to this change, everything else is green.

MichaelOrlov (approved):
LGTM with green CI from buildfarm.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
```diff
@@ -770,14 +770,27 @@ void SqliteStorage::add_topic_to_metadata(
...
-  metadata_.topics_with_message_count.push_back(
-    {
-      {
-        get_or_generate_extern_topic_id(inner_topic_id), topic_name, topic_type, ser_format,
-        offered_qos_profiles, type_hash
-      },
-      static_cast<size_t>(msg_count)
+  const rosbag2_storage::TopicMetadata topic_metadata {
+    static_cast<uint16_t>(inner_topic_id), topic_name, topic_type, ser_format,
+    offered_qos_profiles, type_hash};
```
MichaelOrlov:
```suggestion
  const rosbag2_storage::TopicMetadata topic_metadata {
    get_or_generate_extern_topic_id(inner_topic_id), topic_name, topic_type, ser_format,
    offered_qos_profiles, type_hash};
```
Ups found a bug.

MichaelOrlov (changes requested):
@fujitatomoya Sorry for the late call - but I found a bug in the implementation.
When creating a `TopicMetadata` object, we need to use mapping from the `inner_topic_id` to the `extern_topic_id` for further comparison with an existing object.

rosbag2_storage_sqlite3/src/rosbag2_storage_sqlite3/sqlite_storage.cpp
fujitatomoya: yeah right, that was my oversight sorry about that. addressed in https://github.com/ros2/rosbag2/pull/1725/commits/1708c92b0b6e57358325aef392c7e1548ecef320

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21326/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15639/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/837/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22056/)

MichaelOrlov (approved):
LGTM.

fujitatomoya:
@MichaelOrlov can you merge this?

MichaelOrlov:

PR #1576, opened 2024-02-28
Use std::filesystem instead of rcpputils::fs

r7vme:
Based on PRs https://github.com/ros2/rosbag2/pull/1478 and https://github.com/ros2/rosbag2/pull/1479

Use std::filesystem instead of rcpputils::fs (https://github.com/ros2/rcpputils/issues/164) in all packages of rosbag2.

~~The only place where it is still used is TemporaryDirectoryFixture in https://github.com/ros2/rosbag2/blob/1564687de2d10d5bce32dcb9b38ec5c9239416a4/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp#L34~~

~~Because there is no alternative in the standard library~~

PR #1740, opened 2024-07-09
Remove rcpputils::fs dependencies in rosbag2 packages

MichaelOrlov:
- This PR removes residual dependencies from the `rcpputils::fs` in all rosbag2 packages. 
- Rationale: The `rcpputils::fs` will be deprecated soon, and it is advised that `std::filesystem` be used instead.

Added own `rosbag2_test_common::create_temp_directory(..)` function in replacement of the analog from the `rcpputils::fs` since such functionality doesn't exists in the `std::filesystem`.

### DIFF
--- a/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
@@ -36,7 +36,7 @@ public:
 
   ~TemporaryDirectoryFixture() override
   {
-    rcpputils::fs::remove_all(rcpputils::fs::path(temporary_dir_path_));
+    std::filesystem::remove_all(std::filesystem::path(temporary_dir_path_));
   }
 
   std::string temporary_dir_path_;
--- a/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
@@ -17,72 +17,26 @@
 
 #include <gmock/gmock.h>
 
-#include <cstdio>
-#include <exception>
-#include <filesystem>
-#include <fstream>
-#include <iostream>
-#include <random>
-#include <sstream>
 #include <string>
+
+#include "rcpputils/filesystem_helper.hpp"
 
 using namespace ::testing;  // NOLINT
 
 namespace rosbag2_test_common
 {
 
-/// \brief Construct a uniquely named temporary directory, in "parent", with format base_nameXXXXXX
-/// The output, if successful, is guaranteed to be a newly-created directory.
-/// The underlying implementation keeps generating paths until one that does not exist is found or
-/// until the number of iterations exceeded the maximum tries.
-/// This guarantees that there will be no existing files in the returned directory.
-/// \param[in] base_name User-specified portion of the created directory.
-/// \param[in] parent_path The parent path of the directory that will be created.
-/// \param[in] max_tries The maximum number of tries to find a unique directory (default 1000)
-/// \return A path to a newly-created directory with base_name and a 6-character unique suffix.
-/// \throws std::system_error If any OS APIs do not succeed.
-/// \throws std::runtime_error If the number of the iterations exceeds the maximum tries and
-/// a unique directory is not found.
-std::filesystem::path create_temp_directory(
-  const std::string & base_name,
-  std::filesystem::path parent_path = std::filesystem::temp_directory_path(),
-  size_t max_tries = 1000)
-{
-  // mersenne twister random generator engine seeded with the std::random_device
-  std::mt19937 random_generator(std::random_device{}());
-  std::uniform_int_distribution<> distribution(0, 999999);
-  std::filesystem::path path_to_temp_dir;
-  constexpr size_t kSuffixLength = 7;  // 6 chars + 1 null terminator
-  char random_suffix_str[kSuffixLength];
-  size_t current_iteration = 0;
-  while (true) {
-    snprintf(random_suffix_str, kSuffixLength, "%06x", distribution(random_generator));
-    const std::string random_dir_name = base_name + random_suffix_str;
-    path_to_temp_dir = parent_path / random_dir_name;
-    // true if the directory was newly created.
-    if (std::filesystem::create_directory(path_to_temp_dir)) {
-      break;
-    }
-    if (current_iteration == max_tries) {
-      throw std::runtime_error(
-              "Exceeded maximum allowed iterations to find non-existing directory");
-    }
-    current_iteration++;
-  }
-  return path_to_temp_dir;
-}
-
 class TemporaryDirectoryFixture : public Test
 {
 public:
   TemporaryDirectoryFixture()
   {
-    temporary_dir_path_ = create_temp_directory("tmp_test_dir_").string();
+    temporary_dir_path_ = rcpputils::fs::create_temporary_directory("tmp_test_dir_").string();
   }
 
   ~TemporaryDirectoryFixture() override
   {
-    std::filesystem::remove_all(std::filesystem::path(temporary_dir_path_));
+    rcpputils::fs::remove_all(rcpputils::fs::path(temporary_dir_path_));
   }
 
   std::string temporary_dir_path_;
--- a/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
+++ b/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
@@ -17,26 +17,72 @@
 
 #include <gmock/gmock.h>
 
+#include <cstdio>
+#include <exception>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <random>
+#include <sstream>
 #include <string>
-
-#include "rcpputils/filesystem_helper.hpp"
 
 using namespace ::testing;  // NOLINT
 
 namespace rosbag2_test_common
 {
 
+/// \brief Construct a uniquely named temporary directory, in "parent", with format base_nameXXXXXX
+/// The output, if successful, is guaranteed to be a newly-created directory.
+/// The underlying implementation keeps generating paths until one that does not exist is found or
+/// until the number of iterations exceeded the maximum tries.
+/// This guarantees that there will be no existing files in the returned directory.
+/// \param[in] base_name User-specified portion of the created directory.
+/// \param[in] parent_path The parent path of the directory that will be created.
+/// \param[in] max_tries The maximum number of tries to find a unique directory (default 1000)
+/// \return A path to a newly-created directory with base_name and a 6-character unique suffix.
+/// \throws std::system_error If any OS APIs do not succeed.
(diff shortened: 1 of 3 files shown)


r7vme:
@MichaelOrlov I looked thru the code and it looks ok from my side. I remember there were some Windows related [problems](https://github.com/ros2/rosbag2/pull/1576#issuecomment-2007620604) back, when we tried to do similar change.

rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
```diff
@@ -17,26 +17,72 @@
...
+/// until the number of iterations exceeded the maximum tries.
+/// This guarantees that there will be no existing files in the returned directory.
+/// \param[in] base_name User-specified portion of the created directory.
+/// \param[in] parent_path The parent path of the directory that will be created.
+/// \param[in] max_tries The maximum number of tries to find a unique directory (default 1000)
+/// \return A path to a newly-created directory with base_name and a 6-character unique suffix.
+/// \throws std::system_error If any OS APIs do not succeed.
+/// \throws std::runtime_error If the number of the iterations exceeds the maximum tries and
+/// a unique directory is not found.
+std::filesystem::path create_temp_directory(
```
ahcorde: why you don't use `rcpputils::fs::create_temp_directory` ?

ahcorde (changes requested):
Just a question

rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
clalancette: Yeah, I think we should switch to using the `rcpputils::fs::create_temp_directory` here.  If there are still problems with it, then we should fix them over in `rcpputils`.

MichaelOrlov: @ahcorde @clalancette The reasons why I wrote my own implementation of the `create_temp_directory` is the following:
1. Because [last time](https://github.com/ros2/rosbag2/pull/1576#issuecomment-2007620604) when we tried to use `rcpputils::fs::create_temp_directory` alongside with cleanup via `std::filesystem::remove_all`, the `std::filesystem::remove_all` was failing on Windows and also caused some test failures. See https://github.com/ros2/rosbag2/pull/1576#issuecomment-2006060701 and https://github.com/ros2/rosbag2/pull/1576#pullrequestreview-1950086019 for details. 
I have tried to triage this problem on Windows. It turns out that in failing tests the rosbag2 creating subfolder with files in the temporary created folder from `rcpputils::fs::create_temp_directory`. For some reason, likely because `rcpputils::fs::create_temp_directory` inside using platform-dependent function calls, i.e., `_mktemp_s(..)` and own implementation for `create_directories(const path & p)` the `std::filesystem::remove_all` fails when trying to delete such folder structures. However, if create a temp folder manually or use the temp folder returned by the `std::filesystem::temp_directory_path()` without creating a subfolder in it with `rcpputils::fs::create_temp_directory`, the `std::filesystem::remove_all` is not failing.
2. As far as I understand, it was a motion to deprecate functions from the `rcpputils::fs` and use alternatives from the `std::filesystem`.
3. I don't like the idea of usage `std::filesystem::path` with the `rcpputils::fs::path`. It would be more consistent and more robust to use only `std::filesystem::path`. If we decided to get rid from the `rcpputils::fs` let's do it! We are using  `rcpputils::fs::create_temp_directory` only in two places in the core packages in the `rosbag2` and in tests for the [rcl_logging_spdlog](https://github.com/ros2/rcl_logging/blob/rolling/rcl_logging_spdlog/test/test_logging_interface.cpp#L110).

Let's see how Windows CI will pass. If it will pass without failures, I don't mind to move my `rosbag2_test_common::create_temp_directory(..)` to the `rcpputils::fs` namespace.

clalancette: > * I don't like the idea of usage `std::filesystem::path` with the `rcpputils::fs::path`. It would be more consistent and more robust to use only `std::filesystem::path`. If we decided to get rid from the `rcpputils::fs` let's do it! We are using  `rcpputils::fs::create_temp_directory` only in two places in the core packages in the `rosbag2` and in tests for the [rcl_logging_spdlog](https://github.com/ros2/rcl_logging/blob/rolling/rcl_logging_spdlog/test/test_logging_interface.cpp#L110).

Yes, we are actively doing this right now in https://github.com/ros2/rcpputils/pull/197 and https://github.com/ros2/rcpputils/pull/196 .  That said, `std::filesystem` doesn't have any equivalent to `rcpputils::fs::create_temp_directory`, so we are going to have to carry around that function indefinitely.

MichaelOrlov:
Pulls: ros2/rosbag2#1740
Gist: https://gist.githubusercontent.com/MichaelOrlov/57ea65444857fed55ac99b2bdb80c0a1/raw/c8524ce3ec5433097215036dbc53c6c6f6aaea97/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_storage rosbag2_transport rosbag2_test_common  rosbag2_tests
TEST args:  --packages-above rosbag2_cpp rosbag2_storage rosbag2_transport rosbag2_test_common  rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14211
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21447/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15735/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/943/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22167/)

rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
MichaelOrlov: @clalancette If CI will pass green I would like to move my implementation of the `create_temp_directory` to the `rcpputils::fs` as is but with the new name `create_temporary_directory` and deprecate the original `rcpputils::fs::create_temp_directory` eventually.

clalancette: > @clalancette If CI will pass green I would like to move my implementation of the `create_temp_directory` to the `rcpputils::fs` as is but with the new name `create_temporary_directory` and deprecate the original `rcpputils::fs::create_temp_directory` eventually.

I don't think we should do that.  We are currently in the process of breaking the API for `create_temp_directory` anyway, so I think we should do this all at the same time.  Please take a look at https://github.com/ros2/rcpputils/pull/197 and give feedback on what we think the API should be.

MichaelOrlov: BTW. The implementation in the https://github.com/ros2/rcpputils/pull/197/files has API-breaking changes and still uses platform-dependent code, which is not good.  
Creating a new function with a different name and deprecating the old one would be better.

clalancette: > BTW. The implementation in the https://github.com/ros2/rcpputils/pull/197/files has API-breaking changes and still uses platform-dependent code, which is not good.

Yes, it has API-breaking changes.  See the discussion in https://github.com/ros2/rcpputils/pull/196#pullrequestreview-2167477783 about why we decided to do that.

MichaelOrlov: @clalancette I see. However, I still don't like the idea of doing API-breaking changes, which is require to merge 3 PRs in synch.  
Why not do it sequentially by creating a new function with `create_temporary_directory` name? 
I've tried to look out, and it seems this name is not occupied in the ROS 2 core packages.

MichaelOrlov: @clalancette FYI In Apex.AI, when we are doing synch with upstream ROS 2 by pulling PRs each time, such breaking changes, which require merging multiple PRs in different packages in synch, is a nightmare since it is breaking our CI.

We would be grateful if such breaking changes will be as minimum as possible.

MichaelOrlov:
The CI failed on `rosbag2_storage_sqlite3.test_sqlite_storage`. It looks like a failure in the test's destructor. It could be related to the failure in the `std::filesystem::remove_all`. I will try to reproduce it locally and rerun Windows CI.

MichaelOrlov:
Re-run Windows CI
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22172/)

MichaelOrlov:
Re-run Windows CI after attempting to fix test failure.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22173/)

MichaelOrlov:
Update: On the latest ROS 2 issues triage meeting we decided:
1. [x] Move implementation of the newly added `rosbag2_test_common::create_temp_directory(..)` function to the `rcpputils::fs` with the new name `create_temporary_directory(..)`. Will be done in the https://github.com/ros2/rcpputils/pull/198
2. [x] Deprecate `rcpputils::fs::create_temp_directory(..)` and `rcpputils::fs::temp_directory_path()`
3. [x] Move fix for the failing `throws_on_invalid_pragma_in_config_file` test to the separate PR. Will be handled in the https://github.com/ros2/rosbag2/pull/1742
4. [x] Get back to this PR when https://github.com/ros2/rosbag2/pull/1742 and https://github.com/ros2/rcpputils/pull/198 will be merged.

rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp
```diff
@@ -17,72 +17,26 @@
...
   TemporaryDirectoryFixture()
   {
-    temporary_dir_path_ = create_temp_directory("tmp_test_dir_").string();
+    temporary_dir_path_ = rcpputils::fs::create_temporary_directory("tmp_test_dir_").string();
   }
 
   ~TemporaryDirectoryFixture() override
   {
-    std::filesystem::remove_all(std::filesystem::path(temporary_dir_path_));
+    rcpputils::fs::remove_all(rcpputils::fs::path(temporary_dir_path_));
```
MichaelOrlov: @ahcorde It will be better to use `std::filesystem::remove_all(path)` for deletion. 
Eventually, we want to deprecate and remove `rcpputils::fs::remove_all`

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21517/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15799/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1015/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22233/)

Included https://github.com/ros2/rosbag2/pull/1740/ and https://github.com/ros2/rcl_logging/pull/117

ahcorde (approved):

MichaelOrlov:

PR #1709, opened 2024-06-12
Update CI scripts to use Ubuntu Noble distros and bump action scripts to latest versions

MichaelOrlov:
- Use Ubuntu Noble distros for ci jobs on rolling.
- Bump `actions-ros-lint` to v0.1.3 and `actions/checkout` to v4.
- Exclude `cppcheck` from CI due to a known issue that the new versions are very slow.
See https://github.com/ament/ament_lint/pull/345 for details.

PR #1763, opened 2024-07-26
Fixes for new mypy (Ubuntu 24.04)

r7vme:
Based on https://github.com/ros2/rosbag2/pull/1709

This PR not for merge, but to be pulled by https://github.com/ros2/rosbag2/pull/1709


rosbag2_py/README.md
```diff
@@ -14,7 +14,12 @@ source install/setup.sh
 # Make sure rosbag2_py can be imported
 python3 -c 'import rosbag2_py'
 
-sudo apt update && sudo apt install mypy=0.942-1ubuntu1
+# Ubuntu 24.04
+sudo apt update && sudo apt install mypy
+
+# Older Ubuntu
+# sudo pip3 install -U --break-system-packages mypy==1.9
```
MichaelOrlov, line 21: @r7vme The problem is that on Ubuntu 22.04 the default pip3 version is 22.0.2 and python version is 3.10 and there is no `--break-system-packages` option for pip3 install on those versions.

MichaelOrlov: Another aspect of installing via pip3.
What if the user already has an old version of mypy installed with the
> sudo apt update && sudo apt install mypy

What version of the mypy will be used? It seems like an undefined behavior that we would like to avoid.

r7vme: @MichaelOrlov 

> The problem is that on Ubuntu 22.04 the default pip3 version is 22.0.2 and python version is 3.10 and there is no --break-system-packages option for pip3 install on those versions.

My bad, `--break-system-packages` should be removed

> What version of the mypy will be used? It seems like an undefined behavior that we would like to avoid.

I tested in docker container with ubuntu:jammy and after installing 1.9.0 via pip and 0.942 via apt. I got

```
root@f647688bc269:/# mypy --version
mypy 1.9.0 (compiled: yes)
```

MichaelOrlov: @r7vme Ok it works. 
Only one concern is about `sudo` for `pip3 install` command.
It seems we can install mypy without `sudo`.

MichaelOrlov: Without sudo mypy installing in the user local folder. e.g. `/home/morlov/.local/bin/mypy`

MichaelOrlov (approved):

PR #1773, opened 2024-08-06
Update lz4 to v1.10

mosfet80:
implements multithread support . ->increase speed
a lot of fix https://github.com/lz4/lz4/releases/tag/v1.10.0


clalancette (changes requested):
As with other places, I'd prefer if this stayed the same as the version released in Ubuntu 24.04 (which is 1.9.4).  So I don't think we should take this update.

MichaelOrlov:
- Closing as will not be merged with the rationale explained in the https://github.com/ros2/rosbag2/pull/1773#pullrequestreview-2220944309

PR #1784, opened 2024-08-10
Update zstd lib

mosfet80:
Improvements and fix list
https://github.com/facebook/zstd/releases/tag/v1.5.6


fujitatomoya (approved):
lgtm with green CI.

clalancette (changes requested):
As in other places, I'd prefer to keep this the same as the version installed in Ubuntu 24.04, which is v1.5.5.  So I don't think we should take this.

MichaelOrlov:
- Closing as will not be merged, because we want to use the same version as in the Ubuntu 24.04 distro.

PR #1785, opened 2024-08-10
Update sqlite

mosfet80:
Allocate additional memory from the heap for the SQL parser stack if that stack overflows, rather than reporting a "parser stack overflow" error.

Complete change list::
https://www.sqlite.org/releaselog/3_46_0.html

### DIFF
--- a/sqlite3_vendor/CMakeLists.txt
+++ b/sqlite3_vendor/CMakeLists.txt
@@ -15,8 +15,8 @@ endif()
 ament_vendor(sqlite3_vendor
   SATISFIED ${SQLite3_FOUND}
   VCS_TYPE zip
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
lgtm with green CI

clalancette (changes requested):
As with other vendored packages, I'd prefer to keep this the same as the version in Ubuntu 24.04, which is 3.45.1 (the version already vendored here).  So I don't think we should take this at this time.

fujitatomoya:
@clalancette Ah okay, that is the the policy basically. (unless something comes up) i did not know that, thanks!

MichaelOrlov:
- Closing as will not be merged with explanation in the https://github.com/ros2/rosbag2/pull/1785#pullrequestreview-2245807674

PR #1789, opened 2024-08-14
Add missing throws documentation to  rosbag_cpp writer open() #1788

### LINKED ISSUE #1788: Missing "throws" documentation in rosbag2_cpp writer open()
## Description
The documentation of [rosbag2_cpp/include/writer.hpp](https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/include/rosbag2_cpp/writer.hpp) and [rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp](https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp) are missing an entry that indicates that it may throw.

**Missing documentation:**

https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L78

https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/include/rosbag2_cpp/writer.hpp#L87

### Expected Behavior
 Something like this in `writer.hpp` and `sequential_writer.hpp`:
 
 ```
    *
    * \param storage_options Options to configure the storage
    * \param converter_options options to define in which format incoming messages are stored
+   * \throws runtime_error if 
+   * database directory already exists,
+   * failed to create database directory,
+   * no storage could be initialized,
+   * invalid bag splitting size given,
+   * max cache size less or equal 0 when snapshot mode is enabled.
    **/
   void open(
 ```

### Actual Behavior
Missing documentation.

## Additional context

**The instances where it throws are:**
https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L122
https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L129
https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L135
https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L145
https://github.com/ros2/rosbag2/blob/ce918b786095757a458a71b5b36ccdc818851246/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L150

gDorndorf:
Closes #1788 

Document that open() may throw.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -74,12 +74,7 @@ public:
    * more storage and converter options.
    *
    * \param storage_uri URI of the storage to open.
-   * \throws runtime_error if 
-   * database directory already exists,
-   * failed to create database directory,
-   * no storage could be initialized,
-   * invalid bag splitting size given,
-   * max cache size less or equal 0 when snapshot mode is enabled.
+   * \throws runtime_error if internal error happens.
    **/
   void open(const std::string & uri);
 
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -74,6 +74,12 @@ public:
    * more storage and converter options.
    *
    * \param storage_uri URI of the storage to open.
+   * \throws runtime_error if 
+   * database directory already exists,
+   * failed to create database directory,
+   * no storage could be initialized,
+   * invalid bag splitting size given,
+   * max cache size less or equal 0 when snapshot mode is enabled.
    **/
   void open(const std::string & uri);
 
@@ -83,6 +89,12 @@ public:
    *
    * \param storage_options Options to configure the storage
    * \param converter_options options to define in which format incoming messages are stored
+   * \throws runtime_error if 
+   * database directory already exists,
+   * failed to create database directory,
+   * no storage could be initialized,
+   * invalid bag splitting size given,
+   * max cache size less or equal 0 when snapshot mode is enabled.
    **/
   void open(
     const rosbag2_storage::StorageOptions & storage_options,
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -77,12 +77,10 @@ public:
    *
    * \param storage_options Options to configure the storage
    * \param converter_options options to define in which format incoming messages are stored
-   * \throws runtime_error if 
-   * database directory already exists,
-   * failed to create database directory,
-   * no storage could be initialized,
-   * invalid bag splitting size given,
-   * max cache size less or equal 0 when snapshot mode is enabled.
+   * \throws runtime_error if database directory already exists,
+   *   failed to create database directory, no storage could be initialized,
+   *   invalid bag splitting size given,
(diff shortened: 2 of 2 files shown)


rosbag2_cpp/include/rosbag2_cpp/writer.hpp
```diff
@@ -74,6 +74,12 @@ class ROSBAG2_CPP_PUBLIC Writer
    * more storage and converter options.
    *
    * \param storage_uri URI of the storage to open.
+   * \throws runtime_error if 
+   * database directory already exists,
+   * failed to create database directory,
+   * no storage could be initialized,
+   * invalid bag splitting size given,
+   * max cache size less or equal 0 when snapshot mode is enabled.
```
fujitatomoya: I think exception behavior is dependent on the writer implementation, so how about the following?

```suggestion
   * \throws runtime_error if internal error happens.
```

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -77,6 +77,12 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
    *
    * \param storage_options Options to configure the storage
    * \param converter_options options to define in which format incoming messages are stored
+   * \throws runtime_error if 
+   * database directory already exists,
+   * failed to create database directory,
+   * no storage could be initialized,
+   * invalid bag splitting size given,
+   * max cache size less or equal 0 when snapshot mode is enabled.
```
fujitatomoya: I am not sure if we want to list the all possible exception here, but at least indent is needed as below.

```suggestion
   * \throws runtime_error if database directory already exists,
   *   failed to create database directory, no storage could be initialized,
   *   invalid bag splitting size given,
   *   max cache size less or equal 0 when snapshot mode is enabled.
```

gDorndorf: I agree that introduces some maintenance hazard keeping doc and implementation in sync, but it could be practical to have it there.

I'd also be fine with just a hint at some general throw behavior here. 
So I'd leave it up to any reviewer as I am not too familiar with the style use throughout the rest of the codebase.

MichaelOrlov (changes requested):
@gDorndorf Thanks for PR and attempt to make our Rosbag2 better.
However, in this particular case documenting about possible throwing exceptions is not trivial. Because this is an upper `rosbag2_cpp` layer API which is calling storage plugin API inside. 
Each storage plugin may throw or not and what type of exceptions it throws and in what cases are implementation details for concrete storage plugin.

What we can do is make some notes that certain functions may throw if the underlying rosbag2_storage plugin API also throws. 
Update: we can be more specific and refer to the concrete storage_plugin API interface.

emersonknapp:
Possibly a better API would be a no-throw `bool` API instead of `void`. The `Writer` itself does not throw any exceptions from its own code, but it does have the ability to catch them when making unsafe calls to the implementations

PR #1791, opened 2024-08-22
[Rolling] Release 0.29.0

ahcorde:
Once this is approved, I'll squash merge it onto `rolling`, add the 0.28.1 tag, and bloom-release it.

PR #1796, opened 2024-08-28
Improve the reliability of rosbag2 tests

clalancette:
I initially started this series trying to track down a rare failing flakey test with `test_record__rmw_cyclonedds_cpp`.  That particular flake seems to be able to happen because sometimes discovery takes longer than we expect, and it is possible that the tests "miss" the first publication.  If that's the case, then the rest of the test may fail because it is expecting a certai number of messages.  Along the way, we cleanup the tests a bit:

1.  Remove the unnecessary `wait_until_shutdown` method, which was almost exactly the same as `wait_for_condition`.
2.  Slightly revamp how the async spinner works, so it doesn't need to call `rclcpp::shutdown`.
3.  Do a similar revamp for the custom async spinner in rosbag2_tests.
4.  Wait for topics to be discovered right after calling `recorder->record()`.  This ensures we can't get into the above situation.

With this series in place, the particular flake of `test_record` on `rmw_cyclonedds_cpp` is fixed (or, at least, I can no longer reproduce it).  There are still other flakes that can happen under load, but I think fixes for those will have to come separately.

### DIFF
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -67,6 +67,7 @@ public:
 
   void TearDown() override
   {
+    stop_spinning();
     fs::remove_all(root_bag_path_);
   }
 
@@ -200,8 +201,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_services_only
   recorder->record();
 
   start_async_spin(recorder);
-  auto cleanup_process_handle = rcpputils::make_scope_exit(
-    [&]() {stop_spinning();});
 
   ASSERT_TRUE(service_client_manager->wait_for_service_to_be_ready());
   ASSERT_TRUE(wait_for_subscriptions(*recorder, {"/test_service/_service_event"}));
@@ -227,8 +226,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_services_only
   EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
 
   recorder->stop();
-  stop_spinning();
-  cleanup_process_handle.cancel();
 
   rosbag2_cpp::Info info;
   std::vector<std::shared_ptr<rosbag2_cpp::rosbag2_service_info_t>> ret_service_infos;
@@ -278,8 +275,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_and_se
   recorder->record();
 
   start_async_spin(recorder);
-  auto cleanup_process_handle = rcpputils::make_scope_exit(
-    [&]() {stop_spinning();});
 
   ASSERT_TRUE(
     wait_for_subscriptions(
@@ -314,8 +309,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_and_se
   EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
 
   recorder->stop();
-  stop_spinning();
-  cleanup_process_handle.cancel();
 
   rosbag2_cpp::Info info;
   std::vector<std::shared_ptr<rosbag2_cpp::rosbag2_service_info_t>> ret_service_infos;
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -201,33 +201,36 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_services_only
     std::move(writer), storage_options, record_options);
   recorder->record();
 
-  start_async_spin(recorder);
-  auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
-
-  ASSERT_TRUE(service_client_manager->wait_for_service_to_be_ready());
-  ASSERT_TRUE(wait_for_subscriptions(*recorder, {"/test_service/_service_event"}));
-
   constexpr size_t num_service_requests = 3;
-  for (size_t i = 0; i < num_service_requests; i++) {
-    ASSERT_TRUE(service_client_manager->send_request());
-    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+  start_async_spin(recorder);
+  {
+    auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
+
+    ASSERT_TRUE(service_client_manager->wait_for_service_to_be_ready());
+    ASSERT_TRUE(wait_for_subscriptions(*recorder, {"/test_service/_service_event"}));
+
+    for (size_t i = 0; i < num_service_requests; i++) {
+      ASSERT_TRUE(service_client_manager->send_request());
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+
+    auto & writer_ref = recorder->get_writer_handle();
+    auto & recorder_writer =
+      dynamic_cast<SequentialWriterForTest &>(writer_ref.get_implementation_handle());
+
+    // By default, only client introspection is enabled.
+    // For one request, service event topic get 2 messages.
+    size_t expected_messages = num_service_requests * 2;
+    auto ret = rosbag2_test_common::wait_until_condition(
+      [&recorder_writer, &expected_messages]() {
+        return recorder_writer.get_number_of_written_messages() >= expected_messages;
+      },
+      std::chrono::seconds(5));
+    EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
+
+    recorder->stop();
   }
-
-  auto & writer_ref = recorder->get_writer_handle();
-  auto & recorder_writer =
-    dynamic_cast<SequentialWriterForTest &>(writer_ref.get_implementation_handle());
-
-  // By default, only client introspection is enabled.
-  // For one request, service event topic get 2 messages.
-  size_t expected_messages = num_service_requests * 2;
-  auto ret = rosbag2_test_common::wait_until_condition(
-    [&recorder_writer, &expected_messages]() {
-      return recorder_writer.get_number_of_written_messages() >= expected_messages;
-    },
-    std::chrono::seconds(5));
-  EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
-
-  recorder->stop();
 
   rosbag2_cpp::Info info;
   std::vector<std::shared_ptr<rosbag2_cpp::rosbag2_service_info_t>> ret_service_infos;
@@ -276,42 +279,45 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_and_se
     std::move(writer), storage_options, record_options);
   recorder->record();
 
-  start_async_spin(recorder);
-  auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
-
-  ASSERT_TRUE(
-    wait_for_subscriptions(
-      *recorder,
-      {"/test_service1/_service_event",
-        "/test_service2/_service_event",
-        "/test_topic1",
-        "/test_topic2"}
-    )
-  );
-
   constexpr size_t num_service_requests = 2;
-  for (size_t i = 0; i < num_service_requests; i++) {
-    ASSERT_TRUE(service_client_manager1->send_request());
-    ASSERT_TRUE(service_client_manager2->send_request());
-    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+
+  start_async_spin(recorder);
+  {
+    auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
+
+    ASSERT_TRUE(
+      wait_for_subscriptions(
+        *recorder,
+        {"/test_service1/_service_event",
+          "/test_service2/_service_event",
+          "/test_topic1",
+          "/test_topic2"}
+      )
+    );
+
+    for (size_t i = 0; i < num_service_requests; i++) {
+      ASSERT_TRUE(service_client_manager1->send_request());
+      ASSERT_TRUE(service_client_manager2->send_request());
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+    pub_manager.run_publishers();
+
+    auto & writer_ref = recorder->get_writer_handle();
+    auto & recorder_writer =
+      dynamic_cast<SequentialWriterForTest &>(writer_ref.get_implementation_handle());
+
+    // By default, only client introspection is enabled.
+    // For one request, service event topic get 2 messages.
+    size_t expected_messages = num_service_requests * 2 + 2;
+    auto ret = rosbag2_test_common::wait_until_condition(
+      [&recorder_writer, &expected_messages]() {
+        return recorder_writer.get_number_of_written_messages() >= expected_messages;
+      },
+      std::chrono::seconds(5));
+    EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
+
+    recorder->stop();
   }
-  pub_manager.run_publishers();
-
-  auto & writer_ref = recorder->get_writer_handle();
-  auto & recorder_writer =
-    dynamic_cast<SequentialWriterForTest &>(writer_ref.get_implementation_handle());
-
-  // By default, only client introspection is enabled.
-  // For one request, service event topic get 2 messages.
-  size_t expected_messages = num_service_requests * 2 + 2;
-  auto ret = rosbag2_test_common::wait_until_condition(
-    [&recorder_writer, &expected_messages]() {
-      return recorder_writer.get_number_of_written_messages() >= expected_messages;
-    },
-    std::chrono::seconds(5));
-  EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
-
-  recorder->stop();
 
   rosbag2_cpp::Info info;
   std::vector<std::shared_ptr<rosbag2_cpp::rosbag2_service_info_t>> ret_service_infos;
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -14,8 +14,10 @@
 
 #include <gmock/gmock.h>
 
+#include <atomic>
 #include <chrono>
 #include <filesystem>
+#include <stdexcept>
 #include <string>
 
 #include "rosbag2_cpp/info.hpp"
@@ -81,16 +83,20 @@ public:
   template<class T>
   void start_async_spin(T node)
   {
-    node_spinner_future_ = std::async(
-      std::launch::async,
-      [node, this]() -> void {
-        rclcpp::executors::SingleThreadedExecutor exec;
-        exec.add_node(node);
-        while (rclcpp::ok() && !exit_from_node_spinner_) {
-          exec.spin_some(std::chrono::milliseconds(100));
-        }
-        exec.remove_node(node);
-      });
+    if (!exit_from_node_spinner_.exchange(false)) {
+      node_spinner_future_ = std::async(
+        std::launch::async,
+        [node, this]() -> void {
+          rclcpp::executors::SingleThreadedExecutor exec;
+          exec.add_node(node);
+          while (rclcpp::ok() && !exit_from_node_spinner_) {
+            exec.spin_some(std::chrono::milliseconds(100));
+          }
+          exec.remove_node(node);
+        });
+    } else {
+      throw std::runtime_error("Already spinning a node, can't start a new node spin");
+    }
   }
 
   void stop_spinning()
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -83,16 +83,12 @@ public:
   template<class T>
   void start_async_spin(T node)
   {
-    if (!exit_from_node_spinner_.exchange(false)) {
-      node_spinner_future_ = std::async(
-        std::launch::async,
-        [node, this]() -> void {
-          rclcpp::executors::SingleThreadedExecutor exec;
-          exec.add_node(node);
-          while (rclcpp::ok() && !exit_from_node_spinner_) {
-            exec.spin_some(std::chrono::milliseconds(100));
-          }
-          exec.remove_node(node);
+    if (exec_ == nullptr) {
+      exec_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
+      exec_->add_node(node);
+      spin_thread_ = std::thread(
+        [this]() {
+          exec_->spin();
         });
     } else {
       throw std::runtime_error("Already spinning a node, can't start a new node spin");
@@ -101,9 +97,11 @@ public:
 
   void stop_spinning()
   {
-    exit_from_node_spinner_ = true;
-    if (node_spinner_future_.valid()) {
-      node_spinner_future_.wait();
+    if (exec_ != nullptr) {
+      exec_->cancel();
+      if (spin_thread_.joinable()) {
+        spin_thread_.join();
+      }
     }
   }
 
@@ -153,8 +151,8 @@ public:
 
   // relative path to the root of the bag file.
   fs::path root_bag_path_;
-  std::future<void> node_spinner_future_;
-  std::atomic_bool exit_from_node_spinner_{false};
+  std::unique_ptr<rclcpp::executors::SingleThreadedExecutor> exec_{nullptr};
+  std::thread spin_thread_;
 };
 
 TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_only) {
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -86,9 +86,10 @@ public:
       [node, this]() -> void {
         rclcpp::executors::SingleThreadedExecutor exec;
         exec.add_node(node);
-        while (!exit_from_node_spinner_) {
-          exec.spin_some();
+        while (rclcpp::ok() && !exit_from_node_spinner_) {
+          exec.spin_some(std::chrono::milliseconds(100));
         }
+        exec.remove_node(node);
       });
   }
 
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -91,6 +91,16 @@ public:
         [this]() {
           exec_->spin();
         });
+      // Wait for the executor to start spinning in the newly spawned thread to avoid race condition
+      // with exec_->cancel()
+      using clock = std::chrono::steady_clock;
+      auto start = clock::now();
+      while (!exec_->is_spinning() && (clock::now() - start) < std::chrono::seconds(5)) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(20));
+      }
+      if (!exec_->is_spinning()) {
+        throw std::runtime_error("Failed to start spinning node");
+      }
     } else {
       throw std::runtime_error("Already spinning a node, can't start a new node spin");
     }
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
@@ -103,6 +103,7 @@ public:
       if (spin_thread_.joinable()) {
         spin_thread_.join();
       }
+      exec_ = nullptr;
     }
   }
 
@@ -201,6 +202,7 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_services_only
   recorder->record();
 
   start_async_spin(recorder);
+  auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
 
   ASSERT_TRUE(service_client_manager->wait_for_service_to_be_ready());
   ASSERT_TRUE(wait_for_subscriptions(*recorder, {"/test_service/_service_event"}));
@@ -275,6 +277,7 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_and_se
   recorder->record();
 
   start_async_spin(recorder);
+  auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
 
   ASSERT_TRUE(
     wait_for_subscriptions(
--- a/rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
+++ b/rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
@@ -73,6 +73,7 @@ public:
       if (spin_thread_.joinable()) {
         spin_thread_.join();
       }
+      exec_ = nullptr;
     }
   }
 
--- a/rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
+++ b/rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <gmock/gmock.h>
-
+#include <atomic>
+#include <chrono>
 #include <future>
 #include <memory>
 #include <string>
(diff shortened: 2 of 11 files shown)


clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/77bebfb56ccf2c9d926ce6c8343b0773/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14483
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21759/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16029/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1260/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22456/)

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -126,10 +137,21 @@ TEST_F(RecordIntegrationTestFixture, can_record_again_after_stop)
...
+      return recorder->subscriptions().size() == 1;
+    },
+    std::chrono::seconds(5));
+  ASSERT_TRUE(discovery_ret);
+
   auto & writer = recorder->get_writer_handle();
   auto & mock_writer = dynamic_cast<MockSequentialWriter &>(writer.get_implementation_handle());
 
-  start_async_spin(recorder);
   ASSERT_TRUE(pub_manager.wait_for_matched(string_topic.c_str()));
```
fujitatomoya, line 144: Do we really need to call `wait_until_condition`? here it waits for the one subscription matched to this publisher, that means publisher confirms the corresponding subscriber to send the message. (this goes to some other changes in this PR.)

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
```diff
@@ -104,23 +116,34 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_services_a
...
+  // Wait until recorder discovery is complete, otherwise messages might be missed.
+  // The currently expected topics:
+  // /test_service_1/_service_event
+  // /test_service_2/_service_event
+  auto discovery_ret = rosbag2_test_common::wait_until_condition(
+    [&recorder]() {
+      return recorder->subscriptions().size() == 2;
+    },
+    std::chrono::seconds(5));
+  ASSERT_TRUE(discovery_ret);
```
fujitatomoya: i believe that waiting on this service event subscription discovery makes sense. Although services are discovered and matched, maybe service event subscriptions are not yet available.

fujitatomoya (commented):
overall lgtm, but one comment.

rosbag2_transport/test/rosbag2_transport/test_record.cpp
clalancette: Ah, that's a good point.  For the tests that are strictly using the `pub_manager`, we actually don't need to wait around for the recorder subscription size to be matched; we are getting a similar check from the pub_manager already.  I'll go back and remove these unnecessary ones.

clalancette: And now fixed in 02d5bba

fujitatomoya (approved):
lgtm, @MichaelOrlov can you have a look too? (i do not have maintainer access on rosbag2)

MichaelOrlov:
I will take a look. Anyway, hold off from the merging this for a while. We are trying to make a relase on rolling https://github.com/ros2/rosbag2/pull/1791

rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
```diff
@@ -57,12 +53,20 @@ class RecordIntegrationTestFixture : public Rosbag2TransportTestFixture
...
-      std::launch::async, [node]() -> void {rclcpp::spin(node);});
+      std::launch::async,
+      [node, this]() -> void {
+        rclcpp::executors::SingleThreadedExecutor exec;
+        exec.add_node(node);
+        while (rclcpp::ok() && !done_) {
+          exec.spin_some(std::chrono::milliseconds(100));
+        }
+        exec.remove_node(node);
+      });
```
MichaelOrlov: I would add a sanity check to make sure that we are not spinning already some node.
```
    if (!done_.exchange(false)) {
      future_ = std::async(
        std::launch::async,
        [node, this]() -> void {
          exec.add_node(node);
          while (rclcpp::ok() && !done_) {
            exec.spin_some(std::chrono::milliseconds(100));
          }
          exec.remove_node(node);
        });
    } else {
      throw std::runtime_error("Can't start a new node spin. Already spinning some node.");
    }
```

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
```diff
@@ -86,9 +86,10 @@ class Rosbag2CPPGetServiceInfoTest
       [node, this]() -> void {
         rclcpp::executors::SingleThreadedExecutor exec;
         exec.add_node(node);
-        while (!exit_from_node_spinner_) {
-          exec.spin_some();
+        while (rclcpp::ok() && !exit_from_node_spinner_) {
```
MichaelOrlov: I would add a sanity check to make sure that we are not spinning already some node.
```
if (!exit_from_node_spinner_.exchange(false)) {
      node_spinner_future_ = std::async(
        std::launch::async,
        [node, this]() -> void {
          exec.add_node(node);
          while (rclcpp::ok() && !exit_from_node_spinner_) {
            exec.spin_some(std::chrono::milliseconds(100));
          }
          exec.remove_node(node);
        });
    } else {
      throw std::runtime_error("Can't start a new node spin. Already spinning some node.");
    }
```

MichaelOrlov (changes requested):
@clalancette Thanks for trying to make rosbag2 tests less flaky.
I have a concern about adding Wait for topics to be discovered commit and would like to request to revert it along with the https://github.com/ros2/rosbag2/pull/1796/commits/02d5bba75c486d7af5ddc82d1f656f1fd2f91b03

The reason for removal request is because it is really redundant IMO and not robust. 
- Redundant because in all places we already have either `ASSERT_TRUE(pub_manager.wait_for_matched(expecting_topic_name);` or `ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());`. Which is making sure that recorder have already been subscribed to the topics on which we are going to publish.
- Not robust because in those newly added wait condition - there is no check for the topic names on which recorder subscribed. There is only check for the number of subscriptions. And had assumption that `// /rosout` and `// /events/write_split` topics will always be present and counted, which is not necessarily may be true for all CI configurations and underlying middleware layers.

If aforementioned `ASSERT_TRUE(pub_manager.wait_for_matched(expecting_topic_name);` and `ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());` checks are not enough we have a better way to do a wait for the recorder discovery. In some unit tests we use `MockRecorder` with extended utility methods to facilitate waiting functionality.
For instance:
```
  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(string_topic));
  ASSERT_FALSE(recorder->topic_available_for_recording(string_topic));
```
Please see implementation details in the 
https://github.com/ros2/rosbag2/blob/88c51a133a9a9aa3ef65a851f4d2aed448803fa1/rosbag2_transport/test/rosbag2_transport/mock_recorder.hpp#L41-L73

clalancette:
> * Redundant because in all places we already have either `ASSERT_TRUE(pub_manager.wait_for_matched(expecting_topic_name);` or `ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());`. Which is making sure that recorder have already been subscribed to the topics on which we are going to publish.

The problem is that this is not always true.  There are at least two situations this does not cover: for the event topic associated with services (`/service_name/_events`), and for the event topic associated with the recorder splitting functionality (`/event/split_write`).

`pub_manager_wait_for_matched(expecting_topic_name)` will only wait for the named topic, which is a topic that the pub_manager is managing.  That does not apply to the `/service_name/_events` topic that services use for events.

Additionally, `client_manager->wait_for_service_to_be_ready()` only waits for the *service* topics to be ready, not the event topic.  So again, it does not apply to the `/service_name/_events` topic that services use for events.

That's why adding in the additional waits for recorder subscriptions makes the tests more reliable, since it handles both of these situations.

> If aforementioned `ASSERT_TRUE(pub_manager.wait_for_matched(expecting_topic_name);` and `ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());` checks are not enough we have a better way to do a wait for the recorder discovery. In some unit tests we use `MockRecorder` with extended utility methods to facilitate waiting functionality.

But fair enough, I'll take a look at switching to this.

clalancette:
All right, see 11bb864, which uses the MockRecorder methods instead.

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/08cc831129805c683801a5ebf1f8b33a/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14511
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21786/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16058/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1289/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22485/)

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
```diff
@@ -60,17 +61,22 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
   ASSERT_TRUE(pub_manager.wait_for_matched(array_topic.c_str()));
   ASSERT_TRUE(pub_manager.wait_for_matched(string_topic.c_str()));
 
+  // At this point, we expect that the topics /string_topic, /array_topic, and /events/write_split
+  // are available to be recorded.  However, wait_for_matched() only checks for /string_topic
+  // and /array_topic, so ask the recorder to make sure it has successfully subscribed to all.
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/events/write_split"));
+
```
MichaelOrlov: @clalancette What a point to wait for the `"/events/write_split"` on rosbag2_recorder side?
The `"/events/write_split"` is an internal `rosbag2_recorder` service event that we are not going to use in this test scenario.

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
```diff
@@ -107,20 +113,27 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_services_a
   ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());
   ASSERT_TRUE(client_manager_2->wait_for_service_to_be_ready());
 
+  // At this point, we expect that the services /test_service_1 and /test_service_2, along with
+  // the event topics /test_service_1/_service_event and /test_service_2/_service_event are
+  // available to be recorded.  However, wait_for_service_to_be_ready() only checks the services,
+  // not the event topics, so ask the recorder to make sure it has successfully subscribed to all.
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/test_service_1/_service_event"));
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/test_service_2/_service_event"));
```
MichaelOrlov, line 118: I don't mind to add wait for service event topics here. However, I am curious if we ever got a failure in this test.

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
```diff
@@ -150,6 +163,14 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_topic_and_service_a
 
   ASSERT_TRUE(client_manager_1->wait_for_service_to_be_ready());
 
+  // At this point, we expect that the service /test_service_1, along with the topic /string_topic,
+  // along with the event topic /test_service_1, along with the split topic /events/write_split are
+  // available to be recorded.  However, wait_for_matched() and wait_for_service_to_be_ready() only
+  // check on the service and the topic, not the event or the split topic, so ask the recorder to
+  // make sure it has successfully subscribed to all.
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/test_service/_service_event"));
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/events/write_split"));
```
MichaelOrlov: IMO wait for `"/events/write_split"` is redundant here. We are not going to use it in this test.

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
clalancette: Yes, we have been sporadically seeing this test fail.  Locally, under load, I can make this test fail about 1% of the time.  The reason it fails is because we send two requests below, and each request is expected to make 2 callbacks happen on the `/test_service_<n>/_service_event` topic.  If we send the requests before the topics were matched in the recorder, then line 137 below fails because we never get the callbacks.

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
clalancette: Yeah, I guess that's true.  I'll remove this one.

rosbag2_transport/test/rosbag2_transport/test_record_regex.cpp
```diff
@@ -302,6 +302,14 @@ TEST_F(RecordIntegrationTestFixture, regex_and_exclude_regex_service_recording)
   ASSERT_TRUE(service_manager_b1->wait_for_service_to_be_ready());
   ASSERT_TRUE(service_manager_b2->wait_for_service_to_be_ready());
 
+  // At this point, we expect that the services /still_nice_service and /awesome_nice_service,
+  // along with the event topics /still_nice_service/_service_event
+  // and /awesome_nice_service/_service_event are available to be recorded.  However,
+  // wait_for_service_to_be_ready() only checks the services, not the event topics, so ask the
+  // recorder to make sure it has successfully subscribed to all.
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/still_nice_service/_service_event"));
+  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered("/awesome_nice_service/_service_event"));
```
MichaelOrlov: Nitpick:
```suggestion
  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(v1 + "/_service_event"));
  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(v2 + "/_service_event"));
```

rosbag2_transport/test/rosbag2_transport/test_record_regex.cpp
```diff
@@ -375,6 +383,14 @@ TEST_F(RecordIntegrationTestFixture, regex_and_exclude_service_service_recording
...
```
MichaelOrlov: Nitpick:
```suggestion
  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(v1 + "/_service_event"));
  ASSERT_TRUE(recorder->wait_for_topic_to_be_discovered(v2 + "/_service_event"));
```

MichaelOrlov (commented):
IMO, waiting for "/events/write_split" is redundant since it is an internal service in the rosbag2_recorder and it is not participating in the test scenarios where it was added.

Also found a few nitpicks.

rosbag2_transport/test/rosbag2_transport/test_record_all.cpp
clalancette: Removed in 83b0b6b

MichaelOrlov (approved):
LGTM now. 
Thanks!

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/475d8bf4bfedcf19c2136589ee4451f9/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14514
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21790/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16061/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1292/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22488/)

MichaelOrlov:
I'm curious what a "surprise" failure with rmw_connextdds on CI is.

clalancette:
> I'm curious what a "surprise" failure with rmw_connextdds on CI is.

Yeah, I don't know what is going on with that.  The only change to that test is to wait for the clock topic to show up.  I'll need to look deeper into it, though, since that seems to be a new failure with this PR.

clalancette:
All right, I think I figured it out.  I think the problem is related to the switch to `exec.spin_some()`, and the ongoing bug in rclcpp: https://github.com/ros2/rclcpp/issues/2589

However, it turns out that we actually don't need to use `spin_some` here; using just `spin`, along with a `cancel` later on, is good enough.  So commit eb8db9e does exactly that, and fixes the problems in my local testing.  I'll run another CI here to confirm.

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/72c7e3cff95fa4a9f17085364bd78781/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14517
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21793/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16064/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1295/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22491/)

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/0e69722137e215ba0dee2c7dc96e1a6c/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14518
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21794/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16065/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1296/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22492/)

clalancette:
The only failures here are on Windows, and that has nothing to do with this PR (those will be fixed by https://github.com/ros2/common_interfaces/pull/256).  So I think this is ready for final review now.

MichaelOrlov (commented):
@clalancette Curious what was behind decision to move explict call for the `stop_spinning()` to the `TearDown()` method?

I would prefer to have explicit call to the `stop_spinning()` in the tests where it is needed in pair with `start_spin(node)` or `start_async_spin(node)` with
```cpp
 auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
```
to have a clean and concise logic in tests.

In some tests we have some checks after explictly calling `stop_spinning()` which could rely on the assumption that some handles (node and files) will be free and nothing else will arrive during those extra checks or cleanups.

IMO adding `stop_spinning()` to the `TearDown()` method is a last resort and a sort of hacky way to handle it and could potentially cause another flakiness in tests in the future.

clalancette:
> @clalancette Curious what was behind decision to move explict call for the `stop_spinning()` to the `TearDown()` method?

So the first thing to realize is that because we are putting the spin call in a thread, we *have* to call stop_spinning() (which calls thread.join()), otherwise we get an exception from the `std::thread destructor`.

With that in mind, I could have used `rcpputils::make_scope_exit` to stop the spinning.  However, by default that is actually no different than calling it in `TearDown`, since we wouldn't call that until the end of the function anyway.

I have a prototype locally where I did exactly what you said and added in new scopes, with the `rcpputils::make_scope_exit`, and that works fine.  But it also ends up in a far larger change, because of those additional scopes and the reindentation of a lot of code.  Let me know if you still want me to do that, and I can make the change.

clalancette:
As something of a compromise, 404c78a adds in `rcpputils::make_scope_exit` calls *without* changing the scopes.  This isn't too different from putting it in TearDown (which I've left as a backup), but it would allow us to change scopes later.  Let me know what you think.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
```diff
@@ -222,8 +228,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_services_only
   EXPECT_TRUE(ret) << "Failed to capture " << expected_messages << " expected messages in time";
 
   recorder->stop();
-  stop_spinning();
-  cleanup_process_handle.cancel();
```
MichaelOrlov, line 226: Suggestion to return deleted explict call to `stop_spinning()` and `cleanup_process_handle.cancel();`

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
```diff
@@ -309,8 +312,6 @@ TEST_P(Rosbag2CPPGetServiceInfoTest, get_service_info_for_bag_with_topics_and_se
...
```
MichaelOrlov, line 313: Suggestion to return deleted explict call to `stop_spinning()` and `cleanup_process_handle.cancel();`

rosbag2_transport/test/rosbag2_transport/test_record.cpp
```diff
@@ -64,16 +65,15 @@ TEST_F(RecordIntegrationTestFixture, published_messages_from_multiple_topics_are
...
     [ =, &mock_writer]() {
       return mock_writer.get_messages().size() >= expected_messages;
-    });
+    },
+    std::chrono::seconds(5));
   auto recorded_messages = mock_writer.get_messages();
   EXPECT_TRUE(ret) << "failed to capture expected messages in time" <<
     "recorded messages = " << recorded_messages.size();
   EXPECT_THAT(recorded_messages, SizeIs(expected_messages));
-  stop_spinning();
```
MichaelOrlov, line 76: Suggestion to return deleted explict call to `stop_spinning()` and `cleanup_process_handle.cancel();`

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
```diff
@@ -113,17 +114,17 @@ TEST_F(RecordIntegrationTestFixture, record_all_with_sim_time)
...
     [ =, &mock_writer]() {
       return mock_writer.get_messages().size() >= expected_messages;
-    });
+    },
+    std::chrono::seconds(5));
   auto recorded_messages = mock_writer.get_messages();
-  EXPECT_TRUE(ret) << "failed to capture expected messages in time. " <<
+  ASSERT_TRUE(ret) << "failed to capture expected messages in time. " <<
     "recorded messages = " << recorded_messages.size();
-  stop_spinning();
```
MichaelOrlov, line 124: Suggestion to return deleted explict call to `stop_spinning()` and `cleanup_process_handle.cancel();`

MichaelOrlov (commented):
@clalancette Thanks for adding a safeguard catcher `auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});`
However, my main concern from the previous comment was that the explicit call to the `stop_spinning()` was deleted in some tests. Could you please add them back? 
I added comments in places where it was deleted.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
clalancette: I ended up doing something slightly different, because I think the pattern of calling `stop_spinning` and `cancel` on the scope exit isn't great.  Just for the few instances you pointed out here, I ended up making scopes.  That way, when we leave the scope, `cleanup_process_handle` will always run.  So effectively [e63f8c7](https://github.com/ros2/rosbag2/pull/1796/commits/e63f8c7b27f139e011f4452f9a9fbd795be0e0a2) does what you ask here, just in a slight different way.  Let me know what you think.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
clalancette: Same comment as https://github.com/ros2/rosbag2/pull/1796#discussion_r1752020122

rosbag2_transport/test/rosbag2_transport/test_record.cpp
clalancette: Same comment as https://github.com/ros2/rosbag2/pull/1796#discussion_r1752020122

rosbag2_transport/test/rosbag2_transport/test_record_all_use_sim_time.cpp
clalancette: Same comment as https://github.com/ros2/rosbag2/pull/1796#discussion_r1752020122

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
MichaelOrlov: Yeah, that is even better I think.

MichaelOrlov (approved):
LGTM now.

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/1986838f6b0c6a14c9b1af14a8340640/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14549
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21826/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16096/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1328/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22533/)

rosbag2_transport/test/rosbag2_transport/test_keyboard_controls.cpp
```diff
@@ -203,13 +203,12 @@ TEST_F(RecordIntegrationTestFixture, test_keyboard_controls)
...
+  start_async_spin(recorder);
+  auto cleanup_process_handle = rcpputils::make_scope_exit([&]() {stop_spinning();});
 
   EXPECT_THAT(recorder->is_paused(), true);
   keyboard_handler->simulate_key_press(rosbag2_transport::Recorder::kPauseResumeToggleKey);
   EXPECT_THAT(recorder->is_paused(), false);
   keyboard_handler->simulate_key_press(rosbag2_transport::Recorder::kPauseResumeToggleKey);
   EXPECT_THAT(recorder->is_paused(), true);
-
-  this->stop_spinning();
```
MichaelOrlov, line 214: @clalancette I don't know how, but it seems we got a regression in test_keyboard_controls.cpp file. It fails on RPR job and on Windows CI. I've never seen such a failure before.
Thoughts?

clalancette: Yeah, I saw that.  Both in Rpr and in the Windows CI, those tests are timing out.  I can only imagine this is happening because we are never successfully `join()` on the spinning thread, but I can't reproduce locally at all (either on Linux or on Windows).  I'm going to have to poke at it more.

clalancette:

rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
```diff
@@ -50,21 +47,33 @@ class RecordIntegrationTestFixture : public Rosbag2TransportTestFixture
 
   void TearDown() override
   {
+    stop_spinning();
```
MichaelOrlov, line 50: @clalancette To avoid potential freezes in tests, I think it makes sense to call `stop_spiining()` after the `rclcpp::shutdown()` in this destructor.

MichaelOrlov:
@clalancette I was able to reproduce freeze locally on Ubuntu by increasing discovery wait interval up to the 500mS 
```
rosbag2_transport::RecordOptions record_options =
  {true, false, false, {}, {}, {}, {}, {}, {}, "rmw_format", 500ms};
```
and repeatedly run only those `RecordIntegrationTestFixture, test_keyboard_controls` test.

I see in the debugger that freeze occurs in our modified `stop_spinning()` right on the  **spin_thread_.join();**, exactly as you were suspecting.

It seems the problem is in the underlying SingleThreadedExecutor spin() and cancel() routines. It looks like cancel doen't really work as expected and we are not exiting from the spin().

rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
```diff
@@ -50,21 +47,33 @@ class RecordIntegrationTestFixture : public Rosbag2TransportTestFixture
...
+    if (exec_ == nullptr) {
+      exec_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
+      exec_->add_node(node);
+      spin_thread_ = std::thread(
+        [this]() {
+          exec_->spin();
+        });
+    } else {
+      throw std::runtime_error("Already spinning a node, can't start a new node spin");
+    }
```
MichaelOrlov:
```suggestion
    if (exec_ == nullptr) {
      exec_ = std::make_unique<rclcpp::executors::SingleThreadedExecutor>();
      exec_->add_node(node);
      spin_thread_ = std::thread(
        [this]() {
          exec_->spin();
        });
      // Wait for the executor to start spinning in the newly spawned thread to avoid race condition
      // with exec_->cancel()
      using clock = std::chrono::steady_clock;
      auto start = clock::now();
      while (!exec_->is_spinning() && (clock::now() - start) < std::chrono::seconds(5)) {
        std::this_thread::sleep_for(std::chrono::milliseconds(20));
      }
      if (!exec_->is_spinning()) {
        throw std::runtime_error("Failed to start spinning node");
      }
    } else {
      throw std::runtime_error("Already spinning a node, can't start a new node spin");
    }
```

MichaelOrlov (changes requested):
@clalancette Uff, it was a nasty bug, but I've nailed it down. I've spent a day hunting for it.
I was even start thinking that we have a memory corruption or a race condition in the executor. But not. Everything is more simple.
It is just turns out that we spawn a new `spin_thread_` for `exec_->spin();`. 
However, in some rare cases a context switch happened and we ran throw all test scenario and even call `exec_->cancel();` from the `stop_spinning()`, but context haven't yet got back to the `spin_thread_` and `exec_->spin()`.
The root cause of the freeze was that `exec_->spin()` was executing after we already called `exec_->cancel();`.

I added a comment with proposal to wait untill executor will start spinning before we will get back from the `start_async_spin(node)`. 
I was able to robustly reproduce this issue and verify fix on my local setup.

clalancette:
Thanks for looking into it.  I'm going to poke more at this, because `cancel` should work.

rosbag2_transport/test/rosbag2_transport/record_integration_fixture.hpp
clalancette: Oh!  Fantastic.  Thank you so much for looking into it.  I'm going to test this out.

clalancette:
Pulls: ros2/rosbag2#1796
Gist: https://gist.githubusercontent.com/clalancette/5271579611710ffed73686f516a6ad9e/raw/cb4f2fb2fb0fac274fbef12b77b13699d0b0e232/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_tests rosbag2_transport
TEST args:  --packages-above rosbag2_test_common rosbag2_tests rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14555
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16102/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22540/)

clalancette:
Yep, this seems much happier now.  Thanks again @MichaelOrlov , just need approval again then we can merge this in.

MichaelOrlov (approved):
LGTM

MichaelOrlov:

PR #1664, opened 2024-05-21
Add support for "all" and "exclude" in RecordOptions YAML decoder

MichaelOrlov:
- Fixes https://github.com/ros2/rosbag2/issues/1660
-  This PR adds support for old-style options "all" and "exclude" in the `RecordOptions` YAML parser for backward compatibility with old versions of the config files for the "ros2 bag convert" CLI.

PR #1799, opened 2024-09-04
update documentation for rosbag2 composable player

berndpfrommer:
all: true should read all_topics: true

This was brought to my attention on the robotics stack exchange [here](https://robotics.stackexchange.com/questions/112800/ros2-bag-dropping-some-images-after-recording-for-around-20-seconds/112823?noredirect=1#comment48807_112823)

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ recorder:
   ros__parameters:
     use_sim_time: false
     record:
-      all: true
+      all_topics: true
       is_discovery_disabled: false
       topic_polling_interval:
         sec: 0


fujitatomoya (approved):
lgtm.

https://github.com/ros2/rosbag2/blob/efd8d39843ce47500fde900178755b7550a4be73/rosbag2_transport/src/rosbag2_transport/config_options_from_node_params.cpp#L246

MichaelOrlov (commented):
@berndpfrommer Could you please elaborate why it is should be `all_topics` instead of `all`?

Inspight the fact that we added recently a new CLI options like `--all-topics` and `--all-service` we still have `--all` CLI option which will "Record all topics and services (Exclude hidden topic)"

berndpfrommer:
I had a superficial look at the README, saw only "all_topics" mentioned, and uncritically assumed the failure report on stack exchange was correct.
I wonder if the user is on the latest update. Do I understand correctly that ``all`` will only work if [this commit](https://github.com/ros2/rosbag2/commit/e01e1efe57c4f226eed21e2eefb0b459fd3bd0b8) is being used? That one is fairly recent...

MichaelOrlov:
Ahh you are right, I've already forgotten that we had a compatibility issue in yaml parser and that I've fixed in the https://github.com/ros2/rosbag2/pull/1664.

Anyway, in the current state, everything should be up to date, even on a new Jazzy release and there is no need to update documentation.

berndpfrommer:
OK, closing this PR then.

PR #1804, opened 2024-09-11
Add "--sort" CLI option to the "ros2 bag info" command

### LINKED ISSUE #1797: Sort topics by name when executing `ros2 bag info`
## Description

Currently, when executing ros2 bag info, topics are not ordered alphabetically.
While this is manageable for small bagfiles, we encounter files with hundreds of topics.
Even though we invested time in grouping topics into meaningful namespaces, topics within the same namespace are not gathered together, making the output hard to analyze efficiently.

## Related Issues
It might also be convenient to allow users to sort the output by other criteria, such as topic type or message count.
However, simple alphabetical ordering could become the default option.

[This issue](https://github.com/ros2/rosbag2/issues/51) also seems relevant as moving the console output `ros2 bag info` to the  Python layer would necessarily influence any work on topics ordering.

## Completion Criteria

- `ros2 bag info` outputs topics sorted alphabetically by their names.

Sanoronas:
This PR adds the "--sort" CLI option to the "ros2 bag info" command. With "--sort" CLI option user will be able to sort topics and services by name, topic type or number of recorded messages.

* generate sorted indices for topics from multiple different options (topic name, type, count or serialization format)
* defaults to sorting by topic name
* use sorted indices rather than using strict increasing indices for printing out the topics
* potential use of different sorting options through CLI by passing sort_method argument to format_topics_with_type function

- Closes #1797


rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
```diff
@@ -115,7 +115,8 @@ void format_topics_with_type(
   const std::unordered_map<std::string, uint64_t> & messages_size,
   bool verbose,
   std::stringstream & info_stream,
-  int indentation_spaces)
+  int indentation_spaces,
+  std::string sort_method = "name")
```
fujitatomoya: as you suggested, this option would be nice to configure via CLI so that user can choose the sorting method as they like. and test would be ideal

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
```diff
@@ -139,13 +140,37 @@ void format_topics_with_type(
...
+    [&topics, sort_method](size_t i1, size_t i2) {
+      if (sort_method == "type")
+      {
+        return topics[i1].topic_metadata.type < topics[i2].topic_metadata.type;
+      }
+      if (sort_method == "count")
+      {
+        return topics[i1].message_count < topics[i2].message_count;
+      }
+      if (sort_method == "serialization_format")
```
fujitatomoya: I am not sure about this use case. most likely the serialization format could be the same in the single bag file which is `cdr`?

fujitatomoya (commented):
it would be nice if `format_service_with_type` has the same sorting interface aligned with topics.

this PR is failing to build with the following error and DCO is missing. could you address them?

```bash
07:59:08 /tmp/ws/src/rosbag2/rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp:144:8: error: ‘iota’ is not a member of ‘std’
07:59:08   144 |   std::iota(sorted_idx.begin(), sorted_idx.end(), 0);
07:59:08       |        ^~~~
07:59:08 gmake[2]: *** [CMakeFiles/_storage.dir/build.make:90: CMakeFiles/_storage.dir/src/rosbag2_py/format_bag_metadata.cpp.o] Error 1
07:59:08 gmake[1]: *** [CMakeFiles/Makefile2:284: CMakeFiles/_storage.dir/all] Error 2
07:59:08 gmake: *** [Makefile:146: all] Error 2
```

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
Sanoronas: I was also not sure, if there even is the possibility to have different serialization formats in a bag file. As it was available for output I included it here as option, but can drop it if it is not neccessary

MichaelOrlov (changes requested):
@Sanoronas Thank you for your contribution.
I agree with @fujitatomoya that sorting by the serialization format is unlikely to be useful. Because we are not supporting different serialization formats during recording by design.  
I can only envision a possibility to have a different serilization formats if someone will convert the already recorded bag or record with modifyied rosbag2 recorder or using some different tool for recording.
Therefore I think it makes sense to have options for sorting:
1. By topic or service **name**.
2. By topic or service **type**
3. By topic or service **messages count**

Could you please use C++ enum values for the sorting method?
Also, It will be appreciated if you would add the same sorting algorithm for the services in the `format_service_with_type(..)` function located in the same file.

Sanoronas:
Thanks for the feedback @MichaelOrlov and @fujitatomoya

I added --sort option for CLI, handling sorting of service topics and moved the sorting options to enum.

ros2bag/ros2bag/verb/info.py
```diff
@@ -30,16 +30,21 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
         if args.topic_name and args.verbose:
             print("Warning! You have set both the '-t' and '-v' parameters. The '-t' parameter "
                   'will be ignored.')
         m = Info().read_metadata(args.bag_path, args.storage)
         if args.verbose:
-            Info().print_output_verbose(args.bag_path, m)
+            Info().print_output_verbose(args.bag_path, m, args.sort)
         else:
             if args.topic_name:
                 Info().print_output_topic_name_only(m)
```
MichaelOrlov: For the sake of consistency and more clarity, I would add the `args.sort` parameter to the `print_output_topic_name_only(m)` and print a warning inside those function if `args.sort != name`.

With such approach, it will be clear from the caller level that `print_output_topic_name_only(..)` also do sorting inside.
Currently, it is not clear and need to go inside the function implementation and check if it is really doing any sort.

rosbag2_py/src/rosbag2_py/_info.cpp
```diff
@@ -42,15 +50,29 @@ class Info
     return info_->read_metadata(uri, storage_id);
   }
 
-  void print_output(const rosbag2_storage::BagMetadata & metadata_info)
+  void print_output(
+    const rosbag2_storage::BagMetadata & metadata_info, std::string sorting_method)
```
MichaelOrlov: Please use const reference where possible for input parameters.
```suggestion
  void print_output(
    const rosbag2_storage::BagMetadata & metadata_info, const std::string & sorting_method)
```

rosbag2_py/src/rosbag2_py/_info.cpp
```diff
@@ -42,15 +50,29 @@ class Info
...
-    for (const auto & topic_info : metadata_info.topics_with_message_count) {
+    std::vector<size_t> sorted_idx(metadata_info.topics_with_message_count.size());
+    std::iota(sorted_idx.begin(), sorted_idx.end(), 0);
+    std::sort(
+      sorted_idx.begin(),
+      sorted_idx.end(),
+      [&metadata_info](size_t i1, size_t i2) {
+        std::string topic_name_1 = metadata_info.topics_with_message_count[i1].topic_metadata.name;
+        std::string topic_name_2 = metadata_info.topics_with_message_count[i2].topic_metadata.name;
+        return topic_name_1 < topic_name_2;
```
MichaelOrlov: Using `const auto &` is more efficient and convenient. Or at least `const string &`
```suggestion
      [&metadata_info](size_t i1, size_t i2) {
        const auto & topic_name_1 = metadata_info.topics_with_message_count[i1].topic_metadata.name;
        const auto & topic_name_2 = metadata_info.topics_with_message_count[i2].topic_metadata.name;
        return topic_name_1 < topic_name_2;
```

rosbag2_py/src/rosbag2_py/_info.cpp
```diff
@@ -42,15 +50,29 @@ class Info
...
+      }
+    );
+    for (auto idx : sorted_idx) {
+      auto topic_info = metadata_info.topics_with_message_count[idx];
```
MichaelOrlov:
```suggestion
      const auto & topic_info = metadata_info.topics_with_message_count[idx];
```

rosbag2_py/src/rosbag2_py/_info.cpp
```diff
@@ -61,7 +83,9 @@ class Info
   }
 
   void print_output_verbose(
-    const std::string & uri, const rosbag2_storage::BagMetadata & metadata_info)
+    const std::string & uri,
+    const rosbag2_storage::BagMetadata & metadata_info,
+    std::string sorting_method)
```
MichaelOrlov:
```suggestion
    const std::string & uri,
    const rosbag2_storage::BagMetadata & metadata_info,
    const std::string & sorting_method)
```

rosbag2_py/src/rosbag2_py/format_bag_metadata.hpp
```diff
@@ -23,11 +23,22 @@
...
+/**
+ * \brief Available sorting methods for info output.
+ */
+enum class SortingMethod
+{
+  NAME,
+  TYPE,
+  COUNT,
+};
+
```
MichaelOrlov: Please rename it to the `InfoSortingMethod` and move it to a separate header file. e.g. `InfoSortingMethod.hpp`
Also please use C++ Doxygen comments style. e.g. 
```cpp
/// \brief Available sorting methods for info output.
```

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
```diff
@@ -139,13 +142,29 @@ void format_topics_with_type(
...
+    sorted_idx.begin(),
+    sorted_idx.end(),
+    [&topics, sort_method](size_t i1, size_t i2) {
+      if (sort_method == rosbag2_py::SortingMethod::TYPE) {
+        return topics[i1].topic_metadata.type < topics[i2].topic_metadata.type;
+      }
+      if (sort_method == rosbag2_py::SortingMethod::COUNT) {
+        return topics[i1].message_count < topics[i2].message_count;
+      }
+      return topics[i1].topic_metadata.name < topics[i2].topic_metadata.name;
```
MichaelOrlov: Please use switch-case instead of if-else with returns.
In case of `default:` "unknown value" in switch case, please throw std::runtime_error("switch is not exhaustive");

rosbag2_py/src/rosbag2_py/format_bag_metadata.cpp
```diff
@@ -238,11 +258,27 @@ void format_service_with_type(
...
+    sorted_idx.end(),
+    [&services, sort_method](size_t i1, size_t i2) {
+      if (sort_method == rosbag2_py::SortingMethod::TYPE) {
+        return services[i1]->service_metadata.type < services[i2]->service_metadata.type;
+      }
+      if (sort_method == rosbag2_py::SortingMethod::COUNT) {
+        return services[i1]->event_message_count < services[i2]->event_message_count;
+      }
+      return services[i1]->service_metadata.name < services[i2]->service_metadata.name;
+    }
```
MichaelOrlov: Please use switch-case instead of if-else with returns.
In case of `default:` "unknown value" in switch case, please throw std::runtime_error("switch is not exhaustive");

MichaelOrlov (changes requested):
@Sanoronas, Thank you for addressing comments from the previous review and for adding a CLI option for the sorting method.

I've put some comments and requests for code changes. Also, in addition to that, could you please add sorting to the `format_service_info(..)` function from the `format_service_info.cpp`?
https://github.com/ros2/rosbag2/blob/cd7bd63696604973e23c739afa6387556f3e7781/rosbag2_py/src/rosbag2_py/format_service_info.cpp#L80-L86
The same way you did for services in another place. 
For sorting by "count" you can use the sum of the `si->request_count` and `si->response_count`.
To add sorting for the `format_service_info.cpp`, the declaration of the `enum class SortingMethod` will need to be moved to a separate header file, which makes sense anyway.
Also, I would like to ask you to create the following `info_sorting_method_from_string(str)` helper function
```cpp
SortingMethod info_sorting_method_from_string(std::string str)
{
  static std::unordered_map<std::string, SortingMethod> sorting_method_map = {
    {"name", SortingMethod::NAME},
    {"type", SortingMethod::TYPE},
    {"count", SortingMethod::COUNT}};
  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
  auto find_result = sorting_method_map.find("str");
  if (find_result == sorting_method_map.end()) {
    throw std::runtime_error("Enum value match for \"" + str + "\" string is not found.");
  }
  return find_result->second;
}
```
Use it instead of `rosbag2_py::stringToSortingMethod[sorting_method];` whenever conversion from string to enum is needed. 
You can put a declaration of those `info_sorting_method_from_string(str)` function to the same header file with the enum class. However, the definition should be in the cpp file to not violate the C++ ODR rule. Will need to create one more `InfoSortingMethod.cpp` file for that.

And the last but not least request is to create a few unit tests to cover all these changes. 
I know, creating integration unit tests from scratch sounds like a tedious and non-trivial task. 
But we have got covered you :wink: 
We already have integration tests for the `ros2 bag info`. However, they are in a different package. You can find them in the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.cpp.
All the infrastructure and sample bag files for tests already exist. Perhaps it will not be so difficult to write a few new tests similar to the existing ones.

Sanoronas:
@MichaelOrlov , thanks for your input!

I addressed some of your points in the recent commit (missing switch-case and tests).
However, I was unable to regenerate stub-files and cannot track down the reason for this right now. Any help on that regard would be helpful.

MichaelOrlov:
@Sanoronas Have you tried instruction from the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md?

Sanoronas:
@MichaelOrlov Yes, they were working fine until the recent changes

rosbag2_py/src/rosbag2_py/info_sorting_method.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "info_sorting_method.hpp"
+
+
+namespace rosbag2_py
+{
+
+InfoSortingMethod info_sorting_method_from_string(std::string str)
+{
+  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+  auto find_result = sorting_method_map.find("str");
```
MichaelOrlov: Sorry, it seems I made a mistake originally putting `str` in quotes.
```suggestion
  auto find_result = sorting_method_map.find(str);
```
This is proof of why we need unit tests :wink:

rosbag2_py/src/rosbag2_py/info_sorting_method.cpp
```diff
@@ -0,0 +1,103 @@
...
+    sorted_idx.end(),
+    [&topics, sort_method](size_t i1, size_t i2) {
+      if (sort_method == InfoSortingMethod::TYPE) {
+        return topics[i1].topic_metadata.type < topics[i2].topic_metadata.type;
+      }
+      if (sort_method == InfoSortingMethod::COUNT) {
+        return topics[i1].message_count < topics[i2].message_count;
+      }
+      return topics[i1].topic_metadata.name < topics[i2].topic_metadata.name;
+    }
```
MichaelOrlov: Please use switch-case instead of if-else with returns. As I've mentioned before.
In case of default: "unknown value" in switch case, please throw std::runtime_error("switch is not exhaustive");

MichaelOrlov (changes requested):
@Sanoronas  Thanks for addressing most of my previous comments.
 
I don't think that moving `ServiceInfo` with `ServiceMetadata` to the `rosbag2_storage` package is a good idea because it will cause a lot of confusion about why it is there if it is not used in any way in those package.
I belive we can do better.

First of all, there is some confusion about naming after the implementation of services support in Rosbag2. It seems I overlooked it on the review back then.
We already have a definition for the `rosbag2_service_info_t` in the `rosbag2_cpp` package, where we distinguish the service's requests and responses. 
At the same time, we had a similar structure, `ServiceInfo`, in the `rosbag2_py` package defined in the `format_bag_metada.cpp`. However, in the `ServiceInfo`, we consolidate the service's requests and responses into the one `event_message_count` field. In the Service Introspection feature, we name it as **ServiceEvents**.

To clean up the mess, we need to do the following: 
1. Rename `ServiceInfo` to the `ServiceEventInfo` to avoid further misunderstanding and confusion.
2. Move declaration of the `ServiceMetada` and `ServiceEventIno` in the separate header file `ServiceEventInfo.hpp` and put it inside **rosbag2_py/src**, since we are not going to use it somewhere outside of the `rosbag2_py` package. This way, you will be able to refer to it from the `format_bag_metadata.cpp` and from the newly created `info_sorting_method.hpp{cpp}` files.

MichaelOrlov:
@Sanoronas As regards:
> However, I was unable to regenerate stub-files and cannot track down the reason for this right now. Any help on that regard would be helpful.

I've tried to checkout your latest version and follow up the instruction from the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md
and I've got error message when trying to accomplish step
>  Make sure rosbag2_py can be imported
python3 -c 'import rosbag2_py'

The error message is 
```
 python3 -c 'import rosbag2_py'
  File "<string>", line 1, in <module>
  File "/home/morlov/ros2_rolling_new/install/rosbag2_py/lib/python3.10/site-packages/rosbag2_py/__init__.py", line 35, in <module>
    from rosbag2_py._storage import (
ImportError: /home/morlov/ros2_rolling_new/install/rosbag2_py/lib/python3.10/site-packages/rosbag2_py/_storage.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN10rosbag2_py19generate_sorted_idxERKSt6vectorISt10shared_ptrIN15rosbag2_storage18ServiceInformationEESaIS4_EENS_17InfoSortingMethodE
```
I think this error message will go away as soon as you will move `ServiceInformation` type decalration back to the `rosbag2_py` package as I advised you above. If not, please let me know; I will try to reiterate and help you with pyi stubs regeneration.
Please note that pyi regeneration is usually the last step when all other changes in the rosbag2 packages have already been done.

Sanoronas:
@MichaelOrlov 
Moving ServiceEventInformation inside rosbag2_py didn't help with the pybind-issue. I think it has to do with linking of the source code somehow as I was able to change the import error by adding a resource to _storage element in CMakeLists.
However, I was unable to fully resolve the issue. For now it is working when the implementation of generate_sorted_idx is done directly in format_service_info.cpp. All other locations somehow are fine with the function being included, but not at this position.
If you have an idea what to change please let me know. Having all sorting functions in one file is probably better in the long-run.

Your other comments should also be resolved now.

Regarding tests I used the existing resources for now. For more detailed tests we would probably need to create another resource with differing count, types and names to properly distinguish the available options.

MichaelOrlov (commented):
@Sanoronas I've managed to fix linker issues and was able to regenerate pyi stubs.
I also made small cleanups by moving things around and applying small nitpicks.
Please see my three latest commits on your branch.

However, I didn't have a chance to review unit tests yet.

MichaelOrlov:

MichaelOrlov (changes requested):
@Sanoronas Ok. I reviewed the tests.
Overall looks good. However, it will be more useful to use `bag_with_topics_and_service_events.mcap{db3}` files as test files, since they have many topics and service events. While the `cdr_test.mcap{db3}` has only two topics.

Could you please rewrite tests to use `bag_with_topics_and_service_events.mcap{db3}` files to have better coverage?

Sanoronas:
@MichaelOrlov Thanks for your fixes!

I chose `cdr_test.mcap{db3}` over `bag_with_topics_and_service_events.mcap{db3}` as it has unordered topic names and differing count and types for the contained topics.
Testing ordering on `bag_with_topics_and_service_events.mcap{db3}` is probably not as useful as ordering by name, count and type will always be the same result.

MichaelOrlov (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#1804
Gist: https://gist.githubusercontent.com/MichaelOrlov/8e4617bc24aa9c31e2f2ca79a536429f/raw/c5e2c9724054c21164179d235978ac8bed2b7403/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_py rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_py rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14712
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22004/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16265/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1518/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22739/)

MichaelOrlov:

rosbag2_py/rosbag2_py/_info.pyi
```diff
@@ -2,7 +2,8 @@ import rosbag2_py._storage
 
 class Info:
     def __init__(self) -> None: ...
-    def print_output(self, arg0: rosbag2_py._storage.BagMetadata) -> None: ...
-    def print_output_topic_name_only(self, arg0: rosbag2_py._storage.BagMetadata) -> None: ...
-    def print_output_verbose(self, arg0: str, arg1: rosbag2_py._storage.BagMetadata) -> None: ...
+    def get_sorting_methods(self) -> Set[str]: ...
+    def print_output(self, arg0: rosbag2_py._storage.BagMetadata, arg1: str) -> None: ...
+    def print_output_topic_name_only(self, arg0: rosbag2_py._storage.BagMetadata, arg1: str) -> None: ...
+    def print_output_verbose(self, arg0: str, arg1: rosbag2_py._storage.BagMetadata, arg2: str) -> None: ...
```
MichaelOrlov, line 8: It looks like we are changing API here, which is publicly exposed. Likely we can't backport this PR.
@clalancette Could you please give a second opinion on that?

clalancette: I'm not totally familiar with what is going on here, but I can offer a couple of thoughts:

1.  It looks to me that the API break is only in adding an additional parameter to `print_output`, `print_output_topic_name_only`, and `print_output_verbose`.  Given that this is Python, can we give that final parameter a default value, so that anything depending on the "old" API will continue to work?
2.  The intent of the "no breaks in API" rule is so that third-party code that depends on these APIs continue to work across updates to the core.  If these are considered entirely internal APIs (i.e. there are no external users), then it is probably acceptable to break API here.

MichaelOrlov: @clalancette Thank you for the information I will take it into consideration in the backporting PR https://github.com/ros2/rosbag2/pull/1838.

ros-discourse:

PR #1828, opened 2024-10-11
Snapshot writes to a different file every time it is triggered

cmuehlbacher:
Addresses the issue described in #1099.

This is a simple change adding an argument on the command line to use a split function each time a snapshot is taken. It uses split bagfile function already implemented. Any improvement on this function regarding file names or returns can be simple translated to the snapshot function afterwards. 

It is developed on the Humble branch as we need this feature in humble. But it should be easily ported to rolling. 

There might be one concern the StorageOptions struct has a change (a member was added). This might cause issues with the ABI (depend on if this structure is part of it or not).

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -140,6 +140,10 @@ class RecordVerb(VerbExtension):
                  'the "/rosbag2_recorder/snapshot" service is called.'
         )
         parser.add_argument(
+            '--split-snapshots', action='store_true',
+            help='Splitting the bag file if a snapshot is taken'
+        )
+        parser.add_argument(
             '--ignore-leaf-topics', action='store_true',
             help='Ignore topics without a publisher.'
         )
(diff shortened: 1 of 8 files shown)


MichaelOrlov (changes requested):
@cmuehlbacher Thanks for the initiative, but not this way.
- Please see my comment and discussion in the https://github.com/ros2/rosbag2/issues/1099#issuecomment-2048821972

cmuehlbacher:
> @cmuehlbacher Thanks for the initiative, but not this way.
> 
> * Please see my comment and discussion in the [Snapshot should write to a different file every time it is triggered #1099 (comment)](https://github.com/ros2/rosbag2/issues/1099#issuecomment-2048821972)

Thanks for the quick look. 

As there is no renaming involved only point number 2 is not properly considered in my change. 
I will have a look if this change is possible in Humble, otherwise I will give it a shot in Rolling (with the API constraints in Humble this might be easier). 

But just for my understanding about the code as we have a specific parameter setting (snapshot and cache enabled). 
- In the SequentialWriter use_cache_ is set (line 140 in sequential_writer.cpp). 
- Therefore, we use the circular message cache (line 148-149 in sequential_writer.cpp). Which is also noted within the documentation about the cache. 
- With the CircularMessageCache the function swap_buffers (line 95 in circular_message_cache.cpp) swaps only the buffers if the data are ready. And clears the flag afterwards. 
- This flag is only set if notify_data_ready is called. Which is called when performing a snapshot. 
- Therefore, data are only written once, regardless how often swap buffers is called. 
  So, there should be no race condition/additional data written for this specific case. Right?

cmuehlbacher:
Close this pull request as this feature will be implemented through #1842.

PR #1835, opened 2024-10-14
Remove rolling->master mirroring workflow

### LINKED ISSUE #1834: GitHub CI: broken rolling->master mirroring
There's a GitHub workflow that mirrors `rolling` to `master`: https://github.com/ros2/rosbag2/blob/rolling/.github/workflows/mirror-rolling-to-master.yaml. This was created when the default/Rolling branch was switched from `master` to `rolling`.

The workflow has been failing for a while: https://github.com/ros2/rosbag2/actions/runs/11312031065/job/31458961112. And `master` is now 177 commits behind `rolling`: https://github.com/ros2/rosbag2/tree/master. This is because #1247 was mistakenly merged into `master` instead of `rolling`, although it was subsequently reverted in #1248. However, since `master` can no longer be fast-forwarded to `rolling`, the mirroring action fails.

Options:

1. Force-push `rolling` to `master` to allow the mirroring action to work again
2. Get rid of the mirroring action
    - We could also get rid of the `master` branch, but we don't have to

christophebedard:
Resolves #1834

As discussed in the above issue, remove the mirroring workflow, but keep the `master` branch.


clalancette (approved):
No need to run CI here since it is a GitHub workflow.

christophebedard:
I'll let anyone who has the necessary permissions merge this.

fujitatomoya (approved):
lgtm.

maybe we can do the same thing to ROS 2 core repos gradually.

PR #1836, opened 2024-10-15
Progress bar for ros2 bag play

### LINKED ISSUE #1762: Add progress bar and timestamp when playing back a bag
## Description

Currently there is no timestamp displayed when playing a bag and it doesn't show how far in you are.
This makes analysing specific issues at a certain point in time harder

## Completion Criteria

The current timestamp is shown when playing a bag, together with a completion percentage

Seems straight forward, but I'm not proficient in writing CLI tools

Just play any bag

nicolaloi:
This PR adds a progress bar for `ros2 bag play`, showing the bag time and duration, similar to what is seen in ROS1.
- Closes https://github.com/ros2/rosbag2/issues/1762: (https://github.com/ros/ros_comm/blob/8250c7d434ea34d0589eb8b6eaab5df1b11fd325/tools/rosbag/src/player.cpp#L376-L399).

In ROS1, the progress bar was easy to handle because the `rosbag play` command didn't print additional output during playback. However, in ROS2, many logs can be printed during playback (e.g., keyboard control logs).

The progress bar is printed and continuously updated on the last line of the playback output. I have tested this initial implementation with keyboard controls (play, pause, play next, set rate), Ctrl+C, throwing an exception, and the flags `--log-level debug`, `--delay X`, and `--start-offset X`.


nicolaloi:
https://github.com/user-attachments/assets/0de8b91e-dfab-4373-a95d-4c8f28b2d4ec

nicolaloi:
https://github.com/user-attachments/assets/03fa3ccc-0841-4999-825e-9086cd6a2fec

(at 2.6-2.7 seconds of bag duration, the progress bar is slightly updating while in paused state because I was clicking the left arrow for the _play next_ keyboard control)

Since `ros2 bag play` generates more logs than its ROS1 counterpart, the key aspect of this PR is to print a string shorter than any potential log message not coming from rosbag2_player. For example, with the current draft of the progress bar, this prints nicely:
```bash
[INFO] [1729028986.845909253] [rosbag2_player]: Pausing play.
 Bag Time 1718553782.525239  Duration 1.272325/6.974378 [P]
```
while this longer version of the progress bar doesn't work properly without extra line clearing (in this case, it's easy to clear the line since the logs come from rosbag2_player., but for "external" logs, this becomes more difficult):
```bash
[INFO] [1729028986.845909253] [rosbag2_player]: Pausing play.us:Running]
 Bag Time 1718553782.525239  Duration 1.272325/6.974378 [status:Paused]
```

@MichaelOrlov looking forward to your feedback, for now I have made a test MVP code to check the feasibility of the feature.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -361,7 +373,8 @@ PlayerImpl::PlayerImpl(
   storage_options_(storage_options),
   play_options_(play_options),
   keyboard_handler_(std::move(keyboard_handler)),
-  player_service_client_manager_(std::make_shared<PlayerServiceClientManager>())
+  player_service_client_manager_(std::make_shared<PlayerServiceClientManager>()),
+  disable_progress_bar_(play_options.disable_progress_bar)
```
MichaelOrlov: Please rename `disable_progress_bar` variables to the `enable_progress_bar` for consistency. Since it is going to be disabled by default and will be enabled only when it is needed.

MichaelOrlov (changes requested):
@nicolaloi Thank you for your contribution.
I have a few concerns about the design and implementation.
First, we need to limit the frequency with which we will print in `std::cout`. The best way to do this is to create a local variable, aka `size_t progress_bar_update_freq_` (in Hz). 
e.g. 1 Hz = once per/sec, 2 Hz = twice per second, 3 = three times per second etc..
Then, check the delay between the last call in the `print_status(..)`. If the delay is more than the threshold, print; otherwise, skip.
We also need to move the check for the `if (disable_progress_bar_)` inside `print_status(..)` to make the logic cleaner and in one place.
Calling `double current_time_secs = RCUTILS_NS_TO_S(static_cast<double>(clock_->now()));` each time in the `print_status(..)` is very inefficient and expensive since we are locking mutex inside and transorming current time to the ROS time. We actually know the current ROS time at which we are playing. This is a timestamp from the message. I would suggest passing it to the print function as a parameter. 
i.e., add tiemstamp parameter to the `print_status(..)` function and use `message_ptr->recv_timestamp` as a source of timestamp.

I found the implementation of the `print_status(..)` function a bit messy.
That is a quintessence of the modern C++17 with templates and old style plain C code.
Need to avoid using C code. Please use C++ alternatives. I am not sure if we really need  to flush buffer each time. Flush is an expensive operation since this is a blocking system call. Need to avoid it if possible.

IMO templating `print_status(..)` function is overengineering and make code less readable and less understandable. The cleaner solution would be if using enum class values for different modes. In this case all extra wrappers like `check_and_print_status()` will not be needed. We will be able to call `print_status(PrintStatus::Paused)` directly from where we needed it.

rosbag2_transport/src/rosbag2_transport/player.cpp
nicolaloi: From the CLI options, should I leave it disabled by default, or should I enable it by default with a default update frequency of around 10-20 Hz?

nicolaloi:
@MichaelOrlov thanks for the comments.
- limit printing frequency, mutex and flushing: yes I already had a local change to allow the user to choose the update frequency from the CLI (default 10-20 Hz) to reduce the mutex locking and output stream flushing. But indeed using the timestamp of the last published message is better than calling `clock_->now()`. 
-- The rationale of clock_->now() was to be able to have a consistent time flow if you want to print the progress bar at a fixed rate but messages are not close in time. For example, if the next message is 3 seconds in the future, for the next 3 seconds the progress bar time will appear frozen since it is printing the same timestamp. But yes in practice is a niche case.
-- Regarding the flush, updating the progress bar without flushing will not update the printed time (or anyway it will get flushed automatically after some update calls, but there can be artifacts if other external ros logs are being printed). So instead of calling the progress bar without flushing, I would just reduce the calls to the progress bar printing with the already mentioned user-defined frequency update. I would say that 10-20 Hz can be a good default value? Or even lower to be conservative?
- add timestamp parameter to the `print_status`: for this I would save it as a class attribute and use it within print_status instead of adding it as an input, since in many instances where `print_status` is being called there is no direct easy access to the last message.
- `print_status` implementation: ok I will remove the template and use enums.

Should the progress bar be disabled by default, or enabled by default with a default low update frequency?

nicolaloi:
I have made changes as discussed.

Linked modifications: 

I have modified the `play_next` function to accept a boolean input argument 'is_burst_mode' (default to false); so when `play_next` is called in burst mode, the `print_status` is not called at all, to avoid any delay (but now the progress bar will freeze during burst).

`print_status` will update the progress bar at the defined update frequency if the status is RUNNING; for the other states (or if the new boolean argument 'force_update' is true) it will update it disregarding the update frequency. This is to take care of the istances where a new log message from the rosbag player is being printed (like 'rate changed to 1.1', 'player paused', warnings..., anyway these are not high-frequency logging) to make the progress bar visible again. Otherwise, it will temporarily or permanently (permanently if the player gets paused) disappear since it is overwritten in the terminal by the previous log message.

MichaelOrlov:
@nicolaloi I have not had a chance yet to look deeply into the code after your latest changes. However, I have a few comments.
1. I am ok with a flush; I anticipated that it is likely some constraints as you described.
2. Making the progress bar enabled by default is a good idea. However, the frequency should be of integer type (size_t) in C++, and 10 Hz IMO is to fast. I would put 3-5 Hz at max by default.
3. We don't need flag `play_options.disable_progress_bar` at all we can rely on the `play_options.progress_bar_update rate`. i.e. if it is 0 = disabled.

nicolaloi:
Ok, I'll set the default rate to 3 Hz. 

For 2nd and 3rd points, there is no longer the `--disable_progress_bar` option, but a `--progress_bar` option that saves the frequency rate as a double, and if the value is ==0 the progress bar is disabled, and if <0 it is updated for every message. So the user has the extra option to enable it for each message if they need to, without having to set a large enough frequency rate. If it's not okay I'll change to size_t as you suggested.

nicolaloi:
I have realized the newly added variables `progress_bar_last_time_printed_` and `recv_timestamp_last_published_` in commit 43fa0f8 could have synchronization issues since the function `print_progress_bar_status` can be called contemporaneously from a service callback (like `/play`, `/pause`, `/play_next`), I need to take a better look at them.

nicolaloi:
@MichaelOrlov So, to avoid synchronization issues without using mutexes, etc to not increase overhead, the variable `progress_bar_last_time_printed_`, which tracks the last update time of the progress bar, is now accessed exclusively within the specialized function `update_progress_bar_after_message_published`. This function is called in code sections that are already protected by an existing mutex (`skip_message_in_main_play_loop_mutex_`). The progress bar updates at the specified rate only during the "standard" and recurrent event of a message being published.
For all other "extraordinary" and less common events occurring in separate threads (user triggers pause, resume, change rate, play next, etc, via keyboard controls or services) the function `update_progress_bar` is used. To avoid synchro issues this function does not access `progress_bar_last_time_printed_`. It will always print the progress bar so it appears again immediately after the new log messages; however, it does not update `progress_bar_last_time_printed_`. If I can use a mutex or an atomic int64, I can update `progress_bar_last_time_printed_` for these "extraordinary" events as well (i.e. ensuring the progress bar timing accounts for these extraordinary updates), but I don't think is needed given they are not common.

Regarding how to print the current timestamp: when a message is published, the `update_progress_bar_after_message_published` function is invoked with the message's timestamp as input as suggested. However, for the "extraordinary" events (user triggers pause, resume, change rate, play next, etc, via keyboard controls or services) where the timestamp is not directly available, the saved timestamp of the latest published message (`recv_timestamp_last_published_`) is not used anymore, since it could have synchro issues, but `clock_->now()` is used. This locks the inner mutex of `clock_`, but these events are very rare (or may not occur at all) compared to message publishing.

Extras: 
- fixed printed timestamp when extarordinary events occur while in delayed status (used an atomic bool, but it is accessed only for these rare extraordinary events), related to #1858 (`clock_` is not paused when in delayed status);
- added burst status `[B]`, so in case of a large number of messages being burst, the progress bar will not be frozen at [P]aused status. It will be frozen at [B]urst status, with the [B] conveying the information that messages are currently being burst (progress bar timestamp is frozen because normal continuous updates are blocked during burst to avoid delays);
- reduced the fixed precision of the displayed bag duration, from microseconds to milliseconds (`Duration 1.862/74.434`), to reduce the overall length of the progress bar string. Anyway, the displayed bag time (`Time 1718553783.196209`) has still microsecond precision.

MichaelOrlov:
@nicolaloi I have not had a chance yet to make a new full round of review. However, I have a comment about reducing the precision for output bag duration.
We used to use `sec.nanosec` format for timestamps in the bag and this is an "unified" format for other tools in ROS 2.
IMO make sense to use it without reduction to the milliseconds or nanoseconds. Please see example in the https://github.com/ros2/rosbag2/blob/c8feaea5b64e824bbe76e920f48a3ca39b72f9fc/rosbag2_storage/include/rosbag2_storage/yaml.hpp#L70-L86

The possible use case is that the user may stop playback at the point of interest and then copy past the current timestamp, for instance, to the Foxglove Studio to be able to open the same bag at the current location for further introspection.

nicolaloi:
I wasn't using nanoseconds because the problem with them is that the progress bar message will be longer than other log messages, and artifacts will appear:

> the key aspect of this PR is to print a string shorter than any potential log message not coming from rosbag2_player.

Using nanoseconds for example:
```
[INFO] [1733429800.530785030] [rosbag2_player]: Pausing play.13 [R]    
[INFO] [1733429801.732687906] [rosbag2_player]: Set rate to 1.1 [P]    
 Bag Time 1718553785.447256804 Duration 4.194342136/6.974378413 [P] 
```
Some digits and the final [P] will stay at the end of previous log messages.

But if nanoseconds for both bag time and bag duration are a must, then I can make space for their nanosecond digits by
  - reducing the progress bar string by removing the anyway static ending time info from the progress bar (can be printed before, together with the initial log info messages of the player);
  - reducing the progress bar string by removing/shortening the "Bag Time" and "Duration" words.
 
   So something like (printing an explanation for _t:_ and _d:_ at the beginning, together with the ending time info):
   ```
   [INFO] [1733433036.847650453] [rosbag2_player]: Resuming play.
    Bag t: 1718553786.279276609 d: 5.026361942 [R]
   ```
I will push this now.

MichaelOrlov:
@nicolaloi But why ??
> the key aspect of this PR is to print a string shorter than any potential log message not coming from rosbag2_player.

Why not have a longer length, whatever we need, and whatever will be readable?

MichaelOrlov:
Can we print a separator line in between? Like
```
[INFO] [1733429800.530785030] [rosbag2_player]: Pausing play.13 [R]    
[INFO] [1733429801.732687906] [rosbag2_player]: Set rate to 1.1 [P]    
===================================================================
 Bag Time 1718553785.447256804 Duration 4.194342136/6.974378413 [P] 
```

nicolaloi:
I am using short strings to try to avoid as much as possible 'poisoning' the log messages of other nodes that could print on stdout on the same terminal, given that at the end of the progress bar there is a carriage return '\r'. For example, a long string like:
```
`Bag Time 1718553785.447256804 Duration 4.194342136/16.974378413 [R]` 
```
will be overwritten (due to `\r') by this external log message from the user in the same terminal:
```
[INFO] [1733429801.732687906] [my_node] SLAM i 7, inliers: 25
```
into
```
[INFO] [1733429801.732687906] [my_node] SLAM i 7, inliers: 25413 [R]
```
And now you cannot know if the inliers were 2, 25, 254, 2541, or 25413.
While with a shorter progress bar string:
```
[INFO] [1733429801.732687906] [my_node] SLAM i 7, inliers: 25
 Bag t: 1718553785.447256804 d: 16.194342136 [R]
```
Then, there can always be shorter strings than any possible progress bar, especially if the user just prints using std::cout and not RCLCPP_INFO_STREAM. So this problem could happen whatever the progress bar length is, but using a shorter one makes it less likely. Especially, I was trying to have a progress bar at least shorter than any log message that could be printed by rosbag2_player, so the various "Set rate to 1.2", "Pausing play", etc.. printed by the rosbag2_player would not become something like `Pausing play.13 [R]`.

This was my reason for using short progress bar strings, but if these are not issues or anyway something that the progress bar should not take care of, then yes I can push a progress bar with the long version.

Regarding the separator line, I can try to think about it, maybe it is feasible using only raw ANSI escape codes, maybe it is difficult to implement without side-effects or without using a terminal control library like curses. Especially considering that you don't have control or knowledge of all the messages that could be printed in the terminal.

MichaelOrlov:
@nicolaloi As far as I understand you are warring about output std::cout output mashup due to the calling it from the multiple threads.
I found some explanation on https://stackoverflow.com/questions/14718124/how-to-easily-make-stdcout-thread-safe
```cpp
//Thread 1
std::cout << "the quick brown fox " << "jumped over the lazy dog " << std::endl;

//Thread 2
std::cout << "my mother washes" << " seashells by the sea shore" << std::endl;

//Could just as easily print like this or any other crazy order.
my mother washes the quick brown fox seashells by the sea shore \n
jumped over the lazy dog \n
```
The proposed solution is simply compose your string before you pass it to cout

For example.
```cpp
//There are other ways, but stringstream uses << just like cout.. 
std::stringstream msg;
msg << "Error:" << Err_num << ", " << ErrorString( Err_num ) << "\n"; 
std::cout << msg.str();
```
This way your stings can't be garbled because they are already fully formed, plus its also a better practice to fully form your strings anyway before dispatching them.

nicolaloi:
I am already composing the string with an ostringstream before printing it.
What I meant is related to the carriage return: the progress bar ends with a `\r`, so the next progress bar / rosbag_player log message / other nodes log message will be printed on the same line, not on a new line, effectively overwriting the current line, where the progress bar is. But the new message can overwrite only the first _len(new message)_ characters. If the new log message is longer than the progress bar, the last _len(new message)-len(progress bar)_ characters will stay in the terminal at the end of the new log message, like the example from my previous message.
This could be solved by clearing the line with ANSI code before printing any new log message, but you need to modify all the code that prints to terminal. Which can be done for every RLCPP_X_STREAM within the player.cpp file, but you miss all the RCLCPP_X_STREAM from the files imported by player.cpp, and all the prints effectuated by the user with their nodes:
> while this longer version of the progress bar doesn't work properly without extra line clearing (in this case, it's easy to clear the line since the logs come from rosbag2_player., but for "external" logs, this becomes more difficult):

MichaelOrlov:
@nicolaloi Ideally it would be nice to have something similar to the https://github.com/xqms/rosbag_fancy

@xqms Since you are the author of `rosbag_fancy´, your help with this PR and/or ideas on how to output the status bar without smashing other log messages will be very appreciated.

MichaelOrlov:
@nicolaloi In a short term I agree to shorten the status string to something like 
```
[1718553785.447256804] Duration 4.194342136/16.974378413 [R]
```
i.e. put the current timestamp into the square brackets `[ ]` to save space in line, and/or try to use a delimiter line.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1640,6 +1731,83 @@ const rosbag2_transport::PlayOptions & PlayerImpl::get_play_options()
...
+{
+  const double current_time_secs = RCUTILS_NS_TO_S(static_cast<double>(timestamp));
+  const double progress_secs = current_time_secs - starting_time_secs_;
+  // Print progress bar ending with carriage return '/r' so it will be overwritten by next print
+  // Spaces before '\r' are to clear any previous progress bar in case the new one is shorter,
+  // which can happen when the playback starts a new loop.
+  std::ostringstream oss;
+  oss << " Bag t: " << std::setw(13) << std::fixed << std::setprecision(9) <<
+    current_time_secs << " d: " << std::setprecision(9) << progress_secs <<
+    " [" << static_cast<char>(status) << "]      \r";
```
christophebedard: Having this many digits makes it hard to easily and quickly read the progress. Compare

```
4.194342136/16.974378413
```

with

```
4.19/16.97
```

The user also doesn't really need this amount of precision, right? In ROS 1, rosbag used 6 digits (IIUC), which is less than the 9 digits used here, but I think it's still more than what's really needed.

Something to consider.

MichaelOrlov: I agree. For the duration, we don't need so many digits after the floating point. IMO, 3 digits would be enough, up to millisecond accuracy.

cc: @nicolaloi

MichaelOrlov: Update. After more consideration, I think even 2 digits will be enough. I can't imagine a use case when higher precision would be usefull for the information about duration.

MichaelOrlov:
@nicolaloi General comment: Could you please rename the term `print frequency` and other "frequency" names to the progress bar update rate? 
The frequency and update rate are synonyms in this case.
However, frequency is a more scientific term, while the update the rate with times per second would be more user-friendly and easily understandable for everyone.

rosbag2_transport/src/rosbag2_transport/player.cpp
nicolaloi: Maybe there was a miscommunication yesterday, after https://github.com/ros2/rosbag2/pull/1836#issuecomment-2519646020 I changed the bag duration from milliseconds to nanoseconds, but probably the comment was referring to the bag timestamp only and not the actual bag duration.

Ok so I'll keep the timestamp with nanosecond precision, and change the bag duration to centiseconds (2 decimal digits) precision.

nicolaloi:
Ok based on the comments, I will change the progress bar string to `[1718553785.447256804] Duration 4.19/16.97 [R]`

nicolaloi:
force pushed since I forgot DCO in a previous commit

xqms:
Hey, here's how rosbag_fancy does it:

1. Print a few newline characters (completely clearing the line each time using ANSI control codes). This leaves space for other log messages so that they do not clobber the progress bar / UI
2. Print the progress bar (completely clearing the line each time)
3. Move the cursor back up to the blank lines so that log messages will appear there.

This also ensures that the progress bar is always at the bottom.

Note that this approach assumes that you update the UI more often than log messages appear.

Here's a link to the last step in the code: https://github.com/xqms/rosbag_fancy/blob/1f5ec4d172418a44ee206749182d0bec7cc553f0/rosbag_fancy/src/ui.cpp#L334

However, rosbag_fancy does not have to worry about multi-threading.

MichaelOrlov:
@nicolaloi Could you please rebase your branch and address conflicts one more time?
We pushed another PR recently which overlaps a bit with changes.

nicolaloi:
@xqms thanks for the suggestions. Adding extra lines is indeed a good idea to mitigate the problem. External user's log messages shorter than the progress bar string can still be smashed, but the extra lines give more safety margin. 

> Note that this approach assumes that you update the UI more often than log messages appear.

Due to this, I have added an extra option, `--progress-bar-separation-lines` (default to 2), so for these cases the user can increase the amount of blank lines. I have implemented this in the code so it can be tested. Here's a video showing 3 external user's log messages (**with the second one being a short message**) being printed between two progress bar updates, testing with 4,3,2,1, and 0 separation lines:

https://github.com/user-attachments/assets/c5cff63b-e7b6-42f6-8a8f-02801c3b9afc

From the video, you can see that with 3 external log messages (with one being a short message), you could encounter some issues with 2,1, and 0 separation lines.

General rule: if between two progress bar updates you are expecting N external log messages and at least one is **shorter** than the progress bar strings, then you should set at least N `--progress-bar-separation-lines`. Or, you should increase the progress bar update rate.

However, if the external log messages are longer than the progress bar strings, or there will be less than "default separation lines" (2) external log messages between two progress bar updates, then everything is ok by default. Moreover, this is an issue only if the user has his own log messages on the same output of the playback player and the progress bar is enabled.

@MichaelOrlov I have also added the separator line (`=======`)

nicolaloi:
friendly ping

MichaelOrlov:
@nicolaloi Sorry for the delay with the review.  
Overall, while the functionality is working, the code quality is still far from the production level and needs to be improved/redesigned.  Also lack of unit tests.
I will provide more details in a couple of days.

MichaelOrlov (changes requested):
Hi @nicolaloi. Thanks for your contributions.  
The code needs to be refactored to adhere to the single responsibility principle.  
The progress bar is a good candidate to be a separate self-contained class and software component with its own properties and behavior which should be covered by unit tests.

I've made such refactoring as well as many other fixes as I think make sense for progress bar implementation. 
Please review and feel free to cherry-pick my changes in a separate draft PR here https://github.com/ros2/rosbag2/pull/1902/files/c4a723136795a54ce09865655fb32ce8a597d9f6..42e5134e76c2eed227e3797a660ec9340c9f2a77.
Please note that https://github.com/ros2/rosbag2/pull/1902/commits/c4a723136795a54ce09865655fb32ce8a597d9f6 commit is all your commits from this PR squashed in one and rebased on top of latest master and the rest are mine.

Also please implement the missing unit test for the progress bar mentioned as TODO here https://github.com/ros2/rosbag2/pull/1902/commits/42e5134e76c2eed227e3797a660ec9340c9f2a77. The progress bar now can output to a provided `std::ostringstream` during construction. That should simplify writing unit tests to be able to deterministically verify output from it.

nicolaloi:
Yes at the very beginning, the progress bar was a separate class (not a pointer to implementation though) but then I integrated it in the player class in commit https://github.com/ros2/rosbag2/pull/1836/commits/4ba1a8c0a9bfb02eac79320a9b05820cb27654d6. Regarding the tests, I was waiting for a definite "progress bar style".
Thanks for the updates.

nicolaloi:
- I wrote the unit tests using loose/short regex to avoid multiple repetitions of "=== progress bar ===" or "\x1B[2K\n\x1B[2K..." in half of the tests. These substrings are specifically tested in a few tests only. Is this okay, or should the regex be more strict?
- Could you clarify the target string being tested in `update_with_limited_rate_with_zero_timestamp`? Also, what should be the difference in the tested string between `update_with_limited_rate_update_progress` and eg `update_with_limited_rate_with_negative_update_rate`?
- The DCO test is failing because I fast-forwarded your changes with a merge to preserve authorship, but it looks like I should have added my sign-off to the commits. Should I force-push after running `git rebase HEAD~X --signoff` to keep these commits, or should I force-push a single squashed commit instead?

MichaelOrlov:
@nicolaloi I am trying to verify tests and a simple scenario
```
    auto progress_bar = std::make_unique<PlayerProgressBar>(std::cout, 0, 0, 1, 2);
    progress_bar->update(PlayerProgressBar::PlayerStatus::DELAYED);
    std::cout << "Some log message !! \n";
    progress_bar->update(PlayerProgressBar::PlayerStatus::RUNNING);
```
gave me the following output
```


====== Playback Progress ======
[  0.000000000] Duration 0.00/0.00 [D]      Some log message !! 


====== Playback Progress ======
[  0.000000000] Duration 0.00/0.00 [R]
```
I would expect to see `Some log message !! ` printed out above the first `====== Playback Progress ======`.
Could you please explain why it is not?

nicolaloi:
@MichaelOrlov is the ouput you posted from a new cpp file or from a google test? Because in the google tests you see the whole buffer string, which has ANSI escape codes (`\033]3F`) to print the "log msg" before the progress bar, but in the actual buffer the "log msg" is after the progress bar, since it get std::couted after it.

MichaelOrlov:
So yes. I put
```
    auto progress_bar = std::make_unique<PlayerProgressBar>(std::cout, 0, 0, 1, 2);
    progress_bar->update(PlayerProgressBar::PlayerStatus::DELAYED);
    std::cout << "Some log message !! \n";
    progress_bar->update(PlayerProgressBar::PlayerStatus::RUNNING);
```
in the unit test and see output after the run.

nicolaloi:
Then it is okay. To see the real "terminal-like behaviour", those commands should be tested in a new cpp file without google tests. Gtest shows the actual raw output, which is ordered based on the order of the ostream calls: it does not show the ANSI post-processed "terminal-like" output.

docs/design/rosbag2_record_replay_service.md
```diff
@@ -150,6 +150,10 @@ Added or changed parameters:
 
     Rename from `--exclude`. Exclude topics and services containing provided regular expression.
 
+- `--progress-bar [ProgressBarFrequency]`
+
+    Print a progress bar of the playback player at a specific frequency in Hz. Negative values mark an update for every published message, while a zero value disables the progress bar. Default to a positive low value. 
+
```
MichaelOrlov: That needs to be moved appropriate section in the readme.md file

MichaelOrlov (changes requested):
Hi @nicolaloi I've made one more full review, debug, bugfixes and cleanups.

Please review my changes on your branch in the following commits:
- [Cleanup mess with ANSI control codes](https://github.com/ros2/rosbag2/pull/1836/commits/16f4308eb60815104b9454d2ccc1d60d06fcf4f4)
- [Remove padding before leading seconds in timestamp value](https://github.com/ros2/rosbag2/pull/1836/commits/5f743445a6eb313f4779cf710b41466a9e19f01a)
- [Remove redundant test cases and extra scope parenthesis](https://github.com/ros2/rosbag2/pull/1836/commits/e289e3f11b91f3d37fdfee6dbefe9bc95becb4a0)
- [Implement "update_with_limited_rate_with_zero_timestamp" test](https://github.com/ros2/rosbag2/pull/1836/commits/aa5cc5306bcbc81bbf460e1ad7cef3fc577792d2)
- [Sweep cleanup in test_player_progress_bar.cpp](https://github.com/ros2/rosbag2/pull/1836/commits/c5d99cbaedef34c7b815dbfa060528d1ab32e1af)

nicolaloi:
@MichaelOrlov Thanks for the review. I will be out for 7-8 days. I’ll have a look and manually test everything again, update the readme, and resolve the new conflicts with rolling in a week.

MichaelOrlov:
> @MichaelOrlov Thanks for the review. I will be out for 7-8 days. I’ll have a look and manually test everything again, update the readme, and resolve the new conflicts with rolling in a week.

Hi @nicolaloi, I have a branch from my review PR, https://github.com/ros2/rosbag2/pull/1902, which is a clean, rebased version on top of the latest Rolling.  
I can try to reset your branch to the latest Rolling and cherry-pick commits from my branch, from https://github.com/ros2/rosbag2/pull/1902 PR and then force-push them on your branch. But in this case, your earliest commits will be squashed in one.
Let me know if you are okay with it.

nicolaloi:
@MichaelOrlov ok 👍

MichaelOrlov:
@nicolaloi I rebased your branch and resolved all conflicts as we agreed.

nicolaloi:
@MichaelOrlov I have tested the new changes.

- Now I have the progress bar printing
  ```
  Progress bar [?]: [R]unning, [P]aused, [B]urst, [D]elayed, [S]topped
  [INFO] [1739648391.243081310] [rosbag2_player]: Playback until timestamp: -1
  
  
  ====== Playback Progress ====== 0.00/45.87 [S]
  [1718568699.685641766] Duration 34.33/45.87 [R]
  ```
  i.e. at line `====== Playback Progress ====== 0.00/45.87 [S]` the separator line is printed on top of the previous duration, which should be caused by the removal of the extra line clearing `\033[2K` in 
  [https://github.com/ros2/rosbag2/pull/1836/commits/d804c6cf1a00b21eba169d2dbb60a80608176f23#diff-824d4b9789dc0e75765483df3d74aad35cc0985ca5f2ffa93cefe70c3e7f1d95L59](https://github.com/ros2/rosbag2/pull/1836/commits/d804c6cf1a00b21eba169d2dbb60a80608176f23#diff-824d4b9789dc0e75765483df3d74aad35cc0985ca5f2ffa93cefe70c3e7f1d95L59)

Some comments on specific use cases:

- https://github.com/ros2/rosbag2/pull/1836/commits/2b75b6ff6a604d08832b2526b52d0cd60939af50 removed the immediate progress bar update when `play_next` is called. My idea was that during manual debugging, when pressing the right arrow key multiple times in a row to play the next message, the progress bar would display the correct timestamp immediately. Otherwise, the user now has to press the key slower than the progress bar update rate to see the correct timestamp. If it is possible to add an immediate update of the progress bar when `play_next` is called specifically by the keyboard command. However, it's true that the user could just play the bag with the progress bar enabled at a higher rate to mimic an immediate progress bar update when pressing the right arrow key.

- https://github.com/ros2/rosbag2/pull/1836/commits/1d4391827ec5e36c00cbd2ca55c65df14406d117 removed the padding from the timestamp. My idea was to minimize artifacts when a bag is recorded with simulated time starting, for example, from Unix time 0, and then played back with the loop flag enabled.
E.g. at the end of the bag `[1000.687641766] Duration 1000.69/1000.69 [R]` could become `[0.000000000] Duration 0.00/1000.69 [R].69 [R]` when the bag starts a new loop.

  Similarly, the extra final spaces in https://github.com/ros2/rosbag2/blob/a6d02d21b6233d90916d8f57e898da9fe870f796/rosbag2_transport/src/rosbag2_transport/player_progress_bar_impl.hpp#L145  are used to help limit such artifacts when looping is enabled (also possible due to the duration progress substring shrinking as the bag restarts the loop, e.g., from `10000.00/10000.00` to `0.0/10000.00`).
I can modify the code to precompute in the constructor the necessary number of extra final spaces (if the loop flag is enabled), taking into account how the length difference between the timestamp and duration substrings will change when the loop restarts.

MichaelOrlov:
@nicolaloi Thank you for your review and comments.
**As regards:**
> Now I have the progress bar printing
Progress bar [?]: [R]unning, [P]aused, [B]urst, [D]elayed, [S]topped
[INFO] [1739648391.243081310] [rosbag2_player]: Playback until timestamp: -1


> ====== Playback Progress ====== 0.00/45.87 [S]
[1718568699.685641766] Duration 34.33/45.87 [R]
i.e. at line ====== Playback Progress ====== 0.00/45.87 [S] the separator line is printed on top of the previous duration, which should be caused by the removal of the extra line clearing \033[2K in
https://github.com/ros2/rosbag2/commit/d804c6cf1a00b21eba169d2dbb60a80608176f23#diff-824d4b9789dc0e75765483df3d74aad35cc0985ca5f2ffa93cefe70c3e7f1d95L59

Using the `oss_clear_and_move_cursor_down << "\033[2K";` in the constructor and trailing spaces to make sure that the progress bar would be printed without leftovers from a previous output was a very **hacky** and "unclear" way to achieve that goal.

The more obvious and cleaner way to make sure that the progress bar is printed out on a clean space would be to use a special `\032[2K` ANSI code when forming progress bar output strings. Please see my new https://github.com/ros2/rosbag2/pull/1836/commits/9554b72bd19a847b7018c959f71c8b6edcc683df commit with these changes.

**As regards:**
> https://github.com/ros2/rosbag2/commit/1d4391827ec5e36c00cbd2ca55c65df14406d117 removed the padding from the timestamp. My idea was to minimize artifacts when a bag is recorded with simulated time starting, for example, from Unix time 0, and then played back with the loop flag enabled.
E.g. at the end of the bag [1000.687641766] Duration 1000.69/1000.69 [R] could become [0.000000000] Duration 0.00/1000.69 [R].69 [R] when the bag starts a new loop.

The `[1000.687641766] Duration 1000.69/1000.69 [R] could become [0.000000000] Duration 0.00/1000.69 [R].69 [R] when the bag starts a new loop` is expected and consistent behavior. The leftovers from the previous output will be removed with the `\032[2K` ANSI code in a simple and consistent way. No need to count and add trailing spaces anymore.

MichaelOrlov:
@nicolaloi As reagrds:
> https://github.com/ros2/rosbag2/commit/2b75b6ff6a604d08832b2526b52d0cd60939af50 removed the immediate progress bar update when play_next is called. My idea was that during manual debugging, when pressing the right arrow key multiple times in a row to play the next message, the progress bar would display the correct timestamp immediately. Otherwise, the user now has to press the key slower than the progress bar update rate to see the correct timestamp. If it is possible to add an immediate update of the progress bar when play_next is called specifically by the keyboard command. However, it's true that the user could just play the bag with the progress bar enabled at a higher rate to mimic an immediate progress bar update when pressing the right arrow key.

Good and valid point. I addressed it in the new https://github.com/ros2/rosbag2/pull/1836/commits/9a0f4f8f62d94016fccabd5cfd8bdae328e5d047 commit.

MichaelOrlov:
@nicolaloi I removed line clear ANSI code from the separation lines.

- Rationale:
By clearing separation lines we may occasionally erase important log messages. The separation lines are intended to be as a
 "buffer" in case if logging happened during drawing the progress bar. In most cases the separation lines will be empty by default but when not, we want to preserve those logging information.

nicolaloi:
> By clearing separation lines we may occasionally erase important log messages.

@MichaelOrlov what problem did you encounter? Because with the last commit that removes the line clearing I now have this behavior (multiple `=== Playback Progress ===` in the way):
img1
img2
instead of the "clean" terminal behavior of the preceding commits.

MichaelOrlov (approved):
@nicolaloi Sorry, my bad. I've reverted [Remove line clear ANSI code from separation lines](https://github.com/ros2/rosbag2/pull/1836/commits/47c1cf0ea41db11c32302d78040e8ec7d883d7d1) commit.  
I was relying on the wrong assumption that when we jump down for separation lines, there will be a clean, empty space by default. However, this is not true since we are not erasing the progress bar when jumping back to the upper lines after printing it.

I also made a few commits to update the wording in the readme and play.py files.

I tested it locally. It looks good to me now, and I think we are ready to go if CI passes green.

MichaelOrlov:
Pulls: ros2/rosbag2#1836
Gist: https://gist.githubusercontent.com/MichaelOrlov/278777b21b1a43b2b86f4d03fddbbb1d/raw/320d88bae8458e4bb42620db830cd5a6ccf9d221/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15239
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22584/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16827/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2099/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23386/)

MichaelOrlov:
Re-run CI jobs after fixing Windows test failures
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15240
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22585/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16828/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2100/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23387/)

MichaelOrlov:
CI is green. The failure on [Rpr](https://build.ros2.org/job/Rpr__rosbag2__ubuntu_noble_amd64/542/) job is a known issue reported in the https://github.com/ros2/rosbag2/issues/1914.

Merging than.

MichaelOrlov:
@nicolaloi My congratulations :fireworks:  the PR was finally merged.  
Thank you for your contribution!

rosbag2_transport/include/rosbag2_transport/player_progress_bar.hpp
```diff
@@ -0,0 +1,96 @@
...
+  /// \param timestamp Timestamp of the last published message. If timestamp is less than 0 the
+  /// last published timestamp will be taken from the inner progress bar cache.
+  void update(const PlayerStatus & status, const rcutils_time_point_value_t & timestamp = -1);
+
+  /// \brief Getter function for current player status.
+  /// \return Current player status.
+  PlayerStatus get_player_status();
+
+private:
+  std::unique_ptr<PlayerProgressBarImpl> pimpl_;
```
fujitatomoya, line 91: this generates the linage warning with windows. (the CI actually detected it as warning.)

- https://ci.ros2.org/job/ci_windows/23386/msbuild/
- https://ci.ros2.org/job/ci_windows/23387/msbuild/

The warning is happening because the `PlayerProgressBarImpl` class is not marked with the `ROSBAG2_TRANSPORT_PUBLIC` macro, which is required for classes used in DLL interfaces.

MichaelOrlov: @fujitatomoya 
- The warnings you are mentioning was discovered last week and raised in the separate issue https://github.com/ros2/rosbag2/issues/1926.  
- I've merged fix this morning in the PR https://github.com/ros2/rosbag2/pull/1927

fujitatomoya: @MichaelOrlov 👍 okay, i did not know that thanks!

ros-discourse:

adivardi:
@MichaelOrlov 
Could this be backported to Jazzy please?

PR #1839, opened 2024-10-17
Feature/snapshot write different file each trigger rolling

cmuehlbacher:
This ports the pull request #1828.
It also addresses the race condition mention in [@reinzor That is one more very hacky way to implement this feature.](https://github.com/ros2/rosbag2/issues/1099#issuecomment-2048821972), which was also mention in #1828 
- Relates #1099

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -177,6 +177,9 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
              'the "/rosbag2_recorder/snapshot" service is called. e.g. \n '
              'ros2 service call /rosbag2_recorder/snapshot rosbag2_interfaces/Snapshot')
     parser.add_argument(
+        '--split-snapshots', action='store_true',
+        help='Splitting the bag file if a snapshot is taken')
+    parser.add_argument(
         '--log-level', type=str, default='info',
         choices=['debug', 'info', 'warn', 'error', 'fatal'],
         help='Logging level.')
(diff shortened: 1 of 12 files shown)


MichaelOrlov (commented):
@cmuehlbacher Thank you for your contribution.
We need to make snapshot writing to the new file by default without providing CLI options.
Allowing multiple snapshots to be written in one file was a mistaken design decision from the very beginning. It has no value. The file with multiple snapshots is useless and difficult to handle during playback or introspection.
Imagine a situation when a few snapshot recorded in one file with interval of several hours.
The playback of such file will take also a several hours with huge delays when nothing will be published. Also it would be difficult to find when one snapshot beginnig and another is started during introspection.

In your current implementation possible at least two data races. The one when updating `metadata_` in `switch_to_next_storage()` and the second one when re-registering topics in the newly opened storage in the same function, because at the same time subscription callback can be triggered and `writer::write(message)` could be called.
To avoid data races we need to either add synchronization primitives and partially block `writer::write(message)` when we are updating `metadata_` and creating and switching to a new storage or make snapshot as blocking call the same as `writer::write(message)`.

Another point is that this PR with adding a new field to the `storage_options.hpp` and a new virtual function to the `message_cache_interface.hpp` is not backportable to other ROS 2 distros because this is an ABI breaking changes.

- I have took into account all concerns and created another smaller PR #1842. I cherrypicked your test with minor modifications and added you as co-author. Please review it and let me know what do you think.

cmuehlbacher:
Close this pull request as this feature will be implemented through #1842.

PR #1840, opened 2024-10-17
Add record, stop, start_discovery, stop_discovery and is_discovery_stopped services for rosbag2_transport::Recorder

### LINKED ISSUE #1634: Add record, stop, start_discovery and stop_discovery services for rosbag2_transport::Recorder
## Description
Expose rosbag2_transport::Recorder `record, stop, start_discovery and stop_discovery` APIs as a services

## Related Issues
- Relates https://github.com/ros2/rosbag2/issues/926

## Completion Criteria

Services for the rosbag2_transport::Recorder `record, stop, start_discovery and stop_discovery` exists

It is pretty trivial to add the same way as other services in the `rosbag2_transport::Recorder`.

- Will need to add more unit tests for the newly added services to the existent https://github.com/ros2/rosbag2/blob/rolling/rosbag2_transport/test/rosbag2_transport/test_record_services.cpp

sharminramli:
Adds the following services to rosbag2_transport::Recorder:
- record
- stop
- start_discovery
- stop_discovery
- is_discovery_stopped

Fixes some issues with starting record again after stopping:
- Restarts the event publisher thread
- Initialize stop_discovery_ to false and invert the check in start_discovery so that discovery is restarted when starting record again

Closes https://github.com/ros2/rosbag2/issues/1634


rosbag2_transport/test/rosbag2_transport/test_record_services.cpp
```diff
@@ -226,3 +247,37 @@ TEST_F(RecordSrvsTest, pause_resume)
...
+  auto & mock_writer = dynamic_cast<MockSequentialWriter &>(writer.get_implementation_handle());
+
+  EXPECT_FALSE(mock_writer.closed_was_called());
+
+  successful_service_request<Stop>(cli_stop_);
+  EXPECT_TRUE(mock_writer.closed_was_called());
+
+  successful_service_request<Record>(cli_record_);
+  EXPECT_FALSE(mock_writer.closed_was_called());
+}
```
MichaelOrlov, line 283: Please add recorder start after the stop and then again stop to make sure that we can start again after the stop

rosbag2_transport/include/rosbag2_transport/recorder.hpp
```diff
@@ -154,6 +159,10 @@ class Recorder : public rclcpp::Node
   ROSBAG2_TRANSPORT_PUBLIC
   bool is_paused();
 
+  /// Return the current discovery state
+  ROSBAG2_TRANSPORT_PUBLIC
+  bool is_discovery_stopped();
```
MichaelOrlov, line 164: Please rename to the `is_discovery_running()`

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -135,15 +138,20 @@ class RecorderImpl
...
   rclcpp::Service<rosbag2_interfaces::srv::Snapshot>::SharedPtr srv_snapshot_;
   rclcpp::Service<rosbag2_interfaces::srv::SplitBagfile>::SharedPtr srv_split_bagfile_;
+  rclcpp::Service<rosbag2_interfaces::srv::StartDiscovery>::SharedPtr srv_start_discovery_;
+  rclcpp::Service<rosbag2_interfaces::srv::Stop>::SharedPtr srv_stop_;
+  rclcpp::Service<rosbag2_interfaces::srv::StopDiscovery>::SharedPtr srv_stop_discovery_;
 
   std::mutex start_stop_transition_mutex_;
   std::mutex discovery_mutex_;
-  std::atomic<bool> stop_discovery_ = false;
+  std::atomic<bool> stop_discovery_ = true;
```
MichaelOrlov, line 154: stop_discovery_ shall be false by default. Otherwise, we will immediately exit from `RecorderImpl::topics_discovery()` when running it the first time.
```suggestion
  std::atomic<bool> stop_discovery_ = false;
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -294,6 +302,56 @@ void RecorderImpl::record()
...
     });
 
+  srv_start_discovery_ = node->create_service<rosbag2_interfaces::srv::StartDiscovery>(
+    "~/start_discovery",
+    [this](
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::StartDiscovery::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::StartDiscovery::Response>/* response */)
+    {
+      start_discovery();
```
MichaelOrlov, line 312: Need to check `if (in_recording_)` before calling start discovery. Otherwise, it will be an undefined behavior since the writer may not be opened yet, but discovery can try to create a subscription with a callback calling `writer->write(message)`.

MichaelOrlov (changes requested):
@sharminramli Thank you for your contribution.
IMO it will be more consistent and understandable if rename `is_discovery_stopped()` to the `is_discovery_running()`.
Also it would be more usefull if service request such as `start_discovery`, `stop_dicovery`, `record` and `stop` will return true or false as result.
Please note that `record` and `stop` could throw an exception. It does make sense to wrap them in try-catch and return fail in case of catching exception.
The service requests like `record` and `stop` shall be initialized in constructor rather than in the `record()` method. Otherwise how we will call `record` service request. 
We have some other service requests initializations inside record method only because they depend from opened writer and assumption that recorder running only once.
Since now we can run recorder multiple times need to safeguard them with check if writer is open. And consider moving them in constructor as well.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
sharminramli: stop_discovery_ will be exchanged for false in start_discovery(), and if it was previously true, then topics_discovery will be started. it should not exit because of the exchange.

subsequently stop_discovery() will exchange stop_discovery_ to true and when start_discovery() is called again it can again exchange stop_discovery_ for false and start topics_discovery()

sharminramli:
@MichaelOrlov  Thank you for the review. I've unfortunately been busy this week but I'm starting to work on the changes.

I have a question regarding:
> Also it would be more usefull if service request such as start_discovery, stop_dicovery, record and stop will return true or false as result.

I am right now reusing the existing Stop.srv in rosbag2_interfaces which is being used by the player. If I create a new srv e.g. StopRecorder.srv to have the boolean response, I would also rename the existing Stop.srv to e.g. StopPlayer.srv to avoid confusion. That however might break a lot of existing code for users. Alternatively I could add the bool to the existing Stop.srv (and also set it in the player when called). What do you think would be the best way?

MichaelOrlov:
@sharminramli Adding a return value to the existing service requests looks like more safer option.

MichaelOrlov:
- Closing as wouldn't be merged in favor of the replacement #2248. Since this is a stale PR with no feedback from the author.

PR #1842, opened 2024-10-26
Make snapshot writing into a new file each time it is triggered

### LINKED ISSUE #1099: Snapshot should write to a different file every time it is triggered

MichaelOrlov:
This PR will make snapshot writing into a new file each time it is triggered.
- Relates #1099 
- This PR is a replacement for the #1839

Note. Snapshot now becomes a blocking call and mutually exclusive with the `writer::write(message)` method to avoid race conditions, i.e., blocking the same `writer_mutex_`. 
The `set_data_ready()` method from #1839 is not needed since "snapshot" has now become a blocking call; therefore, there are no race conditions, and we will not dump data from the cyclic buffer twice because `data_ready` will be settled false after the first dump with `message_cache_->notify_data_ready();`.
https://github.com/ros2/rosbag2/blob/786c3c4b8ab05271630371cf515130ba02a9cde8/rosbag2_cpp/src/rosbag2_cpp/cache/circular_message_cache.cpp#L95-L106
However, we still need to call `split_bagfile_local(true)` to trigger callbacks and open a new storage file.

This PR could be backported. There are no API/ABI breaking changes in it.

### DIFF
--- a/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
+++ b/rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
@@ -29,6 +29,7 @@
 #include "rosbag2_cpp/writer.hpp"
 
 #include "rosbag2_storage/storage_options.hpp"
+#include "rosbag2_storage/ros_helper.hpp"
 
 #include "mock_converter_factory.hpp"
 #include "mock_metadata_io.hpp"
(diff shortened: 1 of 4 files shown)


cmuehlbacher:
Thank you for all the work on this feature. 

Looks very promising. And great news that it can be backported. 

As far as I can tell there should be no data races in there although, there will be several calls of write_messages function with an empty array. 

There is only one small remark:
The test snapshot_writes_to_new_file_with_file_compression closes the bag file after taking a snapshot while snapshot_writes_to_new_file does not. It would make sense to have both tests behave similarly.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -29,6 +29,7 @@
 #include "rosbag2_cpp/writer.hpp"
 
 #include "rosbag2_storage/storage_options.hpp"
+#include "rosbag2_storage/ros_helper.hpp"
```
christophebedard: simple reorder:

```suggestion
#include "rosbag2_storage/ros_helper.hpp"
#include "rosbag2_storage/storage_options.hpp"
```

christophebedard (commented):
from what I understand, the tests write messages into the cache, then trigger a single snapshot, and then validate everything.

I believe we should test that this works with more than 1 snapshot. Could we check the metadata too, e.g., `starting_time`, `duration`, and `message_count` too?

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -624,6 +625,82 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_file_com
...
+  initializeFakeFileStorage();
+  // Expect a single write call when the snapshot is triggered
+  EXPECT_CALL(
+    *storage_, write(
+      An<const std::vector<std::shared_ptr<const rosbag2_storage::SerializedBagMessage>> &>())
+  ).Times(1);
+
+  rosbag2_compression::CompressionOptions compression_options {
+    DefaultTestCompressor,
+    rosbag2_compression::CompressionMode::FILE,
```
christophebedard, line 643: what happens if we use `::MESSAGE` here? Looking at the logic and interaction between `SequentialCompressionWriter` and `SequentialWriter`, it should probably work just fine, but it might be worth testing it too.

christophebedard:
> Note. Snapshot now becomes a blocking call and mutually exclusive with the writer::write(message) method to avoid race conditions, i.e., blocking the same writer_mutex_.

I think this makes sense.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
MichaelOrlov: With CompressionMode::MESSAGE, the compression happens before messages will get to the cyclic buffer.
The bag split will be almost the same as in the base class. See
https://github.com/ros2/rosbag2/blob/786c3c4b8ab05271630371cf515130ba02a9cde8/rosbag2_compression/src/rosbag2_compression/sequential_compression_writer.cpp#L376-L393

IMO there is no value to test this scenario. It is basically covered by another test from the `test_sequential_writer`.

christophebedard: OK, all good then.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -409,7 +409,7 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
     is_first_message_ = false;
   }
 
-  if (should_split_bagfile(message_timestamp)) {
+  if (!storage_options_.snapshot_mode && should_split_bagfile(message_timestamp)) {
```
fujitatomoya, line 414: IMO this is obvious and expected behavior, but it would be probably nice to print the warning just once that this cannot be done for user application?

fujitatomoya (approved):
lgtm with green CI, just one minor comment that does not block this PR.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: It doesn't make sense to check that we `should_split_bagfile(..)` in snapshot mode because it could be a relatively heavy operation (for instance, for SQLite3). And I don't check it now. 
To print out the logging message, we will have to call `should_split_bagfile(..)` unconditionally first.

MichaelOrlov:
@christophebedard I added a check for metadata in the test. To cover the case if snapshot could be called twice - I created a separate test case.

@cmuehlbacher As regards
> As far as I can tell there should be no data races in there although, there will be several calls of write_messages function with an empty array.

No it is not. I double-checked in debuuger and we have constraints in tests that write messages will be called only once per snapshot.

As regards:
> The test snapshot_writes_to_new_file_with_file_compression closes the bag file after taking a snapshot while > snapshot_writes_to_new_file does not. It would make sense to have both tests behave similarly.

The reason why I use `writer->close()` in the `snapshot_writes_to_new_file_with_file_compression` is because otherwise without close it will be a race conditions in tests. The `close` method will wait for the compressor threads to finish.
I don't want to call `writer->close()`  in another `snapshot_writes_to_new_file_with_bag_split` test because want to check that file will be finalized when calling snapshot, without calling `writer->close()` method.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -580,6 +580,206 @@ TEST_F(SequentialWriterTest, snapshot_mode_zero_cache_size_throws_exception)
...
+  ASSERT_EQ(closed_files.size(), 2);
+
+  for (size_t i = 0; i < opened_files.size(); i++) {
+    const auto expected_closed = fs::path(storage_options_.uri) /
+      (bag_base_dir_ + "_" + std::to_string(i));
+    const auto expected_opened = fs::path(storage_options_.uri) /
+      (bag_base_dir_ + "_" + std::to_string(i + 1));
+    ASSERT_STREQ(closed_files[i].c_str(), expected_closed.generic_string().c_str());
+    ASSERT_STREQ(opened_files[i].c_str(), expected_opened.generic_string().c_str());
+  }
```
christophebedard, line 780: You don't want to check the metadata using `v_intercepted_update_metadata_`?

MichaelOrlov: I checked it in the previous test `snapshot_writes_to_new_file_with_bag_split`. 
This one to make sure that snapshot could be triggered twice and written in a separate files.
Adding all checks in one test is usually a bad practice. The rule of thumb that one test should check only one test case or action. 
IMO writing one big tests which will write twice and try to check metadata for all files will overcomplicate the logic in the test and will be difficult to understand and/or change in the future.

christophebedard: alright sure, that's fine.

christophebedard (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#1842
Gist: https://gist.githubusercontent.com/MichaelOrlov/127b9b542e0ed83d235de01e67be6919/raw/6dec255f6b87621d4194067c15f7fa94ebac1f63/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_tests
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14765
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16325/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1588/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22803/)

MichaelOrlov:
https://github.com/Mergifyio backport jazzy humble

reinzor:
Great! Thanks for this feature!

ros-discourse:

PR #1844, opened 2024-10-28
[WIP] The time bounded snapshot feature

### LINKED ISSUE #900: Snapshot based on duration
## Description

There was a recent addition of the snapshot feature that allows a circular buffer of fixed size. In our case, it is not possible to specify a buffer zise, but we would like to specify a duration. It would be great to be able to trigger a snapshot for X seconds in the past and Y seconds in the future.

## Related Issues

#663 

## Completion Criteria

- Able to trigger a snapshot based on time instead of memory

Extra goals would be support for multiple overlapping snapshots

MichaelOrlov:
This PR adds the "--snapshot-duration" CLI option.

The snapshot duration CLI option is specified as a floating point value in seconds.
The default value 0.0 indicates that the snapshot will be limited by the `--max-cache-size` parameter only. If the value is more than 0.0, the cyclic buffer for the snapshot will be limited by both the series of messages duration and the maximum cache size parameter. To override the upper bound by total messages size, the `--maximum-cache-size` CLI option can be settled to 0.

- Relates #900


MichaelOrlov:
@reinzor I've made good progress on this PR. I would say I've made all the required changes for this feature. However, I didn't have time to test it and write unit tests.

The help here would be really appreciated. Feel free to cherry-pick commits from this PR or create a fork above it.
Need to review https://github.com/ros2/rosbag2/pull/1844, test it locally, write unit tests, and update documentation.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2021 Amazonhe t.com, Inc. or its affiliates. All Rights Reserved.
```
fujitatomoya: accidental typo?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -25,25 +25,44 @@ namespace rosbag2_cpp
...
-MessageCacheCircularBuffer::MessageCacheCircularBuffer(size_t max_cache_size)
-: max_bytes_size_(max_cache_size)
+MessageCacheCircularBuffer::MessageCacheCircularBuffer(
+  size_t max_cache_size,
+  int64_t max_cache_duration_ns)
+: max_bytes_size_(max_cache_size), max_cache_duration_(max_cache_duration_ns)
 {
+  if (max_bytes_size_ == 0 && max_cache_duration_ == 0) {
+    ROSBAG2_CPP_LOG_ERROR_STREAM("Invalid arguments for the MessageCacheCircularBuffer. "
+                                 "Both max_bytes_size and max_cache_duration are zero.");
```
fujitatomoya: if this is debug message, it should include some extra debug information that the following exception cannot throw as in the message? otherwise, it is the same information with the following exception, so maybe debugging message here is not really useful since user can see the exact same message via exception?

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -145,16 +145,18 @@ void SequentialWriter::open(
     throw std::runtime_error{error.str()};
   }
 
-  use_cache_ = storage_options.max_cache_size > 0u;
+  use_cache_ = storage_options.max_cache_size > 0u ||
+    (storage_options.snapshot_mode && storage_options.snapshot_duration.nanoseconds() > 0);
```
fujitatomoya: checking `storage_options.snapshot_mode` is redundant?

```suggestion
  use_cache_ = storage_options.max_cache_size > 0u || storage_options.snapshot_duration.nanoseconds() > 0;
```

rosbag2_transport/test/resources/player_node_params.yaml
```diff
@@ -49,4 +49,7 @@ player_params_node:
       max_cache_size: 9898
       storage_preset_profile: "resilient"
       snapshot_mode: false
+      snapshot_duration:
+        sec: 1
+        nsec: 500000000
       custom_data: ["key1=value1", "key2=value2"]
```
fujitatomoya: Requires a newline at the end of the file?

fujitatomoya (commented):
@reinzor FYI, i left a few comments for your reference when you take this over.

ndmmdn:
Is this still in development? It seems close to finished, what is blocking its merge?

reinzor:
Sorry, did not have time to work on tests and testing this feature. Afaik that is what still is missing.

ndmmdn:
Understood, thanks for the quick reply. Do you approximately know if/when you will be able to get to it?

reinzor:
When this gets on our radar again and I have time to work on this. I cannot commit to a specific date.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
MichaelOrlov: @fujitatomoya The reason why I added a "duplicated" logging message is that in most cases if it fails, it is going to be a process termination with an unhandled exception. In this case the body of the exception may not appear in the log or std::out/std::err.

rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
MichaelOrlov: Yeah. it seems my cat :cat: made his own contribution :)

rosbag2_transport/test/resources/player_node_params.yaml
MichaelOrlov: Added a new line to the end of file

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov: Removed extra check for the `storage_options.snapshot_mode`

MichaelOrlov:
- Closing as wouldn't be merged in favor of the superseeding PR https://github.com/ros2/rosbag2/pull/2289

PR #1846, opened 2024-10-29
Add in a library prefix for lz4 from conda on Windows.

clalancette:
The comment explains why we need to do this.

Opening as a draft for right now until I run CI with Conda.

### DIFF
--- a/liblz4_vendor/cmake/Modules/Findlz4.cmake
+++ b/liblz4_vendor/cmake/Modules/Findlz4.cmake
@@ -7,7 +7,7 @@ endif()
 # When using lz4 from conda on Windows, and cmake before 3.29.0,
 # we need to add the library prefix "lib" so that the library can
 # be successfully found.
-if(MSVC and CMAKE_VERSION VERSION_LESS "3.29.0")
+if(MSVC AND CMAKE_VERSION VERSION_LESS "3.29.0")
   set(CMAKE_FIND_LIBRARY_PREFIXES "" "lib")
 endif()
 
--- a/liblz4_vendor/cmake/Modules/Findlz4.cmake
+++ b/liblz4_vendor/cmake/Modules/Findlz4.cmake
@@ -4,6 +4,13 @@ if(NOT lz4_ROOT_DIR AND DEFINED ENV{lz4_ROOT_DIR})
       "lz4 base directory location (optional, used for nonstandard installation paths)")
 endif()
 
+# When using lz4 from conda on Windows, and cmake before 3.29.0,
+# we need to add the library prefix "lib" so that the library can
+# be successfully found.
+if(MSVC and CMAKE_VERSION VERSION_LESS "3.29.0")
+  set(CMAKE_FIND_LIBRARY_PREFIXES "" "lib")
+endif()
+
 # Search path for nonstandard package locations
 if(lz4_ROOT_DIR)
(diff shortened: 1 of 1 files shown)


liblz4_vendor/cmake/Modules/Findlz4.cmake
```diff
@@ -4,6 +4,13 @@ if(NOT lz4_ROOT_DIR AND DEFINED ENV{lz4_ROOT_DIR})
       "lz4 base directory location (optional, used for nonstandard installation paths)")
 endif()
 
+# When using lz4 from conda on Windows, and cmake before 3.29.0,
+# we need to add the library prefix "lib" so that the library can
+# be successfully found.
+if(MSVC and CMAKE_VERSION VERSION_LESS "3.29.0")
```
christophebedard:
```suggestion
if(MSVC AND CMAKE_VERSION VERSION_LESS "3.29.0")
```

clalancette: Good call, thanks.  Fixed in 287d4ba.

clalancette:
This seems to have worked (see https://ci.ros2.org/view/All/job/test_ci_windows/590), so I'm marking it as "Ready for review" and I'll run full CI on it.

clalancette:
Pulls: https://github.com/ros2/rosbag2/pull/1846
Gist: https://gist.githubusercontent.com/clalancette/cf78cbb5356352239e74d7046884b1b2/raw/2fe4b8f9b36cc0cd7f27a74310fe3d75c12d4298/ros2.repos
BUILD args:  --packages-above-and-dependencies liblz4_vendor
TEST args:  --packages-above liblz4_vendor
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14755
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22058/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16314/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1576/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22791/)

liblz4_vendor/cmake/Modules/Findlz4.cmake
```diff
@@ -4,6 +4,13 @@ if(NOT lz4_ROOT_DIR AND DEFINED ENV{lz4_ROOT_DIR})
       "lz4 base directory location (optional, used for nonstandard installation paths)")
 endif()
 
+# When using lz4 from conda on Windows, and cmake before 3.29.0,
+# we need to add the library prefix "lib" so that the library can
+# be successfully found.
+if(MSVC AND CMAKE_VERSION VERSION_LESS "3.29.0")
+  set(CMAKE_FIND_LIBRARY_PREFIXES "" "lib")
+endif()
```
christophebedard, line 12: I'm guessing this works fine on Windows without Conda?

christophebedard (approved):

liblz4_vendor/cmake/Modules/Findlz4.cmake
clalancette: > I'm guessing this works fine on Windows without Conda?

Not exactly.  With our current setup, we don't actually install liblz4 anywhere, and we just always build it.  But since we are trying to be more like a Linux distribution with Conda, we prefer to use the system version there.  So this would have always failed with liblz4 (at least, prior to CMake 3.29.0), we just never noticed before.

PR #1848, opened 2024-11-01
Add support for replaying multiple bags

christophebedard:
This PR adds support for replaying multiple bags with `ros2 bag play` and the underlying classes (`rosbag2_transport::Player`/`rosbag2_py::Player`).

To replay multiple bags, use the new `-i,--input` CLI option.

```
$ ros2 bag play -i bag1 -i bag2 -i bag3 [storage_id]
```

Like with the `ros2 bag convert` option, the `-i,--input` option takes an URI to the bag folder or dedicated bag file and an optional explicit storage ID. This option can be used multiple times to be able to specify multiple bags. We keep the `bag_path` positional argument for backward compatibility and simplicity of usage in cases where users need to play just one bag file. However, the standalone `--storage` CLI option has been deprecated. In the future, users will need to use `-i bag [storage_id]` if they need to specify the storage ID explicitly.

The `rosbag2_transport::Player` now uses a `std::priority_queue`, which required changing the logic and interaction between `PlayerImpl::play_messages_from_queue()` and `PlayerImpl::play_next()`. In short, message publication is now all handled by `PlayerImpl::play_messages_from_queue()`. `PlayerImpl::play_next()` simply tells it to play the next message and then waits for the result. While the queue is not lock-free anymore, this change does simplify the interaction between `PlayerImpl::play_messages_from_queue()` and `PlayerImpl::play_next()` a bit. Note: the priority queue depth is defined by the `--read-ahead-queue-size` CLI option, which is set to **1000** elements by default.

Messages are played in order, based on their `recv_timestamp` (recorder reception timestamp). However, with the priority queue, this could eventually be changed (e.g., through `PlayOptions`) to the `send_timestamp` (original publication timestamp).

Other notes for reviewers:
1. Like `ros2 bag convert`, I added the `-i, --input` option that takes a bag URI and an optional explicit storage ID. This option can be used multiple times for multiple bags.
2. I kept the `bag_path` positional argument (but made it optional) because it is quite nice to be able to just do `ros2 bag play my-bag`. However, I deprecated the storage ID option (`--storage`); users will have to use `-i, --input` to specify an explicit storage ID for an input bag.
3. **This PR doesn't introduce API/ABI breaking changes and can be backported to the `jazzy` branch**. Some APIs needed to be added (mainly constructors) to support N input bags.
    - We can consider deprecating some APIs in another PR.


rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -240,14 +247,17 @@ class Player
...
-    const rosbag2_storage::StorageOptions & storage_options,
+    const std::vector<rosbag2_storage::StorageOptions> & storage_options,
     PlayOptions & play_options,
     bool burst = false,
     size_t burst_num_messages = 0)
   {
     install_signal_handlers();
     try {
-      auto reader = rosbag2_transport::ReaderWriterFactory::make_reader(storage_options);
+      std::vector<rosbag2_transport::Player::ReaderStorageOptionsPair> bags{};
```
MichaelOrlov: I think `readers` or `readers_with_options` is more sutable for this name rather than `bags`. 
Bags a bit unclear what it is. Even though it is really a readers with options.

ros2bag/ros2bag/api/__init__.py
```diff
@@ -157,14 +158,94 @@ def check_not_negative_int(arg: str) -> int:
         raise ArgumentTypeError('{} is not the valid type (int)'.format(value))
 
 
-def add_standard_reader_args(parser: ArgumentParser) -> None:
+def add_standard_reader_args(parser: ArgumentParser, multi: bool = False) -> None:
+    """
+    Add arguments for one of multiple input bags.
+
+    :param parser: the parser
+    :param multi: whether this is part of a verb that takes in multiple input bags
```
MichaelOrlov: General comment about `multi` term. It is a bit unclear what it is, especially in different functions and contexts. 
I would like to ask to use a name that is more clear to understand what it is without getting to know the context.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -885,8 +904,15 @@ Player::callback_handle_t PlayerImpl::get_new_on_play_msg_callback_handle()
 
 void PlayerImpl::wait_for_filled_queue() const
 {
+  const auto read_ahead_queues_filled = [this]() {
+      bool filled = true;
+      for (std::size_t i = 0; i < input_bags_.size(); i++) {
+        filled &= message_queue_.size_approx(i) >= play_options_.read_ahead_queue_size;
+      }
+      return filled;
+    };
```
MichaelOrlov:
```suggestion
  const auto read_ahead_queues_filled =
    [this]() {
      for (std::size_t i = 0; i < input_bags_.size(); i++) {
        if (message_queue_.size_approx(i) < play_options_.read_ahead_queue_size) {
          return false;
        }
      }
      return true;
    };
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,26 +927,37 @@ void PlayerImpl::load_storage_content()
...
-      enqueue_up_to_boundary(queue_upper_boundary);
-    } else {
-      lk.unlock();
-      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+      if (message_queue_.size_approx(bag_index) < queue_lower_boundary) {
+        enqueue_up_to_boundary(queue_upper_boundary, bag_index);
+      } else {
+        lk.unlock();
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+      }
```
MichaelOrlov: We should sleep only when we iterrate over all readers and found that all queues are full.

Also if we will decide to switch to the `std::priority_queue` it would be better to iterate over each reader one by one in a round-robin fashion, rather than calling `enqueue_up_to_boundary(queue_upper_boundary, bag_index);` for each reader.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,26 +927,37 @@ void PlayerImpl::load_storage_content()
...
-    if (!reader_->has_next()) {
+  for (size_t i = message_queue_.size_approx(bag_index); i < boundary; i++) {
+    if (!input_bags_[bag_index].first->has_next()) {
       break;
     }
-    message = reader_->read_next();
-    message_queue_.enqueue(message);
+    message = input_bags_[bag_index].first->read_next();
+    message_queue_.enqueue(message, bag_index);
   }
```
MichaelOrlov: Suggestion to improve the readability of the code
```
for (size_t i = message_queue_.size_approx(bag_index); i < boundary; i++) {
    auto & reader = input_bags_[bag_index].first;
    if (!reader->has_next()) {
      break;
    }
    message = reader->read_next();
    message_queue_.enqueue(message, bag_index);
  }
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1092,7 +1131,11 @@ void PlayerImpl::prepare_publishers()
...
 
   // Create topic publishers
-  auto topics = reader_->get_all_topics_and_types();
+  std::vector<rosbag2_storage::TopicMetadata> topics{};
+  for (const auto & [reader, _] : input_bags_) {
+    auto bag_topics = reader->get_all_topics_and_types();
+    topics.insert(topics.begin(), bag_topics.begin(), bag_topics.end());
```
MichaelOrlov: Inserting to the end may save us from an extra copy of other elements.
```suggestion
    topics.insert(topics.end(), bag_topics.begin(), bag_topics.end());
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1092,7 +1131,11 @@ void PlayerImpl::prepare_publishers()
   }
 
   // Create topic publishers
-  auto topics = reader_->get_all_topics_and_types();
+  std::vector<rosbag2_storage::TopicMetadata> topics{};
+  for (const auto & [reader, _] : input_bags_) {
+    auto bag_topics = reader->get_all_topics_and_types();
```
MichaelOrlov, line 1194: Potentially we could have the same topic in multiple input bags. 
As far as I can see, this situation handling properly further on. 
However, I would like to ask to add a comment about it and mention in what place we handling this case.

MichaelOrlov (changes requested):
Hi @christophebedard, Thank you for your contribution.

After thoroughly reviewing and considering the newly added class `MetaPriorityQueue`.
I think it will not be much more efficient than `std::priority_queue` protected with mutex from the STL.
For a few reasons:
1. The MetaPriorityQueue **push(message)** and **pop()** became blocking calls and also had to be protected by a mutex.
2. We impose a constraint on the `MetaPriorityQueue` that items queued into the same queue shall be in order of priority. That might not be true if we need to switch playback to use `send_timestamp`.
3. If considering performance, the `std::priority_queue` has log(n) runtime complexity for insertion and extraction operations and constant for peaking maximum element. While we are trying to achieve a constant time performance when using lock-free queues for relatively small numbers, the log(n) overhead is going to be very insignificant. For instance, we currently use 1000 elements for the queue by default. If we try to read out from 5 files and increase the maximum queue size by a factor of 5, i.e., to be 5000 elements, the **log(5000) = 8.5**. In the worst-case scenario, we will require 8.5 pointers swaps or iterations in the array. In comparison to the current design with `MetaPriorityQueue`, we will have at least 5 iterations over 5 queues trying to find and update the current minimum index, and also, the overall underlying implementation of the lock-free queue, is not free and has some extra operations inside. The difference is going to be in the numbers of the **5-8** iterations over the array or swaps. Which is IMO negligible in this case.
I would prefer to have a more cleaner and straightforward impelmentation based on the `std::priority_queue`.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: fixed

ros2bag/ros2bag/api/__init__.py
christophebedard: changed to `multiple_readers` and also made the names of some variables more clear.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: I just added a comment mentioning this and saying that this is correctly handled later

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: I've changed the logic to make it more similar to how it was before. This should hopefully fix the unlock+sleep.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: fixed, thanks for the suggestion.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: changed to `readers_with_options`.

rosbag2_transport/test/rosbag2_transport/test_play_seek.cpp
```diff
@@ -212,7 +212,7 @@ TEST_P(RosBag2PlaySeekTestFixture, seek_forward) {
   await_received_messages.get();
 
   auto replayed_topic1 = sub_->get_received_messages<test_msgs::msg::BasicTypes>("/topic1");
-  EXPECT_THAT(replayed_topic1, SizeIs(expected_number_of_messages));
+  ASSERT_THAT(replayed_topic1, SizeIs(expected_number_of_messages));
```
christophebedard, line 215: Since we're accessing `replayed_topic1` after this, we should assert its size instead of just expecting it. Same thing with the other instances below.

rosbag2_transport/test/rosbag2_transport/test_play_seek.cpp
```diff
@@ -103,7 +103,7 @@ class RosBag2PlaySeekTestFixture
 };
 
 TEST_P(RosBag2PlaySeekTestFixture, seek_back_in_time) {
-  const size_t expected_number_of_messages = num_msgs_in_bag_ + 2;
+  const size_t expected_number_of_messages = num_msgs_in_bag_ + 4;
```
christophebedard, line 106: We expect all remaining messages to be played after resuming, so this should be 4, even though it passes with 2 because we simply stop receiving messages once we get the number of expected messages.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -682,35 +712,26 @@ bool PlayerImpl::play_next()
...
-    message_queue_.pop();
-    message_ptr = peek_next_message_from_queue();
-  }
-  return next_message_published;
+  // Wait for play next to be done, and then return the result
+  std::unique_lock<std::mutex> lk(finished_play_next_mutex_);
+  // Temporarily resume clock to force wakeup in clock_->sleep_until(time),
+  // then return in pause mode to preserve original state of the player
+  clock_->resume();
+  clock_->pause();
```
christophebedard, line 745: I will check if we can wake up `clock_->sleep_until()` without resuming and pausing the clock, maybe through a new clock API.

Then I could remove the check for `!play_next_.load()` in `play_messages_from_queue()`.

christophebedard:
Changed to `std::priority_queue`.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: Looks like just calling `std::condition_variable::notify_all()` on the clock's condition variable should wake it up from its sleep (i.e., [`std::condition_variable::wait_for()`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for) and [`std::condition_variable::wait_until()`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_until)):

> The thread will be unblocked when [notify_all()](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all) or [notify_one()](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) is executed, or abs_time is reached. It may also be unblocked spuriously.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -16,16 +16,16 @@
...
 #include <memory>
+#include <queue>
 #include <regex>
 #include <string>
 #include <unordered_map>
 #include <utility>
 #include <vector>
 #include <thread>
 
-#include "moodycamel/readerwriterqueue.h"
```
christophebedard, line 27: I don't think anything else in rosbag2 uses this. Are there any reasons not to remove the `shared_queues_vendor` package completely?

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: I opened a separate PR for this: #1869.

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
```diff
@@ -0,0 +1,106 @@
...
+  T take()
+  {
+    rcpputils::unique_lock<std::mutex> lk(queue_mutex_);
+    if (queue_.empty()) {
+      return empty_element_;
+    }
+    const T e = queue_.top();
+    queue_.pop();
+    return e;
+  }
```
MichaelOrlov: In very rare cases, I advocate the use of `std::optional`, and I am really trying to avoid it.
But in this particular case usage of the `std::optional<T>` for take return type is well justified.

christophebedard: Changed to `std::optional<T>`: 939eca31da71a0b0f5bf9472b68efc36290d557b

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -74,6 +74,10 @@ class Player : public rclcpp::Node
   /// delete_on_play_message_callback.
   using callback_handle_t = uint64_t;
 
+  /// \brief Utility type for a pair of rosbag2_cpp::Reader and rosbag2_storage::StorageOptions.
+  using ReaderStorageOptionsPair = \
```
MichaelOrlov: Suggesting to rename for consistency since this is an inner-class definition.
```suggestion
  using reader_storage_options_pair_t = \
```

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include "rcpputils/thread_safety_annotations.hpp"
+#include "rcpputils/unique_lock.hpp"
+
+/// \brief `std::priority_queue` wrapper with locks.
+/// \tparam T the element type
+/// \tparam Container the underlying container type
+/// \tparam Compare the comparator
+/// \see std::priority_queue
+template<typename T, typename Container, typename Compare>
```
MichaelOrlov: Suggestion to add a default values to the template parameters
```suggestion
template<typename T, typename Container = std::vector<T>,
  typename Compare = std::less<typename Container::value_type>>
```

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
```diff
@@ -0,0 +1,104 @@
...
+  /// \brief Remove the top element.
+  void pop()
+  {
+    rcpputils::unique_lock<std::mutex> lk(queue_mutex_);
+    queue_.pop();
+  }
+
+  /// \brief Check if the queue is empty.
+  /// \return whether the queue is empty
+  bool empty() const
```
MichaelOrlov: Nitpick from clangtidy
```suggestion
  [[nodiscard]] bool empty() const
```

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+  /// \brief Try to take the top element from the queue.
+  /// \return the top element, or `std::nullopt` if the queue is empty
+  std::optional<T> take()
+  {
+    rcpputils::unique_lock<std::mutex> lk(queue_mutex_);
+    if (queue_.empty()) {
+      return std::nullopt;
+    }
+    const T e = queue_.top();
```
MichaelOrlov: Out of curiosity, why define `e` as const?
The const qualifier may trigger an extra copy for compilator because we returning non- const value and the compilator likely will not be able to use copy-ellision optimisation.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -16,16 +16,16 @@
 
 #include <algorithm>
 #include <chrono>
+#include <limits>
 #include <memory>
+#include <queue>
```
MichaelOrlov: NItpick: the include of the `queue` became redundant
```suggestion
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -87,12 +99,12 @@ class PlayerImpl
 public:
   using callback_handle_t = Player::callback_handle_t;
   using play_msg_callback_t = Player::play_msg_callback_t;
+  using ReaderStorageOptionsPair = Player::ReaderStorageOptionsPair;
```
MichaelOrlov: Please use snake style naming as above for consistency

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -76,6 +77,17 @@ rclcpp::QoS publisher_qos_for_topic(
...
+/// Comparator for SerializedBagMessageSharedPtr to order chronologically by recv_timestamp.
+struct BagMessageRecvTimestampComparator
+{
+  bool operator()(
+    const rosbag2_storage::SerializedBagMessageSharedPtr & l,
+    const rosbag2_storage::SerializedBagMessageSharedPtr & r) const
+  {
+    return l->recv_timestamp > r->recv_timestamp;
+  }
+};
```
MichaelOrlov: It would be better to move `BagMessageRecvTimestampComparator` inside `PlayerImpl` class and rename it with a snake style
```suggestion
/// Comparator for SerializedBagMessageSharedPtr to order chronologically by recv_timestamp.
struct
{
  bool operator()(
    const rosbag2_storage::SerializedBagMessageSharedPtr & l,
    const rosbag2_storage::SerializedBagMessageSharedPtr & r) const
  {
    return l->recv_timestamp > r->recv_timestamp;
  }
} BagMessageRecvTimestampComparator;
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -353,48 +382,54 @@ class PlayerImpl
...
+  std::vector<ReaderStorageOptionsPair> && input_bags,
   std::shared_ptr<KeyboardHandler> keyboard_handler,
-  const rosbag2_storage::StorageOptions & storage_options,
   const rosbag2_transport::PlayOptions & play_options)
-: owner_(owner),
-  storage_options_(storage_options),
+: input_bags_(std::move(input_bags)),
+  owner_(owner),
   play_options_(play_options),
+  message_queue_(BagMessageRecvTimestampComparator{}),
```
MichaelOrlov:
```suggestion
  message_queue_(BagMessageRecvTimestampComparator),
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+        !clock_->sleep_until(message_ptr->recv_timestamp))
+      {
+        // Stop sleeping if cancelled
+        if (std::atomic_exchange(&cancel_wait_for_next_message_, false)) {
+          break;
+        }
       }
-      publish_message(message_ptr);
+
+      std::lock_guard<std::mutex> lk(skip_message_in_main_play_loop_mutex_);
```
MichaelOrlov: Nitpick from clangtidy. Rename to avoid shadowed variable names in the inner scope with another `lk` definition.
```suggestion
      std::lock_guard<std::mutex> lock_skip_message(skip_message_in_main_play_loop_mutex_);
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+      if (rclcpp::ok()) {
+        // This means that the queue's top() has changed, so we need to update our message
```
MichaelOrlov:
```suggestion
        // This means that we either invalidated queue from seek() and need to take a fresh element
        // from the top of the queue, or trying to exit from the main playback loop in stop()
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+        if (skip_message_in_main_play_loop_) {
+          skip_message_in_main_play_loop_ = false;
+          cancel_wait_for_next_message_ = false;
+          message_ptr = take_next_message_from_queue();
+          continue;
```
MichaelOrlov, line 1011: We shouldn't even try to take a new message if we are trying to escape when `stop_playback_ == true`
```suggestion
          if (stop_playback_.load()) {
            break;
          } else {
            message_ptr = take_next_message_from_queue();
            continue;
          }
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+            finished_play_next_ = true;
+            play_next_result_ = true;
+            finished_play_next_cv_.notify_all();
+          }
+        }
+      }
+      message_ptr = take_next_message_from_queue();
+    }
+
+    // At this point, we're at the end of the queue, there are no more messages to play
```
MichaelOrlov:
```suggestion
    // At this point, we're at the end of the playback round, there are no more messages to play
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+    // If we're still trying to play next or just got a request, we did not succeed
+    if (play_next_.exchange(false)) {
+      std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
+      finished_play_next_ = true;
+      play_next_result_ = false;
+      finished_play_next_cv_.notify_all();
+    }
+
+    // While we're in pause state, make sure we don't return
+    // However, start playing over if we're trying to play next
```
MichaelOrlov: Need to add some clarity why we need all these checks and `sleep_until(..)`
```suggestion
    // While we're in a pause state, make sure we don't return if we happen to be at the end of the
    // queue or playback round. However, if we get a request for play next during the sleep_until(..)
    // we need to proceed to be able to handle the play next request.
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +922,125 @@ void PlayerImpl::load_storage_content()
...
+  while (rclcpp::ok() && !stop_playback_) {
+    // If we had run out of messages before but are starting to play next again, e.g., after a
+    // seek()
+    if (play_next_.load()) {
+      std::lock_guard<std::mutex> lk(skip_message_in_main_play_loop_mutex_);
+      skip_message_in_main_play_loop_ = false;
+      cancel_wait_for_next_message_ = false;
+      // Only take again if we need
+      if (nullptr == message_ptr) {
+        message_ptr = take_next_message_from_queue();
```
MichaelOrlov: I think it would be better to move this block to the end of the loop to have more context as to why we need it.

MichaelOrlov (changes requested):
@christophebedard Thanks for the updates and one more iteration.
Overall seems logically correct. See my comments and requests for changes inline.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: But how about rosbag2 recorder?

christophebedard: it does not use it.

`rosbag2_cpp` declared a dependency on `shared_queues_vendor`, but it did not actually use it.

rosbag2_transport/include/rosbag2_transport/player.hpp
christophebedard: renamed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
christophebedard: changed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
christophebedard: I've added `[[nodiscard]]` to `empty()` in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26. If we do that, we might as well add it to `size()` and especially `take()` too, so I've done that too.

rosbag2_transport/src/rosbag2_transport/locked_priority_queue.hpp
christophebedard: removed `const` in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: removed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: renamed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: I've made those changes in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26.

However, I disagree with changing

```c
struct $name
{
};
```

to

```c
struct
{
} $name;
```

just to be able to change `$name{}` to `$name` when it's used in the constructor, because it doesn't mean the same thing. The first one is a struct type definition, while the second one is a struct type definition (anonymous) + instantiation. It becomes an attribute. Since the goal is to be able to choose a comparator (`recv_timestamp`, `send_timestamp`, or some other comparator), we do not need an instance of each. We just need the struct definitions and then we can instantiate the right one depending on `PlayOptions`, for example.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26, but please see my comment here: https://github.com/ros2/rosbag2/pull/1848#discussion_r1861071671

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: renamed in ca23e7fae501e6a2950f529ef0dafa2f37dd1f26

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: reworded in c6d2ef5951782a6405fc9fbe2d71e769cec7cac6

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in c6d2ef5951782a6405fc9fbe2d71e769cec7cac6

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: reworded in c6d2ef5951782a6405fc9fbe2d71e769cec7cac6

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: reworded in c6d2ef5951782a6405fc9fbe2d71e769cec7cac6

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: great idea, thank you. I believe I can then remove the `if (nullptr == message_ptr)` check, because that was meant to guard against taking again if we just took a message at the beginning of `play_messages_from_queue()`, before the `while`s. Changed in c6d2ef5951782a6405fc9fbe2d71e769cec7cac6

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Ok, feel free to remove it.

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -301,7 +354,7 @@ class Player : public rclcpp::Node
   ROSBAG2_TRANSPORT_PUBLIC
   /// \brief Getter for the currently stored storage options
   /// \return Copy of the currently stored storage options
-  const rosbag2_storage::StorageOptions & get_storage_options();
+  std::vector<rosbag2_storage::StorageOptions> get_storage_options();
```
MichaelOrlov: @christophebedard If we add a new API instead of changing the existent `get_storage_options(),` this PR will likely become API/ABI compatible, and we will be able to backport it to Jazzy.

The idea for the name: `get_all_storage_options()`

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -349,52 +366,69 @@ class PlayerImpl
...
+  /// Comparator for SerializedBagMessageSharedPtr to order chronologically by recv_timestamp.
+  struct
+  {
+    bool operator()(
+      const rosbag2_storage::SerializedBagMessageSharedPtr & l,
+      const rosbag2_storage::SerializedBagMessageSharedPtr & r) const
+    {
+      return l->recv_timestamp > r->recv_timestamp;
+    }
+  } bag_message_chronological_recv_timestamp_comparator;
```
MichaelOrlov, line 384: @christophebedard Could you please move this struct definition to the top of this class? Right before the first constructor. somewhere before or after using declarations.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: @christophebedard Hold off from using changes from the #1869 in this PR. It is a better way to apply them after merging this PR to be able to backport this PR to Jazzy as is.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +921,129 @@ void PlayerImpl::load_storage_content()
...
-  {
-    // Do not move on until sleep_until returns true
-    // It will always sleep, so this is not a tight busy loop on pause
-    while (rclcpp::ok() && !clock_->sleep_until(message_ptr->recv_timestamp)) {
-      if (std::atomic_exchange(&cancel_wait_for_next_message_, false)) {
-        break;
+  rosbag2_storage::SerializedBagMessageSharedPtr message_ptr = take_next_message_from_queue();
+
+  // While we haven't stopped playing
+  while (rclcpp::ok() && !stop_playback_) {
```
MichaelOrlov: I think it does make sense to add a comment that we need this outer loop for the cases when playback in pause i.e. need to wait on `clock.sleep_until(..)` and for the play next operation

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +921,129 @@ void PlayerImpl::load_storage_content()
...
+
+      std::lock_guard<std::mutex> lk_skip_message(skip_message_in_main_play_loop_mutex_);
+      if (rclcpp::ok()) {
+        // This means that the message we took from the queue's top was invalidated after a seek(),
+        // so we need to take a fresh element from the top of the queue, unless we have to stop
+        if (skip_message_in_main_play_loop_) {
+          skip_message_in_main_play_loop_ = false;
+          cancel_wait_for_next_message_ = false;
+          if (stop_playback_.load()) {
+            break;
```
MichaelOrlov: If we are exiting by by `stop_playback_ == true` and `play_next_` also `true` if we will not notify `play_next` thread he will stuck and wait forever.
However, it will be better to add such notification at the very end of the function and consider wrapping the main loop in a try-catch to avoid locking the `play_next` thread forever in case of an exception in the main playback loop.
```
  if (play_next_.exchange(false)) {
      std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
      finished_play_next_ = true;
      play_next_result_ = false;
      finished_play_next_cv_.notify_all();
    }
```

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: removed in dee9bc7aa212c991237aadde736f89281ffa0d67

rosbag2_transport/include/rosbag2_transport/player.hpp
christophebedard: I've changed it back to `const rosbag2_storage::StorageOptions & get_storage_options()` and added `std::vector<rosbag2_storage::StorageOptions> get_all_storage_options()`. I marked `get_storage_options()` as deprecated, but we can leave that out when backporting to Jazzy.

38403090993b817b3cdc8d91abcb9f8e94d3e166

christophebedard:
I'll rebase and fix the DCO & merge conflicts once the new changes look good.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +921,129 @@ void PlayerImpl::load_storage_content()
...
+      break;
+    }
+
+    // If we had run out of messages before but are starting to play next again, e.g., after a
+    // seek(), we need to take
+    if (play_next_.load()) {
+      std::lock_guard<std::mutex> lk(skip_message_in_main_play_loop_mutex_);
+      skip_message_in_main_play_loop_ = false;
+      cancel_wait_for_next_message_ = false;
+      message_ptr = take_next_message_from_queue();
```
MichaelOrlov, line 1059: To make sure that we are in `play_next` after the `seek` need to add a check under the mutex
`if (skip_message_in_main_play_loop_) {..}`
But in this case, it seems will be exact the same logic as in the lines: 984-997. It seems it will be the same logic if we would remove this block with `if (play_next_.load())` and just let the loop continue.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +921,129 @@ void PlayerImpl::load_storage_content()
...
+    // we need to stop waiting here and proceed to handle the play next request by doing another
+    // loop of this while().
+    while (!stop_playback_ && is_paused() && !play_next_.load() && rclcpp::ok()) {
+      clock_->sleep_until(clock_->now());
+    }
+    // If we ran out of messages and are not in pause state, it means we're done playing,
+    // unless play_next() is resuming and pausing the clock in order to wake us up
+    if (!is_paused() && !play_next_.load()) {
+      break;
+    }
```
MichaelOrlov, line 1051: As far as I remember we have guards and situation when `play_next == true` and `is_paused == false` is not possible.
Then it seems we can simplify logic and only check.
```cpp
    if (!is_paused()) {
      break;
    }
```

MichaelOrlov (commented):
@christophebedard Thank you for the next round of changes and improvements.
I started liking the way how it is implemented.
I've made one more round of review. Mostly in the rosbag2_transport player classes. See my comments inline.

I also noticed that if we will not change `get_storage_options()` API in the player.hpp file, this PR will likely become API/ABI compatible, and we will be able to backport it to Jazzy.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: but then it becomes public. There is no need to make it public, as we discussed yesterday.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: we cannot, because of what the comment is saying:

> unless play_next() is resuming and pausing the clock in order to wake us up

There's a race:

1. `play_messages_from_queue()` waits in this `while` if it is paused (which just means that the clock is paused) and there is no play next request:
    ```cpp
    while (!stop_playback_ && is_paused() && !play_next_.load() && rclcpp::ok()) {
      clock_->sleep_until(clock_->now());
    }
    ```
2. `play_next()` makes a request:
    ```cpp
     play_next_ = true;
    ```
3.
    - `play_next()` wakes up `play_messages_from_queue()` from `sleep_until()` by resuming the clock:
        ```cpp
        // Temporarily resume clock to force wakeup in clock_->sleep_until(time),
        // then return in pause mode to preserve original state of the player
        clock_->resume();
        ```
    - and `play_messages_from_queue()` exits from the `while` from (1) because `!play_next_.load()` is now `false`
5. `play_messages_from_queue()` then `break`s out of the outer `while` and exits because the clock is now resumed
    ```cpp
    // If we ran out of messages and are not in pause state, it means we're done playing,
    // unless play_next() is resuming and pausing the clock in order to wake us up
    if (!is_paused()) { // && !play_next_.load()) {
      break;
    }
    ```
5. `play_next()` then pauses the clock again, since its goal was just to stop `sleep_until()` by resuming the clock:
    ```cpp
    clock_->pause();
    ```
6. Now `play_messages_from_queue()` has exited, and no play next request will succeed again

I was able to reproduce this a few times in tests. The solution was to add `!play_next_.load()` to the `if` from (4). However, after merging #1869, we can indeed remove `!play_next_.load()` because we won't be using `resume()`+`pause()` to wake up.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: I would just change it back, because then that situation will get correctly handled

```diff
-          if (stop_playback_.load()) {
-            break;
-          } else {
-            message_ptr = take_next_message_from_queue();
-            continue;
-          }
+         message_ptr = take_next_message_from_queue();
+         continue;
```

no need to make it more complex

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: sorry, I don't understand your comment. `play_next_` cannot turn to `false` after being `true` if `play_messages_from_queue()` doesn't notify `play_next()`.

christophebedard: there is no need to check `if (skip_message_in_main_play_loop_) {..}` here, because it doesn't have a valid message at that point, and it _needs_ to take a new one anyway.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: added in f26ee05b2533fa263ef2c35784081f540f4f6b8e

rosbag2_transport/include/rosbag2_transport/player.hpp
MichaelOrlov: Thanks!

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Okay. Thanks for the detailed explanation. I was misled about `is_paused()`. I was under the impression that we rely on some private member variable for pause detection. But we are actually relying on the underlying `clock_` state, which we resume temporarily.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Okay, we agreed that we don't need `if`, because we need to take a new message anyway and it is safe to do so in all cases.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Okay. We've synched with @christophebedard offline and decided that we don't need wrapping the outer playback loop in try-catch with notification to unblock the play_next thread, and we also don't need to add 
```
    if (play_next_.exchange(false)) {
      std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
      finished_play_next_ = true;
      play_next_result_ = false;
      finished_play_next_cv_.notify_all();
    }
```
at the end of the outer play back loop in this function. Basically, this block is going to be without changes as it is now.
Because we have a try-catch on the upper level in the `play()` method and, we are going to add
```
    if (play_next_.exchange(false)) {
      std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
      finished_play_next_ = true;
      play_next_result_ = false;
      finished_play_next_cv_.notify_all();
    }
```
at the end of the `playback_thread_` right before leaving it.

Also, we will need to add. 
```
  if (!is_in_playback_) {
    RCLCPP_WARN_STREAM(owner_->get_logger(), "Called play next, but not in a playback state.");
    return false;
  }
```
at the beginning of the `bool PlayerImpl::play_next()` method.

christophebedard: please see c0fe94dc527d3698c31737c3704cc3c02b19bd64 for the changes.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: AI describes the second variant as
> struct
 {
 } name;
Anonymous struct:
This syntax defines an anonymous struct. It means the struct doesn't have a name, but you create an instance of it directly with the variable name.
Limited usage:
Anonymous structs are less commonly used. They are useful when you only need a single instance of the struct and don't need to reference it by name elsewhere in your code.

However, this is exactly our use case and the way how we are going to use it. i.e. we need only a single instance of it when passing it to the `message_queue_` constructor.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Okay, I Agree; currently, it is in the right place. a few lines upper from where it is going to be used.

rosbag2_transport/include/rosbag2_transport/player.hpp
christophebedard: I removed the deprecation notice in b22e27b73009b24cd3ae97064b203c838922e698. This way we can just backport this PR to Jazzy easily. I'll do the deprecation in another PR for Rolling, since we might want to deprecate other things too.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: but the `Player` class doesn't need to have an instance of this comparator struct. It's only needed by the priority queue.

christophebedard: anyway, as I said above, I've made the change.

ros2bag/ros2bag/api/__init__.py
```diff
@@ -157,14 +158,95 @@ def check_not_negative_int(arg: str) -> int:
...
+def add_standard_reader_args(parser: ArgumentParser, multiple_readers: bool = False) -> None:
+    """
+    Add arguments for one of multiple input bags.
+
+    :param parser: the parser
+    :param multiple_readers: whether this is part of a verb that takes in multiple input bags
+    """
+    # If multiple readers, let user provide an input bag path using an optional positional arg, but
+    # require them to use --input to provide an input bag with a specific storage ID
+
```
MichaelOrlov: Please create a separate `add_standard_multi_reader_args(parser: ArgumentParser)` function.
The rationale for that request is a difference in behavior. i.e., We deprecate `--storage` and provide `--input` parameters only for the `ros2 bag play` verb. The logic in the combined function became overcomplicated and not clear what is applied and when.

ros2bag/ros2bag/verb/play.py
```diff
@@ -196,11 +198,36 @@ def main(self, *, args):  # noqa: D102
             topic_remapping.append('--remap')
             topic_remapping.append(remap_rule)
 
-        storage_options = StorageOptions(
-            uri=args.bag_path,
-            storage_id=args.storage,
-            storage_config_uri=storage_config_file,
-        )
+        # Add bag from optional positional arg, then bag(s) from optional flag
```
MichaelOrlov, line 206: IMO., it does make sense to add storage options either from the positional argument or from the dedicated `--input` parameter, but not from both of them, to avoid a mess.
Also, makes sense to print an error if both parameters are using

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -240,14 +247,18 @@ class Player
...
+    const std::vector<rosbag2_storage::StorageOptions> & storage_options,
     PlayOptions & play_options,
     bool burst = false,
     size_t burst_num_messages = 0)
   {
     install_signal_handlers();
     try {
-      auto reader = rosbag2_transport::ReaderWriterFactory::make_reader(storage_options);
+      std::vector<rosbag2_transport::Player::reader_storage_options_pair_t> readers_with_options{};
+      for (const auto & options : storage_options) {
```
MichaelOrlov: reserve space for vector
```suggestion
      std::vector<rosbag2_transport::Player::reader_storage_options_pair_t> readers_with_options{};
      readers_with_options.reserve(storage_options.size());
      for (const auto & options : storage_options) {
```

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -151,6 +155,55 @@ class Player : public rclcpp::Node
     const std::string & node_name = "rosbag2_player",
     const rclcpp::NodeOptions & node_options = rclcpp::NodeOptions());
 
+  /// \brief Constructor which will construct Player class with provided parameters and default
+  /// rosbag2_cpp::reader and KeyboardHandler classes.
```
MichaelOrlov:
```suggestion
  /// \brief Constructor which will construct a Player class with multiple readers and provided
  /// storage options for each reader.
```

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -298,14 +351,19 @@ class Player : public rclcpp::Node
   ROSBAG2_TRANSPORT_PUBLIC
   size_t get_number_of_registered_on_play_msg_post_callbacks();
 
+  /// \brief Getter for the first of the currently stored storage options
+  /// \return Copy of the first of the currently stored storage options
```
MichaelOrlov:
```suggestion
  /// \return Reference to the first item in the vector of the StorageOptions.
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -174,6 +174,68 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
           ElementsAre(40.0f, 2.0f, 0.0f)))));
 }
 
+TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics_multi)
```
MichaelOrlov:
```suggestion
TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics_from_three_bags)
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -174,6 +174,68 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+{
+  auto msg = get_messages_basic_types()[0];
+  msg->int32_value = 42;
+
+  auto topic_types = std::vector<rosbag2_storage::TopicMetadata>{
+    {1u, "topic1", "test_msgs/BasicTypes", "", {}, ""},
+    {2u, "topic2", "test_msgs/BasicTypes", "", {}, ""},
```
MichaelOrlov: Need to avoid ambiguity in message types. It could be a services yet.
```suggestion
    {1u, "topic1", "test_msgs/msg/BasicTypes", "", {}, ""},
    {2u, "topic2", "test_msgs/msg/BasicTypes", "", {}, ""},
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -174,6 +174,68 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+  };
+
+  // Make sure each reader's/bag's messages are ordered by time
+  // However, do interlace messages across bags
+  constexpr std::size_t num_bags = 3u;
```
MichaelOrlov:
```suggestion
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -174,6 +174,68 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+  messages_list.emplace_back(std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>>{
+    serialize_test_message("topic1", 1, msg),
+    serialize_test_message("topic2", 4, msg),
+    serialize_test_message("topic1", 7, msg),
+    serialize_test_message("topic2", 9, msg),
+    serialize_test_message("topic1", 11, msg),
+    serialize_test_message("topic2", 15, msg)});
+  std::vector<rosbag2_transport::Player::reader_storage_options_pair_t> bags{};
+  std::size_t total_messages = 0u;
+  for (std::size_t i = 0u; i < num_bags; i++) {
```
MichaelOrlov:
```suggestion
  for (std::size_t i = 0u; i < messages_list.size(); i++) {
```

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -174,6 +174,68 @@ TEST_F(RosBag2PlayTestFixture, recorded_messages_are_played_for_all_topics)
...
+    prepared_mock_reader->prepare(messages_list[i], topic_types);
+    bags.emplace_back(
+      std::make_unique<rosbag2_cpp::Reader>(std::move(prepared_mock_reader)), storage_options_);
+  }
+  ASSERT_GT(total_messages, 0u);
+
+  auto player = std::make_shared<rosbag2_transport::Player>(std::move(bags), play_options_);
+  std::size_t num_played_messages = 0u;
+  rcutils_time_point_value_t last_timetamp = \
+    std::numeric_limits<rcutils_time_point_value_t>::lowest();
```
MichaelOrlov:
```suggestion
  rcutils_time_point_value_t last_timetamp = 0;
```
Zero is enough - we are not using timestamps less than 1 in this unit test.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -901,69 +938,126 @@ void PlayerImpl::load_storage_content()
...
+
+        const bool message_published = publish_message(message_ptr);
+        // If we tried to publish because of play_next(), jump the clock
+        if (play_next_.load()) {
+          clock_->jump(message_ptr->recv_timestamp);
+          // If we successfully played next, notify that we're done, otherwise keep trying
+          if (message_published) {
+            play_next_ = false;
+
+            std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
```
MichaelOrlov: Nitpick: Remove extra empty line
```suggestion
            play_next_ = false;
            std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1519,7 +1622,16 @@ void PlayerImpl::publish_clock_update(const rclcpp::Time & time)
...
 const rosbag2_storage::StorageOptions & PlayerImpl::get_storage_options()
 {
-  return storage_options_;
+  return input_bags_[0].second;
+}
+
+std::vector<rosbag2_storage::StorageOptions> PlayerImpl::get_all_storage_options()
+{
+  std::vector<rosbag2_storage::StorageOptions> storage_options{};
+  for (const auto & [_, options] : input_bags_) {
```
MichaelOrlov: Preallocate memory for the `storage_options`
```suggestion
  std::vector<rosbag2_storage::StorageOptions> storage_options{};
  storage_options.reserve(input_bags_.size());
  for (const auto & [_, options] : input_bags_) {
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1546,8 +1658,10 @@ Player::Player(const std::string & node_name, const rclcpp::NodeOptions & node_o
 
   auto reader = ReaderWriterFactory::make_reader(storage_options);
 
+  std::vector<reader_storage_options_pair_t> input_bags{};
```
MichaelOrlov: Please rename `input_bags` to the `readers_with_options`

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1560,6 +1674,28 @@ Player::Player(
     storage_options, play_options, node_name, node_options)
```
MichaelOrlov: The better way to use another new constructor 
```
: Player(std::vector{storage_options}, play_options, node_name, node_options)
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1560,6 +1674,28 @@ Player::Player(
...
+  std::shared_ptr<KeyboardHandler> keyboard_handler;
+  if (!play_options.disable_keyboard_controls) {
+    keyboard_handler = std::make_shared<KeyboardHandler>();
+  }
+
+  std::vector<reader_storage_options_pair_t> input_bags{};
+  for (const auto & options : storage_options) {
+    input_bags.emplace_back(std::make_unique<rosbag2_cpp::Reader>(), options);
+  }
+  pimpl_ = std::make_unique<PlayerImpl>(
```
MichaelOrlov:
```suggestion
  std::vector<reader_storage_options_pair_t> input_bags{};
  for (const auto & options : storage_options) {
    input_bags.emplace_back(ReaderWriterFactory::make_reader(options), options);
  }
  pimpl_ = std::make_unique<PlayerImpl>(
```
Also, please rename `input_bags` to the `readers_with_options` for clarity

MichaelOrlov (changes requested):
@christophebedard I've verified previous changes and made one more round of full review.
I have some suggestions for improvements but they are not require changes in the logic.
One of the generic suggestions is to rename local variables and constructor parameters `input_bags` as well as private class member variable `input_bags_` to the `readers_with_options`.
IMO it is more clear what it is from the name. It will be helpful for future code supportability. at least for the local member variables and ctor parameters.

Another comment is to create a separate `add_standard_multi_reader_args(parser: ArgumentParser)` function.
The rationale for that request is a difference in behavior. i.e., We deprecate --storage and provide --input parameters only for the ros2 bag play verb. The logic in the combined function became overcomplicated and not clear what is applied and when.
Also, see my other comments inline.

ros2bag/ros2bag/api/__init__.py
christophebedard: changed in a1ad63dc4670633ea5c102af5f7c8d12613825d3

ros2bag/ros2bag/verb/play.py
christophebedard: changed in a1ad63dc4670633ea5c102af5f7c8d12613825d3

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/include/rosbag2_transport/player.hpp
christophebedard: reworded in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/include/rosbag2_transport/player.hpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/test/rosbag2_transport/test_play.cpp
christophebedard: changed in fd349cc793eab28feb6ef089c20ad1a6329a4001

rosbag2_transport/test/rosbag2_transport/test_play.cpp
christophebedard: changed in fd349cc793eab28feb6ef089c20ad1a6329a4001

should this be done for all other tests then? Maybe in a separate PR?

rosbag2_transport/test/rosbag2_transport/test_play.cpp
christophebedard: changed in fd349cc793eab28feb6ef089c20ad1a6329a4001

rosbag2_transport/test/rosbag2_transport/test_play.cpp
christophebedard: changed in fd349cc793eab28feb6ef089c20ad1a6329a4001

rosbag2_transport/test/rosbag2_transport/test_play.cpp
christophebedard: changed in fd349cc793eab28feb6ef089c20ad1a6329a4001

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: changed in 82814f011d9f938889bc50d4d9060b95b1de0689

rosbag2_transport/test/rosbag2_transport/test_play.cpp
MichaelOrlov: Yes. we do this for all other tests in Apex.AI and then going to upstream changes.

MichaelOrlov (approved):
@christophebedard Thank you for addressing all comments.
Now looks good to me.
Please rebase and fix DCO.

christophebedard:
I just squashed to fix the DCO and rebased.

I also fixed a linter issue: https://github.com/ros2/rosbag2/compare/a82f2597188a67cb11752815481d130fec18a9bb..eeca3aad929d4b5d7c6882d47ddc128af4a09181

MichaelOrlov:
@christophebedard It seems we've got a regression in rosbag2_transport tests on RPR job: https://build.ros2.org/job/Rpr__rosbag2__ubuntu_noble_amd64/448/testReport/

christophebedard:

MichaelOrlov:
It seems it was a false alarm. The RPR job passed green from the second start. I also was not able to reproduce any failures on my local setup. Running full CI then.

MichaelOrlov:
Pulls: ros2/rosbag2#1848
Gist: https://gist.githubusercontent.com/MichaelOrlov/a8e3f30612a2ef94072b2023ba655197/raw/89ae0fbb84ac37308f77b3376800249f2a8c1bf8/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_py ros2bag rosbag2
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_py ros2bag rosbag2
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14877
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22189/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16441/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1710/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22939/)

MichaelOrlov:
Two warnings on the Windows build are in the StaticSingleThreaded executor
```
'rclcpp::executors::StaticSingleThreadedExecutor::~StaticSingleThreadedExecutor': Use rclcpp::executors::SingleThreadedExecutor (compiling source file C:\ci\ws\src\ros2\rclcpp\rclcpp\test\rclcpp\executors\test_executors.cpp)
```
and unrelated to the changes from this PR. Merging then to the Rolling.

cc: @christophebedard I changed a bit PR description and title to be more descriptive for others.

christophebedard:
> It seems it was a false alarm. The RPR job passed green from the second start. I also was not able to reproduce any failures on my local setup. Running full CI then.

I ran the `rosbag2_transport` `test_play_*` tests under `stress --cpu $num_cpu` for ~5 hours yesterday and couldn't reproduce.

christophebedard:
Thanks for the review, @MichaelOrlov

MichaelOrlov:

christophebedard:
@MichaelOrlov should I proceed with marking some of the old APIs (i.e., single input bag) as deprecated? In Rolling only, of course.

MichaelOrlov:
@christophebedard Yes please. I would appreciate a follow-up PR.
During my previous code review, I had in mind deprecating some constructors and put them on a separate branch for follow-up. https://github.com/ros2/rosbag2/compare/rolling...morlov/deprecate-some-player-ctor.
Please feel free to cherry-pick or consider taking them for your PR.

ros-discourse:

Issue #1847, opened 2024-10-31
Add Support for Selective Playback of Topics with Mixed Serialization Formats in MCAP Files


jespersmith:
### Summary

When logging data with a custom logger, I store both ROS2 data and additional custom data in MCAP files. Currently, attempting to play back these files with `ros2 bag play` results in an error due to mixed serialization formats. It would be useful if we could selectively play back compatible topics without encountering this issue.

### Current Behavior

Attempting playback with `ros2 bag play` returns the error:

> "Topics with different RMW serialization formats have been found. All topics must have the same serialization format."

Even if specific topics are specified with the `--topics` flag, `ros2 bag play` checks all topics, rather than just those specified.

### Desired Behavior

1. **Ignore topics with unknown serialization formats**: `ros2 bag play` should play back topics with compatible formats (like `cdr`) and skip topics with unknown formats.
2. **Selective topic playback**: When specific topics are listed with the `--topics` flag, only these topics should be validated and played back, ignoring the others.

### Completion Criteria

- The solution should allow for playback of compatible topics without errors.
- Topics specified with `--topics` should be the only ones validated for serialization format.

### Suggested Implementation

1. **Option 1**: Update `mcap_storage.cpp` to generate metadata only for topics with "known" formats, ensuring playback compatibility.
   - **Note**: This approach may need updates if new formats are introduced in the future.
   
2. **Option 2**: Modify the behavior of the `--topics` flag in `ros2 bag play` to validate only the specified topics.

### Testing

1. Generate an MCAP file with a mix of valid `cdr` messages and messages in an unknown format.
2. Confirm that `ros2 bag play`:
   - Ignores unknown formats and plays compatible data.
   - Plays only specified topics when using the `--topics` flag, without checking non-specified topics.

fujitatomoya:
IMO this sounds reasonable and more reliable behavior.

> Ignore topics with unknown serialization formats: ros2 bag play should play back topics with compatible formats (like cdr) and skip topics with unknown formats.

probably it should not ignore, but it can print the warning that this topic data is unknown serialization formats or something like that.

MichaelOrlov:
I agree that it is nice to have feature/behavior. 
As regards the implementation proposal. It should be generic and not specific to some storage plugin.
Therefore only option 2 is valid in this case.
> 2. Option 2: Modify the behavior of the --topics flag in ros2 bag play to validate only the specified topics.

PRs are welcome.

jespersmith:
[...]

PR #1851, opened 2024-11-04
[humble] Handle unsupported serialization formats in rosbag2's. 

jespersmith:
[Relevant issue](https://github.com/ros2/rosbag2/issues/1847)

This PR allows ros2 bag play to playback rosbags that have unsupported topics inside. 

The implementation finds the first topic with a supported format, and ignores all other topics. 

Currently based on humble, as that unblocks us immediately. Once the general approach is approved, we can rebase onto rolling.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
@@ -187,6 +187,8 @@ private:
   std::shared_ptr<SerializationFormatConverterFactoryInterface> converter_factory_{};
 
   bag_events::EventCallbackManager callback_manager_;
+
+  std::string storage_serialization_format;
 };
 
 }  // namespace readers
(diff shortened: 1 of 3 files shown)


jespersmith:
I got a branch for rolling, but haven't had the time to test it yet (still on Ubuntu 22.04/Humble).

https://github.com/jespersmith/rosbag2/tree/feature/support-unsupported-serialization-formats-rolling

MichaelOrlov (changes requested):
Hi @jespersmith, Thank you for the PR. 
Unfortunately, we can't do API/ABi breaking changes on the stable release like Humble.
Please refer to the https://acodersjourney.com/20-abi-breaking-changes/ for details about ABI breaking changes.
I also see that multiple tests failed during CI run.

Could you please create a new PR against Rolling branch for further review and considerations?

jespersmith:
@MichaelOrlov Thank you for the response. It's helpful to know it won't make it in humble, saves me doing unnecessary work.

I'll look into getting this into rolling. I might need some help with the test cases(specifically what their test goal is), I'll reach out if that is the case.

PR #1860, opened 2024-11-21
fixed compilation issues for the backport to humble of issue 1842

cmuehlbacher:
Fixed the compilation problems with using time_stamp instead of recv_timestamp because recv_timestamp is not available in humble and time_stamp is set to the time of reception.
Removed any metadata checks in the tests as there is no callback for the updated metadata.
Fixed issues with using fs::path (Humble does not use std::fs)
Fixed issues within the tests to create topics (topic metadata is different from rolling)

Hopefully this will support the backport.


cmuehlbacher:
This should fix the compilation problems in #1850

fujitatomoya (changes requested):
overall lgtm, but missing DCO: https://github.com/ros2/rosbag2/pull/1860/checks?check_run_id=33324746330

cmuehlbacher:
DCO is now fixed

PR #1861, opened 2024-11-21
Publish clock after delay is over and disable delay on next loops

### LINKED ISSUE #1858: Rosbag2 play --clock option ignores delay
## Description

Even if delay option is specified, clock option ignores delay and publishes `/clock`.
Also, after the time of delay elapses, clock time returns to the start time of bag.

### Expected Behavior

`/clock` is published after the delay time as well as the topic in the bag.

### Actual Behavior

`/clock` is published ignoring delay

## To Reproduce

1. Prepare a bag file that does not contain `/clock`
2. play bag with `--clock` and `--delay` 
3. echo `/clock` before delay time elapses

demo

```shell
# use bag in the sample dataset https://tier4.github.io/driving_log_replayer/quick_start/setup/
ros2 bag play input_bag --delay 10 --clock 200
# open another terminal
ros2 topic echo /clock
```

https://github.com/user-attachments/assets/bcb4bf8f-7fc6-4e7c-9773-4e9648a8fde6

demo bag start Apr  5 2022 15:07:31.594430720 (1649138851.594430720)
clock 1649138851 -> 1649138861 -> 1649138851(start time)

[clock_reverse_after_10sec.txt](https://github.com/user-attachments/files/17795034/clock_reverse_after_10sec.txt)

 - OS: Ubuntu 22.04
 - ROS 2 Distro: humble
 - Install Method: APT
 - Version: ros-humble-rosbag2/jammy,now 0.15.12-1jammy.20240730.153849

## Additional context

nicolaloi:
- Fixes #1858.

`clock_publish_timer` is now initialized with `autostart = false`, so it doesn't immediately start publishing the `/clock` topic. It is "activated" later with `clock_publish_timer->reset()` at the beginning of the playback thread, ensuring that the (optional) user-defined delay period has passed.

~Additionally, the `clock_` now starts in a paused state if a delay time is specified by the user. It resumes after the delay period has ended.~


fujitatomoya (approved):
lgtm with green CI.

@MichaelOrlov could you have 2nd review?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -488,6 +489,12 @@ bool PlayerImpl::play()
             RCLCPP_INFO_STREAM(owner_->get_logger(), "Sleep " << delay.nanoseconds() << " ns");
             std::chrono::nanoseconds delay_duration(delay.nanoseconds());
             std::this_thread::sleep_for(delay_duration);
+            if (!play_options_.start_paused) {
+              clock_->resume();
+            }
```
MichaelOrlov: I can predict that after some time someone will complain and raise a new issue that when using a player with `--loop` ~~and `--start-paused`~~ `delay` options and sending the pause service call request to the player somewhere at the end of the playback the request is ignored and player still will continue to playback from the beginning.
And we will have a hard time to debug this race condition, because of this `resume` call.
cc: @fujitatomoya @hayato-m126

MichaelOrlov (changes requested):
@nicolaloi Thanks for the PR. Overall looks good among one undefined behavior / race introducing with the
```
            if (!play_options_.start_paused) {
              clock_->resume();
            }
```
in Play method.

nicolaloi:
Ah yes, the clock will resume automatically at the next loop regardless of user requests. 

Then to not overcomplicate things, I would suggest removing the "pause clock until delay is over" feature from this PR, and keeping the "publish clock after the delay is over" feature only, which is the actual fix for issue #1858. Pausing the clock is not required to fix this specific issue since anyway the `clock_publish_timer_` was starting to publish after the clock was resumed.

I was pausing the clock as a comprehensive "inner" change for the `Player` class when in delayed status, but it is not required for the "outer" behavior of the timed published clock. If this "pause clock until delay is over" feature is desired, it can be a separate PR.

nicolaloi:
> I was pausing the clock as a comprehensive "inner" change for the `Player` class when in delayed status, but it is not required for the "outer" behavior of the timed published clock. If this "pause clock until delay is over" feature is desired, it can be a separate PR.

Removing 
```
if (!play_options_.start_paused) {
       clock_->resume();
}
```
to keep the "pause clock until delay is over" feature only, but if it is not okay I can reintegrate and try to fix this "pause clock until delay is over" feature in this PR too.

Extra change: now before sleeping for the delay period, the `clock_publish_timer_->cancel()` is called. This does not affect the first standard loop (since `clock_publish_timer_` has `autostart=false`), however for the subsequent (optional) loops this prevents the clock from being published until the player exits again the subsequent delayed periods (to be consistent with the first delay period).

fujitatomoya:
> keeping the "publish clock after the delay is over" feature only, which is the actual fix for issue https://github.com/ros2/rosbag2/issues/1858

i prefer this approach, one thing at a time. and easier to backport to downstream distros.

and the question here is that clock needs to be paused in delay duration? i am not quite sure about what state `delay` is...
in the implementation, it is `playing` state in `PlayerImpl::play` during duration period, but we cannot pause the sleep duration at all.

MichaelOrlov (approved):
@nicolaloi Thanks for addressing the comments from the review.
It looks good to me in the current form. However, I would move the delay out of the playback loop. IMO, there is no value in making a delay in the loop after we have finished the playback round.  
Delay needs for the purporse to be able to wait for other nodes (subscribers and other dependend nodes in the pipeline) to start and be discoverable. When we have run the first playback round all other nodes still up un running. Thereofre, there is no need to make a delay on a consequent rounds of replay.

nicolaloi:
Okay, I will incorporate the change to the delay loop behavior in this PR. 

Need to invoke `clock_->jump(starting_time_)`  before the loop, before starting the `clock_publish_timer`, but I am not sure if locking the mutex during the call is needed also in this pre-loop case.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -483,12 +483,19 @@ bool PlayerImpl::play()
...
       try {
+        if (delay > rclcpp::Duration(0, 0)) {
+          RCLCPP_INFO_STREAM(owner_->get_logger(), "Sleep " << delay.nanoseconds() << " ns");
+          std::chrono::nanoseconds delay_duration(delay.nanoseconds());
+          std::this_thread::sleep_for(delay_duration);
+        }
+        {
+          std::lock_guard<std::mutex> lk(reader_mutex_);
+          clock_->jump(starting_time_);
+        }
```
MichaelOrlov: We have already do `clock_->jump(..) ` just a few lines down. On line 502.
This one is redundant and should be removed IMO.

MichaelOrlov (commented):
@nicolaloi Thanks for the updates.
I think now it would be safe to init clock in pause mode in constructor and resume it after the delay. i.e.
```
 clock_ = std::make_unique<rosbag2_cpp::TimeControllerClock>(
      starting_time_, std::chrono::steady_clock::now,
      std::chrono::milliseconds{100},
      play_options_.start_paused || play_options_.delay > rclcpp::Duration(0, 0));
```

Also may I ask you in future do a separate commit after a round of review instead of push force the branch with the the same changed commit? 
We are squashing all commits before merging to the target branch after review anyway automatically. However, the separate commits allow to keep history of changes during review.

rosbag2_transport/src/rosbag2_transport/player.cpp
nicolaloi: I copied the jump before the loop because in this commit it needs to be called before starting the `clock_publisher_timer` to avoid publishing a non-monotonic clock. But yes if I revert to pausing the clock until the delay is over, then it is not needed.

nicolaloi: @MichaelOrlov I think there could be issues with pausing the clock and resuming it after the delay. If the user commands a direct "pause" of the clock (via keyboard control or service) during the delay period, the clock will resume automatically after the delay regardless of the user's intent (https://github.com/ros2/rosbag2/pull/1861#discussion_r1854934160). Another linked issue is if I remove the `clock_->jump` call before starting `clock_publisher_timer`, and the user resumes the clock during the delay period; then when the delay ends and the `clock_publisher_timer` starts, the clock publisher could publish "future" timestamps before going back to the `starting_time_` timestamp with the `clock_->jump` call inside the loop.

MichaelOrlov: @nicolaloi Sorry for the late response. After a few iterations, it became harder to follow the logic.
However, I think the solution would be if we will do the following:
1. Don't force to pause `clock_` in the constructor if the delay non-zero.
2. Don't call extra jump before do-while loop
3. Move 
```
        if (clock_publish_timer_ != nullptr) {
          clock_publish_timer_->reset();
        }
```
Inside do-while loop right after the `clock_->jump(starting_time_);`, but outside of the `reader_mutex`. It seems it will not be an issue if `clock_publish_timer_->reset();` will be inside do-while loop.

MichaelOrlov (changes requested):
@nicolaloi Could you please also rebase your branch on top of the latest rolling branch?
We made some changes. Now we support multiple readers.

rosbag2_transport/src/rosbag2_transport/player.cpp
nicolaloi: Ok I'll move the `clock_publish_timer_->reset();` inside the loop. I left it outside because if it is inside, I have noticed this `reset` call of the clock publisher can freeze it for something like 1/clock_freq seconds, and it's like "losing a beat" at the beginning of each loop. It is noticeable for low clock frequencies like 3 Hz (clock publisher will send the very first new clock message of the loop after ~0.6 seconds instead of ~0.3 seconds from the last clock message of the previous loop). But yes it is not a big problem, and for more common higher clock frequency it is not even noticeable.

fujitatomoya: if it resets the timer, i will call `rcl_timer_reset` that resets the sleep time based on the current time and period. that is the reason that we can see `losing a beat a bit` situation for each loop. but i do not think resetting the timer for each loop is bad design, in that way we can simulate the similar behavior both data playback and clock publisher for each loop.

MichaelOrlov: @nicolaloi As regards:
>  have noticed this reset call of the clock publisher can freeze it for something like 1/clock_freq seconds, and it's like "losing a beat" at the beginning of each loop. 

I am curious if calling `clock_publish_timer_->cancel();` before `clock_publish_timer_->reset();` will help with `"losing a beat"`?

MichaelOrlov (approved):
@nicolaloi Thanks for the PR and for addressing the review comments. LGTM now.

rosbag2_transport/src/rosbag2_transport/player.cpp
nicolaloi: Now I have tested calling `clock_publish_timer_->cancel()` before `clock_publish_timer_->reset()`, but it is still losing a beat.

MichaelOrlov:
Pulls: ros2/rosbag2#1861
Gist: https://gist.githubusercontent.com/MichaelOrlov/b0a96eb6a82f324d25e2f267ed68985d/raw/a84f0c1d7aca1838d0241fbdfc44c78d16a91c18/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14914
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16486/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1748/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22991/)

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Ok it is, what it is, will be a current design limitation.

MichaelOrlov:
Re-run Windows CI job since previous failure is unrelated 
```
ERROR: Failed building wheel for PyQt5-sip
```
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22993/)

MichaelOrlov:

PR #1867, opened 2024-11-25
[Rolling] Release 0.30.0

clalancette:
Once this is approved, I will squash-merge it, tag the resulting commit on the `rolling` branch, and do a `bloom-release` of `rosbag2`.


rosbag2/CHANGELOG.rst
```diff
@@ -5,6 +5,11 @@ Changelog for package rosbag2
 0.29.0 (2024-09-03)
 -------------------
 
+Forthcoming
+-----------
+* [Rolling] Release 0.28.1 (`#1791 <https://github.com/ros2/rosbag2/issues/1791>`_)
+* Contributors: Alejandro Hernández Cordero
+
```
MichaelOrlov: Curious how this appeared in the release notes?
IMO it shouldn't be there since we have already have record about 0.28.0 release a few lines down.

MichaelOrlov (commented):
@clalancette I found one unexpected record in the changelist. See my comment inline.

rosbag2/CHANGELOG.rst
clalancette: Bah, I missed this one.  Because the previous tags aren't on the `rolling` branch, the tool pulled all changes since 0.28.0 (the last tag that is actually on `rolling`).  I manually fixed the rest of them, but I missed this one.  I'll fix it.

clalancette: And now fixed in dfd08c7

MichaelOrlov: Oh. I see we have record for the 0.28.**0** release but we missed for the 0.28.**1**.

Anyway the 
```
Forthcoming
-----------
```
should be above the latest record
```
0.29.0 (2024-09-03)
-------------------
```

clalancette: Yeah, exactly.  It should all be fixed now.

MichaelOrlov: @clalancette Getting back to our discussion today on the PMC meeting if should keep RPs for release. It seems it is worth keeping it for a while. Since I keep catching some issues.

MichaelOrlov (approved):

PR #1869, opened 2024-11-25
Add PlayerClock::wakeup() to interrupt sleeping

christophebedard:
This adds a new `wakeup()` method to `PlayerClock`. It wakes up the clock, which is (only) useful if it is currently sleeping, i.e., in a `sleep_until()` call. Just calling `std::condition_variable::notify_all()` on the clock's condition variable wakes it up from its sleep (i.e., [`std::condition_variable::wait_for()`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for) and [`std::condition_variable::wait_until()`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_until)):

> The thread will be unblocked when [notify_all()](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all) or [notify_one()](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) is executed, or abs_time is reached. It may also be unblocked spuriously.

This can then replace the current workaround in `Player::stop()` to force-wakeup the clock, which is to resume the clock and then immediately pause it. Since we don't actually want to pause the clock, doing this may have unintended consequences, e.g., if the player is woken up and then thinks it's not in pause mode anymore.

I want to use this in #1848, but I thought a separate PR would be easier to review.

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
@@ -78,6 +78,19 @@ public:
   virtual bool sleep_until(rclcpp::Time time) = 0;
 
   /**
+   * \return whether the clock is currently sleeping.
+   */
+  ROSBAG2_CPP_PUBLIC
+  virtual bool is_sleeping() = 0;
+
+  /**
+   * \brief Wake up the clock if it is sleeping.
+   * \note This will wake any waiting `sleep_until`.
+   */
+  ROSBAG2_CPP_PUBLIC
+  virtual void wakeup() = 0;
+
+  /**
    * Change the rate of the flow of time for the clock.
    * \param rate new rate of clock playback
    * \bool false if rate is invalid for the clock implementation
--- a/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
@@ -123,6 +123,7 @@ public:
   std::condition_variable cv RCPPUTILS_TSA_GUARDED_BY(state_mutex);
   double rate RCPPUTILS_TSA_GUARDED_BY(state_mutex) = 1.0;
   bool paused RCPPUTILS_TSA_GUARDED_BY(state_mutex) = false;
+  bool is_sleeping RCPPUTILS_TSA_GUARDED_BY(state_mutex) = false;
   TimeReference reference RCPPUTILS_TSA_GUARDED_BY(state_mutex);
 };
 
@@ -165,12 +166,16 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
   {
     rcpputils::unique_lock<std::mutex> lock(impl_->state_mutex);
     if (impl_->paused) {
+      impl_->is_sleeping = true;
       impl_->cv.wait_for(lock, impl_->sleep_time_while_paused);
+      impl_->is_sleeping = false;
     } else {
       const auto steady_until = impl_->ros_to_steady(until);
       // wait only if necessary for performance
       if (steady_until > impl_->now_fn()) {
+        impl_->is_sleeping = true;
         impl_->cv.wait_until(lock, steady_until);
+        impl_->is_sleeping = false;
       }
     }
     if (impl_->paused) {
@@ -185,6 +190,17 @@ bool TimeControllerClock::sleep_until(rcutils_time_point_value_t until)
 bool TimeControllerClock::sleep_until(rclcpp::Time until)
 {
   return sleep_until(until.nanoseconds());
+}
+
+bool TimeControllerClock::is_sleeping()
+{
+  std::lock_guard<std::mutex> lock(impl_->state_mutex);
+  return impl_->is_sleeping;
+}
+
+void TimeControllerClock::wakeup()
+{
+  impl_->cv.notify_all();
 }
 
 bool TimeControllerClock::set_rate(double rate)
--- a/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
@@ -211,6 +211,32 @@ TEST_F(TimeControllerClockTest, interrupted_sleep_returns_false_immediately)
   EXPECT_FALSE(thread_sleep_result);
 }
 
+TEST_F(TimeControllerClockTest, wakeup_sleep_returns_immediately)
+{
+  // Use a bigger value for sleep_time_while_paused
+  auto sleep_time_while_paused = std::chrono::seconds(30);
+  rosbag2_cpp::TimeControllerClock clock(ros_start_time, now_fn, sleep_time_while_paused);
+  clock.pause();
+  EXPECT_FALSE(clock.is_sleeping());
+  auto steady_start = std::chrono::steady_clock::now();
+  std::atomic_bool thread_sleep_result{true};
+  auto sleep_long_thread = std::thread(
+    [&clock, &thread_sleep_result]() {
+      bool sleep_result = clock.sleep_until(clock.now() + RCUTILS_S_TO_NS(10));
+      thread_sleep_result.store(sleep_result);
+    });
+  // Wait for the thread to be sleeping for the purposes of this test
+  while (!clock.is_sleeping()) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+  clock.wakeup();
+  sleep_long_thread.join();
+  EXPECT_FALSE(clock.is_sleeping());
+  // Check that the clock slept less than the sleep_time_while_paused value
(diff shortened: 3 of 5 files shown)


rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
```diff
@@ -187,6 +187,12 @@ bool TimeControllerClock::sleep_until(rclcpp::Time until)
   return sleep_until(until.nanoseconds());
 }
 
+void TimeControllerClock::wakeup()
+{
+  std::lock_guard<std::mutex> lock(impl_->state_mutex);
+  impl_->cv.notify_all();
```
MichaelOrlov: Not 100% sure, but it seems we don't need to lock a mutex here.

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
```diff
@@ -211,6 +211,28 @@ TEST_F(TimeControllerClockTest, interrupted_sleep_returns_false_immediately)
...
+  clock.pause();
+  auto steady_start = std::chrono::steady_clock::now();
+  std::atomic_bool thread_sleep_result{true};
+  auto sleep_long_thread = std::thread(
+    [&clock, &thread_sleep_result]() {
+      bool sleep_result = clock.sleep_until(clock.now() + RCUTILS_S_TO_NS(10));
+      thread_sleep_result.store(sleep_result);
+    });
+  // Wait enough time for the thread to start sleeping before we try to wake it up
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
MichaelOrlov: Ahhh. Unfortunately, this sleep going to cause none-determinism and flakiness in the test.
To make it deterministic will need to add one more API. aka `bool TimeControllerClock::is_in_sleep()` and add one more private variable to detect a state when the timer is in sleep mode.

rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp
christophebedard: That does seem to be right: https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s). Doing so may be a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock, though some implementations recognize the pattern and do not attempt to wake up the thread that is notified under lock.

Removed: 9464665401d9ab05e7cd9980accafafa98b02803

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
christophebedard: I added `PlayerClock::is_sleeping()`: 190649f9dfc7949b683f96d2ba2dfa113cdd6dec

rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
```diff
@@ -77,6 +77,18 @@ class PlayerClock
   ROSBAG2_CPP_PUBLIC
   virtual bool sleep_until(rclcpp::Time time) = 0;
 
+  /**
+   * \return whether the clock is currently sleeping.
+   */
+  virtual bool is_sleeping() = 0;
```
MichaelOrlov, line 84: Missed `ROSBAG2_CPP_PUBLIC` ?

MichaelOrlov (commented):
@christophebedard Thanks for addressing my previous comments.
Now, overall looks good. However, it seems I found one missed `ROSBAG2_CPP_PUBLIC` qulifier.

rosbag2_cpp/include/rosbag2_cpp/clocks/player_clock.hpp
christophebedard: Fixed: 49a0e9369ff3aa39ed82d1c258a760e7df25d622

and rebased on `rolling`

MichaelOrlov (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#1869
Gist: https://gist.githubusercontent.com/MichaelOrlov/2c001ee0885d084435cdba39959dbf63/raw/87904b154664acc97ea6b9ca957eeab0ee8fe9b3/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14875
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22187/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16439/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1708/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22937/)

christophebedard:
As mentioned in https://github.com/ros2/rosbag2/pull/1848#discussion_r1861132705, we'll merge this after merging #1848.

christophebedard:
> As mentioned in [#1848 (comment)](https://github.com/ros2/rosbag2/pull/1848#discussion_r1861132705), we'll merge this after merging #1848.

It has been merged, so I rebased this PR on `rolling` and adapted this PR for changes from #1848, see 347585d502f22ca9186f6b69fef5df19e4caf8fb.

MichaelOrlov (approved):
LGTM.

BTW. we likely will be able to backport it to Jazzy if we will remove `virtual` keyword for newly added functions during backporting.

MichaelOrlov:
Pulls: ros2/rosbag2#1869
Gist: https://gist.githubusercontent.com/MichaelOrlov/9e91a8d8d83d3812cff842102f69b308/raw/87904b154664acc97ea6b9ca957eeab0ee8fe9b3/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14890
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22202/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16456/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22952/)

MichaelOrlov:
Two new warnings on Windows are in the `StaticSingleThreadedExecutor` and unrelated to the changes from this PR
```
'rclcpp::executors::StaticSingleThreadedExecutor::~StaticSingleThreadedExecutor': Use rclcpp::executors::SingleThreadedExecutor (compiling source file C:\ci\ws\src\ros2\rclcpp\rclcpp\test\rclcpp\executors\test_executors.cpp)
```

MichaelOrlov:

rosbag2_cpp/test/rosbag2_cpp/test_time_controller_clock.cpp
```diff
@@ -211,6 +211,32 @@ TEST_F(TimeControllerClockTest, interrupted_sleep_returns_false_immediately)
...
+  auto sleep_long_thread = std::thread(
+    [&clock, &thread_sleep_result]() {
+      bool sleep_result = clock.sleep_until(clock.now() + RCUTILS_S_TO_NS(10));
+      thread_sleep_result.store(sleep_result);
+    });
+  // Wait for the thread to be sleeping for the purposes of this test
+  while (!clock.is_sleeping()) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+  clock.wakeup();
```
fujitatomoya, line 232: the assumption for the dispatcher here is that sleep_long_thread should be already calling `cv.wait_for(lock, impl_->sleep_time_while_paused)` with this duration. but that is not just guaranteed by system (soooo unlikely but still), since `clock.wakeup();` does not acquire the mutex lock, this wakeup call could be missed and this test fails.

having mutex lock in `wakeup()` method solves this racy condition to provide consistent behavior for this particular case.

MichaelOrlov: @fujitatomoya No. The test is correct and there are no possibilities for the race conditions in it.
We wait for `sleep_long_thread` to go to sleep with calling `cv.wait_for(lock, impl_->sleep_time_while_paused)` by calling
```cpp
  // Wait for the thread to be sleeping for the purposes of this test
  while (!clock.is_sleeping()) {
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
  }
```
The `clock.is_sleeping()` implemented as 
```
bool TimeControllerClock::is_sleeping()
{
  std::lock_guard<std::mutex> lock(impl_->state_mutex);
  return impl_->is_sleeping;
}
```
Locking the same mutex. Therefore `is_sleeping()` will return true only when `cv.wait_for(lock, impl_->sleep_time_while_paused)` will go to the sleep state and unlock the `impl_->state_mutex`.

fujitatomoya: Ah i see. you are right, while look for `is_sleeping` makes sure that clock is sleep state. thank you for letting me know that!

PR #1884, opened 2024-12-16
It should be testing humble, and should be using jammy.

clalancette:
- This PR fixes GitHub workflow to use Humble base images instead of Rolling.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ jobs:
     name: ament_${{ matrix.linter }}
     runs-on: ubuntu-latest
     container:
-      image: rostooling/setup-ros-docker:ubuntu-focal-ros-rolling-ros-base-latest
+      image: rostooling/setup-ros-docker:ubuntu-jammy-ros-humble-ros-base-latest
     strategy:
       fail-fast: false
       matrix:
@@ -18,7 +18,7 @@ jobs:
     - uses: ros-tooling/action-ros-lint@v0.1
       with:
         linter: ${{ matrix.linter }}
-        distribution: rolling
+        distribution: humble
         package-name: |
             ros2bag
             rosbag2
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -38,17 +38,17 @@ jobs:
     name: ament_${{ matrix.linter }}
     runs-on: ubuntu-latest
     container:
-      image: rostooling/setup-ros-docker:ubuntu-focal-ros-rolling-ros-base-latest
+      image: rostooling/setup-ros-docker:ubuntu-jammy-ros-humble-ros-base-latest
     strategy:
       fail-fast: false
       matrix:
-          linter: [cppcheck, cpplint, uncrustify]
```
fujitatomoya, line 45: Why cppcheck needs to be removed from here?

fujitatomoya (commented):
@clalancette thank you very much for the fix.

so technically we should do the same thing to `jazzy` branch as well?

.github/workflows/lint.yml
clalancette: It's because cppcheck always fails on Jammy/Humble.  The problem is that cppcheck 2.X is *very* slow, so slow that it makes our CI times ridiculous.  So starting with Ubuntu 22.04, we disabled it, and we no longer run it in CI (honestly we should probably just remove it, but that's a different change).

fujitatomoya (approved):
lgtm with github workflow.

@MichaelOrlov @christophfroehlich either of you can take a look? thanks in advance.

christophebedard (approved):
I assume you meant to tag me :grin:

fujitatomoya:
@christophebedard @christophfroehlich thanks for both of you! and sorry 😓

fujitatomoya:

https://build.ros2.org/job/Hpr__rosbag2__ubuntu_jammy_amd64/303/testReport/junit/rosbag2_transport/RecordIntegrationTestFixture/write_split_callback_is_called/ is obviously unrelated to this PR, these tests are just unstable?

CC: @MichaelOrlov

MichaelOrlov (approved):
LGTM

MichaelOrlov:
@fujitatomoya Yes, changes are unrelated to the test failures I belive.

MichaelOrlov:

fujitatomoya:
> so technically we should do the same thing to jazzy branch as well?

ah i see that jazzy does not have problem like this.

clalancette:
> @fujitatomoya Yes, changes are unrelated to the test failures I belive.

Yes, agreed.  I can't see how the github workflow would affect the Hpr job.  So with that, and the approvals, I'm going to go ahead and merge this.  Thanks everyone for reviewing!

Issue #1886, opened 2024-12-23
Snapshot mode throws away transient local topics when the buffer is full (in other words, tf_static is lost)


doisyg:
## Description
Not completely a bug, but due to the current design of `MessageCacheCircularBuffer`, all incoming topics are using the same buffer regardless of their QoS.
Hence, when the buffer is full, when we start erasing older messages, we could be erasing messages with transient local that would have been useful to keep.
An example is the `/tf_static` topic, that will be typically received by the snapshot bag recorder at the very beginning and hence throw away as soon as the buffer is full. And this topic will be missing when replaying the bag.

### Expected Behavior
Support the `/tf_static` use case, i.e. save and replay `/tf_static` properly (and other transient local messages)

### Actual Behavior
Cannot replay `/tf_static`

## To Reproduce

1. Start a snapshot bag record with transient local topics, like `/tf_static`.
2. Wait long enough for the buffer to be full.
3. Trigger the snapshot.
4. Stop everything and try to replay the bag. `/tf_static` will be missing.

## System (please complete the following information)
 - Same behavior under Iron and Jazzy

## Additional context
I will push a draft PR to discuss how to fix this asap.
Edit: here https://github.com/ros2/rosbag2/pull/1887

tonynajjar:
@MichaelOrlov We implemented a practical solution here: https://github.com/botsandus/rosbag2/pull/2
I don't know enough about rosbag2 so it might not be the cleanest implementation but I'm happy to hear your feedback about it and hopefully contribute it back upstream. I think it goes in the direction of your comment [here](https://github.com/ros2/rosbag2/pull/1887#issuecomment-3680982212).

> For snapshot mode, the straightforward approach would be to keep "latched" messages in a separate buffer and dump it to the disk before the circular buffer with other messages

doisyg:
@tonynajjar maybe a PR your practical solution from: https://github.com/botsandus/rosbag2/pull/2 to upstream/this repo for visibility ?
@emersonknapp fyi @tonynajjar's workaround is what we ended up using to have our transient local topics (tf_static and robot_description) in our snapshot bags.

tonynajjar:
Needed to adapt it for rolling but here it is: https://github.com/ros2/rosbag2/pull/2279

MichaelOrlov:
[...]

Issue #1159, opened 2022-11-11
Support `--repeat-latched` in ROS 2


amacneil:
## Description
ROS 1 Noetic added the `--repeat-latched` feature (https://github.com/ros/ros_comm/pull/1850) to repeat latched topics at the start of each bag split. 

As far as I can tell, rosbag2 does not have this feature. Is there any reason we should not add it?

MichaelOrlov:
@amacneil Yes we have the reason to not do this for the sake to have deterministic replay.
This is duplicate of the https://github.com/ros2/rosbag2/pull/1051  please see my detailed explanation in https://github.com/ros2/rosbag2/pull/1051#pullrequestreview-1059614100

amacneil:
I'll admit I am not super familiar with how tf_static and `transient local` works in ROS 2 (I am more familiar with latched topics in ROS 1). https://github.com/ros2/ros2/issues/464 seems to have some details.

We have users who would like to open one file from a split bag and have the transforms work correctly. Your comment in that issue seems to be based on the fact that you don't have a way in rosbag2 to "play" a split bag file, but we do support opening a single file in Foxglove. I think it would be good to get more feedback from users on whether they need this feature.

MichaelOrlov:
If users want something it doesn't means that we should jump in and implement what they want without considering other cases and drawbacks. 

My main point not in the case that rosbag2 doesn't support playing a split bag rather that we need to have determinism in replay. Basically we shall not inject or copy messages which we didn't receive on the wires in any cases.

We have a lot another cases when transient local durability uses for instance camera images or LIDAR streams - and what? We should copy all messages from camera topic from previous file to the next file when doing split?

If someone want to repeat `tf_static` messages published once at the beginning before playing back bag after split, one can use rosbag2 play with duration option and topic filtering for playing back just messages from `tf_static` then right after playback normally bag which was recorded after split. 
Another option is to use `ros2 bag convert` to create new bag where will be messages from `tf_static` from first bag and the rest will be content from bag after split.

jhurliman:
> We have a lot another cases when transient local durability uses for instance camera images or LIDAR streams - and what? We should copy all messages from camera topic from previous file to the next file when doing split?

My understanding of transient local durability is there is a user-configured history size, so it would only copy all messages from the previous file for a topic if the previous file was very short or the history size was unusually large. The default history size is 10, and 5 for the sensor QoS profile.

MichaelOrlov:
[...]

PR #1887, opened 2024-12-23
Handle transient local messages in snapshot mode when the buffer is full

doisyg:
Draft to discuss ways of solving https://github.com/ros2/rosbag2/issues/1886

Naive solution which throws away the first message that is not transient local when the buffer is full, as long as the number of messages at the beginning of the buffer accounts for less than 10% of the total number of messages in the buffer.

That will not solve completely the `/tf_static` use case if other transient local messages are recorded and periodically published: at some point there will be more than 10% of transient local messages in the buffer, and the first ones (i.e `/tf_static`) will be removed.

Interested in debating and suggestions on how to solve this properly :
- Different buffers for each transient local topic ? But then how to control their max size ?
- A buffer dedicated to `/tf_static`? But then loss of generality.
- Something else ?

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/circular_message_cache.hpp
@@ -20,6 +20,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <unordered_map>
 
 #include "rcpputils/thread_safety_annotations.hpp"
 
@@ -28,6 +29,7 @@
 #include "rosbag2_cpp/cache/cache_buffer_interface.hpp"
 #include "rosbag2_cpp/visibility_control.hpp"
 
+#include "rosbag2_storage/rosbag2_storage/bag_metadata.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
 
 // This is necessary because of using stl types here. It is completely safe, because
@@ -52,7 +54,9 @@ class ROSBAG2_CPP_PUBLIC CircularMessageCache
   : public MessageCacheInterface
 {
 public:
-  explicit CircularMessageCache(size_t max_buffer_size);
+  explicit CircularMessageCache(
+    size_t max_buffer_size, const std::unordered_map<std::string,
+    rosbag2_storage::TopicInformation> & topics_names_to_info);
 
   ~CircularMessageCache() override;
 
--- a/rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp
@@ -18,9 +18,12 @@
 #include <deque>
 #include <memory>
 #include <vector>
+#include <unordered_map>
+#include <string>
 
 #include "rosbag2_cpp/visibility_control.hpp"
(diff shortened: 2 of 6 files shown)


doisyg:
(PR non functional for Rolling, I was originally targeting Iron, will fix)

MichaelOrlov:
@doisyg The transient local messages usually appears somewhere at the beginning of the recording like `/tf_static`. 
It seems make sense to consider adding an option for repeating some number of the transient local messages for each new recording bag rather than throwing away some percentage of the messages.  
I have formulated a design proposal in another task https://github.com/ros2/rosbag2/issues/1159#issuecomment-2183281437. Please take it into consideration.

tonynajjar:
> @doisyg The transient local messages usually appears somewhere at the beginning of the recording like `/tf_static`. It seems make sense to consider adding an option for repeating some number of the transient local messages for each new recording bag rather than throwing away some percentage of the messages. I have formulated a design proposal in another task [#1159 (comment)](https://github.com/ros2/rosbag2/issues/1159#issuecomment-2183281437). Please take it into consideration.

@MichaelOrlov are you sure that your proposal would also solve what this PR is trying to solve? The bug described in the ticket has nothing to do with splitting, you can reproduce by:

1- Starting a a bag in snapshot with a relatively low --max-cache-size. It should record `tf_static` along with other topics
2- Publish a transient local topic, like `/tf_static`
3- Publish enough so that the bag size exceeds --max-cache-size
4- Call the snapshot service
5- See that /tf_static topic is not in the bag

Do you have any proposal to fix or workaround that?

MichaelOrlov:
@tonynajjar As from my other comment https://github.com/ros2/rosbag2/issues/1159#issuecomment-2902812644
The additional CLI options would be 
```bash
$ ros2 bag record -a --repeat-latched /map=1 /tf_static=5
```

For snapshot mode, the straightforward approach would be to keep "latched" messages in a separate buffer and dump it to the disk before the circular buffer with other messages. Trying to reduce the circular buffer on a fly when latched mesages arrives seems is non-trivial. The feature would be more consistent with another feature for snapshot https://github.com/ros2/rosbag2/pull/1844 to limit the bag file by duration rather than by bag size.

MichaelOrlov (changes requested):
Hi @tonynajjar, I've finally found a time for review and more thorough consideration about how to gracefully handle the repetition of the transient-local messages during recording in general.

First of all, I would like to say you a big thank you for the prototyping and your attempt to solve this complicated task.
Although I found that your design approach is not suitable for the cases when we need to do a regular bag split, and is not capable of handling some corner cases like message duplicates in the snapshot buffer and transient-local buffer as well as some performance aspects and concurrency.

I've gone ahead and spent some substantial time creating a design document https://github.com/ros2/rosbag2/pull/2327 for this transient-local (“latched”) messages in Rosbag2 recorder feature. I would encourage you to read it and participate in the review.

tonynajjar:
Hi @MichaelOrlov , thanks for jumping on that! No worries about my PR, I'm pretty sure it's not the cleanest one conceptually and architecturally - I just needed something quick to fix our issue and kickstart the discussion with a practical first step.
I'll take a look next week at what you came up with!

MichaelOrlov:
- Closing thi PR in favor of the 
- https://github.com/ros2/rosbag2/pull/2385
- https://github.com/ros2/rosbag2/pull/2386
- https://github.com/ros2/rosbag2/pull/2387

PR #1889, opened 2024-12-27
Increase timeout to 180s for test_rosbag2_record_end_to_end

Yadunund:
This PR increases timeout to 180s for `test_rosbag2_record_end_to_end`.

Running the test with `rmw_zenoh` takes a bit longer since it takes a couple seconds to initialize the ROS context. With this change, I was able to get the test to pass.


fujitatomoya (approved):
i think that is okay. @MichaelOrlov what do you think?

MichaelOrlov (approved):
If it helps to pass tests on CI with rmw_zenoh, I don't mind increasing test's timeout.

PR #1894, opened 2025-01-09
Bugfix: Recorder discovery does not restart after being stopped

### LINKED ISSUE #1893: Recorder discovery does not restart after being stopped
The discovery is currently not restarted properly in `record()` when previously stopped through `stop()`. This implies that no new topics are subscribed to on subsequent calls to resume and no topics at all when simulation time is used due to how these topics are initialized.

The cause is the following comparison, which needs to be inverted:
https://github.com/ros2/rosbag2/blob/ffd8c7da4ebd5aca9a1a5d860c9a10db28576e31/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L440-L449
The stop_discovery_ atomic should then also be initialized to true in order for the first start_discovery call to succeed.

oysstu:
This is a bug fix for a situation when the discovery in Recorder does not restart after being stopped.

Repeated calls to start_discovery should not start additional topics discovery tasks without stop_discovery calls.

- This was originally reported in #1893, as discovery task would not be spawned if a previous stop_discovery call was made. This only works currently because the atomic is initialized to false, thus making the "correct" choice on the first call after construction.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -143,7 +143,7 @@ private:
 
   std::mutex start_stop_transition_mutex_;
   std::mutex discovery_mutex_;
-  std::atomic<bool> stop_discovery_ = true;
+  std::atomic<bool> discovery_running_ = false;
   std::atomic_uchar paused_ = 0;
   std::atomic<bool> in_recording_ = false;
   std::shared_ptr<KeyboardHandler> keyboard_handler_;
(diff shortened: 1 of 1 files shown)


oysstu:
The can_record_again_after_stop is failing because it records one more message than expected. If discovery now works between the record calls, it makes sense that it records something else as well, e.g., rosout?

MichaelOrlov (changes requested):
@oysstu thank you for PR and attempt to fix a wrong behavior.  
The naming is hard! It seems the root cause of the issue is in unclear naming fo the variable `stop_discovery_`.  
The logic would be more readable and much easy to understand if `stop_discovery_` would be renamed to the `discovery_runing_`.

Could you please do such rename in the code and change conditions accordingly?

oysstu:
I've made the change, that should clarify the logic a lot. ~Any thoughts on the failing tests?~

Edit: seemed to pass tests this time

fujitatomoya (approved):
lgtm with green CI (backport required for jazzy)

fujitatomoya:
Pulls: ros2/rosbag2#1894
Gist: https://gist.githubusercontent.com/fujitatomoya/93c6ebe95670806a2726bdeb3399771a/raw/a9f1000e30b41bab61c31daa3415b6801c810d1f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15057
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22373/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16636/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1905/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23166/)

MichaelOrlov (approved):
LGTM

MichaelOrlov:

PR #1896, opened 2025-01-13
Add options to conditionally stop the recording

### LINKED ISSUE #1857: Specify recording time for a ros2 bag
## Description

I would like to have a flag that allows to specify the time length of a record of a ros2 bag.

## Related Issues
I cannot specify the time length when I want to record a bag file. I need it since I will run the record from a bash file and the command `timeout 10 ros2 bag record /topic_name` doesn't work. If I run the same command directly from the terminal works.
 
## Completion Criteria

- have a flag that allow to specify the recording lenght

nicolaloi:
Resolves #1857.

Add `--max-recording-duration`, `--max-recording-size`, `--max-recording-messages` CLI options to conditionally stop the recording when it reaches a specified duration, size, or number of messages. The conditions consider all the possible split bags.

A new event message is published when the recording stops.

This is a first working implementation trying to follow the `split_bag` processing and event generation in [sequential_writer](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp).

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,18 @@ Messages written to the bag will use the latest received value of `/clock` for t
 Note: Until the first `/clock` message is received, the recorder will not write any messages.
 Before that message is received, the time is 0, which leads to a significant time jump once simulation time begins, making the bag essentially unplayable if messages are written first with time 0 and then time N from `/clock`.
 
+#### Recording with termination conditions
+
+rosbag2 can be configured to stop recording after a certain size, duration or number of messages is reached (considering **all** the bag files of the current recording). By default the recording will continue until an external signal is stopping it, but this can be changed using the CLI options.
+
+_Stopping by size_: `ros2 bag record -a --max-recording-size 100000` will stop the recording after it becomes greater than 100 kilobytes. This option defaults to `0`, which means the recording will not stop based on size.
+
+_Stopping by duration_: `ros2 bag record -a --max-recording-duration 9000` will stop the recording after `9000` seconds. This option defaults to `0`, which means the recording will not stop based on duration.
+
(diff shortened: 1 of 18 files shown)


nicolaloi:
Now the main computations of this PR are in [sequential_writer](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp), since it contains the data required to process the conditional stopping (eg current size, duration, message count). However I think the logic should be moved to [recorder.cpp](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_transport/src/rosbag2_transport/recorder.cpp) (before/after the `writer_->write` within the [lambda functions](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L568-L618)), but given the hierarchy [recorder.cpp](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_transport/src/rosbag2_transport/recorder.cpp) ->  [writer.cpp](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_cpp/src/rosbag2_cpp/writer.cpp) -> [sequential_writer.cpp](https://github.com/ros2/rosbag2/blob/0.30.0/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp), sharing the required data "up" to `recorder.cpp` requires to change the various intermediate interfaces.

@MichaelOrlov @fujitatomoya do you have suggestions or new ideas?

PR #1903, opened 2025-01-28
Upstream quality changes from Apex.AI part 1

MichaelOrlov:
This PR consolidates fixes related to the code quality and stability after backporting Rosbag2 PRs to Apex.AI. **PART 1**

### List of changes:
1. Fix for memory leaks in tests
   Details: Need to use custom deleter when releasing rcl serialized buffer to a shared pointer.
1. Move wait_for_playback_to_start() to public section
   - Rationale: Need to use in another package
   -  Also added timeout parameter for the wait_for_playback_to_start(..) to make it non-blocking call optionally and return boolean value.
1. Gracefully handle exceptions from on-play callbacks in player
  Rationale: Avoid abnormal termination in case of unhandled exceptions happening in callbacks.
1. Address Axivion warnings in rosbag2_py
   Description: 
   1. MisraC++2023-8.1.1 [required]: _transport.cpp:286 This variable shall not be implicitly captured in a lambda expression.; Field: []
   2. MisraC++2023-8.2.3 [required]: _transport.cpp:50 Cast removes const qualification; Field: [const char*->char*]
1. Fix the clang16 warning that the timestamp comparator is not initialized
1. Make sure to start with a clean subscriptions list in Recorder::record()
    Rationale: It could be some residual subscriptions in the list in case if the previous call to Recorder::record() failed for some reason and Recorder::stop() wasn't called.
1. Call discovery_future_.get() if it's ready in the `stop_discovery()`
1. Use absolute path instead of relative path in the `InfoEndToEndTestFixture`
1. Don't keep trying to publish the next message if we fail the first time. The `Player::play_next()` should return false right away if, for some reason, failed to publish the current message.

### DIFF
--- a/rosbag2_py/src/rosbag2_py/_transport.cpp
+++ b/rosbag2_py/src/rosbag2_py/_transport.cpp
@@ -47,13 +47,13 @@ public:
     std::for_each(
       arguments_.begin(), arguments_.end(),
       [this](const std::string & arg) {
-        pointers_.push_back(const_cast<char *>(arg.c_str()));
+        pointers_.push_back(arg.c_str());
       }
     );
     pointers_.push_back(nullptr);
   }
 
-  char ** argv()
+  const char ** argv()
   {
     return arguments_.empty() ? nullptr : pointers_.data();
   }
@@ -65,7 +65,7 @@ public:
 
 private:
   std::vector<std::string> arguments_;
-  std::vector<char *> pointers_;
+  std::vector<const char *> pointers_;
 };
 
 rclcpp::QoS qos_from_handle(const py::handle source)
@@ -283,7 +283,7 @@ protected:
       player->play();
 
       auto wait_for_exit_thread = std::thread(
-        [&]() {
+        [this, player]() {
           std::unique_lock<std::mutex> lock(wait_for_exit_mutex_);
           wait_for_exit_cv_.wait(lock, [] {return rosbag2_py::Player::exit_.load();});
           player->stop();
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -374,10 +374,10 @@ private:
 
   std::shared_ptr<PlayerServiceClientManager> player_service_client_manager_;
 
-  BagMessageComparator get_bag_message_comparator(const MessageOrder & order);
+  static BagMessageComparator get_bag_message_comparator(const MessageOrder & order);
 
   /// Comparator for SerializedBagMessageSharedPtr to order chronologically by recv_timestamp.
-  struct
+  static inline const struct
   {
     bool operator()(
       const rosbag2_storage::SerializedBagMessageSharedPtr & l,
@@ -388,7 +388,7 @@ private:
   } bag_message_chronological_recv_timestamp_comparator;
 
   /// Comparator for SerializedBagMessageSharedPtr to order chronologically by send_timestamp.
-  struct
+  static inline const struct
   {
     bool operator()(
       const rosbag2_storage::SerializedBagMessageSharedPtr & l,
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -1054,14 +1054,11 @@ void PlayerImpl::play_messages_from_queue()
         // If we tried to publish because of play_next(), jump the clock
         if (play_next_.load()) {
           clock_->jump(get_message_order_timestamp(message_ptr));
-          // If we successfully played next, notify that we're done, otherwise keep trying
-          if (message_published) {
-            play_next_ = false;
-            std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
-            finished_play_next_ = true;
-            play_next_result_ = true;
-            finished_play_next_cv_.notify_all();
-          }
+          play_next_ = false;
+          std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
+          finished_play_next_ = true;
+          play_next_result_ = message_published;
+          finished_play_next_cv_.notify_all();
         }
       }
       message_ptr = take_next_message_from_queue();
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -1364,16 +1364,26 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr 
   auto pub_iter = publishers_.find(message->topic_name);
   if (pub_iter != publishers_.end()) {
     bool message_published = false;
+    bool pre_callbacks_failed = true;
     try {
       // Calling on play message pre-callbacks
       run_play_msg_pre_callbacks(message);
-
-      pub_iter->second->publish(rclcpp::SerializedMessage(*message->serialized_data));
-      message_published = true;
+      pre_callbacks_failed = false;
     } catch (const std::exception & e) {
-      RCLCPP_ERROR_STREAM(
-        owner_->get_logger(), "Failed to publish message on '" << message->topic_name <<
+      RCLCPP_ERROR_STREAM(owner_->get_logger(),
+        "Failed to call on play message pre-callback on '" << message->topic_name <<
+        "' topic. \nError: " << e.what());
+    }
+
+    if (!pre_callbacks_failed) {
+      try {
+        pub_iter->second->publish(rclcpp::SerializedMessage(*message->serialized_data));
+        message_published = true;
+      } catch (const std::exception & e) {
+        RCLCPP_ERROR_STREAM(owner_->get_logger(),
+          "Failed to publish message on '" << message->topic_name <<
           "' topic. \nError: " << e.what());
+      }
     }
 
     try {
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -586,7 +586,7 @@ bool PlayerImpl::play()
             ready_to_play_from_queue_cv_.notify_all();
           }
         } while (rclcpp::ok() && !stop_playback_ && play_options_.loop);
-      } catch (std::runtime_error & e) {
+      } catch (const std::exception & e) {
         RCLCPP_ERROR(owner_->get_logger(), "Failed to play: %s", e.what());
         load_storage_content_ = false;
         if (storage_loading_future_.valid()) {storage_loading_future_.get();}
@@ -1357,10 +1357,11 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr 
 {
   auto pub_iter = publishers_.find(message->topic_name);
   if (pub_iter != publishers_.end()) {
-    // Calling on play message pre-callbacks
-    run_play_msg_pre_callbacks(message);
     bool message_published = false;
     try {
+      // Calling on play message pre-callbacks
+      run_play_msg_pre_callbacks(message);
+
       pub_iter->second->publish(rclcpp::SerializedMessage(*message->serialized_data));
       message_published = true;
     } catch (const std::exception & e) {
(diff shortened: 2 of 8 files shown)


MichaelOrlov:
There are a few failures on CI that need to be addressed. Turning this PR back to draft.  
Sorry it was premature to ask for review.

MichaelOrlov:
Ok. Now the CI is green and this PR is ready for review.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -196,9 +196,11 @@ class PlayerImpl
   /// \return Shared pointer to the inner clock_publisher
   rclcpp::Publisher<rosgraph_msgs::msg::Clock>::SharedPtr get_clock_publisher();
 
-  /// \brief Blocks and wait on condition variable until first message will be taken from read
-  /// queue
-  void wait_for_playback_to_start();
+  /// \brief Waits on the condition variable until first message will be taken from read queue
+  /// @param timeout Maximum time in the fraction of seconds to wait for player to start.
```
christophebedard, line 200: Why "fraction of seconds"? We shouldn't need to care about whether `std::chrono::duration` is seconds, nanoseconds, etc. Users could just provide `std::chrono::nanoseconds(10)`, right?

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1357,10 +1357,11 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr
 {
   auto pub_iter = publishers_.find(message->topic_name);
   if (pub_iter != publishers_.end()) {
-    // Calling on play message pre-callbacks
-    run_play_msg_pre_callbacks(message);
     bool message_published = false;
     try {
+      // Calling on play message pre-callbacks
+      run_play_msg_pre_callbacks(message);
```
christophebedard, line 1370: This should be in its own `try`/`catch` and log a `Failed to call on play message pre-callback on ...` error, otherwise an exception in `run_play_msg_pre_callbacks()` will be reported as a publish failure.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard, line 1: You could drop the references to Axivion or MISRA in the commit message/body when contributing here :grin:

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -267,6 +267,7 @@ void RecorderImpl::record()
     throw std::runtime_error("No serialization format specified!");
   }
 
+  subscriptions_.clear();
```
christophebedard, line 271: Is this worth adding a test for?

christophebedard (changes requested):
Minor changes.

christophebedard:
Also, you don't really need to mention/frame the title of this PR as a "sync with Apex.AI"

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: I'm referring to c5e49176d85c04252db76d91d163c1c6c6fc3c40

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: I don't think so. It is more like defensive programming. No real issue so far.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: Why not? There is an explanation with reasonig why the changes were made. Otherwise, someone can start arguing or revert it.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: The idea was in a specific logic:
if the preconditions in terms of the pre-callbacks are not met, i.e. if an exception is thrown in one of the pre-callbacks,  then the publish method shouldn't be called.  
As far as I recall, we rely on this logic in one of our downstream packages, which uses the player directly.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Well, because this function is intended to be used and uses from unit tests.  
No one is going to use it with `wait_for_playback_to_start(std::chrono::nanoseconds(10))` because it doesn't make sense. The real wait time on CI ranges from hundreds of milliseconds to a few seconds, and I don't want to overcomplicate it with templates.
Therefore, a fraction of a second was chosen as the most suitable dimension for timeout.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: Good point, but this logic could still be achieved using separate `try`/`catch` blocks (e.g., using a simple `bool`) so that each one has its own error message. Something like this:

```cpp
bool do_publish = false;
try {
  run_play_msg_pre_callbacks(message);
  do_publish = true;
} catch (e) {
  // Log pre-callback error
}

if (do_publish) {
    try {
      // publish()
    } catch (e) {
      // Log publish error
  }
}
```

It has the same behaviour (skips publishing if pre-callback throws) but logs a helpful error message that directly mentions that a pre-callback threw.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: Even if no one is going to use `std::chrono::nanoseconds()`, the input type is `std::chrono::duration`, not `std::chrono::nanseconds` or `std::chrono::seconds`, so the method documentation should not treat it as "seconds." It should just treat it as a "duration," which could be anything (s/ms/ns/etc.), even if the caller really only wants to use seconds.

If it _only_ makes sense to provide a timeout in seconds, then the input type should be `std::chrono::seconds`.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: I simply meant that the changes have merit on their own; no need to mention where they're coming from. But this is just a comment, feel free to ignore.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: According to the [Google](https://www.google.com/search?q=std%3A%3Achrono%3A%3Aduration%3Cdouble%3E&sca_esv=c7459735fc04b658&ei=RqqaZ5TaEvKgur8P_cTe6Qk&ved=0ahUKEwiUlb2F_JuLAxVykO4BHX2iN50Q4dUDCBM&uact=5&oq=std%3A%3Achrono%3A%3Aduration%3Cdouble%3E&gs_lp=Egxnd3Mtd2l6LXNlcnAiHXN0ZDo6Y2hyb25vOjpkdXJhdGlvbjxkb3VibGU-SNsnUJwPWPsgcAJ4AJABAJgBLKAB8wKqAQE5uAEDyAEA-AEBmAICoAIGwgIKEAAYsAMY1gQYR5gDAIgGAZAGCJIHATKgB-kH&sclient=gws-wiz-serp), ChatGPT and cppreference 
**std::chrono::duration<double> represents a time duration in seconds with floating-point precision.**
You can specify different time units using a second template parameter (e.g., std::milli, std::micro).
```
    std::chrono::duration<double> sec(2.5); // Represents 2.5 seconds

    // Convert to milliseconds
    std::chrono::duration<double, std::milli> millisec = sec;
    std::cout << "Duration: " << millisec.count() << " milliseconds" << std::endl;
```
I wouldn't change anything in this regard. IMO code and description are correct.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Okay, I will rewrite it this way and will add a unit test for the case when `play_next` returns false if `pre_callback` throws.

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: It's not about being able to create a `std::chrono::duration<double>` object from a `double` value in seconds. That's just because the `std::chrono::duration` constructor accepts a `double` value, and, by default, it's in seconds. It's about being able to pass any kind of time value. For example, you can still do this:

```cpp
wait_for_playback_to_start(std::chrono::hours(1.0));
```

In which case it doesn't make sense to mention "seconds" in the documentation. That's it.

christophebedard: anyway, if you prefer it this way, then alright

MichaelOrlov: @christophebedard No, it doesn't work this way.
The `wait_for_playback_to_start(std::chrono::hours(1.0));` will not compile.
I have tried it from unit tests. It can't convert to the underlying type.
The build fails with an error message like:
```
/rosbag2_transport/test_play.cpp:987:60: error: no matching function for call to ‘std::chrono::duration<long int, std::ratio<3600> >::duration(double)’
  987 |   player->wait_for_playback_to_start(std::chrono::hours(1.0));
      |                                                            ^
```

However, I found mistakes in wording in the doxygen section for the `wait_for_playback_to_start`. I simply copy-pasted the description from the other `wait_for_playback_to_finish(..)` method and forgot to adopt it to start. 
Fixed it in the latest commit.

MichaelOrlov:
@christophebedard Could you please give a one more round of review? 
I've addressed review comments and also added a few more fixes
> 7. Call discovery_future_.get() if it's ready in the stop_discovery()
> 8. Use absolute path instead of relative path in the InfoEndToEndTestFixture
> 9. Don't keep trying to publish the next message if we fail the first time. The Player::play_next() should return false right away if, for some reason, failed to publish the current message.

rosbag2_transport/src/rosbag2_transport/player.cpp
clalancette: > The `wait_for_playback_to_start(std::chrono::hours(1.0));` will not compile.

That's true, but I think it is only true because I think that the type is over-specified to be a `std::chrono::duration<double>`.  If you changed the API to be either:

```
bool wait_for_playback_to_start(std::chrono::milliseconds timeout = std::chrono::milliseconds(-1))
```

or

```
bool wait_for_playback_to_start(std::chrono::duration<std::chrono::milliseconds::rep, std::milli> timeout = std::chrono::milliseconds(-1))
```

then you'd be able to pass arbitrary times in.  In other words, you probably don't need fractions of a tick, since you probably don't want to sleep for shorter than a millisecond.  In that case, you don't need the duration specified with a double, and you can instead use the "standard" types.

christophebedard: > No, it doesn't work this way.
The `wait_for_playback_to_start(std::chrono::hours(1.0));` will not compile.

This compiles:

```cpp
  function(std::chrono::duration<double>(1.0));
  function(std::chrono::duration<double>(1.5));
  function(std::chrono::seconds(2));
  function(std::chrono::minutes(2));
  function(std::chrono::hours(2));
```

See: https://godbolt.org/z/vnc96cEr1

christophebedard (approved):
If you really want to proceed as-is with the `wait_for_playback_to_start()` docs, this looks good to me with green CI.

MichaelOrlov:
Pulls: ros2/rosbag2#1903
Gist: https://gist.githubusercontent.com/MichaelOrlov/09606b1727cd650b8ad81063eca5060f/raw/2aab4e80b4711be3bfd625549652c96db827c46d/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_transport rosbag2_py rosbag2_storage_mcap rosbag2_test_common rosbag2_tests
TEST args:  --packages-above rosbag2_transport rosbag2_py rosbag2_storage_mcap rosbag2_test_common rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15128
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22460/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16711/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1986/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23247/)

MichaelOrlov:
@christophebedard Thank you for review.

MichaelOrlov:

PR #1924, opened 2025-02-27
Upstream quality changes from Apex.AI part-2

MichaelOrlov:
This PR consolidates fixes related to the code quality and stability after backporting Rosbag2 PRs to Apex.AI. PART 2

List of changes:

1. Fix for `record_if_topic_list_service_list_and_all_are_specified`
  Details: Flush output from Python print when printing warnings. Some tests rely on the warning output to be in the std::cout. Without flushing, the warnings may not be captured by tests.
1. Fix for flaky `test_play_services`
  Details: Increase timeouts for service calls in rosbag2_transport `test_play_services` tests
1. Address unused variable warning on the clang16 build
1. Mark `MessageProducer<T>` as final to address clang16 warnings "destructor called on non-final 'msg_utils::MessageProducer<T>' that has virtual functions but non-virtual destructor"
1. Fix linter warnings in the `message_definition_encoding.md`
1. Mitigate flakiness in `record_all_include_unpublished_false`
  Details: Increase timeout when waiting to stop discovery at least to 0.5 sec. The `record_all_include_unpublished_false_includes_later_published` tends to be flaky due to the insufficient timeout for waiting to stop the discovery thread in the recorder.
1. Use `RESOURCES_PATH` from envar in the `test_transport.py`
  Rationale: Fix to avoid failure with remote test execution and to be consistent with other similar tests.
1. Reduce runtime for the `PlayerTestFixture.playing_respects_delay`
1. Fix for flaky `play_respect_messages_timing_after_play_next`.
 Details: Take the start timestamp before calling resume for the player.
1. Add debug info to flaky `test_play_cancel`
1. Print warnings instead of errors for not-found type definitions
1. Adjust `test_rosbag2_storage_api.cpp` for bounded strings.
  Details:  Reduce the size of each `std_string_msg` to fit in a 256-byte boundary.
1. Address CPPLint warnings in the `rosbag2_storage_sqlite3`
1. Fix CPP_LINT warnings about missing include headers
1. Address flakiness in `test_play_services`.
  Details: Increase timeout for `expect_messages(..)`. Also add `player_->wait_for_playback_to_start();` in `setup_player()`
1. Fix for incorrectly handling exit code in `wait_until_completion(..)`

### DIFF
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -242,7 +242,7 @@ def check_necessary_argument(args):
 def validate_parsed_arguments(args, uri) -> str:
     if args.topics_positional:
         print(print_warn('Positional "topics" argument deprecated. '
-                         'Please use optional "--topics" argument instead.'))
+                         'Please use optional "--topics" argument instead.'), flush=True)
         args.topics = args.topics_positional
 
     if not check_necessary_argument(args):
@@ -274,17 +274,17 @@ def validate_parsed_arguments(args, uri) -> str:
                            'or --regex')
 
     if (args.all or args.all_services) and args.services:
-        print(print_warn('--all or --all-services will override --services'))
+        print(print_warn('--all or --all-services will override --services'), flush=True)
 
     if (args.all or args.all_topics) and args.topics:
-        print(print_warn('--all or --all-topics will override --topics'))
+        print(print_warn('--all or --all-topics will override --topics'), flush=True)
 
     if (args.all or args.all_actions) and args.actions:
         print(print_warn('--all or --all-actions will override --actions'))
 
     if (args.all or args.all_topics or args.all_services or args.all_actions) and args.regex:
         print(print_warn('--all, --all-topics --all-services or --all-actions will override '
-                         '--regex'))
+                         '--regex'),  flush=True)
 
     if os.path.isdir(uri):
         return print_error("Output folder '{}' already exists.".format(uri))
--- a/rosbag2_cpp/src/rosbag2_cpp/info.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/info.cpp
@@ -17,8 +17,11 @@
 #include <filesystem>
 #include <unordered_map>
 #include <unordered_set>
+#include <vector>
+#include <memory>
 #include <stdexcept>
 #include <string>
+#include <utility>
 
 #include "rmw/rmw.h"
 #include "rosidl_typesupport_cpp/message_type_support.hpp"
--- a/rosbag2_cpp/src/rosbag2_cpp/info.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/info.cpp
@@ -12,26 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "rosbag2_cpp/info.hpp"
-
 #include <filesystem>
-#include <unordered_map>
-#include <unordered_set>
-#include <vector>
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <unordered_map>
+#include <unordered_set>
 #include <utility>
+#include <vector>
 
 #include "rmw/rmw.h"
-#include "rosidl_typesupport_cpp/message_type_support.hpp"
-#include "service_msgs/msg/service_event_info.hpp"
-
 #include "rosbag2_cpp/action_utils.hpp"
+#include "rosbag2_cpp/info.hpp"
 #include "rosbag2_cpp/service_utils.hpp"
-
 #include "rosbag2_storage/metadata_io.hpp"
 #include "rosbag2_storage/storage_factory.hpp"
+#include "rosidl_typesupport_cpp/message_type_support.hpp"
+#include "service_msgs/msg/service_event_info.hpp"
 
 namespace fs = std::filesystem;
 
--- a/rosbag2_cpp/test/rosbag2_cpp/test_service_utils.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_service_utils.cpp
@@ -16,6 +16,7 @@
 #include <string>
 #include <tuple>
 #include <vector>
+#include <utility>
 
 #include "rosbag2_cpp/service_utils.hpp"
 #include "rosbag2_test_common/memory_management.hpp"
--- a/rosbag2_cpp/test/rosbag2_cpp/test_service_utils.cpp
+++ b/rosbag2_cpp/test/rosbag2_cpp/test_service_utils.cpp
@@ -15,8 +15,8 @@
 
 #include <string>
 #include <tuple>
-#include <vector>
 #include <utility>
+#include <vector>
 
 #include "rosbag2_cpp/service_utils.hpp"
 #include "rosbag2_test_common/memory_management.hpp"
--- a/rosbag2_py/src/rosbag2_py/_info.cpp
+++ b/rosbag2_py/src/rosbag2_py/_info.cpp
@@ -15,6 +15,7 @@
 #include <iostream>
 #include <memory>
 #include <string>
+#include <vector>
 #include <algorithm>
 
 #include "info_sorting_method.hpp"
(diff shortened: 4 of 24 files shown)


rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -163,10 +163,10 @@ TEST_F(PlayerTestFixture, playing_rate_negative)
 
 TEST_F(PlayerTestFixture, playing_respects_delay)
 {
-  rclcpp::Duration delay_margin(1, 0);
+  rclcpp::Duration delay_margin(0, static_cast<uint32_t>(RCUTILS_S_TO_NS(0.5)));
```
Timple: For readability I would include chrono `s` and just write `delay_margin{0.5s}`.

MichaelOrlov: Thanks for the comment and review. Fixed in the https://github.com/ros2/rosbag2/pull/1924/commits/0022efc4568ae6deef1736c881a8c123f481a0e7

rosbag2_cpp/src/rosbag2_cpp/info.cpp
```diff
@@ -17,8 +17,11 @@
 #include <filesystem>
 #include <unordered_map>
 #include <unordered_set>
+#include <vector>
+#include <memory>
 #include <stdexcept>
 #include <string>
+#include <utility>
```
christophebedard: Very minor, but these could be ordered alphabetically. Same for all other files.

ros2bag/ros2bag/verb/record.py
```diff
@@ -242,7 +242,7 @@ def check_necessary_argument(args):
 def validate_parsed_arguments(args, uri) -> str:
     if args.topics_positional:
         print(print_warn('Positional "topics" argument deprecated. '
-                         'Please use optional "--topics" argument instead.'))
+                         'Please use optional "--topics" argument instead.'), flush=True)
```
christophebedard, line 245: We usually set the `PYTHONUNBUFFERED` env var to `1` for tests to get this behaviour. This way we don't need to flush (which can be expensive) for normal uses of the tool.

Could we instead set `PYTHONUNBUFFERED=1` in the relevant tests?

MichaelOrlov: > Could we instead set PYTHONUNBUFFERED=1 in the relevant tests?

I don't think so. The failure is happening in the integration tests when we are runing `ros2 bag record` in a separate process from the C++ code.

MichaelOrlov: The failure happens in the `record_if_topic_list_service_list_action_list_and_all_are_specified` test

rosbag2_cpp/src/rosbag2_cpp/info.cpp
MichaelOrlov: Okay, fixed in the https://github.com/ros2/rosbag2/pull/1924/commits/b6984d8f711d151113d2914c7a8aa3c5a1f4b548

ros2bag/ros2bag/verb/record.py
christophebedard: You can set `PYTHONUNBUFFERED=1` in the child process after forking here, but before executing the command: https://github.com/ros2/rosbag2/blob/dcaca47599611a378d20fc897fdfc14df8813c92/rosbag2_test_common/include/rosbag2_test_common/process_execution_helpers_unix.hpp#L83-L96

If you use [`execvpe()`](https://linux.die.net/man/3/execvpe), it should be easy to set an environment variable.

I'm suggesting this because you wouldn't have to set `flush=True` for _all_ prints, which seems more scalable to me. It would also only flush in tests, and not for normal uses. Your choice.

rosbag2_cpp/src/rosbag2_cpp/info.cpp
christophebedard: Other files also have unordered includes.

christophebedard: And I didn't necessarily ask to order the existing includes, just the ones you're adding here

rosbag2_cpp/test/rosbag2_cpp/test_service_utils.cpp
```diff
@@ -16,6 +16,7 @@
 #include <string>
 #include <tuple>
 #include <vector>
+#include <utility>
```
christophebedard:
```suggestion
#include <utility>
#include <vector>
```

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
```diff
@@ -14,11 +14,13 @@
...
 
-#include <atomic>
-#include <chrono>
 #include <filesystem>
+#include <chrono>
+#include <vector>
+#include <memory>
 #include <stdexcept>
 #include <string>
+#include <utility>
```
christophebedard:
```suggestion
#include <chrono>
#include <filesystem>
#include <memory>
#include <stdexcept>
#include <string>
#include <utility>
#include <vector>
```

rosbag2_py/src/rosbag2_py/_info.cpp
```diff
@@ -15,6 +15,7 @@
 #include <iostream>
 #include <memory>
 #include <string>
+#include <vector>
 #include <algorithm>
```
christophebedard:
```suggestion
#include <algorithm>
#include <iostream>
#include <memory>
#include <string>
#include <vector>
```

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -29,6 +29,7 @@
 
 #include "rosbag2_transport_test_fixture.hpp"
 
+using namespace std::chrono_literals;  // NOLINT
```
Timple, line 32: Just a tip, but instead of the NOLINT you can also just include only the used operator:

```c++
using std::chrono_literals::operator""s;
```

edit: although I see now this is in line with the `using`s below.

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Sorry, It might have other side effects for other tests if we will set `PYTHONUNBUFFERED=1` by default for any new process runing from tests.
I would prefer to fix the root cause of the problem in the source file rather than the consequences in the tests.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
MichaelOrlov: We may need to use other chrono literals in the same file further on. This is a common practice to put `// NOLINT` there.

christophebedard (approved):
The Rpr job doesn't seem happy

MichaelOrlov:

MichaelOrlov:
Pulls: ros2/rosbag2#1924
Gist: https://gist.githubusercontent.com/MichaelOrlov/fa4c49a68ce6a7e317d8dd4ddb435b04/raw/0b6dd10403dabed361b5810ba1f3b948e5073b1e/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_examples_py rosbag2_py rosbag2_performance_benchmarking rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_examples_py rosbag2_py rosbag2_performance_benchmarking rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15772
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23239/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17457/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2707/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24119/)

MichaelOrlov:

PR #1927, opened 2025-02-28
Address windows warnings in the progress bar class

MichaelOrlov:
This PR addresses MSVS compiler warnings found on nightly CI jobs.
- Closes https://github.com/ros2/rosbag2/issues/1926

PR #1932, opened 2025-03-04
address dll-interface linkage warnings.

fujitatomoya:
address https://github.com/ros2/rosbag2/pull/1836#discussion_r1980171818

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player_progress_bar.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player_progress_bar.hpp
@@ -37,7 +37,7 @@
 
 namespace rosbag2_transport
 {
-class PlayerProgressBarImpl;
+class ROSBAG2_TRANSPORT_PUBLIC PlayerProgressBarImpl;
 
 class ROSBAG2_TRANSPORT_PUBLIC PlayerProgressBar
 {


rosbag2_transport/include/rosbag2_transport/player_progress_bar.hpp
```diff
@@ -37,7 +37,7 @@
 
 namespace rosbag2_transport
 {
-class PlayerProgressBarImpl;
+class ROSBAG2_TRANSPORT_PUBLIC PlayerProgressBarImpl;
```
fujitatomoya, line 40: or maybe `PlayerProgressBarImpl` can be implemented with `PlayerProgressBar`.

fujitatomoya:
@nicolaloi @MichaelOrlov friendly ping.

MichaelOrlov:
- Closing this PR. Since warning has already been fixed in the https://github.com/ros2/rosbag2/pull/1927

rosbag2_transport/include/rosbag2_transport/player_progress_bar.hpp
MichaelOrlov: Nop. Not this way. 
Pimpl shouldn't be exposed to the outside of the DLL boundary by design. Otherwise, it doesn't make sense in the Pimpl pattern.

fujitatomoya: yeah, i thought that too.

> PlayerProgressBarImpl can be implemented with PlayerProgressBar.

how about this like `TimeControllerClockImpl`, so probably we do not need https://github.com/ros2/rosbag2/pull/1927 windows special handling???

MichaelOrlov: Yeah, we have such a pattern when putting implementation details from the pimpl to the actual class implementation. 
We have it in multiple places, including `rosbag2_transport::Player` and `rosbag2_transport::Recorder`.  
I found this approach to be more cumbersome, and the API wrapper cluttering implementation details. 
The source file became very big and very often duplicated methods with the same names, making it difficult to search and navigate to the right place.  And all of this just to ~~satisfy~~ the Microsoft compilator (workaround a bug in compilator)! 
Please note there is no real issue with those warning. The microsoft compilator just not smart enough to see that Impl class not uses directly in the external API.

I decided to try spliting them this time to see if further support will be easy.

fujitatomoya: I see, thanks for sharing thoughts!

PR #1933, opened 2025-03-07
Do not crash on failed writes but skip the file

Timple:
In the current implementation when a write cannot be performed, the recorder node crashes.

But since we have it running in a composable container with other  nodes, this takes down much more than it should.

So I propose to handle this a bit more graceful.

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -333,10 +333,8 @@ void SequentialWriter::switch_to_next_storage()
     metadata_.relative_file_paths.size());
   storage_ = storage_factory_->open_read_write(storage_options_);
   if (!storage_) {
-    std::stringstream errmsg;
-    errmsg << "Failed to rollover bagfile to new file: \"" << storage_options_.uri << "\"!";
-
-    throw std::runtime_error(errmsg.str());
+    ROSBAG2_CPP_LOG_ERROR_STREAM("Failed to rollover bagfile to new file: \"" <<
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
@Timple 

i am not sure if this is the right thing to do.

1st, after printing this error, following code will generate the segmentation fault to access NULL shared pointer. what the difference does it make?
this situation is pretty much similar with `cannot open the storage`, IMO this is really exceptional situation and it cannot proceed anymore. this case, i say exception should be generated.

CC: @MichaelOrlov what do you think?

Timple:
Ah, the *how* is completely up to debate. I'm not that familiar with the codebase.

But crashing nodes bringing down composed containers is an issue.

MichaelOrlov:
@Timple @fujitatomoya Unable to open storage is a fatal unrecoverable error, and an exception should be thrown.  
However,  probably we shouldn't crash other nodes running in the container.  
The reason why crash happened is because an unhandled exception.  
Before adding composable recorder/player it was Okay to not handle unrecoverable exceptions because in this case our CLI recorder or player would exit with the error code. That is expected behavior.

However, for the composable recorder and player, it would be better to intercept all exceptions on the upper layer and terminate only the failed node.
I am curious if there is an option to do this on the composition container layer.   
BTW it seems this is a generic problem for composable nodes. The crash may cause by any other node's unhandled exception not only by the Rosbag2 nodes.

I think we **shall** keep baseline behavior when terminating CLI `ros2 bag record` by unrecoverable exception and returning the error code.

MichaelOrlov:
- Closing as will not be merged. Because unrecoverable exceptions shall be handled on upper layers. Preferably inside a composition container.

PR #1936, opened 2025-03-12
fix: QoS incompatibilites are not expected with rmw_zenoh_cpp

YuanYuYuan:
This PR aims to address https://github.com/ros2/rmw_zenoh/issues/510

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -23,7 +23,6 @@
 #include <utility>
 
 #include "rclcpp/rclcpp.hpp"
-#include "rmw/rmw.h"
 
 #include "rosbag2_test_common/subscription_manager.hpp"
 #include "rosbag2_test_common/service_manager.hpp"
@@ -1245,14 +1244,11 @@ TEST_F(RosBag2PlayQosOverrideTestFixture, topic_qos_profiles_overridden_incompat
     topic_name_, num_msgs_to_wait_for_, qos_request);
   play_options_.topic_qos_profile_overrides = topic_qos_profile_overrides;
 
-  // QoS incompatibilities are not expected with rmw_zenoh_cpp.
-  bool expect_timeout = true;
-  if (std::strcmp(rmw_get_implementation_identifier(), "rmw_zenoh_cpp") == 0) {
-    expect_timeout = false;
-  }
+  // A timeout is expected if QoS is incompatible.
+  bool expect_timeout = (rclcpp::qos_check_compatible(
+    qos_playback_override, qos_request).compatibility != rclcpp::QoSCompatibility::Ok);
 
-  // Fails if it doesn't time out
-  play_and_wait(timeout, expect_timeout /* expect timeout */);
+  play_and_wait(timeout, expect_timeout);
 }
 
 TEST_F(RosBag2PlayQosOverrideTestFixture, playback_uses_recorded_transient_local_profile)
--- a/rosbag2_transport/test/rosbag2_transport/test_play.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play.cpp
@@ -23,6 +23,7 @@
 #include <utility>
 
 #include "rclcpp/rclcpp.hpp"
+#include "rmw/rmw.h"
 
 #include "rosbag2_test_common/subscription_manager.hpp"
 #include "rosbag2_test_common/service_manager.hpp"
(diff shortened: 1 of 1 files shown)


MichaelOrlov:
@YuanYuYuan, Could you please provide more context to explain why this fix is needed and why the test fails with `rmw_zenoh`?

YuanYuYuan:
> @YuanYuYuan, Could you please provide more context to explain why this fix is needed and why the test fails with `rmw_zenoh`?

Hi @MichaelOrlov! This is a fix similar to what we did in rclcpp. https://github.com/ros2/rclcpp/pull/2639. The reason is rmw_zenoh_cpp is not a DDS-backboned RMW. The assumption of QoS incompatibility of reliability doesn't mean the subscriber receives no message. On the other hand, the user can still use [rmw_qos_profile_check_compatible](https://github.com/ros2/rmw_zenoh/blob/c7a80c285540b16730ec9fd7ccf5c9edf8b1900a/rmw_zenoh_cpp/src/rmw_qos.cpp?plain=1#L74-L79) to check the compatibility.

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -1244,8 +1245,14 @@ TEST_F(RosBag2PlayQosOverrideTestFixture, topic_qos_profiles_overridden_incompat
     topic_name_, num_msgs_to_wait_for_, qos_request);
   play_options_.topic_qos_profile_overrides = topic_qos_profile_overrides;
 
+  // QoS incompatibilities are not expected with rmw_zenoh_cpp.
+  bool expect_timeout = true;
+  if (std::strcmp(rmw_get_implementation_identifier(), "rmw_zenoh_cpp") == 0) {
```
ahcorde: what about using here `qos_check_compatible` ? In other test we do somthing like:

```cpp
  if (qos_check_compatible(qos_profile_publisher,
    qos_profile_subscription).compatibility != rclcpp::QoSCompatibility::Ok)
  {
```

ahcorde (changes requested):

rosbag2_transport/test/rosbag2_transport/test_play.cpp
fujitatomoya: this sounds good idea. this API can be called on anywhere before `play_and_wait`.

instead of putting the hard code with specific implementation in the test to be skipped, maybe asking rmw implementation if that is compatible or incompatible qos configuration, and based on that we change the expectation resut. this can be more flexible and maintainability since it does not depend on rmw specific code.

YuanYuYuan: Yes, this is a better modification. Updated in [21d59f5](https://github.com/ros2/rosbag2/pull/1936/commits/21d59f5111dcea46a1e77bd2a6c59615f3ab554c)

MichaelOrlov: Sorry, I have a bit opposed opinion. If `rmw_zenoh` do not expect any QoS incompatibilities it doesn't make sense to run this test at all.   
In my opinion, it would be much cleaner if we would skip this test for the `rmw_zenoh` with an explanation like 
```
// If the RMW implementation is rmw_zenoh_cpp, we do not expect any QoS incompatibilities.
```
rather than trying to work around expectations in this test.

fujitatomoya: IMO, incompatible or not, it needs to call the API to the rmw implementation since that is not what rosbag2 or clients interface can decide. in the case of other rmw implementation support or deprecating rmw implementation in the future, we do not need to come back here to maintain. i think, that is better for the maintenance and integrity of the test. but i am okay with @MichaelOrlov opinion, if anybody agrees with that 👍

YuanYuYuan: I agree with @fujitatomoya. Communicating through the API is better for long-term maintenance. With this change, the test now verifies RMWs that require QoS compatibility. It still makes sense to me.

ahcorde (approved):
LGTM, @MichaelOrlov what do you think ?

MichaelOrlov:
I wouldn't insist. 
Overall, I prefer to adhere to the concept that tests should test what they are intended to be tested. In a regular V&V the tests should cover the requirements and code branches.  
And in my opinion need to avoid chnaging expectations in the tests based on the behavior of other SW component. Otherwise we increase risks for test to be false positive or false negative.

ahcorde:
Pulls: ros2/rosbag2#1936
Gist: https://gist.githubusercontent.com/ahcorde/f7b2b4cc8e42938777a7e65dd012e749/raw/7daf9188694920b1e965ad27c4448be9327603f5/ros2.repos
BUILD args:  --packages-up-to rosbag2_transport  --packages-above-and-dependencies rosbag2_transport
TEST args:  --packages-select rosbag2_transport  --packages-above rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15377
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22732/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16974/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2249/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23607/)

fujitatomoya (approved):
lgtm, thanks for the discussion everyone!

fujitatomoya (approved):
lgtm, thanks for the discussion everyone!

ahcorde:

PR #1928, opened 2025-02-28
[Design] Add design doc for actions recording and replay

Barry-Xu-2018:
The design for recording and replay action.

PR #1955, opened 2025-04-02
Implement rosbag2 action play

Barry-Xu-2018:
Address https://github.com/ros-infrastructure/rep/pull/405

Design https://github.com/ros2/rosbag2/pull/1928

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,15 @@ Options:
 
 For more options, run with `--help`.
 
+#### Playback action messages as action client
+
+If you want Rosbag2 to replay recorded action messages in the role of an action client, you need to specify the --send-actions-as-client parameter.
+```
+$ ros2 bag play --send-actions-as-client <bag>
+```
+Rosbag2 will send recorded goal request, cancel request and result request to action server.  
+For more information, please refer to https://github.com/ros2/rosbag2/blob/rolling/docs/design/rosbag2_record_replay_action.md.
+
 #### Controlling playback via services
 
 The Rosbag2 player provides the following services for remote control, which can be called via `ros2 service` commandline or from your nodes,
--- a/README.md
+++ b/README.md
@@ -167,6 +167,8 @@ Options:
   Space-delimited list of topics to play.
 * `--services`:
   Space-delimited list of services to play.
+* `--actions`:
+  Space-delimited list of actions to play.
 * `-e,--regex`:
   Play only topics and services matches with regular expression.
 * `-x,--exclude-regex`:
@@ -175,6 +177,8 @@ Options:
   Space-delimited list of topics not to play.
 * `--exclude-services`:
   Space-delimited list of services not to play.
+* `--exclude-actions`:
+  Space-delimited list of actions not to play.
 * `--message-order {received,sent}`:
   The reference to use for bag message chronological ordering.
   Choices: reception timestamp (`received`), publication timestamp (`sent`).
@@ -282,12 +286,15 @@ output_bags:
   topic_types: []
   all_services: false
   services: []
+  all_actions: false
+  actions: []
   rmw_serialization_format: ""  # defaults to using the format of the input topic
   regex: ""
   exclude_regex: ""
   exclude_topics: []
   exclude_topic_types: []
   exclude_services: []
+  exclude_actions: []
   compression_mode: ""
   compression_format: ""
   compression_queue_size: 1
--- a/ros2bag/ros2bag/verb/burst.py
+++ b/ros2bag/ros2bag/verb/burst.py
@@ -49,6 +49,10 @@ class BurstVerb(VerbExtension):
             help='services to replay, separated by space. At least one service needs to be '
                  "specified. If this parameter isn\'t specified, all services will be replayed.")
         parser.add_argument(
+            '--actions', type=str, default=[], nargs='+',
+            help='actions to replay, separated by space. At least one action needs to be '
+                 "specified. If this parameter isn\'t specified, all actions will be replayed.")
+        parser.add_argument(
             '--qos-profile-overrides-path', type=FileType('r'),
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.')
         parser.add_argument(
@@ -97,6 +101,8 @@ class BurstVerb(VerbExtension):
         play_options.topics_to_filter = args.topics
         # Convert service name to service event topic name
         play_options.services_to_filter = convert_service_to_service_event_topic(args.services)
+        # Covert action name to action related topic names
+        play_options.actions_to_filter = args.actions
         play_options.topic_qos_profile_overrides = qos_profile_overrides
         play_options.loop = False
         play_options.topic_remapping_options = topic_remapping
--- a/ros2bag/ros2bag/verb/burst.py
+++ b/ros2bag/ros2bag/verb/burst.py
@@ -101,7 +101,6 @@ class BurstVerb(VerbExtension):
         play_options.topics_to_filter = args.topics
         # Convert service name to service event topic name
         play_options.services_to_filter = convert_service_to_service_event_topic(args.services)
-        # Covert action name to action related topic names
         play_options.actions_to_filter = args.actions
         play_options.topic_qos_profile_overrides = qos_profile_overrides
         play_options.loop = False
--- a/ros2bag/ros2bag/verb/play.py
+++ b/ros2bag/ros2bag/verb/play.py
@@ -61,17 +61,23 @@ class PlayVerb(VerbExtension):
             '--services', type=str, default=[], metavar='service', nargs='+',
             help='Space-delimited list of services to play.')
         parser.add_argument(
+            '--actions', type=str, default=[], metavar='action', nargs='+',
+            help='Space-delimited list of actions to play.')
+        parser.add_argument(
             '-e', '--regex', default='',
-            help='Play only topics and services matches with regular expression.')
+            help='Play only topics, services and actions matches with regular expression.')
         parser.add_argument(
             '-x', '--exclude-regex', default='',
-            help='regular expressions to exclude topics and services from replay.')
+            help='regular expressions to exclude topics, services and actions from replay.')
         parser.add_argument(
             '--exclude-topics', type=str, default=[], metavar='topic', nargs='+',
             help='Space-delimited list of topics not to play.')
         parser.add_argument(
             '--exclude-services', type=str, default=[], metavar='service', nargs='+',
             help='Space-delimited list of services not to play.')
+        parser.add_argument(
+            '--exclude-actions', type=str, default=[], metavar='action', nargs='+',
+            help='Space-delimited list of actions not to play.')
         parser.add_argument(
             '--qos-profile-overrides-path', type=FileType('r'),
             help='Path to a yaml file defining overrides of the QoS profile for specific topics.')
@@ -163,11 +169,19 @@ class PlayVerb(VerbExtension):
             '--publish-service-requests', action='store_true', default=False,
             help='Publish recorded service requests instead of recorded service events')
         parser.add_argument(
+            '--send-actions-as-client', action='store_true', default=False,
+            help='Send the send_goal request, cancel_goal request, and get_result request '
+                 'respectively based on the recorded send_goal, cancel_goal, and get_result '
+                 'event messages. Note that the messages from action\'s "status topic" and '
+                 '"feedback topic" will not be sent because they are expected to be sent from '
+                 'the action server side.')
+        parser.add_argument(
             '--service-requests-source', default='service_introspection',
             choices=['service_introspection', 'client_introspection'],
             help='Determine the source of the service requests to be replayed. This option only '
-                 'makes sense if the "--publish-service-requests" option is set. By default,'
-                 ' the service requests replaying from recorded service introspection message.')
+                 'makes sense if the "--publish-service-requests" or "--send-actions-as-client" '
+                 'option is set. By default, the service requests replaying from recorded '
+                 'service introspection message.')
         parser.add_argument(
             '--message-order', default='received',
             choices=['received', 'sent'],
@@ -271,12 +285,17 @@ class PlayVerb(VerbExtension):
         # Convert service name to service event topic name
         play_options.services_to_filter = convert_service_to_service_event_topic(args.services)
 
+        play_options.actions_to_filter = args.actions
+
         play_options.regex_to_filter = args.regex
 
         play_options.exclude_regex_to_filter = args.exclude_regex
 
         play_options.exclude_service_events_to_filter = \
             convert_service_to_service_event_topic(args.exclude_services)
+
+        play_options.exclude_actions_to_filter = \
+            args.exclude_actions if args.exclude_actions else []
 
         play_options.topic_qos_profile_overrides = qos_profile_overrides
         play_options.loop = args.loop
@@ -299,6 +318,7 @@ class PlayVerb(VerbExtension):
         play_options.wait_acked_timeout = args.wait_for_all_acked
         play_options.disable_loan_message = args.disable_loan_message
         play_options.publish_service_requests = args.publish_service_requests
+        play_options.send_actions_as_client = args.send_actions_as_client
         if not args.service_requests_source or \
                 args.service_requests_source == 'service_introspection':
             play_options.service_requests_source = ServiceRequestsSource.SERVICE_INTROSPECTION
--- a/rosbag2_storage_mcap/src/mcap_storage.cpp
+++ b/rosbag2_storage_mcap/src/mcap_storage.cpp
@@ -235,8 +235,6 @@ private:
                  rosbag2_storage::storage_interfaces::IOFlag io_flag,
                  const std::string & storage_config_uri);
 
-  static bool is_topic_name_a_service_event(const std::string_view topic_name);
-
   /// \brief Check if topic match with the selection criteria by the white list or regex during
   /// data read.
   /// \details There is assumption that by default all topics shall be selected if none of the
@@ -566,7 +564,9 @@ bool MCAPStorage::read_and_enqueue_message()
   return true;
 }
 
-bool MCAPStorage::is_topic_name_a_service_event(const std::string_view topic_name)
+namespace
+{
+bool is_topic_name_a_service_event(const std::string_view topic_name)
 {
   // The origin definition is RCL_SERVICE_INTROSPECTION_TOPIC_POSTFIX
   static const char * service_event_topic_postfix = "/_service_event";
@@ -581,6 +581,24 @@ bool MCAPStorage::is_topic_name_a_service_event(const std::string_view topic_nam
   }
   return true;
 }
+
+// The postfix of the action internal topics and service event topics
+const std::vector<std::string> ActionInterfacePostfix = {
+  "/_action/send_goal/_service_event", "/_action/cancel_goal/_service_event",
+  "/_action/get_result/_service_event", "/_action/feedback", "/_action/status"};
+
+bool is_topic_name_related_to_action(const std::string_view topic_name)
+{
+  for (const auto & postfix : ActionInterfacePostfix) {
+    if (topic_name.length() > postfix.length() &&
+        topic_name.compare(topic_name.length() - postfix.length(), postfix.length(), postfix) ==
+          0) {
+      return true;
+    }
+  }
+  return false;
+}
+}  // namespace
 
 template <typename T>
 bool MCAPStorage::is_topic_selected_by_white_list_or_regex(const std::string_view topic_name,
@@ -644,18 +662,25 @@ void MCAPStorage::reset_iterator()
   options.readOrder = read_order_;
 
   auto topic_filter = [this](std::string_view topic) {
-    bool topic_a_service_event = is_topic_name_a_service_event(topic);
+    std::vector<std::string> * include_list = nullptr;
+    std::vector<std::string> * exclude_list = nullptr;
 
-    const auto & include_list =
-      topic_a_service_event ? storage_filter_.services_events : storage_filter_.topics;
+    if (is_topic_name_related_to_action(topic)) {
+      include_list = &storage_filter_.actions_interfaces;
+      exclude_list = &storage_filter_.exclude_actions_interfaces;
+    } else if (is_topic_name_a_service_event(topic)) {
+      include_list = &storage_filter_.services_events;
+      exclude_list = &storage_filter_.exclude_service_events;
+    } else {
+      include_list = &storage_filter_.topics;
+      exclude_list = &storage_filter_.exclude_topics;
+    }
 
-    const auto & exclude_list = topic_a_service_event ? storage_filter_.exclude_service_events
-                                                      : storage_filter_.exclude_topics;
     // if topic not found in exclude list or regex_to_exclude
-    if (!is_topic_in_black_list_or_exclude_regex(topic, exclude_list,
+    if (!is_topic_in_black_list_or_exclude_regex(topic, *exclude_list,
                                                  storage_filter_.regex_to_exclude)) {
       // if topic selected by include list or regex
-      if (is_topic_selected_by_white_list_or_regex(topic, include_list, storage_filter_.regex)) {
+      if (is_topic_selected_by_white_list_or_regex(topic, *include_list, storage_filter_.regex)) {
         return true;
       }
     }
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -30,6 +30,7 @@ find_package(ament_cmake_ros REQUIRED)
 find_package(keyboard_handler REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_action REQUIRED)
 find_package(rclcpp_components REQUIRED)
 find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
@@ -48,6 +49,7 @@ add_library(${PROJECT_NAME} SHARED
   src/rosbag2_transport/play_options.cpp
   src/rosbag2_transport/player_progress_bar.cpp
   src/rosbag2_transport/player_service_client.cpp
+  src/rosbag2_transport/player_action_client.cpp
   src/rosbag2_transport/reader_writer_factory.cpp
   src/rosbag2_transport/recorder.cpp
   src/rosbag2_transport/record_options.cpp
@@ -59,6 +61,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC
 target_link_libraries(${PROJECT_NAME} PUBLIC
   keyboard_handler::keyboard_handler
   rclcpp::rclcpp
+  rclcpp_action::rclcpp_action
   rosbag2_cpp::rosbag2_cpp
   ${rosbag2_interfaces_TARGETS}
   rosbag2_storage::rosbag2_storage
@@ -105,6 +108,7 @@ ament_export_targets(export_${PROJECT_NAME})
 ament_export_dependencies(
   keyboard_handler
   rclcpp
+  rclcpp_action
   rosbag2_cpp
   rosbag2_interfaces
   rosbag2_storage
--- a/rosbag2_transport/include/rosbag2_transport/play_options.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/play_options.hpp
@@ -30,7 +30,7 @@ namespace rosbag2_transport
 {
 enum class ServiceRequestsSource : int8_t
 {
-  SERVICE_INTROSPECTION = 0,
+  SERVER_INTROSPECTION = 0,
   CLIENT_INTROSPECTION = 1
 };
 
@@ -59,6 +59,11 @@ public:
   // If list is empty, the filter is ignored and all messages of services are played.
   std::vector<std::string> services_to_filter = {};
 
+  // Action names to whitelist when playing a bag.
+  // Only messages matching these specified actions will be played. If list is empty,
+  // the filter is ignored and all messages of actions are played.
+  std::vector<std::string> actions_to_filter = {};
+
   // Regular expression of topic names and service name to whitelist when playing a bag.
   // Only messages matching these specified topics and services will be played.
   // If list is empty, the filter is ignored and all messages are played.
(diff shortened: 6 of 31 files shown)


Barry-Xu-2018:
@MichaelOrlov @fujitatomoya 

Please review this PR.

ros2bag/ros2bag/verb/burst.py
```diff
@@ -97,6 +101,8 @@ def main(self, *, args):  # noqa: D102
         play_options.topics_to_filter = args.topics
         # Convert service name to service event topic name
         play_options.services_to_filter = convert_service_to_service_event_topic(args.services)
+        # Covert action name to action related topic names
+        play_options.actions_to_filter = args.actions
```
fujitatomoya, line 104: (note for my review) we just cannot change the names here like service events above, action needs to be managed with special care later.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
```diff
@@ -30,7 +30,7 @@ namespace rosbag2_transport
 {
 enum class ServiceRequestsSource : int8_t
 {
-  SERVICE_INTROSPECTION = 0,
+  SERVER_INTROSPECTION = 0,
```
fujitatomoya, line 33: why do we need to rename this? i think `SERVICE_INTROSPECTION` can be okay? just checking, we do not need to roll this back. but if the application built on top of `rosbag2_transport`, this will break their build. i am not sure if this is worth to do that during distro transition.

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -566,7 +564,9 @@ bool MCAPStorage::read_and_enqueue_message()
   return true;
 }
 
-bool MCAPStorage::is_topic_name_a_service_event(const std::string_view topic_name)
+namespace
+{
+bool is_topic_name_a_service_event(const std::string_view topic_name)
```
fujitatomoya, line 569: i think this is typical favors for modern c++ approach, compared to static free functions.

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -339,6 +341,19 @@ class Player : public rclcpp::Node
...
   std::unordered_map<std::string, std::shared_ptr<rclcpp::GenericClient>> get_service_clients();
 
+  /// \brief Getter for clients corresponding to each action name
+  /// \return Hashtable representing action name to client
+  ROSBAG2_TRANSPORT_PUBLIC
+  std::unordered_map<std::string, std::shared_ptr<rclcpp_action::GenericClient>>
+  get_action_clients();
+
+  /// \brief Goal handle for action client is in processing.
+  /// \return true if goal handle is in processing, otherwise false.
```
fujitatomoya, line 351:
```suggestion
  /// \brief Check if goal ID is in processing.
  /// \return true if goal id still is in processing, otherwise false.
```

rosbag2_transport/include/rosbag2_transport/player_action_client.hpp
```diff
@@ -0,0 +1,179 @@
...
+  /// \brief Extract the request data from the get_result event message and asynchronously
+  // send the extracted request data.
+  void async_send_result_request(
+    const std::shared_ptr<uint8_t[]> & type_erased_get_result_service_event);
+
+  /// Wait until sent service requests will receive responses from service servers.
+  /// \param timeout - Timeout in fraction of seconds to wait for.
+  /// \return true if service requests successfully finished, otherwise false.
+  bool wait_for_sent_requests_to_finish(
+    std::chrono::duration<double> timeout = std::chrono::seconds(5));
```
fujitatomoya: i may be missed, but i do not see the implementation for this, do we need to declare this method?

rosbag2_transport/include/rosbag2_transport/player_action_client.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+  bool gold_handle_in_processing(const GoalID & goal_id);
+
+private:
+  rclcpp_action::GenericClient::SharedPtr client_;
+  std::string action_name_;
+  std::string action_type_;
+  const rclcpp::Logger logger_;
+
+  // Warning output once exceeding this value
```
fujitatomoya: actually this prints every time it hits the limit.

```suggestion
  // Warning output when exceeding this value
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -194,6 +197,16 @@ class PlayerImpl
   /// \return Hashtable representing service name to client map
   std::unordered_map<std::string, std::shared_ptr<rclcpp::GenericClient>> get_service_clients();
 
+  /// \brief Getter for clients corresponding to actions
+  /// \return Hashtable representing action name to client map
+  std::unordered_map<std::string, std::shared_ptr<rclcpp_action::GenericClient>>
+  get_action_clients();
+
+  /// \brief Goal handle for action client is in processing.
+  /// \return true if goal handle is in processing, otherwise false.
```
fujitatomoya:
```suggestion
  /// \brief Check if goal id is in processing.
  /// \return true if goal id is in processing, otherwise false.
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -951,6 +988,27 @@ std::unordered_map<std::string,
...
+  }
+  return topic_to_action_client_map;
+}
+
+bool
+PlayerImpl::goal_handle_in_process(std::string action_name, const rclcpp_action::GoalUUID & goal_id)
+{
+  if (action_name.empty() || action_clients_.find(action_name) == action_clients_.end()) {
+    throw(std::invalid_argument("Action client for action '" + action_name + "' does not exist."));
+  }
```
fujitatomoya: i think `action_name.empty()` can be invalid argument and we can raise the exception.
but it is expected that goal_id might not be found on the list as API design? in that case, i would print the warning that it cannot find the goal_id on the list, and return false to avoid raising the exception for the application.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1169,34 +1227,62 @@ rcutils_time_point_value_t PlayerImpl::get_message_order_timestamp(
 
 namespace
 {
+enum class TopicKind
+{
+  TOPIC,
```
fujitatomoya: how about this?

```suggestion
  GENERIC_TOPIC,
```

rosbag2_transport/test/rosbag2_transport/mock_player.hpp
```diff
@@ -75,6 +77,26 @@ class MockPlayer : public rosbag2_transport::Player
...
+    for (const auto & client : this->get_action_clients()) {
+      action_cli_list.push_back(
+        static_cast<rclcpp_action::ClientBase *>(
+          client.second.get()));
+    }
+
+    return action_cli_list;
+  }
+
+  // Before calling this function, make sure that goal has been accepted by the action server.
```
fujitatomoya, line 92: as i mentioned before, we can avoid pushing this constraints for the user application by system. i think that would be better for user application experience.

ros2bag/ros2bag/verb/play.py
```diff
@@ -60,18 +60,24 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            help='regular expressions to exclude topics, services and actions from replay.')
         parser.add_argument(
             '--exclude-topics', type=str, default=[], metavar='topic', nargs='+',
             help='Space-delimited list of topics not to play.')
         parser.add_argument(
             '--exclude-services', type=str, default=[], metavar='service', nargs='+',
             help='Space-delimited list of services not to play.')
+        parser.add_argument(
+            '--exclude-actions', type=str, default=[], metavar='action', nargs='+',
+            help='Space-delimited list of actions not to play.')
```
fujitatomoya, line 80: test would be ideal for these additional arguments to make sure that does not break cli.

ros2bag/ros2bag/verb/play.py
```diff
@@ -162,12 +168,20 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--publish-service-requests', action='store_true', default=False,
             help='Publish recorded service requests instead of recorded service events')
+        parser.add_argument(
+            '--send-actions-as-client', action='store_true', default=False,
+            help='Send the send_goal request, cancel_goal request, and get_result request '
+                 'respectively based on the recorded send_goal, cancel_goal, and get_result '
+                 'event messages. Note that the messages from action\'s "status topic" and '
+                 '"feedback topic" will not be sent because they are expected to be sent from '
+                 'the action server side.')
```
fujitatomoya, line 177: actually there is no test for `--publish-service-requests` as cli, probably we can follow this up as service and action playback test with another PR?

fujitatomoya (commented):
@Barry-Xu-2018 1st review, can you check my comments?

in the meantime, i will start CI to see if other warnings during review.

fujitatomoya:
Pulls: ros2/rosbag2#1955
Gist: https://gist.githubusercontent.com/fujitatomoya/b5abc904756413ea7d3f28e439e0f919/raw/f55ebf7922ad03a51da906c027c3cb5b0d28c84b/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15575
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23007/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17233/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2494/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23882/)

MichaelOrlov:
Hi @Barry-Xu-2018, @fujitatomoya,
I am currently on vacation and can't review this PR.
I will be back Sunday.

fujitatomoya:
@MichaelOrlov thanks for replying to this on your vacation 🍹 i think we can iterate before your review, and we will have 10 days until API freeze 🧊 i believe we can get this in next week 🙏🙏🙏 anyway, let's discuss when you get back! thanks and enjoy ✌️

fujitatomoya:
@Barry-Xu-2018 CI looks fine, please address my comments above, and i can start CI again after that.

ros2bag/ros2bag/verb/burst.py
Barry-Xu-2018: Sorry, the previous note was incorrect, and I should remove it.  
Currently, the actions still contain the user-inputted action name rather than the converted action interface name.

rosbag2_transport/include/rosbag2_transport/play_options.hpp
Barry-Xu-2018: The enum class ServiceRequestsSource indicates where the introspection information is coming from—either the service server or the service client.  
There are two defined values, with the other being CLIENT_INTROSPECTION. Therefore, I think SERVER_INTROSPECTION would be more appropriate than SERVICE_INTROSPECTION. So I want change it in this PR.

rosbag2_transport/include/rosbag2_transport/player_action_client.hpp
Barry-Xu-2018: For actions, this function is not needed. I will remove it.

rosbag2_transport/src/rosbag2_transport/player.cpp
Barry-Xu-2018: Okay. Update this.

rosbag2_transport/test/rosbag2_transport/mock_player.hpp
Barry-Xu-2018: This is a function used for testing, so I just added a comment to clarify the requirements for calling it.

ros2bag/ros2bag/verb/play.py
Barry-Xu-2018: Currently, this play command is not tested directly; instead, the tests are placed where the filter actually works, such as in test_mcap_topic_filter.cpp and test_sqlite_topic_filter.cpp.

Barry-Xu-2018: Just like ros2bag/test/test_recorder_args_parser.py, a test file for play_args needs to be added.

ros2bag/ros2bag/verb/play.py
Barry-Xu-2018: Yes

Barry-Xu-2018:
@MichaelOrlov @fujitatomoya 

In the current implementation, the goal ID is derived from recorded data. I'm considering whether it should be replaced with a new created goal ID.  I would like to hear your thoughts.   

If the action server is continuously running and the user repeatedly uses `ros2 bag play` to replay action requests, there will be an issue because the goal ID sent each time is the same.

fujitatomoya:
@Barry-Xu-2018 

i think that,

- as bag information, we need to store the original goal ID for sure. so that user can match the goal ID with debug messages or something like that.
- but when it is doing playback, i think goal ID needs to be replaced with new global unique identification, otherwise, you cannot play the action date to the action server concurrently.

fujitatomoya:
Pulls: ros2/rosbag2#1955
Gist: https://gist.githubusercontent.com/fujitatomoya/7fe3dc7bcfa57ed1effaab8db0c1c8fa/raw/f55ebf7922ad03a51da906c027c3cb5b0d28c84b/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15621
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23068/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17288/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2547/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23946/)

Barry-Xu-2018:
@fujitatomoya 

> i think that,
>
> - as bag information, we need to store the original goal ID for sure. so that user can match the goal ID with debug messages or something like that.
> - but when it is doing playback, i think goal ID needs to be replaced with new global unique identification, otherwise, you cannot play the action date to the action server concurrently.

Thanks. I also think so.   
I will update code.

rosbag2_storage_mcap/src/mcap_storage.cpp
```diff
@@ -582,6 +582,24 @@ bool MCAPStorage::is_topic_name_a_service_event(const std::string_view topic_nam
...
 
+// The postfix of the action internal topics and service event topics
+const std::vector<std::string> ActionInterfacePostfix = {
+  "/_action/send_goal/_service_event", "/_action/cancel_goal/_service_event",
+  "/_action/get_result/_service_event", "/_action/feedback", "/_action/status"};
+
+bool is_topic_name_related_to_action(const std::string_view topic_name)
+{
+  for (const auto & postfix : ActionInterfacePostfix) {
+    if (topic_name.length() > postfix.length() &&
```
MichaelOrlov, line 593: The condition
```cpp
if (topic_name.length() < postfix.length() return false;
```
can be moved out of the for-loop to reduce the number of comparisons.

fujitatomoya (approved):

fujitatomoya:
Pulls: ros2/rosbag2#1955
Gist: https://gist.githubusercontent.com/fujitatomoya/9ecdcc6e32fe9e0732c3227bea6d1b02/raw/f55ebf7922ad03a51da906c027c3cb5b0d28c84b/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15633
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23086/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17305/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2563/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23963/)

rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -339,6 +341,19 @@ class Player : public rclcpp::Node
...
+  /// \exception std::exception No action client is created for this action name.
+  ROSBAG2_TRANSPORT_PUBLIC
+  bool
+  goal_handle_in_process(std::string action_name, const rclcpp_action::GoalUUID & goal_id);
```
MichaelOrlov: I would like to ask to rename this function a  bit for the sake of clarity and better readability.
```suggestion
  ROSBAG2_TRANSPORT_PUBLIC
  bool
  is_goal_handle_in_processing(std::string action_name, const rclcpp_action::GoalUUID & goal_id);
```

rosbag2_transport/include/rosbag2_transport/player_action_client.hpp
```diff
@@ -0,0 +1,187 @@
...
+  ServiceEventType get_service_event_type(
+    const std::shared_ptr<uint8_t[]> & type_erased_service_event,
+    ServiceInterfaceInAction service_type_in_action);
+
+  rclcpp_action::GenericClient::SharedPtr generic_client()
+  {
+    return client_;
+  }
+
+  bool gold_handle_in_processing(const GoalID & goal_id);
```
MichaelOrlov: Typo. **gold**. I think it was expected to be a `goal`.
```suggestion
  bool goal_handle_in_processing(const GoalID & goal_id);
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1169,34 +1234,62 @@ rcutils_time_point_value_t PlayerImpl::get_message_order_timestamp(
...
   auto & regex = storage_filter.regex;
   auto & regex_to_exclude = storage_filter.regex_to_exclude;
 
-  if (is_service) {
-    if (!exclude_services.empty()) {
-      auto it = std::find(exclude_services.begin(), exclude_services.end(), topic_name);
-      if (it != exclude_services.end()) {
-        return false;
+  // Check if topic is in the exclude list
+  switch (topic_type) {
```
MichaelOrlov, line 1259: Please add a `default` section to switches with throwing exception in the `allow_topic` function

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1251,10 +1381,28 @@ void PlayerImpl::prepare_publishers()
   rosbag2_storage::StorageFilter storage_filter;
   storage_filter.topics = play_options_.topics_to_filter;
   storage_filter.services_events = play_options_.services_to_filter;
+  if (!play_options_.actions_to_filter.empty()) {
```
MichaelOrlov: The check for the emptiness of the actions_to_filter list is redundant here. I would recommend removing it to reduce the complexity

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1251,10 +1381,28 @@ void PlayerImpl::prepare_publishers()
...
+        storage_filter.actions_interfaces.end(),
+        std::make_move_iterator(action_interfaces.begin()),
+        std::make_move_iterator(action_interfaces.end()));
+    }
+  }
   storage_filter.regex = play_options_.regex_to_filter;
   storage_filter.regex_to_exclude = play_options_.exclude_regex_to_filter;
   storage_filter.exclude_topics = play_options_.exclude_topics_to_filter;
   storage_filter.exclude_service_events = play_options_.exclude_services_to_filter;
+  if (!play_options_.exclude_actions_to_filter.empty()) {
```
MichaelOrlov: The same here, the check for the emptiness of the `exclude_actions_to_filter` list is redundant here. I would recommend removing it to reduce the complexity

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1414,31 +1602,223 @@ void PlayerImpl::run_play_msg_post_callbacks(
...
+      break;
+    case rosbag2_cpp::ActionInterfaceType::GetResultEvent:
+      type_erased_service_event =
+        action_client->deserialize_get_result_service_event(*message->serialized_data);
+      service_event_type = action_client->get_service_event_type(
+        type_erased_service_event,
+        PlayerActionClient::ServiceInterfaceInAction::GET_RESULT_SERVICE);
+      break;
+    default:
+      // Never go here
```
MichaelOrlov, line 1779:
```suggestion
       throw std::logic_error("Wrong action interface type.");
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1449,92 +1829,48 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr
         "Failed to call on play message post-callback on '" << message->topic_name <<
```
MichaelOrlov:
```suggestion
        "Failed to call on play action post-callback on '" << message->topic_name <<
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1414,31 +1602,223 @@ void PlayerImpl::run_play_msg_post_callbacks(
...
+        "' service. \nError: " << e.what());
+    return false;
+  }
+
+  // Calling on play message pre-callbacks
+  run_play_msg_pre_callbacks(message);
+
+  bool message_published = false;
+  try {
+    service_client->async_send_request(service_event);
```
MichaelOrlov: @Barry-Xu-2018 We have a "contract" in player API that if pre-callback fails we shouldn't publish the current message. i.e. this way user can filter messages via pre-callback.
Could you please guard pre-callback and try-catch block with publishing the similar way as in the `publish_message_by_player_publisher`?
```cpp
bool pre_callbacks_failed = true;
  try {
    // Calling on play message pre-callbacks
    run_play_msg_pre_callbacks(message);
    pre_callbacks_failed = false;
    RCLCPP_ERROR_STREAM(owner_->get_logger(),
      "Failed to call on play message pre-callback on '" << message->topic_name <<
      "' topic. \nError: " << e.what());
  }

  if (!pre_callbacks_failed) {
    try {
      player_publisher->publish(rclcpp::SerializedMessage(*message->serialized_data));
```
And the same for the `publish_message_by_play_action_client`.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1414,31 +1602,223 @@ void PlayerImpl::run_play_msg_post_callbacks(
...
+    }
+  } else {
+    // ServiceRequestsSource::CLIENT_INTROSPECTION
+    if (service_event_type == PlayerActionClient::ServiceEventType::REQUEST_RECEIVED) {
+      // Ignore request message from service server
+      return true;
+    }
+  }
+
+  try {
```
MichaelOrlov, line 1816: Need to remove this large scope outer try-catch, and use the small only when it is really needed in the 
```cpp
 switch (action_interface_type) {
      case rosbag2_cpp::ActionInterfaceType::SendGoalEvent:
        action_client->async_send_goal_request(type_erased_service_event);
```

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1414,31 +1602,223 @@ void PlayerImpl::run_play_msg_post_callbacks(
...
+    if (!action_client->generic_client()->wait_for_action_server(
```
MichaelOrlov: Can we avoid calling `wait_for_action_server(..)` inside the publishing method?
It could significantly delay other messages. We want to be close to the deterministic and reproducible behavior. 
This wait with a timeout will ruin reproducibility and determinism.

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1449,92 +1829,48 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr
...
-          "' service. \nError: " << e.what());
-      return false;
+  // Try to publish message as action request
+  auto action_interface_type = rosbag2_cpp::get_action_interface_type(message->topic_name);
+  if (action_interface_type != rosbag2_cpp::ActionInterfaceType::Unknown) {
+    if (action_interface_type == rosbag2_cpp::ActionInterfaceType::Feedback ||
+      action_interface_type == rosbag2_cpp::ActionInterfaceType::Status)
+    {
+      // Ignore messages for feedback and status
+      return true;
```
MichaelOrlov: I think it makes sense to add a logging message about ignoring messages at least in `RCLCPP_DEBUG_STREAM`

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -1449,92 +1829,48 @@ bool PlayerImpl::publish_message(rosbag2_storage::SerializedBagMessageSharedPtr
...
-          rosbag2_cpp::service_event_topic_name_to_service_name(message->topic_name) <<
-          "' service. \nError: " << e.what());
+    auto action_name = rosbag2_cpp::action_interface_name_to_action_name(message->topic_name);
+    auto action_client_iter = action_clients_.find(action_name);
+    if (action_client_iter != action_clients_.end()) {
+      return publish_message_by_play_action_client(action_client_iter->second,
+          action_interface_type, message);
+    } else {
+      // Ignored action message
+      return true;
```
MichaelOrlov: The same here. I think it makes sense to add a logging message about ignoring messages at least in `RCLCPP_DEBUG_STREAM`

rosbag2_transport/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ find_package(ament_cmake_ros REQUIRED)
 find_package(keyboard_handler REQUIRED)
 find_package(rcl REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_action REQUIRED)
```
MichaelOrlov, line 33: Please add new dependencies to the `package.xml` file as well

README.md
MichaelOrlov, line 1: I think we need to add some explanations about `send_actions_as_client` to the readme file as well.

MichaelOrlov (changes requested):
@Barry-Xu-2018 Here are my findings so far.
However, I haven't thoroughly reviewed `play_action_clients` yet.
Thanks for the PR - the code quality became way better this time.

Barry-Xu-2018:
@MichaelOrlov 

Thank you for your review. I will address your review comments.

rosbag2_storage_mcap/src/mcap_storage.cpp
Barry-Xu-2018: This cannot be moved out of the for-loop since it depend on postfix.length().   
The strings being compared are fixed, so I can reduce computation by not calculating their lengths each time.

MichaelOrlov: Ahh, thanks for the clarification. I missed it. Sorry, my bad.
Do not need to cache string length. according to the ChatGPT 
> In modern C++ (C++11 and later), std::string::length() (which is the same as std::string::size()) is required to run in constant time, O(1). This is because the string’s size is maintained as part of its internal state rather than being computed each time you call length().

MichaelOrlov (approved):
@Barry-Xu-2018 Thank you for addressing the review comments looks good to me now. :+1:

fujitatomoya:
@MichaelOrlov @Barry-Xu-2018 thanks 👍 i will start the CI now.

fujitatomoya:
Pulls: ros2/rosbag2#1955
Gist: https://gist.githubusercontent.com/fujitatomoya/e18e606861b500e8ad5890a100e3d71b/raw/f55ebf7922ad03a51da906c027c3cb5b0d28c84b/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_storage_sqlite3 rosbag2_test_common rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15658
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23115/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17332/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2588/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23990/)

fujitatomoya:
@MichaelOrlov @Barry-Xu-2018 CI is green and this is merged, thank you so much for your effort. (i will be working on the documentation for next week.)

Barry-Xu-2018:
@fujitatomoya  

I think Michael's review is not finished yet. There should be additional comments. 

@MichaelOrlov Please continue to put your comments here. I will handle them as a new PR.

fujitatomoya:
@Barry-Xu-2018 @MichaelOrlov oh sorry, i thought this is ready since approved. anyway, if we need to follow up more comments, please open an another PR for that. thanks for letting me know.

MichaelOrlov:
@Barry-Xu-2018 @fujitatomoya Sorry for the confusion.  
I got approval and think we should merge this PR sooner rather than later because
1. I've reviewed the essential implementation part and after addressing the review comments it looks good to me
2. I don't have the capacity to review it more this week, and we need to merge this PR before the feature freeze on Monday.
3. If I or we find something in the `play_action_clients` - we can fix it as a bugfix after the feature freeze.

fujitatomoya:
@MichaelOrlov appreciate your consideration, and totally understood.

we will address the follow-ups for sure, we are not going anywhere 👍

PR #1961, opened 2025-04-13
Add message sequence number to the messages write API

MichaelOrlov:
This PR brings the ability to specify an optional sequence number of the message. 
If non-zero, sequence numbers should be unique per channel (per topic and per publisher) and increasing over time.

The published message sequence number could be very useful when detecting the cause of message drop.

While this is not a feature since we are still not populating the correct message sequence number, this is rather a placeholder in the API for future feature backports.


MichaelOrlov:
cc: @cottsay I would like to merge these API changes as a placeholder for future feature implementation and backports.

christophebedard:
> If non-zero, sequence numbers should be unique per channel (per topic and per publisher) and increasing over time.

It might be worth adding more information regarding how & when these sequence numbers are expected to be generated: Is this using one of the sequence numbers from `rmw` (e.g., reception sequence number), or will it be assigned right before the `write()` call?

By the way, sequence numbers in `rmw` are 64 bits as opposed to 32 bits, so it might be worth using 64 bits here.

MichaelOrlov:
As regards
> It might be worth adding more information regarding how & when these sequence numbers are expected to be generated: Is this using one of the sequence numbers from rmw (e.g., reception sequence number), or will it be assigned right before the write() call?

That is already a future feature implementation detail.
As regards:
> By the way, sequence numbers in rmw are 64 bits as opposed to 32 bits, so it might be worth using 64 bits here.

The reason why it is `uint32_t` in the rosbag2 API is because it is the same type in the mcap spec https://mcap.dev/spec#message-op0x05. 
The way how we will be converting `uint64_t` RMW sequence number to the `uint32_t` - is also an feature implementation details and to be decided in the future.

christophebedard (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#1961
Gist: https://gist.githubusercontent.com/MichaelOrlov/c9914b2efb59035f267f4369e2255a6b/raw/06f93e746bcd4835e2f7358f1d2a6e16b1dad4a3/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_tests
TEST args:  --packages-above rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_mcap rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15688
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23148/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17363/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24022/)

fujitatomoya (approved):
i think that is okay to have this optional argument in the place holder. a couple of questions.

- sqlite3 does not use this optional sequence number field, because it does not support sequence number?
- how can we tell the publishers? i think we also need to have publisher identification with this sequence number to tell this specific message is from which endpoint? are we going to integrate that in the sequence identification or we already have the field to tell which publisher endpoint in the topic?

MichaelOrlov:
> - sqlite3 does not use this optional sequence number field, because it does not support sequence number?

- yes. this is correct.

> - how can we tell the publishers? i think we also need to have publisher identification with this sequence number to tell this specific message is from which endpoint? are we going to integrate that in the sequence identification or we already have the field to tell which publisher endpoint in the topic?

In message info we do have publisher GUID and publishing sequence number. As a first attempt we can potentially fuse them together like taking LSB 16 bit from 64 bit sequence number and other MSB 16 bit use for the publisher ID. However, will need somehow compress publisher GUID to the 16 bit.
The appropriate implementation would be to add support for recording by channels (per publisher per topic) rather than by topic name as we do currently. Then, there would be no need to fuse publisher GUID with sequence number since we will write per channel, which is already unique for the publisher.

PR #1970, opened 2025-04-21
Document start_time_ns and end_time_ns

berndpfrommer:
Document the start and end time filter for the "convert" verb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,9 @@ output_bags:
   max_bagfile_duration: 0
   storage_preset_profile: ""
   storage_config_uri: ""
+  # optional filter for msg time t [nsec since epoch]:  start_time_ns <= t <= end_time_ns
+  # start_time_ns: 1744227144744197147
+  # end_time_ns: 1744227145734665546
   all_topics: false
   topics: []
   topic_types: []


README.md
```diff
@@ -310,6 +310,9 @@ output_bags:
   compression_threads: 0
   include_hidden_topics: false
   include_unpublished_topics: false
+  # optional filter for msg time t [nsec since epoch]:  start_time_ns <= t <= end_time_ns
+  # start_time_ns: 1744227144744197147
+  # end_time_ns: 1744227145734665546
```
fujitatomoya: i am not sure if we need to list all the possible options in the README, there are some other properties that are not listed here. besides, above description says `StorageOptions` and `RecordOptions`... i do not have any objection here.

but if we add them, probably we want to move these to right after `storage_config_uri` that is where `StorageOptions` belongs.

berndpfrommer: moved documentation of start/end_time_ns to right after storage_config_uri.
I decided to submit this PR because slicing a bag by time (to reduce size) is a common operation and currently requires reading the source code to figure out how to do.
[Here](https://robotics.stackexchange.com/questions/97734/how-can-i-split-an-existing-bag-in-ros2) is the first time this came up on StackExchange.

MichaelOrlov (commented):
@berndpfrommer Could you please sign your commit?
Please see details about how to sign commit here https://github.com/ros2/rosbag2/pull/1970/checks?check_run_id=41049894044

MichaelOrlov (changes requested):
It seems it would be a good candidate to go into the bottom of the readme doc to the 
https://github.com/ros2/rosbag2?tab=readme-ov-file#tips-and-tricks section with a few more wordings. Something like if you need to slice a bag with some certain timestamps do this and this..

berndpfrommer:
Commit has been signed

berndpfrommer:
> It seems it would be a good candidate to go into the bottom of the readme doc to the https://github.com/ros2/rosbag2?tab=readme-ov-file#tips-and-tricks section with a few more wordings. Something like if you need to slice a bag with some certain timestamps do this and this..

We already have a section [here](https://github.com/ros2/rosbag2?tab=readme-ov-file#converting-bags-merge-split-etc-) on how to split by size. I feel that documentation for splitting by time should be nearby rather than at the end. Also the "tips and tricks" so far cover how to use shell/launch methods, and splitting by time stamp does not fall into that category. 

Lastly, I wanted to keep things concise. The proposed addition to the documentation IMO is sufficient for most users to successfully use the feature without resorting to piecing things together from source code.

MichaelOrlov (approved):
Okay, Approved.

MichaelOrlov:

PR #1993, opened 2025-04-26
Update sqlite minor version

mosfet80:
update minor version
changelog:
https://github.com/sqlite/sqlite/compare/version-3.45.0...version-3.45.3


christophebedard:
This should match the version available from Ubuntu 24.04, so there is no need to bump this: https://packages.ubuntu.com/noble/sqlite3

MichaelOrlov:
@mosfet80 Is there a particular reason for the update?

MichaelOrlov:
- Closing as wouldn't be done, at least for a while. The version should match the version available from Ubuntu 24.04.

PR #2001, opened 2025-05-01
Address flakiness in tests where need to spin a node

MichaelOrlov:
This PR addresses some flaky tests in Rosbag2
 1. Address flakiness in tests where need to spin a node
    - Wait for the executor to start spinning in the newly spawned thread to avoid race conditions. The same approach as from the (ros2/rosbag2#1796)
 2. Address flakiness in `write_split_callback_is_called` test
    - Reduce subscriptions polling interval in recorder to 10ms
    - Start recording after creating publishers to let recorder opportunity to subscribe to the test topic before starting topics discovery.

### DIFF
--- a/rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp
+++ b/rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp
@@ -17,6 +17,7 @@
 #include <cstdlib>
 #include <filesystem>
 #include <future>
+#include <iostream>
 #include <memory>
 #include <string>
 #include <vector>
@@ -55,6 +56,17 @@ public:
       [this]() {
         exec_->spin();
       });
+
+    // Wait for the executor to start spinning in the newly spawned thread to avoid race conditions
+    using clock = std::chrono::steady_clock;
+    auto start = clock::now();
+    while (!exec_->is_spinning() && (clock::now() - start) < std::chrono::seconds(5)) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    }
+    if (!exec_->is_spinning()) {
+      std::cerr << "Failed to start spinning node" << std::endl;
+      throw std::runtime_error("Failed to start spinning node");
+    }
   }
 
   ~PlayEndToEndTestFixture() override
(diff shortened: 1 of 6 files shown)


rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp
```diff
@@ -55,6 +56,17 @@ class PlayEndToEndTestFixture : public Test, public WithParamInterface<std::stri
...
+    // Wait for the executor to start spinning in the newly spawned thread to avoid race conditions
+    using clock = std::chrono::steady_clock;
+    auto start = clock::now();
+    while (!exec_->is_spinning() && (clock::now() - start) < std::chrono::seconds(5)) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    }
+    if (!exec_->is_spinning()) {
+      std::cerr << "Failed to start spinning node" << std::endl;
+      throw std::runtime_error("Failed to start spinning node");
+    }
```
christophebedard: This looks like it would make a good test function (e.g., in a base test fixture class) that could be used in multiple places.

I'll let you decide whether or not to do that.

christophebedard (approved):

fujitatomoya (approved):
as @christophebedard mentioned, these are redundant logics. but since they are already there as in each test, i am good to go with this.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_play_end_to_end.cpp
MichaelOrlov: Thanks for catching this.  
I replaced some boilerplates with 
```cpp
if (!wait_until_condition([this]() {return exec_->is_spinning();}, std::chrono::seconds(5))) {
```

MichaelOrlov:
Pulls: ros2/rosbag2#2001
Gist: https://gist.githubusercontent.com/MichaelOrlov/33740b0c2634adebcd636df4f6c7433b/raw/de3be63636da806c049acc068e8447475eddcfef/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_test_common rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_test_common rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16077
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23752/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17846/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3186/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24445/)

christophebedard (approved):

MichaelOrlov:

PR #2006, opened 2025-05-09
Add design doc for "Statistics about messages lost during recording" feature

MichaelOrlov:
This PR adds a design document for the "Statistics about messages lost during recording" feature.

This design proposes adding a new feature to Rosbag2 that will provide statistics about the number of lost and recorded messages and the number of bytes recorded per topic in real time and during post-processing.

- Relates https://github.com/ros2/rosbag2/issues/2007

### DIFF
--- a/docs/design/statistics_about_lost_and_recorded_messages.md
+++ b/docs/design/statistics_about_lost_and_recorded_messages.md
@@ -29,43 +29,50 @@ and during post-processing.
 
 ## Design proposals and implementation details consideration
 
-<a name="dds-lost-stat"></a>
-1. To facilitate functionality for gathering statistics about number messages lost on the DDS 
+### 1. DDS Transport Layer Messages Loss Tracking
+
+   To facilitate functionality for gathering statistics about number messages lost on the DDS 
    transport layer, we can specify `message_lost_callback` in the 
    `rclcpp::SubscriptionOptions::event_callbacks` when creating the subscription in the 
    `rosbag2_transport::recorder` by providing `rclcpp::SubscriptionOptions` parameter when 
    creating `GenericSubscription`. The `message_lost_callback` will provide a reference to the 
    `rclcpp::QOSMessageLostInfo` which is mapped to the `rmw_message_lost_status_s` struct.
-      ```cpp
-        typedef struct RMW_PUBLIC_TYPE rmw_message_lost_status_s
-        {
-          /// Total number of messages lost.
-          size_t total_count;
-          /// Number of messages lost since last callback.
-          size_t total_count_change;
-        } rmw_message_lost_status_t;
-      ```
-<a name="recorder-lost-stat"></a>
-2. To facilitate functionality for gathering statistics about number of messages lost on the Rosbag2
+   ```cpp
+     typedef struct RMW_PUBLIC_TYPE rmw_message_lost_status_s
+     {
+       /// Total number of messages lost.
+       size_t total_count;
+       /// Number of messages lost since last callback.
+       size_t total_count_change;
+     } rmw_message_lost_status_t;
+   ```
+
+### 2. Recorder-Side Messages Loss Tracking
+
+  To facilitate functionality for gathering statistics about number of messages lost on the Rosbag2
   recorder side will need to add a new `messages_lost_callback` to the `rosbag2_cpp::writer` class.
   The `messages_lost_callback` will be called when the `rosbag2_cpp::writer` is not able to 
   keep up with the incoming messages and the internal buffer is full.
   The `messages_lost_callback` will have a parameter as a return value defined as a const 
   reference to the `std::vector` of the `rosbag2_cpp::MessageLostInfo` structure that will 
   contain the following fields:
-     - `topic_name` (string): The name of the topic on which messages lost happened.
-     - `num_messages_lost` (uint64_t): The number of messages lost.
+   - `topic_name` (string): The name of the topic on which messages lost happened.
+   - `num_messages_lost` (uint64_t): The number of messages lost.
   The `messages_lost_callback` shall be defined in the `rosbag2_cpp::bag_events` namespace and
   added to the `WriterEventCallbacks` struct similar to the `write_split_callback`.
-<a name="cli-option-update-rate"></a>
-3. Will need to add a CLI option to the Rosbag2 recorder to publish statistics on the predefined
+
+### 3. CLI Option for Statistics Update Rate
+
+  Will need to add a CLI option to the Rosbag2 recorder to publish statistics on the predefined
   topic with the rate not higher than specified.
-    - We can name this option as `--statistics-update-rate` and define it as integer . It 
-      will be used to 
-      specify the maximum update rate in times per second at which the statistics will be published.
-      The default value will be 1Hz. The value of 0 will disable the publishing of statistics.
-<a name="messages-lost-notifier"></a>
-4. Will need to add a new `MessagesLostNotifier` class to the `rosbag2_transport::recorder` that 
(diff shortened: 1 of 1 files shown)


docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+of lost and recorded messages and the number of bytes recorded per topic in real time and during
+post-processing.
+
+## Use cases
+
+1. Real-time monitoring of the performance of recording.
+    1. Overall statistics about the number of messages lost and bytes written on a per-topic
+       basis from recording start should be available in real-time by service request.
+    2. Incremental statistics about the number of messages lost and bytes written on a per-topic
+       basis should be published on a specific topic when messages lost happened.
```
fujitatomoya, line 26: per-topic statistics are still useful for sure, but it would be nicer if it can print the statistics for per-publisher basis? my assumption for user perspective is that user can see the statistics and know that there are some messages lost. and then probably what user wants to do is that, adjust the publisher rate or rosbag2 storage cache size and so on?
AFAIK, i think this is not doable with current rosbag2, because it does not recognize the publisher endpoints for the topic?

docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+2. Statistics about the number of lost messages and bytes written to the current bag file on a
+   per-topic basis should be saved in metadata at the end of the recording.
+
+## Design proposals and implementation details consideration
+
+### 1. DDS Transport Layer Messages Loss Tracking
```
fujitatomoya: I would avoid using DDS here, instead probably RMW? besides, we have non-DDS based RMW zenoh in tier 1 now 😄

docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+
+   To facilitate functionality for gathering statistics about the number of messages lost on the
+   DDS transport layer, we can specify `message_lost_callback` in the 
+   `rclcpp::SubscriptionOptions::event_callbacks` when creating the subscription in the
+   `rosbag2_transport::recorder` by providing `rclcpp::SubscriptionOptions` parameter when
+   creating `GenericSubscription`. The `message_lost_callback` will provide a reference to the
+   `rclcpp::QOSMessageLostInfo`, which is mapped to the `rmw_message_lost_status_s` struct.
+   ```cpp
+     typedef struct RMW_PUBLIC_TYPE rmw_message_lost_status_s
```
fujitatomoya, line 41: this RMW interface does not have per-publisher based endpoint information either. that is said, even though rosbag2 recognize the publisher endpoint for the topic, it would be impossible to get the event from which publisher endpoint via rmw interface. unless we update the rmw interface to recognize the publisher endpoint.

docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+  the following fields:
+   - `topic_name` (string): The name of the topic on which the messages were lost.
+   - `num_messages_lost` (uint64_t): The number of messages lost.
+  The `messages_lost_callback` shall be defined in the `rosbag2_cpp::bag_events` namespace and
+  added to the `WriterEventCallbacks` struct, similar to the `write_split_callback`.
+
+### 3. CLI Option for Statistics Update Rate
+
+  Will need to add a CLI option to the Rosbag2 recorder to publish statistics on the predefined 
+  topic at a rate that is not higher than specified.
```
fujitatomoya, line 70: since this is statistics, i can assume that it still publishes the statistics on the topic even if no messages dropped during the rate, correct? or it publishes the statistics only if message lost happens in the time window?

docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+         - Array of the following fields:
+         - `topic_name` (string): The name of the topic on which the messages were lost.
+         - `messages_lost_in_transport` (uint64_t): The number of messages lost since the last 
+           event on a DDS transport layer.
+         - `messages_lost_in_recorder` (uint64_t): The number of messages lost since the last 
+           event in the Rosbag2 recorder.
+         - `bytes_written` (uint64_t): The total number of bytes written on this topic.
+      3. The messages lost event shall not be published more often than the user-specified event 
+         update 
+         rate to avoid excessive resource usage.
```
fujitatomoya: maybe this can be single line?

```suggestion
         update rate to avoid excessive resource usage.
```

docs/design/statistics_about_lost_and_recorded_messages.md
```diff
@@ -0,0 +1,206 @@
...
+      4. The Messages lost event shall **not** include topics with zero number of lost messages.
+         To get full statistics about the number of recorded and lost messages, the user shall use
+         the service request `GetRecorderStatistics` described in the
+         [[5. New Service Request To Get Statistics]](#5-new-service-request-to-get-statistics).
```
fujitatomoya, line 102: ah okay, this is more like message lost event, not exactly the statistics on the recorder? i was expecting that we can still get the event that tells message has not lost as statistics, but that is out of scope for this event topic. i think this is also good for efficiency, and if user wants to take that statistics, user application should call the following service.

fujitatomoya (commented):
overall lgtm, i do not have any objections. this can be useful for user to adjust the storage cache and other parameters for recording.

docs/design/statistics_about_lost_and_recorded_messages.md
MichaelOrlov: Yes, that is correct. There is no per-publisher information about lost messages. This is probably because message drops could happen on the subscriber side as well, and in this case, it would be irrelevant to any publisher.  
This is why your request from the first comment to have per-publisher statistics is not doable.

docs/design/statistics_about_lost_and_recorded_messages.md
MichaelOrlov: It publishes the statistics only if a message loss occurs, but not more often than specified. This is designed to mitigate the performance burden during recording.

docs/design/statistics_about_lost_and_recorded_messages.md
MichaelOrlov: Yes, your understanding is correct. I am trying to mitigate the extra performance burden during the recording.  
If the user wants to see statistics regularly and doesn't care about an extra performance burden, it is possible to call the service on a regular basis with some predefined interval.

MichaelOrlov:
@fujitatomoya Thank you for your review. I've answered your questions and addressed your comments.

MichaelOrlov:
@fujitatomoya Friendly ping here to reiterate on the review here.

fujitatomoya (approved):
lgtm.

while this statistics are dedicated to rosbag2 recording, it is practical and useful for actual recoding use case in the production. (missing data during recording is always concern and problem for data analysis and AI development)

(TLDR; Just a note for generic statistics in ROS 2, only available on rclcpp) from the study of this PR. i think this message lost statistics is also important for generic subscription, which is still missing. (probably more important statistics can be added deadline miss and so on.) besides that, this generic statistics is only published as topic, this is also the latency problem for the production grade application. for low latency, the application should call the statistics API on the subscription and reconfigure the parameters right away, instead of collecting the statistics topic data which is private class in subscription in rclcpp. and that statistics topic includes many unconcerned statistics for the application, this can be huge disturbance for the application too. maybe having statistics service instead of topic as this PR proposes would be better, so that user (system observability application) can manage frequency and priority to collect the statistics as they need. after all, i think current statistics in generic subscription is not useful, we never used before.

PR #2016, opened 2025-05-24
Fix for flaky `playing_respects_delay` test

MichaelOrlov:
## Description
The `rosbag2_transport::playing_respects_delay` test tends to be flaky under the heavy load on CI or on some platforms when C++ compiler doesn't provide a precise enough steady_clock.

This PR addresses flakiness in the `PlayerTestFixture::playing_respects_delay` test by using `std::chrono::steady_clock::now()` for measuring time difference in the test and adjusting expectations for `replay_time` to be in microseconds instead of nanoseconds.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
@@ -164,19 +164,32 @@ TEST_F(PlayerTestFixture, playing_rate_negative)
 
 TEST_F(PlayerTestFixture, playing_respects_delay)
 {
-  rclcpp::Duration delay_margin(0.5s);
+  rclcpp::Duration upper_bound_tolerance(1s);
+  rclcpp::Duration lower_bound_tolerance(2us);
 
-  // Sleep 1.0 seconds before play
-  play_options_.delay = rclcpp::Duration(1, 0);
-  auto lower_expected_duration = message_time_difference + play_options_.delay;
-  auto upper_expected_duration = message_time_difference + play_options_.delay + delay_margin;
+  // Sleep 5.0 seconds before play
+  play_options_.delay = rclcpp::Duration(5, 0);
+  // Don't expect accuracy more than a few microseconds
+  auto lower_expected_duration = (message_time_difference + play_options_.delay -
+    lower_bound_tolerance).to_chrono<std::chrono::microseconds>();
+  auto upper_expected_duration = (message_time_difference + play_options_.delay +
+    upper_bound_tolerance).to_chrono<std::chrono::microseconds>();
+
+  // Sanity check for steady clock accuracy
+  ASSERT_TRUE(std::chrono::duration_cast<std::chrono::microseconds>(
+      typename std::chrono::steady_clock::duration(1)).count() <= 1) <<
+    "C++ compiler doesn't provide a precise enough steady_clock.";
+  static_assert(std::ratio_less_equal_v<std::chrono::steady_clock::period, std::micro>,
+    "C++ compiler doesn't provide a precise enough steady_clock.");
+
   auto player = std::make_shared<rosbag2_transport::Player>(
     std::move(reader), storage_options_, play_options_);
 
-  auto start = clock.now();
+  auto start = std::chrono::steady_clock::now();
   player->play();
   player->wait_for_playback_to_finish();
-  auto replay_time = clock.now() - start;
+  auto replay_time =
+    std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::steady_clock::now() - start);
 
   EXPECT_THAT(replay_time, Gt(lower_expected_duration));
   EXPECT_THAT(replay_time, Lt(upper_expected_duration));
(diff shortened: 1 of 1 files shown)


MichaelOrlov:

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -164,19 +164,32 @@ TEST_F(PlayerTestFixture, playing_rate_negative)
...
+    lower_bound_tolerance).to_chrono<std::chrono::microseconds>();
+  auto upper_expected_duration = (message_time_difference + play_options_.delay +
+    upper_bound_tolerance).to_chrono<std::chrono::microseconds>();
+
+  // Sanity check for steady clock accuracy
+  ASSERT_TRUE(std::chrono::duration_cast<std::chrono::microseconds>(
+      typename std::chrono::steady_clock::duration(1)).count() <= 1) <<
+    "C++ compiler doesn't provide a precise enough steady_clock.";
+  static_assert(std::ratio_less_equal_v<std::chrono::steady_clock::period, std::micro>,
+    "C++ compiler doesn't provide a precise enough steady_clock.");
```
christophebedard: these two checks seem redundant to me. Do we need both of them?

MichaelOrlov: No, they are different checks.

christophebedard: can you explain how they are different? I think making the two strings (for `ASSERT_TRUE` and `static_assert`) more specific would help.

MichaelOrlov: It has been a while since I wrote this fix, and I don't have a fresh memory of it.

Here is an explanation from ChatGPT:
In C++, within the <chrono> library, the types `std::chrono::steady_clock::duration` and `std::chrono::steady_clock::period` are related, but they represent different concepts:
### 1. `std::chrono::steady_clock::duration`
 - Definition:
`steady_clock::duration` is a type alias for a specialization of `std::chrono::duration`. It represents the **duration type** used by the `steady_clock`.
 - Purpose:
It is used to measure time intervals in terms of ticks. Each duration has a **tick count** (an integer value) and a **tick period** (the time span of each tick).
 - Example:
```cpp
std::chrono::steady_clock::duration d = std::chrono::seconds(5);
```
### 2. `std::chrono::steady_clock::period`
 - Definition:
`steady_clock::period` defines the `ratio` (as a compile-time constant) representing the length of one tick of the `steady_clock` in seconds.
- Purpose:
It indicates the time interval represented by a single tick of the clock. The `period` is defined using `std::ratio`, specifying the clock resolution.
- Example (typical implementation):
```cpp
using period = std::ratio<1, 1000000000>; // nanoseconds
```
This means each tick is 1/1,000,000,000 seconds (1 nanosecond).

I've added a bit more wording in the sanity checks warnings for both checks. Please see https://github.com/ros2/rosbag2/pull/2016/commits/7e587b976ae0f40dad9721935194b831bbb322e4 commit for details.

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
```diff
@@ -164,19 +164,34 @@ TEST_F(PlayerTestFixture, playing_rate_negative)
...
+    upper_bound_tolerance).to_chrono<std::chrono::microseconds>();
+
+  // Sanity check for steady clock accuracy
+  ASSERT_TRUE(std::chrono::duration_cast<std::chrono::microseconds>(
+      typename std::chrono::steady_clock::duration(1)).count() <= 1) <<
+    "The C++ compiler doesn't provide a precise enough steady_clock. The steady_clock duration"
+    " shall have a resolution of at least 1 microsecond.";
+  static_assert(std::ratio_less_equal_v<std::chrono::steady_clock::period, std::micro>,
+    "The C++ compiler doesn't provide a precise enough steady_clock. The steady_clock period shall"
+    " be at least 1 microsecond.");
```
christophebedard, line 185: Another comment: should the test simply be skipped in that case? For platforms with a clock that isn't precise enough, this test will just fail and nothing can be done.

christophebedard (approved):

rosbag2_transport/test/rosbag2_transport/test_play_timing.cpp
MichaelOrlov: Thanks for the review. I think an explicit test failure will be better than just automatically skipping it.  
It is better to adjust the skip manually if the test will fail due to insufficient precision in the clock.

MichaelOrlov:
Pulls: ros2/rosbag2#2016
Gist: https://gist.githubusercontent.com/MichaelOrlov/a81c110723797a5790504eb0b282cd58/raw/6f4e2309813039674769109e002423d22cbbd795/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16094
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23776/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17864/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3206/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24464/)

MichaelOrlov:
The Linux-aarch64 CI build automatically restarted and passed green
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17865/)

PR #2034, opened 2025-06-05
Update index.ros.org/p/ links for rosbag2_storage_mcap

christophebedard:
## Description

`index.ros.org/p/$pkg/` links with the GitHub repo at the end no longer work and result in a 404.

No.

### Additional Information

Consider backporting this to non-EOL distro branches.

### DIFF
--- a/rosbag2_storage_mcap/README.md
+++ b/rosbag2_storage_mcap/README.md
@@ -3,10 +3,7 @@
 This package provides a [storage plugin](https://github.com/ros2/rosbag2#storage-format-plugin-architecture) for rosbag2 which extends it with support for the [MCAP](https://mcap.dev) file format.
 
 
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
lgtm.

btw do we want to add jazzy and kilted instead, and remove E.O.L ones?

christophebedard:
what about having 1 link that doesn't include any specific distro? That's more maintainable.

christophebedard:
TIL about https://shields.io/badges/ros-package-index. It requires a specific distro, so I just kept the Rolling one. However, it doesn't seem to work: ROS Rolling version. Both the Rolling and Humble ones don't work currently: https://github.com/ros2/rosbag2/tree/rolling/rosbag2_storage_mcap#rosbag2_storage_mcap. Maybe the shields.io backend needs to be updated? Should we just go with this for now?

MichaelOrlov (approved):
LGTM

christophebedard:
Since this only changes a README, I'll skip ci.ros2.org.

Issue #2046, opened 2025-06-21
Provide Python API for the Rosbag2 player and recorder to be able to pause, stop, resume and seek.


MichaelOrlov:
## Description
Provide Python API for the Rosbag2 player and recorder classes to be able to pause, stop, resume and seek.

## Related Issues

N/A

## Completion Criteria

1. [x] The `rosbag2_py.Player` exposes `stop()`, `pause()`, `resume()` and `seek` API.
2. [x] The `rosbag2_py.Recorder` exposes `stop()`, `pause()`, and `resume()` API.

## Implementation Notes / Suggestions

With the current design of the `rosbag2_py.Player` and `rosbag2_py.Recorder` classes, it is difficult to implement the requested API because player and recorder instances are created only during the `Play(storage_options, play_options)` or `Record(storage_options, record_options)` API calls. We don't have references to the player or recorder instances outside of the `Play(..)` and `Record(..)` methods.  

To facilitate the requested new API, we will need to redesign `rosbag2_py.Player` and `rosbag2_py.Recorder` classes by providing constructors with storage and playback options to construct player and recorder instances inside the constructors and keep them alive during the class lifetime by storing them in the private member variable. 
The existing `Play(storage_options, play_options)` and `Record(storage_options, record_options)` API calls needs to be replaced with the `Play()` and `Record()`. i.e, without options. Since options shall be provided to the constructor.
To keep backward compatibility, it is proposed to keep old constructors and the existing `Play(storage_options, play_options)` or `Record(storage_options, record_options)` API, but deprecate them.

## Testing Notes / Suggestions

Basic unit tests for a newly added API could be added to the `rosbag2_py` package.

christophebedard:
I'm having a hard time testing and just working with the existing Python Player `play()` wrapper method, and I think users would also find it quite limiting and awkward to use.

The current Python `Player` wrapper class' `play()` method does more than the underlying `rosbag2_transport` C++ `Player` class does:

1. It creates an executor and spins the player node (which you're expected to do yourself with the C++ class)
2. It calls the underlying `play()` C++ method
3. (It potentially calls `burst()`, which is simply a separate method in C++)
4. It waits for playback to finish by calling `wait_for_playback_to_finish()`
5. It has a cancel mechanism

Also, the Python `Player` wrapper class initializes the global context directly in its constructor, which makes doing anything else before/after the `Player` object in Python annoying (e.g., create a node, subscription, etc.).

[...]

PR #2047, opened 2025-06-21
Change Python player and recorder to be more as a classes

MichaelOrlov:
## Description

This PR redesigns `rosbag2_py.Player` and `rosbag2_py.Recorder` classes to be self-contained by moving player and recorder instance construction into the newly added constructors with storage and player/recorder options.

- Relates #2046

Yes. 
It is recommended for users to migrate to the newly added API with initialization during construction.

Yes. I am using GitHub Copilot, GPT-4.1 in my workflow.

### Additional Information

To keep backward compatibility and not break the existing API, the old constructors and the existing `Play(storage_options, play_options)`, `Record(storage_options, record_options)` API are preserved but will be deprecated.

This PR doesn't have API/ABI breaking changes and could be backported.


rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -339,44 +357,66 @@ class Recorder
...
-      if (!record_options.disable_keyboard_controls) {
-#ifndef _WIN32
-        // Instantiate KeyboardHandler explicitly with disabled signal handler inside,
-        // since we have already installed our own signal handler
-        keyboard_handler = std::make_shared<KeyboardHandler>(false);
-#else
-        // We don't have signal handler option in constructor for windows version
-        keyboard_handler = std::shared_ptr<KeyboardHandler>(new KeyboardHandler());
-#endif
-      }
```
christophebedard, line 373: the `rosbag2_transport::Recorder` constructor that we will now be calling (without the `KeyboardHandler`) doesn't have any `#ifndef _WIN32`, it just creates a `std::make_shared<KeyboardHandler>()` if `!record_options.disable_keyboard_controls`: https://github.com/ros2/rosbag2/blob/2f54e8c6bf1365f79998c69478b05adef3ee8bd1/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L791

is that intentional? Just in case this would cause issues.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -615,33 +657,187 @@ PYBIND11_MODULE(_transport, m) {
...
-    "play",
-    py::overload_cast<const rosbag2_storage::StorageOptions &, PlayOptions &>(
-      &rosbag2_py::Player::play),
+    // Deprecated default constructor
+  .def(py::init([]()
+    {
+      PyErr_WarnEx(PyExc_DeprecationWarning, "Player() is deprecated. Use the constructor with "
+          "full configuration parameters instead.", 1);
+      return new rosbag2_py::Player();
+    }), "Deprecated: Use constructor with full options.")
```
christophebedard, line 666: What is the `"Deprecated: Use constructor with full options."` string doing here; where is it going?

rosbag2_py/test/test_transport.py
```diff
@@ -48,43 +48,54 @@ def test_options_qos_conversion():
...
 
     invalid_log_level = 'xxx'
     with pytest.raises(RuntimeError):
-        rosbag2_py.Player(invalid_log_level)
+        rosbag2_py.Player(storage_options, play_options, invalid_log_level)
+
 
+@pytest.mark.parametrize('storage_id', TESTED_STORAGE_IDS)
+def test_recoder_log_level(tmp_path, storage_id):
+    bag_path = tmp_path / 'test_record_cancel'
```
christophebedard:
```suggestion
    bag_path = tmp_path / 'test_recoder_log_level'
```

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: Though, this is supposed to be a deprecation notice for the default `Player()` constructor (without parameters) to be visible in the docstring.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: That is correct behavior. Here is an explanation:
Before we were calling `install_signal_handlers();` then creating CPP Player class instance. 
```cpp
 void play_impl(bool burst = false, size_t burst_num_messages = 0)
  {
    if (!player_) {
      throw std::runtime_error("Player is not initialized. Please use constructor with "
        "storage and play options.");
    }
    install_signal_handlers();
```
The default `keyboard_handler = std::make_shared<KeyboardHandler>();` constructor installs its own signal handlers by overriding others already installed if any. To avoid this, we were disabling its inner signal handler by calling the constructor with a `false` value for the parameter.

However, in the new workflow we always instantiating C++ Player class before we are calling `    install_signal_handlers();` in the ` void play_impl(bool burst = false, size_t burst_num_messages = 0)`.  
The KeybordHandler's signal interception will be overridden in the `play_impl(..)` anyway.  Therefore, it is safe to call a `rosbag2_transport::Player` constructor, which creates `KeyboardHandler` with the default arguments.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: I see

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: makes sense, thanks for the explanation.

christophebedard (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#2047
Gist: https://gist.githubusercontent.com/MichaelOrlov/f9ad7d2dd8defe7db1038a839db35693/raw/1df4c3fd40edd847b94729b639e74cbf5e6237e2/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_py rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_py rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16316
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24092/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18165/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3513/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24708/)

christophebedard:
The ci_linux and ci_linux-rhel jobs had infra failures. You might be able to find the automatically-retried jobs.

MichaelOrlov:
Okay, auto-restarted CI jobs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24096/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3515/)

MichaelOrlov:
The CI is green - merging than.

MichaelOrlov:

PR #2061, opened 2025-07-01
Add send_timestamp to python interface for reading serialized messages

### LINKED ISSUE #2022: Provide a full interface to the serialized message on python
The current read interface for rosbags in python is:
https://github.com/ros2/rosbag2/blob/aee5ae1b0a66fb2b11e72d5c9c39c5e01da16f52/rosbag2_py/src/rosbag2_py/_reader.cpp#L48-L56

But this is missing the send_timestamp.

Which is available to cpp:
https://github.com/ros2/rosbag2/blob/aee5ae1b0a66fb2b11e72d5c9c39c5e01da16f52/rosbag2_cpp/src/rosbag2_cpp/reader.cpp#L69-L72

omsv135:
## Description

Fixes #2022

The cpp interface for reading rosbags provides two timestamps, namely `recv_timestamp` and `send_timestamp`, from the `read_next` function. However, the python interface only returns `recv_timestamp`. This PR extends the Python API to return both timestamps.

Related changes:
- Python stub files are generated as mentioned in [this readme](https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md).
- Relevant tests have also been updated in rosbag2_py.

Yes.
Calling the `rosbag2_py.SequentialReader.read_next` and `rosbag2_py.SequentialCompressionReader.read_next` functions now give a DeprecationWarning indicating use of `read_next_ext` that provides the complete interface.

No.

### Additional Information

- (Encountered while updating tests) There seems to be different behaviors of the writer when using `sqlite3` and `mcap` storage plugins, which causes the `send_timestamp` to have different values. This has been accounted for in the updated tests. This could be removed if tests seems too rigid since it complicates testing.


MichaelOrlov (commented):
@omsv135 Thank you for your contribution.

The reason why `sqlite3` storage plugin always returns **0** for published timestamp is because it doesn't save it internally.

Could you please use more meaningful names for received and published timestamps instead of the `rt, st` when getting return values from the `reader.read_next_ext()`?
For instance in the
```python
(topic, data, rt, st) = reader.read_next_ext()
```

omsv135:
> The reason why sqlite3 storage plugin always returns 0 for published timestamp is because it doesn't save it internally.

Understood.

> Could you please use more meaningful names for received and published timestamps instead of the rt, st when getting return values from the reader.read_next_ext()?

Changed the names - renamed `rt` to `recv_ts` and `st` to `send_ts`, following the names in the serialized messages returned in cpp.

MichaelOrlov (approved):
Thank you. LGTM now.

MichaelOrlov:

MichaelOrlov:
Pulls: ros2/rosbag2#2061
Gist: https://gist.githubusercontent.com/MichaelOrlov/89cb62b115b76d7cdbc51f5549e65ef0/raw/1cddf81058ea6872b272fc0cba4ad99d6400acfa/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py
TEST args:  --packages-above rosbag2_py
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16391
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24180/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18260/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24789/)

EmmanuelMess:
I appreciate I am late, but I feel its important to note that it looks like all the tests for the original `read_next` were removed in favor of testing `read_next_ext`, which might be a bad idea as many scripts still use the original `read_next`.

MichaelOrlov:
@EmmanuelMess Well, we didn't touch logic of the original `read_next()`. It should still work the same way as before.  
However, it has now become deprecated and will be removed in the next release.
At the end it is just a thin wrapper around the same cpp API as in the `read_next_ext(..)`.

PR #2062, opened 2025-07-01
Expose more of the player/recorder API through Python

christophebedard:
## Description

This exposes more of the player's and recorder's API through Python with `rosbag2_py.Player` and `rosbag2_py.Recorder`.

Part of #2046  
Follow-up to #2047

New APIs. No breaking changes.

Yes, GitHub Copilot 1.336.0 with GPT-4o to document some Python APIs (pybind11).

### Additional Information

The current `rosbag2_py.{Recorder,Player}` classes were mostly used for the `ros2 bag {record,play,burst}` verbs. For instance, its `play()` method was more of a utility around the underlying `rosbag2_transport::Player` class' playback functionality and not a trivial `play()` API wrapper: it is a synchronous/blocking function that creates an executor to spin the player node, bursts/plays all messages and waits until it finishes, and also deals with signal handlers.

However, what we want as part #2046 is a full API so that users can programmatically control the player in Python. Therefore, this PR continues what #2047 by making the `rosbag2_py.{Recorder,Player}` classes simple Python wrappers for the important `rosbag2_transport::{Recorder,Player}` APIs.

I kept the existing functionality: ~~I "renamed" the existing _blocking_ `record()`/`play()`/`burst()` methods to `{record,play,burst}_sync()` and updated `ros2 bag {record,play,burst}` to use those.~~ Then eventually we can remove the `rosbag2_py.{Recorder,Player}` constructors that do not take any record/play/storage options and remove the `record()`/`play()`/`burst()` (sync) methods that take record/play/storage options, since they were replaced by the `*_sync()` methods. And then we can also remove all the checks for `!recorder_`/`!player_` everywhere.

For performance reasons, I added a C++ executor to the `rosbag2_py.{Recorder,Player}` classes to spin the underlying recorder/player node. Users can start it with `start_spin()`.

Note: The new simple APIs do not deal with signal handling at all, unlike the old deprecated methods.

In the Player, we added the following APIs:
1. `bool is_paused`
1. `pause()`
1. `bool play_next()`
1. `resume()`
1. `seek(time_point)`
1. `start_spin()`
1. `stop()`
1. `stop_spin()`
1. `bool wait_for_playback_to_finish(timeout)`
1. `bool wait_for_playback_to_finish_exclusively(timeout)`
1. `bool wait_for_playback_to_start(timeout)`
1. `bool wait_for_playback_to_start_exclusively(timeout)`

In the Recorder, we added the following APIs:
1. `bool is_paused`
1. `pause()`
1. `stop()`
1. `stop_spin()`
1. `start_spin()`
1. `stop_spin()`


christophebedard:
Marking this as a draft first to get some higher-level feedback on the implementation. If the solution is deemed to be acceptable, I can either do the same thing for the recorder in this PR or do it in a follow-up PR.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,27 +190,116 @@ class Player
...
+  }
+
+  void seek(int64_t time_point)
+  {
+    if (!player_) {
+      throw std::runtime_error("Player is not initialized. Please use constructor with "
+        "storage and play options.");
+    }
+    player_->seek(static_cast<rcutils_time_point_value_t>(time_point));
+  }
```
christophebedard: As you can see with all the `if (!player_) ...` checks, I feel like we really have two Python `Player` classes now: the existing utility wrapper for `ros2 bag play` and the more trivial wrapper for the full-ish player API (with C++ global context init and C++ executor).

I think it would be better to split them up, but I don't know if we can change the existing API: I would ideally make these the `rosbag2_py.Player` class and move the utility methods for `ros2 bag play` to another class, e.g., `rosbag2_py.PlayUtils` or something.

rosbag2_py/test/test_transport.py
```diff
@@ -126,6 +127,89 @@ def test_record_cancel(tmp_path, storage_id):
...
+
+    assert wait_for(
+        lambda: sub.get_publisher_count() == 1,
+        timeout=rclpy.duration.Duration(seconds=5),
+    ), sub.get_publisher_count()
+
+    player.play()
+    assert player.wait_for_playback_to_start()
+    assert player.play_next()
+    assert player.play_next()
```
MichaelOrlov:
```suggestion
    player.play()
    assert player.play_next()
    assert player.play_next()
```
In the test need to make sure that the regular playback scenario works well without calling
`assert player.wait_for_playback_to_start()`

rosbag2_py/test/test_transport.py
```diff
@@ -126,6 +127,89 @@ def test_record_cancel(tmp_path, storage_id):
...
+
+    play_options = rosbag2_py.PlayOptions()
+    play_options.start_paused = True
+    play_options.topics_to_filter = ['topic']
+
+    player = rosbag2_py.Player(storage_options, play_options, 'info', 'rosbag2_player_test')
+
+    ctx = rclpy.Context()
+    ctx.init()
+    node = rclpy.create_node('test_player_play_pause_resume_seek_stop', context=ctx)
```
MichaelOrlov: Please use a more meaningful name for the node. Maybe `subscribers_node`?

rosbag2_py/test/test_transport.py
```diff
@@ -126,6 +127,89 @@ def test_record_cancel(tmp_path, storage_id):
...
+    player.play()
+    assert player.wait_for_playback_to_start()
+    assert player.play_next()
+    assert player.play_next()
+    assert 2 == player.burst(2)
+    player.seek(0)
+    # This requires that the player node be spun by an executor
+    resume_client.call(Resume.Request(), 5.0)
+    assert not player.play_next()
+    assert player.wait_for_playback_to_finish()
```
MichaelOrlov: Need to use some timeout as a parameter for the `wait_for_playback_to_finish(timeout)` to avoid the test from hangout.

MichaelOrlov (commented):
@christophebedard The approach seems to look good. 
What is unclear to me is how to gracefully handle `SIGINT` and `SIGTERM` on the Python level with the new API.

As regards your other comment about 
> I think it would be better to split them up, but I don't know if we can change the existing API: I would ideally make these the rosbag2_py.Player class and move the utility methods for ros2 bag play to another class, e.g., rosbag2_py.PlayUtils or something.

I agree that it would be good to split them out. However, we can't break the API.  
Otherwise, all scripts that use the `rosbag2_py.Player` class will be broken. At least RQT_Bag will be broken for sure.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,27 +190,116 @@ class Player
...
+    exec_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+    exec_->add_node(player_);
+    spin_thread_ = std::thread(
+      [this]() {
+        exec_->spin();
+      });
+    // Wait for the executor to start spinning to avoid race conditions
+    while (!exec_->is_spinning()) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
```
MichaelOrlov: @christophebedard After some consideration. I have a bit of concern about starting the executor to spin the node in the constructor of the Player.

Before we adhered to the paradigm **"You are not paying for what you are not using"**, but with this change, continuously spinning the player node throughout its whole lifetime, it is a completely opposite approach.
We really do need to spin the player node only when it is in the playback. 

Could you please clarify the rationale for this change?
If it is just for the sake of having the same Player behaviour as a C++ counterpart and making `Player.play()` as a non-blocking call, I wouldn't go this way.
If you are worried about that, in the original design it was needed to run `Player.play()` in a separate thread from the Python side, and we are actually creating one more extra thread above. That is not a problem because:
1. No extra performance burden, those extra thread will sleep and wait on a condition variable until the end of the playback. 
2. From the usability point of view not much difference in comparison to the regular node. Since with regular node the user has to spin the node in a separate thread as well.

Also, making `Player.play()` as a non-blocking call will require changes in the workflow and logic in the already written upstream packages, which are using the Player class. That is more error-prone and has more significant changes.

cc: @emersonknapp - Do you have an opinion on this change?

rosbag2_py/test/test_transport.py
christophebedard: Can you explain why it needs to work without calling `wait_for_playback_to_start()`?

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: I can move this to a `start_spin()` method (with a `stop_spin()` method that automatically gets called in the destructor) so that:

1. Users can use it _if_ they need it
2. It's a C++ executor for previously-mentioned performance reasons (https://github.com/ros2/rosbag2/issues/2046#issuecomment-3020564467)

What do you think?

christophebedard: > Also, making Player.play() as a non-blocking call will require changes in the workflow and logic in the already written upstream packages, which are using the Player class. That is more error-prone and has more significant changes.

I planned to keep the current `play(storage_options, play_options)` (synchronous wrapper, since you want to keep it for `ros2 bag play`) and make `play()` a simple wrapper around `rosbag2_transport::Player::play()` (which changes what you did in #2047). That would probably mean not deprecating `play(storage_options, play_options)` in favour of `play()` (same for constructors), since we want both. Then that wouldn't affect `ros2 bag play`/wouldn't break API at all and we would still have the full player API (#2046).

christophebedard: Maybe I should just create a separate Python wrapper class with a different name, e.g., `rosbag2_py.PlayerCpp` or something. Then we can have both the nice wrapper for `ros2 bag play` and the full player API without having to use a specific constructor depending on which version we want.

rosbag2_py/test/test_transport.py
christophebedard: I changed it to `$testname_node` (with the correct test name), since the node contains a subscription and a client.

rosbag2_py/test/test_transport.py
christophebedard: done

christophebedard:
I've added the equivalent for the recorder.

rosbag2_py/src/rosbag2_py/_transport.cpp
MichaelOrlov: @christophebedard Exposing `start_spin()` and `stop_spin()` method actually a good idea. I assume you will remove the automatic spin from the constructor.

As regards.
> I planned to keep the current play(storage_options, play_options) (synchronous wrapper, since you want to keep it for ros2 bag play) and make play() a simple wrapper around rosbag2_transport::Player::play() (which changes what you did in https://github.com/ros2/rosbag2/pull/2047). That would probably mean not deprecating play(storage_options, play_options) in favour of play() (same for constructors), since we want both. Then that wouldn't affect ros2 bag play/wouldn't break API at all and we would still have the full player API (https://github.com/ros2/rosbag2/issues/2046).

I would like to deprecate old API and approach and safely migrate to the new one via deprecation warnings and **temporarily** keeping backward compatibility with the old one. If we will go with the proposed changes as non-blocking `player.play()` method and explicit calls for the `start_spin()` and `stop_spin()` from Python API, I want to use it everywhere in the Rosbag2 instead of the old approach. Just keep old API for some time to let others some time for migration.

As regards:
> Maybe I should just create a separate Python wrapper class with a different name, e.g., rosbag2_py.PlayerCpp or something. Then we can have both the nice wrapper for ros2 bag play and the full player API without having to use a specific constructor depending on which version we want.

I agree that creating a new class would have a cleaner implementation at least in a short term untill we will remove the old one. 
However, as usual, the naming is hard. I found that `rosbag2_py.PlayerCpp` is a confusing name, and I don't have a good alternative in mind.  
I would propose to go with the current `rosbag2_py.Player`  naming. The implementation would be less clean and have some extra checks, but hopefully, that is a temporary inconvenience. We will remove the old API and relevant logic after a year, i.e., after a next release.

In my opinion, it is currently a good time to make structural and behavioral changes (but keeping the old one for compatability) since I have already changed the constructors and we haven't released it yet.

MichaelOrlov: Or perhaps we can even backport to the Kilted and Jazzy with the deprecation notices and remove the old API from Rollig sooner than waiting for the next release. Maybe half a year or so...

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -190,7 +190,20 @@ class Player
...
+    stop_spin();
+    rclcpp::shutdown();
+  }
 
+  void start_spin()
+  {
+    if (!player_) {
+      throw std::runtime_error("Player is not initialized. Please use constructor with "
+        "storage and play options.");
+    }
```
MichaelOrlov: Need to avoid calling `start_spin` twice if it is already spinning.

rosbag2_py/test/test_transport.py
```diff
@@ -80,6 +81,170 @@ def test_recoder_log_level(tmp_path, storage_id):
...
+    metadata_io = rosbag2_py.MetadataIo()
+    assert wait_for(
+        lambda: metadata_io.metadata_file_exists(str(bag_path)),
+        timeout=rclpy.duration.Duration(seconds=10))
+    metadata = metadata_io.read_metadata(str(bag_path))
+    bag_topics = [
+        topic_info.topic_metadata.name
+        for topic_info in metadata.topics_with_message_count
+    ]
+    assert '/topic' in bag_topics, str(bag_topics)
```
christophebedard, line 214: I believe the metadata file should exist & contain the `/topic` topic after `recorder.stop()`: actually publishing on the topic isn't really required here because the topic will exist in the metadata, since the topic was discovered by the recorder and a writer was created.

I am not checking the _number_ of messages here to avoid races or awkwardly recording & stopping multiple times until the metadata contains the correct number of messages.

MichaelOrlov:
@christophebedard As regards
> I've added the equivalent for the recorder.

With the recorder, things are a bit more complicated.  
In the Apex.AI version, we do not use the executor for the Recorder at all. Instead, we use a `waitset` in the Recorder C++ class directly for the sake of performance. This difference was very well aligned with the old design. 
However, it is difficult to match with the new API. 
Please take a look at the `rosbag2_py::Recorder::Record(..)` implementation in the Apex.AI repo to see if we can somehow workaround the difference in the implementation to have a more common API.

christophebedard:
I think it's fine. Our internal version of the C++ recorder starts a waitset thread in `rosbag2_transport::RecorderImpl::record()` and stops it in `rosbag2_transport::RecorderImpl::stop()`, so we simply won't need an executor thread in the `rosbag2_py.Recorder`. We can just omit the `start_spin()`/`stop_spin()` methods or have empty functions instead.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: see e1fc20bd91a694a73fcb5cdbbeb4976f1d59fedc

christophebedard:
OK, I ended up "renaming" the existing _blocking_ `record()`/`play()`/`burst()` methods to `*_sync()` and updated `ros2 bag {record,play,burst}` to use those. Then eventually we can remove the constructors that do not take any record/play/storage options and remove the `record()`/`play()`/`burst()` (sync) methods that take record/play/storage options, since they were replaced by the `*_sync()` methods. And then we can also remove all the checks for `!recorder_`/`!player_` everywhere.

Please let me know what you think.

christophebedard:
We just had a short offline discussion. I now understand that the intention is to completely move away from the existing blocking wrappers at the Python level (`rosbag2_py`): we want `rosbag2_py.{Recorder,Player}` to be trivial wrappers around the underlying `rosbag2_transport::{Recorder,Player}` APIs (which I've done), _and_ we want to fully get rid of the blocking record/play wrappers.

This means that we do not want the `*_sync()` methods. We want to deprecate the existing record/play methods (which I've done) and move the signal handling logic up to the Python layer, e.g., in the implementation of the `ros2 bag {record,play}` verbs. This will also serve as examples for people wanting to use the `rosbag2_py.{Recorder,Player}` APIs and deal with signals.

rosbag2_py/src/rosbag2_py/_transport.cpp
christophebedard: > and remove the old API from Rollig sooner than waiting for the next release. Maybe half a year or so...

We can deprecate something now in Rolling so that it is deprecated in ROS 2 Lyrical and then remove it from Rolling after Lyrical. However, we can't remove it from Rolling before Lyrical, otherwise we wouldn't follow the usual [tick-tock deprecation strategy](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#deprecation-strategy): it would completely disappear from Kilted to Lyrical.

MichaelOrlov:
@christophebedard Please don't forget to protect `start_spin()/stop_spin()` with mutex to avoid potential race conditions.

MichaelOrlov:
@christophebedard Also one more request. Please bring `Recorder.is_paused`, `Player.is_paused` API to the Python.
It will be very helpful in writing unit tests to verify resume/pause service requests.

christophebedard:
I moved the wait and signal handling logic up to the Python `ros2 bag record` implementation, but I'm still getting this error at the end, and it exits with an exit code == 2:

```
Error. Can't return old SIGINT handler, someone override our signal handler
```

I don't think Python signal handling plays well with underlying C++ signal handling (from the `rclcpp` context). I need to look into it more.

Using Rolling binaries (exits with 0):

```console
$ ros2 bag record --all-topics 
[INFO] [1752270918.036608316] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1752270918.038883718] [rosbag2_recorder]: Listening for topics...
[INFO] [1752270918.038914232] [rosbag2_recorder]: Event publisher thread: Starting
[INFO] [1752270918.039422861] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1752270918.039793752] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
[INFO] [1752270918.039839230] [rosbag2_recorder]: Recording...
[INFO] [1752270919.843539696] [rosbag2_recorder]: Pausing recording.
[INFO] [1752270919.843625814] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1752270919.844991206] [rosbag2_recorder]: Event publisher thread: Exiting
[INFO] [1752270919.845110130] [rosbag2_recorder]: Recording stopped
[INFO] [1752270919.905567514] [rclcpp]: signal_handler(signum=2)
```

Current Python implementation (exits with 2):

```console
$ ros2 bag record --all-topics
[INFO] [1752272576.202435565] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1752272576.206745971] [rosbag2_recorder]: Listening for topics...
[INFO] [1752272576.206776074] [rosbag2_recorder]: Event publisher thread: Starting
[INFO] [1752272576.210327729] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1752272576.211528277] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
[INFO] [1752272576.211598927] [rosbag2_recorder]: Recording...
[INFO] [1752272577.323583349] [rosbag2_recorder]: Pausing recording.
[INFO] [1752272577.323733519] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1752272577.325677799] [rosbag2_recorder]: Event publisher thread: Exiting
[INFO] [1752272577.325749458] [rosbag2_recorder]: Recording stopped
Error. Can't return old SIGINT handler, someone override our signal handler
```

You can see the lack of `[rclcpp]: signal_handler(signum=2)`. Also, when installing the SIGINT/SIGTERM handlers at the Python levels (after the `rosbag2_py.Recorder` gets created, which means after the `rclcpp` context gets initialized and it installs its signal handlers), I think the old signal handler is just the default Python signal handler.

With `rolling` from source (exits with 0):

```console
$ ros2 bag record --all-topics
[INFO] [1752271698.234136792] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1752271698.239362109] [rosbag2_recorder]: Listening for topics...
[INFO] [1752271698.239469616] [rosbag2_recorder]: Event publisher thread: Starting
[INFO] [1752271698.242759309] [rosbag2_recorder]: Subscribed to topic '/rosout'
[INFO] [1752271698.244121880] [rosbag2_recorder]: Subscribed to topic '/events/write_split'
[INFO] [1752271698.244204910] [rosbag2_recorder]: Recording...
[INFO] [1752271699.558165600] [rosbag2_recorder]: Pausing recording.
[INFO] [1752271699.558282898] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1752271699.560208205] [rosbag2_recorder]: Event publisher thread: Exiting
[INFO] [1752271699.560272078] [rosbag2_recorder]: Recording stopped
[INFO] [1752271699.560693461] [rclcpp]: signal_handler(signum=2)
Error. Can't return old SIGINT handler, someone override our signal handler
```

rosbag2_py/test/test_transport.py
MichaelOrlov: Though it seems I was wrong. It is better to wait. However, looking into the Player logic - it seems it should work correctly even without waiting to start

MichaelOrlov:
@christophebedard The error message
> Error. Can't return old SIGINT handler, someone override our signal handler

Is coming from the `KeyboardHandler` class. The following sequnce of events happened in the current implementation
1. We instantiate the Recorder class. 
2. During the construction we called `rclcpp::init(..)` which is overriding signals handling with its own handler. This is why you see `[INFO] [1752271699.560693461] [rclcpp]: signal_handler(signum=2)` message in log.
3. Then we are calling Recorder construtor which is instantiate `KeyboardHandler` class
4. The `KeyboardHandler` class installs its own signal handler because it needs to return terminal to the canonical input mode upon abnormal termination. However, the `KeyboardHandler` class saving already installed signals handlers i.e. in this case from the rclcpp.
5. Then you override signal handling in the Python code
6. When signal SIGINT arrives, it is first processed in the Python code
7. Then trying to call signal handler from the `KeyboardHandler` class as the deffered signal in Python. Which is restoring terminal input to the canonical mode and calling previously saved signal handler from the `rclcpp`.
8. Then when destruction of the Recorder and KeyboardHanlder happened. The `KeyboardHanlder` is trying to return the old signal handler from the rclcpp. But before that he is trying to make sure that nothing got changed and the current signal handler is still installed to be processed in the `KeyboardHanlder`. That is not true anymore because signal handler was overriden from Python code and hasn't been returned back correctly at the moment at least.
9. That is why you see `"Error. Can't return old SIGINT handler, someone override our signal handler"` error message.

MichaelOrlov (commented):
@christophebedard, I think you are overengineering a bit with deferred signal handling on the Python layer. If we don't install signal handlers when calling `rclcpp::init(..)`, the signals will be propagated up to the Python layer, where they can be handled in a native Python style.

I pushed a couple of commits on your branch directly with simplification and a couple of unit tests. Please review them and continue.
- [Add unit tests for testing SIGINT in Python for player and recorder](https://github.com/ros2/rosbag2/pull/2062/commits/dfdb5490417a329839e47a420d2a554ab3524d7b)
- [Don't install signal handler when calling rclcpp::init in recorder/player](https://github.com/ros2/rosbag2/pull/2062/commits/98a89c5c4be0335da18f10f3184c8ad6e969af3a)

I think it will simplify the workflow a lot. I want to avoid handling signals on the Pybind layer and eventually get rid of the old approach with deprecation.
It is a nightmare to make a cross-platform support including Windows for signal handling.

ros2bag/ros2bag/verb/record.py
```diff
@@ -386,10 +387,20 @@ def main(self, *, args):  # noqa: D102
...
         try:
-            recorder.record()
+            with stop_record_cv:
+                stop_record_cv.wait_for(lambda: stop_record)
         except KeyboardInterrupt:
+            stop_record = True
             pass
+        finally:
+            recorder.stop()
+            recorder.stop_spin()
```
christophebedard: This doesn't work and the use of `stop_record` doesn't really make sense.

MichaelOrlov: @christophebedard can you please elaborate why it doesn't work?
It was working for me and tests was passing locally.
As regards that stop_record doesn't make sense.
It is needed just for condition variable.

christophebedard: Here's what I understand based on what you want/don't want and how Python signal handling works, especially when C++ code is involved:

* You want to gracefully handle SIGINT and SIGTERM, just like before
* You want to leave signal handling up to the Python layer and not really do it in C++ in `rosbag2_py` (if possible/simple)
    * You don't want to use something like existing C++ signal handlers in `rosbag2_py` (`install_signal_handlers()`/`uninstall_signal_handlers()`/`process_deferred_signal()`)
* `KeyboardHandler` handles SIGINT but not SIGTERM
* `KeyboardHandler`'s SIGINT handling doesn't always play nice with other signal handlers
    * See this error: `Error. Can't return old SIGINT handler, someone override our signal handler`
    * But we can disable `KeyboardHandler`'s signal handling when we construct it in `rosbag2_py` with `install_signal_handler = false`, similar to how you disabled global `rclcpp` context signal handling
* If we're waiting in Python and we get a Ctrl+C/SIGINT, Python will raise a `KeyboardInterrupt`, which makes it pretty simple. However, Python will _not_ handle SIGINT if we're in C++ code (and releasing the GIL doesn't help with that), which happens with the player when waiting (`wait_for_playback_to_finish()`). We can manually check for signals with [`PyErr_CheckSignals()`](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_CheckSignals) (in the main thread) and make Python handle them, which will result in a `KeyboardInterrupt` error, making SIGINT handling very easy for the Python layer:
    ```cpp
    if (PyErr_CheckSignals() != 0) {
      throw py::error_already_set();
    }
    ```
    * Manually checking for signals gets a bit complicated when we call a blocking function like `wait_for_playback_to_finish()`, but we can split the wait into smaller waits and periodically check for signals
    * For the player, we could do the C++ wait (`wait_for_playback_to_finish()`) in a non-main Python thread to avoid dealing with this for `wait_for_playback_to_finish()`, but then we'd need to make the main thread wait on that using a condition variable or something, which would make the Python layer a bit more complex
* However, the above doesn't handle SIGTERM, so we need a simple signal handler in the Python layer that calls `recorder.stop()` or `play.stop()`

With the commits you pushed, Ctrl+C/SIGINT didn't do anything for the reason I mentioned above (and we can see that tests are failing here). Please take a look at the new changes.

MichaelOrlov: @christophebedard Thank you for your follow-up and updates.  
Indeed, we need to add a proper SIGTERM handling on the Python layer as well as signals propagation from C++ layer to Python if we are releasing GIL.

However, I would like to keep SIGINT handling in the KeyboardHandler class. The reason for that is to avoid terminal input corruption in case if SIGINT is not properly handled on the Python layer. I would like to avoid many users' complaints about screwed-up terminal input after Player or Recorder abnormal termination due to improper signal handling on Python layer. I've pushed a https://github.com/ros2/rosbag2/pull/2062/commits/7cb6825f8817b3d5fcb59728ccb30bdfbac3ca4c commit to returning it. It will align with the current approach because after processing `SIGINT` in the `KeyboardHandler` the KeyboardHandler's signal handler will call "original" previously installed signal handler which is Python signal handler in most cases. 

Also, I have a concern to the current signal handling on the Python layer in the Player, Recorder and Burst entry points for the Rosbag2 CLI. We can't call some complicated constrains like conditional variables or `Player.Stop()` functions directly from the signal handler. It will cause undefined behavior and in some cases dead locks or freezes.
I reworked a bit of signal handling for those parts to make it simpler and call `threading.Event()` primitive variable, which is safe to set up from the signal handler and do a polling check for it with some small timeout in the main Python thread. Please refer to the  https://github.com/ros2/rosbag2/pull/2062/commits/d5bb7640e45f16eaeeb0b0a5fd03f9382cd80395 commit for details. Unfortunately, there are no alternative safe ways to do it. e.g., wait on conditional variable somehow.

I also introduced `wait_for_playback_to_start_exclusively(timeout)` and `wait_for_playback_to_finish_exclusively(timeout)` - which will not do GIL release inside for the sake of the performance consideration and to be used in the Player ros2bag CLI.
Also, I made some cleanups in the `wait_for_and_check_signals(..)`. Please refer to the https://github.com/ros2/rosbag2/pull/2062/commits/9cbd29ca27282db38dd6be1d1dc4d691245931e8 and https://github.com/ros2/rosbag2/pull/2062/commits/a397127a938e1f2adf2689426723049db725455f commits respectively.

Now is your turn to review my commits :stuck_out_tongue:

MichaelOrlov (approved):
LGTM with green CI.

MichaelOrlov:
Pulls: ros2/rosbag2#2062
Gist: https://gist.githubusercontent.com/MichaelOrlov/48394d4691465e421879f1dbb03433c9/raw/8c209472898a208dc87995904799c18514312ee0/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_py ros2bag rosbag2_tests
TEST args:  --packages-above rosbag2_py ros2bag rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16585
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24430/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18499/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3876/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24992/)

ros2bag/ros2bag/verb/record.py
MichaelOrlov: @christophebedard Friendly ping to review my changes here.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -239,6 +382,51 @@ class Player
...
+      // The GIL release doesn't result in Python checking for signals while we're in C++, so we can
+      // do it here to simplify the Python layer: just periodically check for signals while waiting
+      // Wait for a short period of time
+      {
+        py::gil_scoped_acquire gil;
+        // If a signal (Ctrl+C/SIGINT or SIGTERM) has been received, propagate it to Python
+        if (PyErr_CheckSignals() != 0) {
+          throw py::error_already_set();
+        }
+      }
```
christophebedard, line 418: Don't we already have the GIL here?

MichaelOrlov: Well, a better way to explicitly acquire it.

christophebedard: ah okay, as long as it's fine if we already have the lock before this, then all good.

MichaelOrlov:
Re-run CI after adding deprecation notice for Player and Recorder `cancel()` methods
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16608
Rebuilds build #[16585](https://ci.ros2.org/job/ci_launcher/16585)
Running as SYSTEM
Building on the built-in node in workspace /var/lib/jenkins/workspace/ci_launcher
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24457/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18522/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3903/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25016/)

MichaelOrlov:
The CI Linux job failed due to the inner infrastructure issues, and was auto-restarted
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24465/)

MichaelOrlov:
The failed RHEL CI job auto-restarted and passed green
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3915/)

MichaelOrlov:

PR #2105, opened 2025-07-30
Bugfix for callback not called in writer for MESSAGES_LOST event

MichaelOrlov:
## Description

Bugfix for callback not called in writer for MESSAGES_LOST event

Note: There was a mistake in the if-else logic, and the `MESSAGES_LOST` event callback was omitted to install if `WRITE_SPLIT` event callback, also provisioned.

Fixes # (issue) N/A

No.

Yes. I am using GitHub Copilot, GPT-4.1 in my workflow.

### Additional Information
This PR can't be backported because:
1. It is applied bugfix to the code that exists only on Rolling
2. It has ABI-breaking changes by adding virtual methods to the header files..

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/reader.hpp
@@ -206,6 +206,12 @@ public:
    */
   void add_event_callbacks(bag_events::ReaderEventCallbacks & callbacks);
 
+  /**
+   * \brief Check if a callback is registered for the given event.
+   * \return True if there is any callback registered for the event, false otherwise.
+   */
+  [[nodiscard]] bool has_callback_for_event(bag_events::BagEvent event) const;
+
 private:
   std::unique_ptr<reader_interfaces::BaseReaderInterface> reader_impl_;
 };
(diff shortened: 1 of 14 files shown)


fujitatomoya (approved):
thanks for the fix, lgtm with green CI.

@MichaelOrlov just out of curiosity, how did you find this bug? some tests were failing? or you just found it during code scan?

fujitatomoya:
Pulls: ros2/rosbag2#2105
Gist: https://gist.githubusercontent.com/fujitatomoya/22284dde0d636de96de9a57bdf2cb96d/raw/646e965baba044a67a3c4f9e59743da69c5ea45f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16661
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18581/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3989/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25075/)

MichaelOrlov:
> thanks for the fix, lgtm with green CI.
> 
> 
> 
> @MichaelOrlov just out of curiosity, how did you find this bug? some tests were failing? or you just found it during code scan?

Hi @fujitatomoya, thank you for your review.
It was a test failure with messages lost in internal recorder's queue, and in the log there wasn't output about it on the rosbag2_transport layer as I would expect. I start digging into the source code trying to figure out why there was no output in the log and found this if-else mistake.

Issue #2007, opened 2025-05-10
Statistics about the number of lost and recorded messages


MichaelOrlov:
## Description

The current implementation of Rosbag2 does not provide any statistics about the number of lost and recorded messages in real time or during post-processing. To be more specific, the current implementation only prints out to the log at the time when the current bag file closes the number of messages dropped by the Rosbag2 writer in cases where the writer is not able to keep up with the incoming messages, and messages are dropped due to the internal buffer being full.
However, it does not provide any information about the number of messages that were lost on the DDS transport layer. And this information is not available in the bag file itself or in the runtime during recording.

This is a problem because it makes it difficult to assess the quality of the recorded data and to identify potential issues with the recording process.

## Related Issues
n/a

## Completion Criteria

- [ ] **1.** Overall statistics about the number of messages lost and bytes written on a per-topic basis from recording start should be available in real-time by service request.
- [x] **2.** Incremental statistics about the number of messages lost and bytes written on a per-topic basis should be published on a specific topic when messages are lost.
- [ ] **3.** Statistics about the number of lost messages and bytes written to the current bag file on a per-topic basis should be saved in metadata at the end of the current bag recording.
- [ ] **4.** The `ros2 bag info` command with `--verbose` option shall be extended to include the statistics about the number of lost messages on a per-topic basis by reading this information from saved metadata

## Implementation Notes / Suggestions

Please refer to the design document in the [Add design doc for "Statistics about messages lost during recording" feature #2006](https://github.com/ros2/rosbag2/pull/2006)

## Testing Notes / Suggestions

WIll need to figure out how to emulate messages lost on the DDS transport layer. We might need to use mocking in unit tests.

PR #2144, opened 2025-08-16
Add RecorderEventNotifier class

MichaelOrlov:
## Description

Add recorder event notifier class, which will be responsible for sending events on predefined topics about bag split and statistics about messages lost on transport and inner recorder level.

This PR is needed as part of the #2007 
>  2. Incremental statistics about the number of messages lost and bytes written on a per-topic basis should be published on a specific topic when messages are lost.

The `RecorderEventNotifier` class implemented as an extended replacement of the Messages Lost Notifier described in the 
[statistics_about_lost_and_recorded_messages.md](https://github.com/ros2/rosbag2/blob/7f304136062740cc718502dc4a855f79aa0571a3/docs/design/statistics_about_lost_and_recorded_messages.md?plain=1#L76C1-L108) design document.
```markdown
### 4. Messages Lost Notifier

   Will need to add a new `MessagesLostNotifier` class to the `rosbag2_transport::recorder` that 
   will publish on a predefined topic, incremental statistics about the number of lost messages 
   and bytes written on a per-topic basis when messages lost happened.
   - To facilitate this, the Rosbag2 recorder shall satisfy the following requirements:
      1. The messages lost event shall be published from the Rosbag2 recorder on a dedicated topic.
      e.g., `/events/rosbag2_messages_lost`
      2. The messages lost event shall include per-topic statistics with the number of lost 
        messages. The message type could be named as `rosbag2_interfaces::msg::MessagesLostEvent`
        and defined in the `rosbag2_interfaces` package. The message type will be similar to the 
        `rosbag2_interfaces::msg::WriteSplitEvent` message type and shall contain the following 
         fields:
         - `node_name` (string): The name of the node that is recording.
         - Array of the following fields:
         - `topic_name` (string): The name of the topic on which the messages were lost.
         - `messages_lost_in_transport` (uint64_t): The number of messages lost since the last 
           event on a DDS transport layer.
         - `messages_lost_in_recorder` (uint64_t): The number of messages lost since the last 
           event in the Rosbag2 recorder.
```

Besides the functionality for the message lost notification on a dedicated topic, the `RecorderEventNotifier` also encapsulates functionality for other event notifications, such as notifications about bag split. 
Notice that notifications about bag splits were built into the `RecorderImpl` class itself. Now that functionality was moved with minimum changes to the `RecorderEventNotifier` class. A queue was added for the bag_split events to avoid missing notifications when events happened one-to-one, with minimum delay.

**Note: In this PR, the `RecorderEventNotifier` doesn't implement the core functionality of the `Messages Lost Notifier` described in the [statistics_about_lost_and_recorded_messages.md](https://github.com/ros2/rosbag2/blob/rolling/docs/design/statistics_about_lost_and_recorded_messages.md) design document. However, it is a one-to-one replacement of the current implementation for the logic of what we have in the `RecorderImpl` class. That was made on purpose to be able to backport this PR to Kilted and Jazzy with minimum changes to reduce divergencies in the code base.
We will be able to backport the part related to the moving bag split event notification and callbacks for messages lost in the transport layer. However, the part related to the callbacks from the writer about lost messages will be omitted since the underlying API in the writer is missing.
The rest of the missing functionality for the `Messages Lost Notifier` will be implemented in follow-up PRs.**

Fixes # (issue) N/A
- Relates  #2007

No.

### Additional Information
- Can be backported. Please refer to the notes above.
- It was considered to make one universal `EventNotifier` class for Player and Recorder. However, this design option was not accepted because events are very specific to the Player and Recorder classes, and it will be a cleaner implementation if they are split into separate classes. Having a common `EventNotifierBase` class also will not give much benefits, but rather will introduce extra layers of dependencies, which is unnecessary in this particular case.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder_event_notifier.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder_event_notifier.cpp
@@ -16,147 +16,61 @@
 #include "rclcpp/node.hpp"
 
 #include "rosbag2_transport/recorder_event_notifier.hpp"
+#include "recorder_event_notifier_impl.hpp"
 
 namespace rosbag2_transport
 {
 
 RecorderEventNotifier::RecorderEventNotifier(rclcpp::Node * node)
-: node(node)
 {
-  if (!node) {
-    throw std::invalid_argument("Node pointer cannot be null");
-  }
-  split_event_pub_ =
-    node->create_publisher<rosbag2_interfaces::msg::WriteSplitEvent>("events/write_split", 1);
-
-  // Start the thread that will publish events
-  {
-    std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
-    event_publisher_thread_should_exit_ = false;
-    event_publisher_thread_ = std::thread(&RecorderEventNotifier::event_publisher_thread_main,
-        this);
-  }
+  pimpl_ = std::make_unique<RecorderEventNotifierImpl>(node);
 }
 
 RecorderEventNotifier::~RecorderEventNotifier()
 {
-  if (event_publisher_thread_.joinable()) {
-    {
-      std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
-      event_publisher_thread_should_exit_ = true;
-    }
-    event_publisher_thread_wake_cv_.notify_all();
-    event_publisher_thread_.join();
-  }
+  // Explicitly reset the pimpl_ to ensure the destructor of RecorderEventNotifierImpl is called
+  // only once.
+  pimpl_.reset();
 }
 
-void RecorderEventNotifier::event_publisher_thread_main()
-{
-  RCLCPP_INFO(node->get_logger(), "Event publisher thread: Started");
-  while (!event_publisher_thread_should_exit_.load()) {
-    std::unique_lock<std::mutex> pub_thread_lock(event_publisher_thread_mutex_);
-
-    event_publisher_thread_wake_cv_.wait_for(pub_thread_lock, msgs_lost_statistics_update_period_,
-      [this]() {
-        return write_split_has_occurred_ || event_publisher_thread_should_exit_;
-        }
-    );
-
-    if (write_split_has_occurred_) {
-      write_split_has_occurred_ = false;
-      auto message = rosbag2_interfaces::msg::WriteSplitEvent();
-      message.closed_file = bag_split_info_.closed_file;
-      message.opened_file = bag_split_info_.opened_file;
-      message.node_name = node->get_fully_qualified_name();
-      try {
-        split_event_pub_->publish(message);
-      } catch (const std::exception & e) {
-        RCLCPP_ERROR_STREAM(
-          node->get_logger(),
-          "Failed to publish message on '/events/write_split' topic. \nError: " << e.what());
-      } catch (...) {
-        RCLCPP_ERROR_STREAM(
-          node->get_logger(),
-          "Failed to publish message on '/events/write_split' topic.");
-      }
-    }
-
-//    {
-//      // TODO(morlov): Check if we need to publish statistics about messages lost events
-//      std::unique_lock<std::mutex> statistics_lock(per_topic_messages_lost_statistics_mutex_);
-//      for (const auto &[topic, lost_stats] : per_topic_messages_lost_statistics_) {
-//        const auto &[transport_lost, recorder_lost] = lost_stats;
-//        // Use topic, transport_lost, and recorder_lost to publish statistics if needed
-//      }
-//    }
-  }
-  RCLCPP_INFO(node->get_logger(), "Event publisher thread: Exited");
-}
 
 void RecorderEventNotifier::on_bag_split_in_recorder(
   const rosbag2_cpp::bag_events::BagSplitInfo & bag_split_info)
 {
-  {
-    std::lock_guard<std::mutex> lock(event_publisher_thread_mutex_);
-    bag_split_info_ = bag_split_info;
-    write_split_has_occurred_ = true;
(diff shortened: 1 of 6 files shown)


MichaelOrlov:
@christophebedard, @fujitatomoya I would appreciate your review here.

MichaelOrlov:
Pulls: ros2/rosbag2#2144
Gist: https://gist.githubusercontent.com/MichaelOrlov/0023bd074c23476d281e860b019ce139/raw/ba99cf9cffeb1510530a782cac7cd3e09d9e6263/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16752
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24653/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18678/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4103/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25175/)

MichaelOrlov:
Re-run Windows CI job after attempt to fix build errors.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25176/)

rosbag2_transport/src/rosbag2_transport/recorder_event_notifier.cpp
```diff
@@ -0,0 +1,80 @@
...
+RecorderEventNotifier::RecorderEventNotifier(rclcpp::Node * node)
+{
+  pimpl_ = std::make_unique<RecorderEventNotifierImpl>(node);
+}
+
+RecorderEventNotifier::~RecorderEventNotifier()
+{
+  // Explicitly reset the pimpl_ to ensure the destructor of RecorderEventNotifierImpl is called
+  // only once.
+  pimpl_.reset();
```
kylemarcey, line 33: @MichaelOrlov why do you need to explicitly reset this instead of just letting the unique_ptr dtor do it automatically?

rosbag2_transport/src/rosbag2_transport/recorder_event_notifier_impl.hpp
```diff
@@ -0,0 +1,236 @@
...
+  {
+    {
+      std::unique_lock<std::mutex> pub_thread_lock(event_publisher_thread_mutex_);
+      if (update_rate_hz == 0.0f) {
+        disable_publishing_msgs_lost_statistics_ = true;
+        RCLCPP_DEBUG(node->get_logger(), "Messages lost statistics publishing is disabled");
+      } else if (update_rate_hz > 0.0f) {
+        disable_publishing_msgs_lost_statistics_ = false;
+        msgs_lost_stats_max_publishing_period_ =
+          std::chrono::milliseconds(static_cast<int>(1000 / update_rate_hz));
```
kylemarcey, line 85: @MichaelOrlov is there any validation of `update_rate_hz`? If it is set (probably mistakenly) to more than 1000, the period will be 0 and I think you will busy loop.

rosbag2_transport/src/rosbag2_transport/recorder_event_notifier.cpp
MichaelOrlov: Yeah, it was a surprise for me too.
The simple fist test
```cpp
TEST_F(TestRecorderEventNotifier, default_ctor_dtor)
{
  auto node = std::make_unique<rclcpp::Node>("test_recorder_event_notifier");
  auto notifier = std::make_unique<RecorderEventNotifier>(node_.get());
  // Destructor should not throw
  ASSERT_NO_THROW(notifier.reset());
}
```
Was catching a segmentation fault when exiting from test. i.e. when `notifier` was trying to be destructed since it goes out of scope.
I've tried to debug this phenomenon and it turns out that after the first destruction call via 
```cpp
  ASSERT_NO_THROW(notifier.reset());
```
The inner `pimpl_` is not `nullptr` and still points to the same memory address as before the first destruction call.
And we got a double destruction of the same object.

rosbag2_transport/src/rosbag2_transport/recorder_event_notifier_impl.hpp
MichaelOrlov: @kylemarcey Good catch I will add an extra check and throw an exception

kylemarcey:
@MichaelOrlov ok, LGTM

MichaelOrlov:
@christophebedard @fujitatomoya This PR has already been reviewed. Can some of you additionally approve it to be able to merge it?

MichaelOrlov:
Pulls: ros2/rosbag2#2144
Gist: https://gist.githubusercontent.com/MichaelOrlov/a9f3377b3931536f159c2f7d1f258fef/raw/ba99cf9cffeb1510530a782cac7cd3e09d9e6263/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16768
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24673/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18699/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4121/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25194/)

MichaelOrlov:
The failed Linux CI job was auto-restarted due to the CI infrastructure failure and passed green
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24675/)

christophebedard (approved):

MichaelOrlov:
https://github.com/Mergifyio backport kilted

PR #2166, opened 2025-09-05
Replace zstd_vendor with zstd_cmake_module

ahcorde:
Part of https://github.com/ros2/ros2/issues/1729

Replace `zstd_vendor` with `zstd_cmake_module`


ahcorde:
Pulls: ros2/rosbag2#2166
Gist: https://gist.githubusercontent.com/ahcorde/2ddc068be9ea541eac6f0dea963d71b0/raw/405435ff6411f82a5b93845e64506c3d156fcfbb/ros2.repos
BUILD args:  --packages-above-and-dependencies zstd_cmake_module
TEST args:  --packages-above zstd_cmake_module
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16880
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24794/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18821/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4245/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25307/)

zstd_cmake_module/package.xml
```diff
@@ -1,25 +1,24 @@
...
+  <exec_depend>libzstd-dev</exec_depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
   </export>
-</package>
+</package>
```
MichaelOrlov: @ahcorde General comment. Please don't forget to make sure that there at least one empty line at the end of file. GitHub/GitLab seems not to like it when there is no empty new line at the end of the file.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18850/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4272/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25336/)

ahcorde:

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24836/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18866/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4286/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25350/)

MichaelOrlov (changes requested):
As far as I understand, the whole purpose of removing vendor packages was to reduce the number of supported packages in the core ROS 2.  

But changes like this, where we are removing a vendor package and adding another one, `zstd_cmake_module,` at the same time, defeat the original goal.

Replacing vendor packages with another  `*_cmake_module` package is generally a bad idea.

For instance, if in the future ROS needs to support another build system like `BAZEL`, we will need to get back to the vendor packages and remove this CMake-tailored workaround as a CMake module package.

In Apex.AI, we have moved ahead and fully support builds and tests with `Bazel` on multiple platforms, like QNX and configurations like Yocto. Vendor packages help us a lot to manage differences in the versioning for third-party dependencies.

I would oppose changes like those in this PR in general unless someone provides evidence that they are better than what we already have and will align with future needs.

cc: @tfoote @clalancette

clalancette:
My opinion in general is that if there is a concrete use case for a vendor package, we should keep it, though we should evaluate on a case-by-case basis.

That said, I also think that the ROS 2 core should *not* attempt to vendor every dependency.  Even before these changes, we don't vendor (at least): `bullet`, `eigen`, `asio`, `qt5`, `pyqt`, `git`, `pkg-config`, `python3-empy`, `python3-pytest`, `python3-setuptools`, `acl`, `opencv`, `openssl`, `python3-lark`, `flake8`, `lttng`, many more python packages that I'm glossing over, and probably a handful that I am missing.

In this case, it sounds like there is a concrete use case, and we wouldn't really be getting rid of a package, so I'm fine with rejecting this PR.  It's probably worthwhile to spend some time in one of the ROS PMC meeting going over the vendored list in detail and figuring out which ones we can remove.

MichaelOrlov:
I have a concern in general, unrelated to this concrete PR.  
I see a pattern and general approach for removing one vendor package; we are creating another `name_cmake_module` package.  
That is not really removal of the vendor package - it is a replacement of the vendor package with the CMake module package at the end.

Can we somehow optimize and not create separate CMake module packages for each, removing vendor package?  
Maybe aggregate those CMake-related rules and workarounds somewhere in one place/package?

clalancette:
> Can we somehow optimize and not create separate CMake module packages for each, removing the vendor package? 

I don't think this is going to be necessary in most cases.  Nowadays most C/C++ packages either have CMake find modules available via CMake, via their devel packages, or via another method (often the distribution packages).

> Maybe aggregate those CMake-related rules and workarounds somewhere in one place/package?

Sigh.  We used to have that in ROS 1: https://github.com/ros/cmake_modules .  But I think the general consensus was that it was hard to find, and it is better to have individual packages rather than have a "dumping ground" for CMake-related things.  That's a decision that could be revisited, but again, I don't think we'll need a huge amount of them.

For instance, right now the custom CMake modules for external dependencies we have in the core are:

1.  Eigen3 (https://github.com/ros2/eigen3_cmake_module) - There are CMake files available for it on both Ubuntu 24.04 and RHEL-9.  They seem to be missing in the pixi package, but we may be able to convince @j-rivero to fix that, in which case we could remove it completely.
2. TinyXML2 (https://github.com/ros2/tinyxml2_vendor) - This one can likely be removed with the switch to pixi; there are CMake files available for it on Ubuntu 24.04, RHEL-9, and Windows/Pixi
3. ZStd (this PR) - It looks like there are CMake modules available for zstd in both Ubuntu 24.04 and in Windows/Pixi, but missing in RHEL-9.  Maybe we could convince @cottsay to look at this and maybe update the RHEL-9 package?  If so, we could remove this one.
4.  Pybind11 (https://github.com/ros2/pybind11_vendor) - It looks like there are CMake modules available on Ubuntu 24.04, RHEL-9, and Windows/Pixi.  We can probably remove this one.
5. Libyaml (https://github.com/ros2/libyaml_vendor) - This one is missing CMake modules everywhere, so we would need to keep this.
6. LZ4 (in this repository) - This one is missing CMake modules everywhere, so we would need to keep this.

So it looks to me like we'd need to do a bit of work in upstream packaging, but after that, we would only have two modules; `libyaml_cmake_modules` and `lz4_cmake_modules`.

Thus, in my opinion, we don't currently have much of a need for a centralized place.  But pinging @sloretz for thoughts.

j-rivero:
> Eigen3 (https://github.com/ros2/eigen3_cmake_module) - There are CMake files available for it on both Ubuntu 24.04 and RHEL-9. They seem to be missing in the pixi package, but we may be able to convince @j-rivero to fix that, in which case we could remove it completely.

I'm all ears if you have a good offer :beer: :beer: . Looking into the Conda package, I see some files in the [version 3.4.0 on win-64](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Feigen-3.4.0-h91493d7_0.conda) , I see listed in as installed by the package:

```
....
Library/share/eigen3/cmake/Eigen3Config.cmake
Library/share/eigen3/cmake/Eigen3ConfigVersion.cmake
Library/share/eigen3/cmake/Eigen3Targets.cmake
Library/share/eigen3/cmake/UseEigen3.cmake
...
```
Do we need something more?

clalancette:
> I'm all ears if you have a good offer 🍺 🍺 .

I'll buy you a beer anytime :).


> ```
> ....
> Library/share/eigen3/cmake/Eigen3Config.cmake
> Library/share/eigen3/cmake/Eigen3ConfigVersion.cmake
> Library/share/eigen3/cmake/Eigen3Targets.cmake
> Library/share/eigen3/cmake/UseEigen3.cmake
> ...
> ```
> 
> Do we need something more?

Nope!  I somehow missed this when looking at the package locally.  In that case, we can probably remove `eigen3_cmake_module` completely.  Thanks for taking a look.

zstd_cmake_module/package.xml
```diff
@@ -1,23 +1,22 @@
...
   <maintainer email="michael.orlov@apex.ai">Michael Orlov</maintainer>
-  <maintainer email="geoff@openrobotics.org">Geoffrey Biggs</maintainer>
-  <maintainer email="michel@ekumenlabs.com">Michel Hidalgo</maintainer>
   <maintainer email="me@emersonknapp.com">Emerson Knapp</maintainer>
   <maintainer email="ros-tooling@googlegroups.com">ROS Tooling Working Group</maintainer>
   <license>Apache License 2.0</license>  <!-- the contents of this package are Apache 2.0 -->
-  <license>BSD</license>  <!-- Zstd is BSD -->
 
-  <url type="website">https://facebook.github.io/zstd/</url>
+  <url type="website">https://github.com/lz4/lz4/</url>
```
MichaelOrlov:
```suggestion
  <url type="website">https://facebook.github.io/zstd/</url>
```

zstd_cmake_module/package.xml
```diff
@@ -1,23 +1,22 @@
 <?xml version="1.0"?>
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
-  <name>zstd_vendor</name>
+  <name>zstd_cmake_module</name>
   <version>0.33.0</version>
```
MichaelOrlov, line 5:
```suggestion
  <version>0.1.0</version>
```

MichaelOrlov:
@clalancette Thank you for your answer. At the end, it seems it doesn't look so scary.
Let's try to summarize and clarify some things.
It seems we will need at least `libyaml_cmake_modules`,  `lz4_cmake_modules` and this `zstd_cmake_module`.  
However, I forgot to mention that this `zstd_vendor` package is linked to the 1.5.5 version for Noble and has a patch for it that is missing in the newly created `zstd_cmake_module`. If we go ahead with this PR, we will have a new warning on CI, which is going to be annoying.  
Also, I don't see any advantages in replacing the `zstd_vendor` package with the `zstd_cmake_module` since the resulting `CMakeList.txt ` file is almost the same among missing patches and find for the original zstd library. I would prefer to keep the original `zstd_vendor` package in this case.

Also forgot to mention about `mcap_vendor` package, which is more complicated than others since it has dependencies from the zstd, lz4 and doesn't have the original `CMakeList.txt` file and plus to that we also have a patch for it. - It is going to be much more fun replacing it with the CMake module package.

zstd_cmake_module/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 3.20)
+project(zstd_cmake_module)
+
+find_package(ament_cmake REQUIRED)
+
```
MichaelOrlov, line 5: I may have misunderstood how this scheme should work, but shouldn't we have somewhere at least
```
find_package(zstd REQUIRED)
```
?

MichaelOrlov (changes requested):

zstd_cmake_module/CMakeLists.txt
sloretz: `zstd_cmake_mdule` wouldn't need to find `zstd` because it only exposes a find module. The two important files here are `zstd_cmake_module/zstd_cmake_module-extras.cmake` and `zstd_cmake_module/cmake/Modules/Findzstd.cmake`. A user of this package would call

```cmake
# First find zstd_cmake_module so that the extras file adds the find module to CMAKE_MODULE_PATH
find_package(zstd_cmake_module REQUIRED)
# Find zstd, which will use the find module
find_package(zstd REQUIRED)
```

sloretz (dismissed):
~~Do we still need the zstd find module? I see it comes with a zstdConfig.cmake at least on debian and ubuntu. Seems like we should be able to `find_package(zstd)` on its own.~~ **Edit**: Oops, missed that Chris already looked at this and found RHEL 9 is missing the projectConfig.cmake

https://packages.debian.org/bookworm/amd64/libzstd-dev/filelist

> ZStd (this PR) - It looks like there are CMake modules available for zstd in both Ubuntu 24.04 and in Windows/Pixi, but missing in RHEL-9. Maybe we could convince @cottsay to look at this and maybe update the RHEL-9 package? If so, we could remove this one.

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-september-23-2025/50232/1

zstd_cmake_module/CMakeLists.txt
MichaelOrlov: @sloretz Thank you for the explanation. In this case, we need to make sure that in other places where we were using `zstd_vendor` present `find_package(zstd REQUIRED)`
cc: @ahcorde

zstd_cmake_module/package.xml
MichaelOrlov: The version corresponds to the bloom release, and we do bloom release for all packages in the rosbag2 folder. Let's keep the same versioning, i.e., 0.33.0 numbers, as in other packages.

zstd_cmake_module/CMakeLists.txt
ahcorde: `mcap_vendor` and `rosba2_compression` already include this

MichaelOrlov (approved):
LGTM with green CI.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25056/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19118/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4534/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25509/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25058/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19120/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4535/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25511/)

ahcorde:
Pulls: ros2/rosbag2#2166
Gist: https://gist.githubusercontent.com/ahcorde/b401ddfa8c223853a662fb78dc191ccc/raw/748d19872fb778089c0fdb32397a7e2255fd0dd3/ros2.repos
BUILD args:  --packages-above-and-dependencies zstd_cmake_module
TEST args:  --packages-above zstd_cmake_module
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17079
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25060/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19124/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25513/)

PR #2171, opened 2025-09-12
Fixes for multiple race conditions in player

MichaelOrlov:
## Description

This PR addresses multiple possible race conditions in the player. (Not all of them, but at least bunch of them)

1. Got rid of the extra while loop in the "play_messages_from_queue()".
Reasoning: The logic will still be correct. After exiting from the "play_messages_from_queue()", we will check if we need to repeat from the beginning and restart "play_messages_from_queue()" again after init. All other calls, such as `play_next()` and `seek()`, will preserve their logic and work the same way as before.
2. Fix for race condition with `seek()` at the end of the playback.
3. Wait for `Player->play()` to be called in the `wait_for_playback_to_start()`
Reasoning: To avoid freeze if wait_for_playback_to_start(timeout) is called with a timeout, and we haven't called `play()` yet or if `play()` has already been finished.
4. Fix for the possible deadlock in the `PlaySrvsTest::stop_in_active_play` unit test
Unlock `calls_counter_update_mutex` before calling `stop()` to avoid a deadlock in the callback if we happen to call `stop()` after the next message started being processed in the callback. This is to avoid flakiness in the test and fix the ThreadSanitizer warning.
 
Fixes # (issue) N/A

Fixes for possible undefined behaviour when calling `seek()` or `play_next()` at the very end of the bag when playback transitions from the very end to the beginning in the loop or just finishing playback.

### Additional Information
N/A

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -683,7 +683,7 @@ bool PlayerImpl::play()
       is_in_playback_cv_.notify_all();
 
       // If we get here and still have/just got a play next request, make sure to notify play_next()
-      // After that, requests will be automatically rejected since is_in_playback_ is false
+      // After that, requests will be automatically rejected since is_in_playback_ become false
       if (play_next_.exchange(false)) {
         std::lock_guard<std::mutex> lk(finished_play_next_mutex_);
         finished_play_next_ = true;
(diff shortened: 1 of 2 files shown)


rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -682,7 +683,7 @@ bool PlayerImpl::play()
       is_in_playback_cv_.notify_all();
 
       // If we get here and still have/just got a play next request, make sure to notify play_next()
-      // After that, requests will be automatically rejected since is_in_playback_ is false
+      // After that, requests will be automatically rejected since is_in_playback_ become false
```
fujitatomoya: nit??

```suggestion
      // After that, requests will be automatically rejected since is_in_playback_ becomes false
```

fujitatomoya (approved):
lgtm, and this PR claims no user-facing changes.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: Fixed

MichaelOrlov:
Pulls: ros2/rosbag2#2171
Gist: https://gist.githubusercontent.com/MichaelOrlov/1a14f8cd13595e28e42a343fb35fc686/raw/88ded79fe7ee252cf307cbeed97b18410458ef3a/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17024
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24953/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19049/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4428/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25455/)

MichaelOrlov:

PR #2172, opened 2025-09-13
Make the player respect the original messages' order with the same timestamp

### LINKED ISSUE #2168: No messages are read with --read-ahead-queue-size 1
## Description
When a rosbag is played back with `--read-ahead-queue-size 1`, no messages are played.  This results from the calculation
```
auto queue_lower_boundary =
    static_cast<size_t>(play_options_.read_ahead_queue_size * read_ahead_lower_bound_percentage_);
```
which rounds to 0 in the case of `play_options_.read_ahead_queue_size == 1`

This could be fixed by handling the edge case, setting the lower bound to 1 when it rounds to 0.

### Expected Behavior
A single message is queued at a time, and messages are played on their respective topics.

### Actual Behavior
The player hangs, playing no messages, as it continues waiting for messages to be loaded from the bag but the loading thread never retrieves them.

## To Reproduce
Play back a rosbag with
`ros2 bag play --read-ahead-queue-size 1 <bag_file>`

 - OS: Ubuntu Jammy
 - ROS 2 Distro:  Humble onwards (at least), but more important from Jazzy
 - Install Method: N/A
 - Version: N/A

### LINKED ISSUE #2173: Player does not preserve the order of messages with the same timestamp
## Description
Rosbag2 player does not preserve the order of messages with the same timestamp

### Expected Behavior
If two messages have the same received timestamp (for example, due to recording while using sim time) they will be played back in the same order they were written to the mcap

### Actual Behavior
The order of messages played back is undefined.

## To Reproduce
4. Logs print error '...' **

 - OS: Ubuntu 22.04
 - ROS 2 Distro: Rolling, Kilted
 - Install Method: Build from sources
 - Version: bfce7acfd88d1d8e21e83843209ac7a97cd807fe

### RCA
The Rosbag player was [changed in late 2024](https://github.com/ros2/rosbag2/pull/1848) to use a `PriorityQueue` instead of a simple (lock-free) queue, in order to handle multiple readers. This has the side effect that messages with the same `received_timestamp`, even from the same channel/bag may get reordered during playback. This is most often an issue for bags recorded with simulation time enabled.

### Possible solution
The rosbag's `LockedPriorityQueue` stores the loaded sequence number of each element in the priority queue and uses it in the compare function to break ties. However, it is not fully clear how to handle this for multiple bags.

MichaelOrlov:
## Description

Make the player respect the original messages' order with the same timestamp.
Note: The messages with the same `recv_timestamps` are a common situation when doing a recording with `sim_time`

- Added insertion sequence number to the LockedPriorityQueue to ensure that it can provide a stable runtime sort algorithm.

- Fixes #2173

Yes. Playback will respect the originally saved messages' order even if messages have the same timestamps. 
However, playback from multiple bags can't preserve the original order of messages with the same timestamps in different bags because we don't know which one was first during recording, although the order is at least the same for each playback.

### Additional Information
N/A

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -357,14 +357,9 @@ private:
   Player * owner_;
   rosbag2_transport::PlayOptions play_options_;
   rcutils_time_point_value_t play_until_timestamp_ = -1;
-  using BagMessageComparator = std::function<
-    bool(
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> &,
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> &)>;
-  LockedPriorityQueue<
-    rosbag2_storage::SerializedBagMessageSharedPtr,
-    std::vector<std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t>>,
-    BagMessageComparator> message_queue_;
+  LockedPriorityQueue<rosbag2_storage::SerializedBagMessageSharedPtr> message_queue_;
+  using BagMessageComparator =
+    LockedPriorityQueue<rosbag2_storage::SerializedBagMessageSharedPtr>::Comparator;
   mutable std::future<void> storage_loading_future_;
   std::atomic_bool load_storage_content_{true};
   std::unordered_map<std::string, rclcpp::QoS> topic_qos_profile_overrides_;
@@ -418,16 +413,10 @@ private:
   static inline const struct
   {
     bool operator()(
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> & l,
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> & r) const
+      const rosbag2_storage::SerializedBagMessageSharedPtr & l,
+      const rosbag2_storage::SerializedBagMessageSharedPtr & r) const
     {
-      const auto & [l_msg, l_insertion_seq_num] = l;
-      const auto & [r_msg, r_insertion_seq_num] = r;
-      if (l_msg->recv_timestamp == r_msg->recv_timestamp) {
-        // Earlier insertion come first
-        return l_insertion_seq_num > r_insertion_seq_num;
-      }
-      return l_msg->recv_timestamp > r_msg->recv_timestamp;  // Smaller timestamp come first
+      return l->recv_timestamp > r->recv_timestamp;  // Smaller timestamp comes first
     }
   } bag_message_chronological_recv_timestamp_comparator;
 
@@ -435,16 +424,10 @@ private:
   static inline const struct
   {
     bool operator()(
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> & l,
-      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> & r) const
+      const rosbag2_storage::SerializedBagMessageSharedPtr & l,
+      const rosbag2_storage::SerializedBagMessageSharedPtr & r) const
     {
-      const auto & [l_msg, l_insertion_seq_num] = l;
-      const auto & [r_msg, r_insertion_seq_num] = r;
-      if (l_msg->send_timestamp == r_msg->send_timestamp) {
-        // Earlier insertion come first
-        return l_insertion_seq_num > r_insertion_seq_num;
-      }
-      return l_msg->send_timestamp > r_msg->send_timestamp;  // Smaller timestamp come first
+      return l->send_timestamp > r->send_timestamp;  // Smaller timestamp comes first
     }
   } bag_message_chronological_send_timestamp_comparator;
 };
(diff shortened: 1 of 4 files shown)


MichaelOrlov:
cc: @DaleKoenig

rosbag2_transport/src/rosbag2_transport/player.cpp
```diff
@@ -418,21 +418,33 @@ class PlayerImpl
...
+      const std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t> & r) const
     {
-      return l->recv_timestamp > r->recv_timestamp;
+      const auto & [l_msg, l_insertion_seq_num] = l;
+      const auto & [r_msg, r_insertion_seq_num] = r;
+      if (l_msg->recv_timestamp == r_msg->recv_timestamp) {
+        // Earlier insertion come first
+        return l_insertion_seq_num > r_insertion_seq_num;
+      }
+      return l_msg->recv_timestamp > r_msg->recv_timestamp;  // Smaller timestamp come first
```
christophebedard: Nothing critical, but it seems weird to me that the insertion sequence number belongs to the `LockedPriorityQueue`, but that the user (`Player`) still has to consider it in the comparator it provides. I would expect it to be completely managed by either the queue or the user, not both.

Could the `LockedPriorityQueue` not take care of considering the sequence number itself when ordering? For example, by wrapping the user-provided comparator with an internal comparator that checks if left and right are equal (usually by checking if `compare(left, right)` and `compare(right, left)` give the same result), in which case it orders with the insertion sequence number.

Or we could avoid modifying `LockedPriorityQueue` and just have the `Player` deal with the insertion sequence number by just creating a `LockedPriorityQueue<T = std::pair<rosbag2_storage::SerializedBagMessageSharedPtr, size_t>`.

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -554,6 +554,164 @@ TEST_P(RosBag2PlayTestFixtureMessageOrder, recorded_msgs_are_played_for_all_topi
...
+    };
+  const auto callback = [&](std::shared_ptr<rosbag2_storage::SerializedBagMessage> playing_msg) {
+      // Make sure messages are played in order
+      num_played_messages++;
+      const auto timestamp = get_timestamp(playing_msg);
+      using MessageT = typename decltype(msg)::element_type;
+      const auto deserialized_msg = deserialize_test_message<MessageT>(playing_msg);
+      // The int32_value was set in an increasing order when creating the messages
+      // std::cout << "deserialized_value = " << deserialized_msg->int32_value << ", timestamp = "
+      //   << RCUTILS_NS_TO_MS(timestamp) << ", topic_name = `" << playing_msg->topic_name << "`\n";
```
christophebedard, line 617: Should these lines be deleted?

rosbag2_transport/test/rosbag2_transport/test_play.cpp
```diff
@@ -554,6 +554,164 @@ TEST_P(RosBag2PlayTestFixtureMessageOrder, recorded_msgs_are_played_for_all_topi
...
```
christophebedard, line 702: Should these lines be deleted?

rosbag2_transport/test/rosbag2_transport/test_play.cpp
MichaelOrlov: I would like to keep it. It will be very useful to quickly debug the failing test.

rosbag2_transport/test/rosbag2_transport/test_play.cpp
MichaelOrlov: I would like to keep it. It will be very useful to quickly debug the failing test.

rosbag2_transport/src/rosbag2_transport/player.cpp
MichaelOrlov: @christophebedard I simplified the user API for the comparator by using an inner wrapper for a user-defined comparator. 
However, with the price that the underlying queue type is hard-coded now to be a vector, and the queue will always return the earliest added element in case of equality of the comparing elements. 
I wasn't able to find a way how to generalize these elements.
I don't think that we will ever need to change them, at least the underlying container type.

MichaelOrlov:
Pulls: ros2/rosbag2#2172
Gist: https://gist.githubusercontent.com/MichaelOrlov/e5c50201d418c76d1b8b52e7d8c1eaad/raw/ef3b73d955f50d17110225dd5014a8f403966e61/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17035
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24994/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19070/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25472/)

christophebedard (approved):

rosbag2_transport/src/rosbag2_transport/player.cpp
christophebedard: yeah, I agree. LGTM!

christophebedard:
I found the job that was retriggered for ci_linux-aarch64, but the ci_windows job wasn't automatically retriggered, so I did it manually.

MichaelOrlov:

PR #2158, opened 2025-09-01
Fix for multibag replay stagnation

MichaelOrlov:
## Description

This PR fixes a problem for slow playback when one reader has a lower frequency topic than others

- Added cache for the last element from each reader and taking messages from cache in a chronological order instead of the round-robin order.
- To mitigate "spaggetish" code after the fix, this PR also introduces the `ReadersWrapper` class, which will encapsulate readers with their options and message cache.
- The `ReadersWrapper` class manages multiple `rosbag2_cpp::Reader` instances, allowing for chronological reading of messages across all readers. It maintains a cache of the next message from each reader and provides methods to retrieve the next message in chronological order, seek to a specific timestamp across all readers, and apply filters to all readers. It also provides access to the earliest and latest timestamps across all readers.

- Fixes #2156
- Depends on the #2157 

### Is this user-facing behavior change?
Fixed a bug where playback of the higher frequency topics can slow down when present slow frequency topics in another playing bag.

### Did you use Generative AI?
Yes. I am using GitHub Copilot, GPT-4.1, in my workflow. in particular to help with unit tests and documentation.

### Additional Information
This PR can be backported to the other distros.

PR #2182, opened 2025-09-20
[kilted] Follow-up on "Fix for multibag replay stagnation (#2158)"

MichaelOrlov:
## Description

- Fix for returning a dangling reference from the "Player::get_storage_options()" method
- Also deprecate "Player::get_storage_options()" method in favor of the "get_all_storage_options()"

Fixes # (issue) N/A

No. Because the method is in a protected section and was used only in the unit tests before.

Not in this case. :smile: 

### Additional Information
Note:  The 'get_storage_options()' is a method of the Player class and resides in the protected section. It was originally designed to be used in the unit tests and was recently superseded by the `Player::get_all_storage_options()` method

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -381,6 +381,7 @@ protected:
 
   /// \brief Getter for the first of the currently stored storage options
   /// \return Reference to the first item in the StorageOptions vector
+  [[deprecated("Use rosabg2_transport::Player::get_all_storage_options() instead")]]
   ROSBAG2_TRANSPORT_PUBLIC
   const rosbag2_storage::StorageOptions & get_storage_options();
 
--- a/rosbag2_transport/include/rosbag2_transport/player.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/player.hpp
@@ -381,7 +381,6 @@ protected:
 
   /// \brief Getter for the first of the currently stored storage options
   /// \return Reference to the first item in the StorageOptions vector
-  [[deprecated("Use rosabg2_transport::Player::get_all_storage_options() instead")]]
   ROSBAG2_TRANSPORT_PUBLIC
   const rosbag2_storage::StorageOptions & get_storage_options();
 
--- a/rosbag2_transport/src/rosbag2_transport/player.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/player.cpp
@@ -241,7 +241,7 @@ public:
   size_t get_number_of_registered_on_play_msg_post_callbacks();
 
   /// \brief Getter for the first of the currently stored storage options
-  /// \return Copy of the first of the currently stored storage options
+  /// \return Reference to the first of the currently stored storage options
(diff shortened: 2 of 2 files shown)


rosbag2_transport/include/rosbag2_transport/player.hpp
```diff
@@ -381,6 +381,7 @@ class Player : public rclcpp::Node
 
   /// \brief Getter for the first of the currently stored storage options
   /// \return Reference to the first item in the StorageOptions vector
+  [[deprecated("Use rosabg2_transport::Player::get_all_storage_options() instead")]]
```
fujitatomoya: we usually do not do this for the already-released distribution in the policy of https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#deprecation-strategy? as you mentioned, this is protected method, but could be used by the application already? in that case, we are breaking our own rule... i am not saying we never do this, but would like to know how others feel... 🤔 @christophebedard wdyt?

christophebedard: I don't have a strong opinion, but I would err on the side of caution and simply not deprecate it.

MichaelOrlov: I can remove the deprecation. However, I wouldn't print an error since the method is functional as before.  
I put a deprecation notice as a fair warning not to use it since we will remove it sometime anyway, and an alternative is already present in the target distro.

christophebedard: I still wouldn't mark it as deprecated like this; I think the only valid thing to do would be to report an error at runtime if `get_storage_options()` is called when `get_all_storage_options().size() > 1` .

MichaelOrlov: Okay. I removed the deprecation warning.  
However, I don't think that we will need to print an error if we have more than one storage option.
The Doxygen comment pretty much clears what this function does and what would be the expected behavior 
```cpp
  /// \brief Getter for the first of the currently stored storage options
  /// \return Reference to the first item in the StorageOptions vector
```
and behavior is consistent with expectations.

christophebedard (approved):

MichaelOrlov:
Pulls: ros2/rosbag2#2182
Gist: https://gist.githubusercontent.com/MichaelOrlov/9cd49c4f8aa4e2d3c0d56555952f4cea/raw/3d7a2f5627d6876ce57a8ae704d404a4994c2d14/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17043
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25019/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19079/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4493/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25475/)

MichaelOrlov:

PR #2218, opened 2025-10-30
Feature: Circular logging by split count (--max-bag-files)

### LINKED ISSUE #2217: Feature: Add circular logging (by size/duration) to ros2 bag record
## Description

This feature introduces parameters to `ros2 bag record` to control recording file size or duration limits, enabling both circular logging.

* **`--max-record-size`:** Sets a maximum total size (in bytes) for all bag files. When this limit is exceeded, the oldest file(s) are deleted to maintain the limit. This enables continuous, size-bounded circular logging.
* **`--max-record-duration`:** Sets a maximum total duration (in seconds) spanning all bag files. When this limit is exceeded, the oldest file(s) are deleted. This enables continuous, time-bounded circular logging.
* [optional] **`--max-splits`:** Stops the recording session automatically after a specified number of split files have been created. This is a stopping condition, not a circular logging mechanism.

## Related Issues

* Fulfills common requests for circular/continuous logging (for size and duration). 
* https://github.com/ros2/rosbag2/issues/865
* https://github.com/ros2/rosbag2/issues/1915

## Completion Criteria

* `ros2 bag record` accepts `--max-record-size <SIZE_BYTES>`. When total size is exceeded, the oldest files are deleted, and recording continues.
* `ros2 bag record` accepts `--max-record-duration <DURATION_SEC>`. When total duration is exceeded, the oldest files are deleted, and recording continues.
* All new options are documented.

* Logic for monitoring total size, duration, and split file count should be implemented within the `SequentialWriter` class.
* The `SequentialWriter` must include a mechanism to identify and delete the oldest split files when `--max-record-size` or `--max-record-duration` limits are breached.
* To limit checking and simplify implementation, checking of size and duration only happens when splitting files.

* **`--max-record-size` Test:** Record with a small size limit and file splitting. Verify that total size remains bounded and old files are deleted.
* **`--max-record-duration` Test:** Record with a short duration limit and file splitting. Verify that total duration remains bounded and old files are deleted.
* **`--max-splits` Test:** Record with splitting and `--max-splits 5`. Verify recording stops after exactly 5 files.

luke-alloy:
## Description

Adds circular logging by split count, record size and duration to rosbag2 recording.

- CLI: adds ~~`--max-splits`~~ `--max-bag-files`
- Writer: deletes oldest files when exceeding split count

Fixes #2217

Yes.
- Older recordings are deleted once max split is reached:
  - ~~`--max-splits N`~~ `--max-bag-files`: limit number of retained bagfile splits.
- Examples:
  - `ros2 bag record -a --max-bag-size 100MB --max-bag-files 5`

Yes.
- Tool/model: GPT-5
- Used for: naming suggestions, CLI/help text wording, and drafting PR description.
- Code impact: assisted in scaffolding changes (option wiring, param names); final code implementation was completed, built, run, and manually reviewed and tested locally.

### Additional Information

- The max split limit is checked together after each bag split
- The writer logs a warning when deleting files due to limits. Adjust `--log-level` if needed.
- `metadata.yaml` currently retains aggregate `message_count` and `topics_with_message_count` across the entire capture, even when older bagfiles are pruned by circular logging. File lists and durations reflect pruning, but those two fields still represent the full session totals.
- Circular logging changes are only in sequential_writer.cpp (deleting the oldest file if needed); the rest involve adding StorageOptions parameters, updating the shell UI, and adjusting related unit tests.


.DS_Store
fujitatomoya, line 1: what is this file? should be removed?

ros2bag/ros2bag/verb/record.py
```diff
@@ -159,6 +159,23 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
...
+        help='Maximum total size in bytes for the entire recording across all splits. '
+             'Oldest bagfiles will be deleted when this limit is exceeded (circular logging). '
+             'Default: %(default)d, no limit on total recording size. '
+             'Requires --max-bag-size or --max-bag-duration to be set.')
+    parser.add_argument(
+        '--max-record-duration', type=int, default=0,
+        help='Maximum total duration in seconds for the entire recording across all splits. '
+             'Oldest bagfiles will be deleted when this limit is exceeded (circular logging). '
+             'Default: %(default)d, no limit on total recording duration. '
+             'Requires --max-bag-size or --max-bag-duration to be set.')
```
fujitatomoya: i am not against this PR. but i am not sure if those 2 optional arguments need to be supported by rosbag2... because i think those can be easily calculated by `--max-splits` option above? i think that using `--max-splits` with `--max-bag-size` or `--max-bag-duration` would be more straight-forward for user experience.

besides that,

- (already) too many optional arguments for `ros2 bag record`... this generates more and more complication.
- additional error handling needs to be implemented in the rosbag2 system. e.g) what if user accidentally set `--max-record-size` with smaller value than `--max-bag-size`.

.DS_Store
luke-alloy: Yes. This will be removed. I didn't notice it snuck through.

luke-alloy:
I've committed changes that should address the unit test error. I'm also happy to include @fujitatomoya ’s suggestion to remove `--max-record-size` and `--max-record-duration`, pending confirmation to proceed.

ros2bag/ros2bag/verb/record.py
MichaelOrlov: I didn't have time yet to thoroughly review this PR, but I agree with @fujitatomoya on
> that using --max-splits with --max-bag-size or --max-bag-duration would be more straight-forward for user experience.

Also agree that we need to avoid extra parameters as much as possible in the Rosbag2 CLI - there are already too many of them.

luke-alloy: @fujitatomoya @MichaelOrlov removed `max-bag-size` and `max-bag-duration` flag

ros2bag/ros2bag/verb/record.py
```diff
@@ -159,6 +159,11 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
              'Default: %(default)d, recording written in single bagfile and splitting '
              'is disabled. If both splitting by size and duration are enabled, '
              'the bag will split at whichever threshold is reached first.')
+    parser.add_argument(
+        '--max-splits', type=int, default=0,
+        help='Maximum number of splits to retain before deleting the oldest bagfile '
+             '(circular logging). Default: %(default)d, unlimited. '
+             'Requires --max-bag-size or --max-bag-duration to be set.')
```
MichaelOrlov, line 174:
```suggestion
             'Requires --max-bag-size or --max-bag-duration to be set or usage of the bag split '
             'via direct recorder API or service calls.')
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -326,9 +327,19 @@ void SequentialWriter::switch_to_next_storage()
...
+    ROSBAG2_CPP_LOG_WARN_STREAM(
+      "File index counter has overflowed (wrapped to 0). "
+      "This should not happen in practice, but continuing with index 0. "
+      "If circular logging is enabled, ensure old files are deleted to avoid conflicts.");
+  }
+
   storage_options_.uri = format_storage_uri(
     base_folder_,
-    metadata_.relative_file_paths.size());
+    next_file_index_);
```
MichaelOrlov: This assignment would be better readable in one line
```suggestion
storage_options_.uri = format_storage_uri(base_folder_, next_file_index_);
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -416,6 +427,8 @@ void SequentialWriter::write(std::shared_ptr<const rosbag2_storage::SerializedBa
   if (!storage_options_.snapshot_mode && should_split_bagfile(message_timestamp)) {
     split_bagfile();
     metadata_.files.back().starting_time = message_timestamp;
+    // Delete oldest files if circular buffer limit exceeded (after split creates new file)
+    delete_oldest_files_if_needed();
```
MichaelOrlov:
```suggestion
```
I think it would be better to move this delete_oldest_files call into the `void SequentialWriter::switch_to_next_storage()` right before `storage_->update_metadata(metadata_);`. This way we can avoid extra `storage_->update_metadata(metadata_);` call inside `delete_oldest_files_if_needed();`
Please be aware that our `update_metadata(metadata_)` in mcap is a bit suboptimal, and each time update is called, we are writing a new section with metadata instead of really updating the one.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
MichaelOrlov, line 352:
```suggestion
  // Delete the oldest files if circular buffer limit exceeded (after split creates new file)
  delete_oldest_files_if_needed();
  storage_->update_metadata(metadata_);  
```
"Moved from after the split"

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -499,6 +512,50 @@ bool SequentialWriter::should_split_bagfile(
...
+  // Protect metadata access with mutex to prevent race conditions
+  std::lock_guard<std::mutex> lock(topics_info_mutex_);
+
+  // Delete oldest files until we're under the split-count limit
+  // Note: We just split, so we have at least 2 files and the oldest is definitely not current
+  while (metadata_.files.size() > 1) {
+    // If we're under the limit, stop deleting
+    if (metadata_.files.size() <= storage_options_.max_splits) {
+      break;
+    }
```
MichaelOrlov:
```suggestion
   // Delete the oldest files until we're under the split-count limit
  while (metadata_.files.size() <= storage_options_.max_splits) {
```
Simplify logic.
Note: if `storage_options_.max_splits` were `storage_options_.max_bag_files`, the logic would be easy to understand.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -499,6 +512,50 @@ bool SequentialWriter::should_split_bagfile(
...
+    // Remove from metadata
+    metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
+    metadata_.files.erase(metadata_.files.begin());
+    metadata_.starting_time = metadata_.files.front().starting_time;
+  }
+
+  // Update metadata after deletions
+  if (!metadata_.files.empty()) {
+    storage_->update_metadata(metadata_);
+  }
```
MichaelOrlov:
```suggestion
```
Metadata update should happen once in the `SequentialWriter::switch_to_next_storage()`

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -39,6 +39,11 @@ struct StorageOptions
   // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_duration = 0;
 
+  // The maximum number of bagfile splits to retain before the oldest bagfile is deleted.
+  // A value of 0 indicates that no deletion based on split count will occur.
+  // This feature is only available if either max_bagfile_size or max_bagfile_duration is set.
```
MichaelOrlov:
```suggestion
  // This feature is only available when the bag split is active.
  // Requires --max-bag-size or --max-bag-duration to be set or usage of the bag split via
  // direct recorder API or service calls.set.
```

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -39,6 +39,11 @@ struct StorageOptions
...
+  uint64_t max_splits = 0;
```
MichaelOrlov:
```suggestion
  uint64_t max_bag_files = 0;
```
The naming is hard. I would propose to rename this parameter to `max_bag_files` to avoid confusion and multiple user questions, like:
- Will splits stop after reaching `max_splits`?
- Why is there one file when `max_splits = 1`? I would expect to have at least two files after the split.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -557,6 +557,7 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_msg_comp
   }
 }
 
+
```
MichaelOrlov:
```suggestion
```
The stray empty line should be removed.

MichaelOrlov (changes requested):
@luke-alloy Thank you for your contribution. This is a long-time-wanted feature that will be very useful.

I've made a thorough review, and overall implementation looks good to me. However, I have a few suggestions:
1. The naming is hard. I would propose to rename `--max-splits` parameter to `--max_bag_files` to avoid confusion and multiple user questions, like:
> - Will splits stop after reaching `max_splits`?
> - Why is there one file when `max_splits = 1`? I would expect to have at least two files after the split.
2. I would like to see at least one integration test with real files writing and deletion.
  To facilitate this, we do have a special `TemporaryDirectoryFixture`. Please consider modifying your tests or adding another one. I recently added a similar test in one of my PR. Please, refer to the https://github.com/ros2/rosbag2/pull/2224/commits/81ef9fdb1e0a8e997feff195fce0267e45ee18f8#diff-c14a7d44179cf64868ce86b5e60c7f741f957ef368c04a0477ccfe7ccd89f1f2R138-R164 as an example from the https://github.com/ros2/rosbag2/pull/2224.

luke-alloy:
> @luke-alloy Thank you for your contribution. This is a long-time-wanted feature that will be very useful.
> 
> I've made a thorough review, and overall implementation looks good to me. However, I have a few suggestions:
> 
> 1. The naming is hard. I would propose to rename `--max-splits` parameter to `--max_bag_files` to avoid confusion and multiple user questions, like:
> 
> > * Will splits stop after reaching `max_splits`?
> > * Why is there one file when `max_splits = 1`? I would expect to have at least two files after the split.
> 
> 2. I would like to see at least one integration test with real files writing and deletion.
>    To facilitate this, we do have a special `TemporaryDirectoryFixture`. Please consider modifying your tests or adding another one. I recently added a similar test in one of my PR. Please, refer to the [81ef9fd#diff-c14a7d44179cf64868ce86b5e60c7f741f957ef368c04a0477ccfe7ccd89f1f2R138-R164](https://github.com/ros2/rosbag2/commit/81ef9fdb1e0a8e997feff195fce0267e45ee18f8#diff-c14a7d44179cf64868ce86b5e60c7f741f957ef368c04a0477ccfe7ccd89f1f2R138-R164) as an example from the #2224.

@MichaelOrlov Thanks for the detailed review! I will update the PR soon.

I completely agree regarding the naming confusion—good catch. To maintain consistency with the other CLI flags, would `--max-bag-files` (using hyphens instead of _) be the preferred convention?

MichaelOrlov:
@luke-alloy Yes. Need to use `--max-bag-files` (using hyphens instead of _). 
Sorry for the confusion. It was my mistake, I copied and pasted the name from another post where I was referring to the storage_options

ros2bag/ros2bag/verb/record.py
luke-alloy: Updated.

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
luke-alloy: Removed.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Updated.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Thank you for this suggestion. Moved.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Removed.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Moved.

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
luke-alloy: Updated

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
luke-alloy: Updated.

luke-alloy:
Committed changes to address (most) review comments.
Will address the last comment (ie, add integration test with real file writing and deletion) next.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Good catch! Simplified.

luke-alloy:
@MichaelOrlov Thanks for the review. I've pushed fixes addressing the comments. Ready for a re-review!

MichaelOrlov:
@luke-alloy Thank you for the follow-up.
As regards
> metadata.yaml currently retains aggregate `message_count` and `topics_with_message_count` across the entire capture, even when older bagfiles are pruned by circular logging. File lists and durations reflect pruning, but those two fields still represent the full session totals.

I think it is a sort of inconsistency and would be considered as a bug.
I would suggest adjusting `message_count` and `topics_with_message_count` accordingly when old files are deleted.

Update:
I also think that there are no values in knowing the total messages in a session if messages are not available.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -1108,6 +1108,103 @@ TEST_F(SequentialWriterTest, split_event_calls_on_writer_close)
...
+}
+
+TEST_P(
+  ParametrizedTemporaryDirectoryFixture,
+  circular_logging_deletes_oldest_files_with_real_storage)
+{
+  // Integration test: verify max_bag_files actually deletes files from disk
+  const uint64_t max_bag_files = 3;
+  const size_t total_files_to_create = 6;
+  std::string topic_name = "topic";
```
MichaelOrlov:
```suggestion
  std::string topic_name = "circular_bag_topic";
```

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -1108,6 +1108,103 @@ TEST_F(SequentialWriterTest, split_event_calls_on_writer_close)
...
+    "cdr",
+    {},
+    ""
+  },
+  {
+    "test_msgs/msg/BasicTypes",
+    "ros2msg",
+    "bool bool_value",
+    ""
+  });
```
MichaelOrlov: Nitpick. A bit of simplification
```suggestion

  rosbag2_storage::TopicMetadata topic_metadata {
    0U, topic_name, "test_msgs/BasicTypes", "cdr", {}, ""
  };
  writer.create_topic(topic_metadata);```

File: rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp [hunk 12]
```diff
@@ -1108,6 +1108,103 @@ TEST_F(SequentialWriterTest, split_event_calls_on_writer_close)
...
+    if (i > 0) {
+      writer.split_bagfile();
+    }
+
+    auto msg = std::make_shared<rosbag2_storage::SerializedBagMessage>();
+    msg->topic_name = topic_name;
+    msg->recv_timestamp = static_cast<rcutils_time_point_value_t>(i * 100);
+    msg->send_timestamp = msg->recv_timestamp;
+    uint32_t data = static_cast<uint32_t>(i);
+    msg->serialized_data = rosbag2_storage::make_serialized_message(&data, sizeof(data));
```
MichaelOrlov: We do have a `make_test_msg();` helper function
```suggestion
    auto msg = make_test_msg();
    msg->topic_name = topic_name;
    msg->recv_timestamp = static_cast<rcutils_time_point_value_t>(i * 100);
    msg->send_timestamp = msg->recv_timestamp;```

File: rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp [hunk 13]
```diff
@@ -1108,6 +1108,103 @@ TEST_F(SequentialWriterTest, split_event_calls_on_writer_close)
...
+  writer.close();
+
+  // Count bag files on disk
+  size_t file_count = 0;
+  for (const auto & entry : fs::directory_iterator(storage_options.uri)) {
+    const auto ext = entry.path().extension().string();
+    if (ext == ".mcap" || ext == ".db3") {
+      file_count++;
+    }
+  }
```
MichaelOrlov, line 1209: Nitpick:
```suggestion
  const std::string expected_ext = (GetParam() == "mcap") ? ".mcap" : ".db3";
  for (const auto & entry : fs::directory_iterator(storage_options.uri)) {
    const auto ext = entry.path().extension().generic_string();
    if (ext == expected_ext) {
      file_count++;
    }
  }
```

MichaelOrlov (commented):
@luke-alloy Thank you one more time.

I've made one more round of review. Overall looks good to me, with one concern regarding adjustments of the `message_count` and `topics_with_message_count` in metadata when we are deleting old files, and a few nitpicks findings in the newly added unit test.

MichaelOrlov:
@luke-alloy Can you please also regenerate Python stub files (.pyi) according to the https://github.com/ros2/rosbag2/blob/rolling/rosbag2_py/README.md ?
The [Test rosbag2 / build_and_test (pull_request)](https://github.com/ros2/rosbag2/actions/runs/19905278049/job/57336583407?pr=2218) fials due to this.

luke-alloy:
> @MichaelOrlov Thanks for the review. I've pushed fixes addressing the comments. Ready for a re-review!

Thanks for the feedback! I agree this should be fixed.

**The challenge:** `FileInformation` only stores total `message_count` per file, not per-topic counts. So fixing `topics_with_message_count` requires tracking per-file per-topic counts somewhere.

**Two options:**

| Option | Approach | Trade-off |
|--------|----------|-----------|
| **A** | Add `topic_message_counts` to `FileInformation` | Persisted, but requires metadata version bump (9→10), YAML serialization changes, and touches readers/info tools |
| **B** | Track per-file per-topic counts in-memory in `SequentialWriter` | Simpler & localized, but lost on restart |

**Plan:** I'll go with **Option B** since circular logging is a runtime feature anyway—if the process restarts, the session ends. Smaller blast radius too. Can extend to A later if needed.

Let me know if you'd prefer Option A.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
luke-alloy: Updated.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
luke-alloy: Updated.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
luke-alloy: Updated.

MichaelOrlov:
@luke-alloy Thank you for the follow-up. I agree with proposed option **B**.

luke-alloy:
Hi @MichaelOrlov, just following up — Option B was implemented back in December along with the `.pyi` stub regeneration. I've also rebased on the latest `rolling` today. Would you be able to take another look when you get a chance? Happy to address any further feedback.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -340,6 +342,7 @@ void SequentialWriter::switch_to_next_storage()
   file_info.path = strip_parent_path(storage_->get_relative_file_path());
   metadata_.files.push_back(file_info);
   metadata_.relative_file_paths.push_back(file_info.path);
+  per_file_topic_message_counts_.push_back({});  // Initialize tracking for new file
```
MichaelOrlov: It makes more sense to reset the message counts list via 
```cpp
per_file_topic_message_counts_.clear();
```
rather than pushing back a dummy element
```cpp
per_file_topic_message_counts_.push_back({});  // Initialize tracking for new file
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -505,8 +509,10 @@ bool SequentialWriter::should_split_bagfile(
 
 void SequentialWriter::delete_oldest_files_if_needed()
 {
-  // Only delete if bag file count limit is set
-  // Note: This is only called after split_bagfile(), so max_bagfile_size is guaranteed to be set
+  // Only delete if bag file count limit is set.
+  // Note: This is only called after split_bagfile(), so max_bagfile_size is guaranteed to be set.
+  // This function also adjusts per-topic message counts in topics_names_to_info_, which are
+  // used to compute metadata.message_count and topics_with_message_count during finalization.
```
MichaelOrlov: Please move these notes into the Doxygen comment above the function declaration in the header file.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -194,6 +194,11 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
   // Tracks the next file index to use for sequential numbering (independent of file deletions)
   uint64_t next_file_index_ {1};
 
+  // Per-file per-topic message counts for accurate adjustment during circular deletion.
+  // Each element corresponds to a file in metadata_.files (same index).
+  // Used to decrement topics_names_to_info_ counts when oldest files are pruned.
+  std::vector<std::unordered_map<std::string, size_t>> per_file_topic_message_counts_;
```
MichaelOrlov: I would highly recommend using `std::deque` instead of `vector` to avoid shifting elements on deletion. With deque, it is possible to do `push_back()` and `pop_front()` with O(1) complexity. Anyway, we don't need random access to the elements. We have access to either the front or back element, or iterate over the whole range.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -191,6 +191,17 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
 
   rosbag2_storage::BagMetadata metadata_;
 
+  // Tracks the next file index to use for sequential numbering (independent of file deletions)
+  uint64_t next_file_index_ {1};
+
+  // Per-file per-topic message counts for accurate adjustment during circular deletion.
+  // Each element corresponds to a file in metadata_.files (same index).
+  // Used to decrement topics_names_to_info_ counts when oldest files are pruned.
+  std::vector<std::unordered_map<std::string, size_t>> per_file_topic_message_counts_;
```
MichaelOrlov: I would highly recommend using std::deque instead of vector to avoid shifting elements on deletion. With deque, it is possible to do push_back() and pop_front() with O(1) complexity. Anyway, we don't need random access to the elements. We have access to either the front or back element, or iterate over the whole range.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -314,9 +317,17 @@ void SequentialWriter::switch_to_next_storage()
   }
 
   storage_->update_metadata(metadata_);
```
MichaelOrlov, line 320: Need to add finalize_metadata(); before calling the storage_->update_metadata(metadata_); to update metadata_.topics_with_message_count before saving to the file.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -334,6 +345,10 @@ void SequentialWriter::switch_to_next_storage()
   file_info.path = strip_parent_path(storage_->get_relative_file_path());
   metadata_.files.push_back(file_info);
   metadata_.relative_file_paths.push_back(file_info.path);
+  per_file_topic_message_counts_.push_back({});  // Initialize tracking for new file
+
+  // Delete oldest files if circular buffer limit exceeded (after new file is added)
+  delete_oldest_files_if_needed();
 
   storage_->update_metadata(metadata_);
```
MichaelOrlov, line 355: Need to add finalize_metadata(); before calling the storage_->update_metadata(metadata_); to update metadata_.topics_with_message_count before saving to the file.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -334,6 +345,10 @@ void SequentialWriter::switch_to_next_storage()
   file_info.path = strip_parent_path(storage_->get_relative_file_path());
   metadata_.files.push_back(file_info);
   metadata_.relative_file_paths.push_back(file_info.path);
+  per_file_topic_message_counts_.push_back({});  // Initialize tracking for new file
```
MichaelOrlov: Nitpick: Could be simplified a bit to use emplace back.
```suggestion
  per_file_topic_message_counts_.emplace_back();  // Initialize tracking for new file
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -496,6 +512,51 @@ bool SequentialWriter::should_split_bagfile(
...
+      if (it != topics_names_to_info_.end()) {
+        it->second.message_count -= count;
+      }
+    }
+    per_file_topic_message_counts_.erase(per_file_topic_message_counts_.begin());
+
+    // Remove from metadata
+    metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
+    metadata_.files.erase(metadata_.files.begin());
+    metadata_.starting_time = metadata_.files.front().starting_time;
```
MichaelOrlov: Need to gracefully handle errors on deletion and don't increment counters on errors.
```suggestion
     // Delete file from filesystem
    if (fs::exists(file_path)) {
      const auto file_size = fs::file_size(file_path);
      const auto file_duration_ns = oldest_file.duration.count();
      std::error_code ec;
      bool file_removed = fs::remove(file_path, ec);
      if (!file_removed || ec) {
        ROSBAG2_CPP_LOG_ERROR("Failed to delete oldest bagfile: %s. Error: %s",
                              file_path.generic_string().c_str(), ec.message().c_str());
      } else {
        ROSBAG2_CPP_LOG_INFO("Deleted oldest bagfile: %s (%lu bytes, %lu ns)",
                             oldest_file.path.c_str(), file_size, file_duration_ns);
      }
      // Adjust per-topic message counts to reflect only retained files
      const auto & oldest_topic_counts = per_file_topic_message_counts_.front();
      for (const auto & [topic_name, count] : oldest_topic_counts) {
        auto it = topics_names_to_info_.find(topic_name);
        if (it != topics_names_to_info_.end()) {
          auto & topic_info = it->second;
          topic_info.message_count -= count;
        }
      }
      per_file_topic_message_counts_.erase(per_file_topic_message_counts_.begin());

      // Remove from metadata
      metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
      metadata_.files.erase(metadata_.files.begin());
      metadata_.starting_time = metadata_.files.front().starting_time;
    }
```

MichaelOrlov (commented):
@luke-alloy Thank you for addressing my previous findings and suggestions.  
Please see my new comments inline. Mostly requests for some minor improvements.

I also found it is difficult to see the delta and do code review on the branch in the IDE like CLion or VSCode when the rolling branch was merged multiple times above the changes on the development branch. Next time, I would recommend doing a real rebase on top of rolling. To simplify this process, you can squash all your commits on the branch before doing a rebase or create a new branch on top of rolling and transfer changes via a patch created from all your prior commits at once.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Done.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
luke-alloy: Done — also replaced `.erase(.begin())` with `.pop_front()`.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Done — added before both `update_metadata()` calls.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Done.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Done

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Done — adopted suggested approach.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -197,9 +198,15 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
...
+  std::deque<std::unordered_map<std::string, size_t>> per_file_topic_message_counts_;
 
-  // Delete oldest bagfiles if split count exceeds limit
+  /**
+   * Delete oldest bagfiles if the number of files exceeds the max_bag_files limit.
+   *
+   * Only called after split_bagfile(), so max_bagfile_size is guaranteed to be set.
+   * Adjusts per-topic message counts in topics_names_to_info_, which are used to
+   * compute metadata.message_count and topics_with_message_count during finalization.
+   */
```
MichaelOrlov: Can you please use a C++ style Doxygen commets like `///` instead of `/**..*/`

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -496,6 +514,55 @@ bool SequentialWriter::should_split_bagfile(
...
+          auto & topic_info = it->second;
+          topic_info.message_count -= count;
+        }
+      }
+      per_file_topic_message_counts_.pop_front();
+
+      // Remove from metadata
+      metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
+      metadata_.files.erase(metadata_.files.begin());
+      metadata_.starting_time = metadata_.files.front().starting_time;
```
MichaelOrlov: This block with the " // Adjust per-topic message counts to reflect only retained files" and "// Remove from metadata" should actually be inside the `else` statement of the `if (!file_removed || ec) {`.

Can you please move it there?

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
luke-alloy: Done

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
luke-alloy: Yikes, i missed that. Fixed!

MichaelOrlov (commented):
@luke-alloy Thanks. Now it looks good to me.  

However, the `circular_logging_limits_number_of_files_by_max_bag_files` test hangs out and fails due to the timeout.  
Can you please try to take look why it fails?

MichaelOrlov:
Ahh It hangs out becasue file doesn't seems exists when we are trying to delete it in the `delete_oldest_files_if_needed()`

Need to remove the file from the metadata list even if it doesn't exist.  
i.e. we need to add `else` for the 
```cpp
 // Delete file from filesystem
    if (fs::exists(file_path)) {
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -496,6 +514,55 @@ bool SequentialWriter::should_split_bagfile(
...
+        }
+        per_file_topic_message_counts_.pop_front();
+
+        // Remove from metadata
+        metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
+        metadata_.files.erase(metadata_.files.begin());
+        metadata_.starting_time = metadata_.files.front().starting_time;
+      }
+    }
+  }
```
MichaelOrlov: @luke-alloy Here is a fix for the failing test. Please verify.
```suggestion
  /// Adjust per-topic message counts to reflect only retained files
  auto adjust_per_topic_message_counts_with_deleted_file = [&]()
    {
      // Protect topics_names_to_info_ access with mutex to prevent race conditions
      std::lock_guard<std::mutex> lock(topics_info_mutex_);
      const auto & oldest_topic_counts = per_file_topic_message_counts_.front();
      for (const auto & [topic_name, count] : oldest_topic_counts) {
        auto it = topics_names_to_info_.find(topic_name);
        if (it != topics_names_to_info_.end()) {
          auto & topic_info = it->second;
          topic_info.message_count -= count;
        }
      }
      per_file_topic_message_counts_.pop_front();
    };

  /// Remove the deleted file from metadata_ and update starting_time if needed
  auto remove_deleted_file_from_metadata = [&]()
    {
      metadata_.relative_file_paths.erase(metadata_.relative_file_paths.begin());
      metadata_.files.erase(metadata_.files.begin());
      if (!metadata_.files.empty()) {
        metadata_.starting_time = metadata_.files.front().starting_time;
      }
    };

  // Delete the oldest files until we're under the bag file count limit
  while (metadata_.files.size() > storage_options_.max_bag_files) {
    const auto & oldest_file = metadata_.files.front();
    const auto file_path = fs::path(base_folder_) / oldest_file.path;

    // Delete file from filesystem
    if (fs::exists(file_path)) {
      const auto file_size = fs::file_size(file_path);
      const auto file_duration_ns = oldest_file.duration.count();
      std::error_code ec;
      bool file_removed = fs::remove(file_path, ec);
      if (!file_removed || ec) {
        ROSBAG2_CPP_LOG_ERROR("Failed to delete oldest bagfile: %s. Error: %s",
                              file_path.generic_string().c_str(), ec.message().c_str());
      } else {
        ROSBAG2_CPP_LOG_INFO("Deleted oldest bagfile: %s (%lu bytes, %lu ns)",
                             oldest_file.path.c_str(), file_size, file_duration_ns);
        adjust_per_topic_message_counts_with_deleted_file();
        remove_deleted_file_from_metadata();
      }
    } else {
      ROSBAG2_CPP_LOG_ERROR("Oldest bagfile to delete not found: %s",
                            file_path.generic_string().c_str());
      adjust_per_topic_message_counts_with_deleted_file();
      remove_deleted_file_from_metadata();
    }
  }
```

luke-alloy: Thanks @MichaelOrlov  — apologies for the messy intermediate state, I pushed before finishing the cleanup.

Updated `delete_oldest_files_if_needed()` with two changes:

1. **Merged `adjust_per_topic_message_counts_with_deleted_file` and `remove_deleted_file_from_metadata` into a single `remove_oldest_file_from_tracking` lambda.** They were always called as a pair — every code path that called one immediately called the other. Kept the same lock scoping as before (mutex only around `topics_names_to_info_` access) since `metadata_` is already protected by the upper-level writer mutex.

2. **Moved `remove_oldest_file_from_tracking()` outside the if/else branches so it runs unconditionally.** If `fs::remove` failed (e.g. permissions, locked file), the previous version would retry the same file in a tight loop with no backoff, freezing recording. An orphaned file on disk is preferable to hanging the recorder. The error is still logged.

MichaelOrlov: @luke-alloy Please correct me if I am wrong. But in my opinion, we shall not remove the oldest file from tracking if we 
failed on file deletion, and it still exists
i.e. in the 
```cpp
 if (!file_removed || ec) {
        ROSBAG2_CPP_LOG_ERROR( "Failed to delete oldest bagfile: %s. Error: %s",
                               file_path.generic_string().c_str(), ec.message().c_str());
```

luke-alloy: @MichaelOrlov After some more thought, agreed — metadata should be an accurate representation of the files on disk. Updated `delete_oldest_files_if_needed()`:

- Delete fails (file still exists): keep in tracking and break. This was my earlier concern but I didn't provide a good solution — without the break, a persistent failure would spin the loop and hang the recorder. Retries naturally on the next split.
- File not found on disk: remove from tracking so the loop can proceed (this was the cause of the test hang).
- Delete succeeds: remove from tracking as before.

This follows the structure from your suggestion — the main difference is the break on failure to prevent spinning if `fs::remove` consistently fails.

MichaelOrlov (approved):
@luke-alloy Thank you for the quick fixes and elaboration.
Now it looks good to me with the green CI. Approving!

MichaelOrlov:
Pulls: ros2/rosbag2#2218
Gist: https://gist.githubusercontent.com/MichaelOrlov/2ecee29c5a619f139ab5603909ab332f/raw/80b8f8e5d280fc19d6cfc7665772cada013205a0/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_compression rosbag2_cpp rosbag2_py rosbag2_transport rosbag2_storage rosbag2_storage_sqlite3
TEST args:  --packages-above ros2bag rosbag2_compression rosbag2_cpp rosbag2_py rosbag2_transport rosbag2_storage rosbag2_storage_sqlite3
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18176
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27636/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20693/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7510/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26652/)

PR #2215, opened 2025-10-25
Add input[output]_serialization_format to the RecordOptions

MichaelOrlov:
## Description

This PR adds a new `input[output]_serialization_format` fields to the `RecordOptions` in replacement of the `RecordOptions::rmw_serialization_format`. The new options parameters are needed to avoid possible conversion to the CDR serialization if incoming messages are originally serialized in a different format, which is currently made by default. 

These changes needed to facilitate:
>  4.  -  [ ] Rosbag2 recorder has an option to avoid conversion to the CDR serialization format if the topic has a different serialization format.

from the #2214

The newly added `RecorderOption` fields explicitly specify input and output serialization formats for the messages the recorder processes. 
Please note that if `input_serialization_format` and `output_serialization_format` are different, the Rosbag2 recorder will try to find a plugin and use it for conversion from one serialization format to another.

- Relates #2214

### Is this user-facing behavior change?
Note: The former `RecordOptions::rmw_serialization_format` wasn't removed, but rather a deprecation message was added in the log if it is used.
While the backward compatibility on the API level is preserved, it is advised to use the newly added `RecordOptions::output_serialization_format` instead of the `RecordOptions::rmw_serialization_format` since the latter is expected to be removed after the next ROS 2 distro release.

### Did you use Generative AI?
Yes. I am using GitHub Copilot, GPT-4.1 in my workflow

### Additional Information
Note: The former `RecordOptions::rmw_serialization_format` wasn't removed, but rather a deprecation message was added in the log if it is used.

This PR can't be backported since it introduces ABI-breaking changes.

PR #2221, opened 2025-11-03
Add a direct API for rosbag2_transport::Recorder

### LINKED ISSUE #2214: Direct API for the Rosbag2 recorder
## Description
We need to have the ability to save data directly from the node without publishing it on the topics, while maintaining the same behaviour of the instantiated recorder as other standalone recorder instances.  
The real-world usage would be saving raw data from the driver node to avoid extra CPU burden with unnecessary data copies and serialization/deserialization effort. At the same time, it would be very useful if this directly operated recorder would have the same behavior as other recorders that save data from subscriptions regarding the existing `rosbag2_interfaces`. Meant to be managed remotely via service calls and provide statistics and notifications about events, like other recorders. 
It will be very useful for the distributed recording and for the cases with AI/ML and camera/LIDAR where need to be very efficient and avoid extra data copies as much as possible.

Note: The current `rosbag2_cpp` package provides direct API calls for creating a new topic and saving serialized data. However, we need a direct API to the `rosbag2_transport` layer since a lot of functionality is already there, and we want to have unified behavior and avoid code duplication as much as possible.

## Related Issues

N/A

## Completion Criteria

1. 
    - [x] Rosbag2 recorder has a direct API to be able to create a new topic "channel".
2.
    - [x] Rosbag2 recorder has a direct API to be able to write a serialized message.
3.
    - [x] Rosbag2 recorder support option to not subscribe to any topics, but rather use the direct API.
4.
    -  [x] Rosbag2 recorder has an option to avoid conversion to the CDR serialization format if the topic has a different serialization format.

In essence, the `rosbag2_transport::Recorder` class shall be extended to support direct API calls to be able to create a new channel "topic" and call `record_message(serialized_message)` directly without subscribing to any topics.

The newly added API shall be covered by the unit tests.

MichaelOrlov:
## Description
This PR adds a direct API for rosbag2_transport::Recorder to enable direct use of the recorder without creating any subscriptions. For instance, from device drivers.

- This PR depends on https://github.com/ros2/rosbag2/pull/2215

Added direct recorder API add_channel(..) and write_message(..)
Added direct write_message(..) API with recv_timestamp 
Added direct `on_messages_lost_in_transport(..)` API
   Note: The `on_messages_lost_in_transport(..)` method is expected to be called when messages are lost in the transport layer.
Added `get_total_num_messages_lost_in_transport()` getter method.

- Fixes #2214

No. However, the new API will become available

Yes, I am using GitHub Copilot, GPT-5 in my workflow, mostly to help with writing tests and documentation.

### Additional Information
This PR will be difficult to backport since it depends on the non-backportable other PR #2215

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -63,12 +63,98 @@ public:
 
   ~RecorderImpl();
 
+  /// @brief Start recording.
+  /// The record() method will return almost immediately and recording will happen in background.
   void record();
+
+  /// @brief Add a new channel (topic) to the rosbag2 writer to be recorded.
+  /// \details This is a direct Recorder API equivalent to the rosbag2_cpp::Writer::add_topic().
+  /// \note This method does not require the message definition. The recorder will try to find the
+  /// corresponding message definition by the given topic name and type.
+  /// @param topic_name The name of the topic.
+  /// @param topic_type The type of the topic.
+  /// @param serialization_format The serialization format of the topic.
+  /// @param type_description_hash REP-2011 type description hash of the topic.
+  /// @param offered_qos_profiles The list of offered QoS profiles for the topic.
+  void add_channel(
+    const std::string & topic_name,
+    const std::string & topic_type,
+    const std::string & serialization_format = "memory_view",
+    const std::string & type_description_hash = "",
+    const std::vector<rclcpp::QoS> & offered_qos_profiles = {});
+
+  /// \brief Add a new channel (topic) to the rosbag2 writer to be recorded.
+  /// \details This is a direct Recorder API equivalent to the rosbag2_cpp::Writer::add_topic().
+  /// \param topic_name The name of the topic.
+  /// \param topic_type The type of the topic.
+  /// \param message_definition_encoding The encoding technique used in
+  /// the `encoded_message_definition` e.g. "ros2idl", "ros2msg", "apex_json" or "unknown" if
+  /// encoded_message_definition is empty.
+  /// \param encoded_message_definition The fully encoded message definition for this type.
+  /// \param serialization_format The serialization format of the topic.
+  /// \param type_description_hash REP-2011 type description hash of the topic.
+  /// \param offered_qos_profiles The list of offered QoS profiles for the topic.
+  void add_channel(
+    const std::string & topic_name,
+    const std::string & topic_type,
+    const std::string & message_definition_encoding,
+    const std::string & encoded_message_definition,
+    const std::string & serialization_format = "memory_view",
+    const std::string & type_description_hash = "",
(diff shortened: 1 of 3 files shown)


rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -150,6 +236,7 @@ class RecorderImpl
   KeyboardHandler::callback_handle_t toggle_paused_key_callback_handle_ =
     KeyboardHandler::invalid_handle;
 
+public:
```
christophebedard: minor: can we create `RecorderImpl::get_total_num_messages_lost_in_transport()` and call that in `Recorder::get_total_num_messages_lost_in_transport() ` instead of making this public?

MichaelOrlov: @christophebedard  Possible. I've considered this option when implementing this functionality. However, I didn't find any advantages to this, as it is a one-to-one mapping to `event_notifier_->get_total_num_messages_lost_in_transport();`. It appears that we will get nothing rather than bloating `RecorderImpl` with one more extra getter, which could be avoided.

If you have any advantages or rationalities in mind for doing this, please let me know.

MichaelOrlov: Ahh you likely have a concern that this public qualifier appeared again at the end of the private section.  
Got it. make sense. I've added a public method in a new commit. Please verify.

MichaelOrlov:
Pulls: ros2/rosbag2#2221
Gist: https://gist.githubusercontent.com/MichaelOrlov/df277b214be56f941548fa2ac2b8347b/raw/5c2e131defa568d48f5861a6cf5b27a6d3ec0732/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17427
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25639/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19676/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5198/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25868/)

rosbag2_transport/src/rosbag2_transport/recorder.cpp
christophebedard-apexai: It wasn't about having `public:` then `private:` then `public:` again; it was about exposing an attribute. bfc43cb34dadac65b30ca9e3eca54d2198c5ab1e looks good to me

christophebedard-apexai (approved):

rosbag2_transport/src/rosbag2_transport/recorder.cpp
christophebedard: oops, wrong account

christophebedard (approved):

PR #2233, opened 2025-11-13
Fix macOS build: disable thread-safety annotations in LockedPriorityQ…

idesign0:
## Description

On macOS, thread-safety annotations in `rcpputils` cause compilation
failures in `rosbag2_transport` due to Clang not supporting function-style
macro definitions on the command line.

### Changes in this PR:
- Disables `RCPPUTILS_TSA_GUARDED_BY(x)` on Apple platforms via an `#ifdef`.
- Allows `rosbag2_transport` and other packages depending on `LockedPriorityQueue`
  to build successfully on macOS.
- Keeps Linux behavior unchanged.

### Testing
- Successfully built `rcpputils` and `rosbag2_transport` on macOS (Apple Clang).
- Verified Linux build is unaffected.


MichaelOrlov (approved):
LGTM with green CI.

MichaelOrlov:
@idesign0 I'm curious how disabling thread-safety annotations in one file will help with build, since we are using them all over other places?

MichaelOrlov (changes requested):
If disabling the macro in the header file, it must be re-enabled at the end. Otherwise, it can go viral and lead to unexpected behavior. 

If you want to disable thread annotations globally, please consider making it somewhere in another more generic place.

idesign0:
> If disabling the macro in the header file, it must be re-enabled at the end. Otherwise, it can go viral and lead to unexpected behavior.
> 
> If you want to disable thread annotations globally, please consider making it somewhere in another more generic place.

Thanks for the review and the clarification.

You're right — disabling `RCPPUTILS_TSA_GUARDED_BY` directly inside this header without restoring it can leak the macro to any file that includes `locked_priority_queue.hpp`. That was not intended.

On macOS, Clang/libc++ rejects the thread-safety annotation expansion in this file with the error:

```
error: expected ';' at end of declaration list
size_t insert_sequence_number_{0} RCPPUTILS_TSA_GUARDED_BY(queue_mutex_);
                                     ^
```

To fix the build while keeping the change strictly scoped, I updated the patch to disable the macro only inside this header and restore it afterward, using `push_macro` / `pop_macro`.

This keeps:

- macOS: build passes (annotations disabled only in this file)
- Global annotations: unaffected (no “viral” macro override)

MichaelOrlov (commented):
@idesign0 Okay, good. Can you please retarget your branch to the Rolling?
We need to make changes on Rolling first, then backport it to kilted to be in sync for the future.

Also, please add a comment that it will be ``// no-op` macro.

idesign0:
@MichaelOrlov should i make a new PR? 😅

Issue #1548, opened 2024-01-22
[FEA] Support distributed recording to get ultra high performance.


ZhenshengLee:
## Description

the idea is originally from https://eclipse-ecal.github.io/ecal/applications/rec/recorder_architecture.html#applications-recorder-centralized-distributed-recording

> Centralized recording:
This means that you simply start the eCAL Recorder on one of your many machines. It will subscribe to all topics and record the data to your hard drive. Data from remote machines will be sent via network.
This is the most trivial and easy to use mode, as you only need one application on one machine.
Distributed recording:
When having many eCAL applications, it is generally advisable to run applications that exchange huge amounts of data on the same machine, as eCAL will then use its shared memory transport mechanism, which is way faster than sending data over the network. The recorder can also take advantage of that feature and avoid network congestion while running a recording. For this mode you will have to launch the eCAL Recorder on one of your machines and the eCAL Recorder Client on all of your other machines.
Now each Recorder and Recorder Client will only record topics coming from their respectable machine and record them to their local hard drive. After you finished your test, you can then let the Recorder Clients push their files to your main PC to get a contiguous recording.
Of course, mixed configurations are possible (e.g. some machines will only record their own topics, while other machines will also record topics of some external machines).

The main motivation is performance consideration in AV domain projects.

## Related Issues

there are many efforts to get the desired performance of rosbag in AV domain.

- component rosbag recorder: 
https://github.com/ros2/rosbag2/issues/902
https://github.com/berndpfrommer/rosbag2_composable_recorder

- apex.ai optimization
https://www.apex.ai/post/improvements-in-data-recording-path

## Completion Criteria

this FEA requests that there is a rosbag cli support for distributed recording

## Implementation Notes / Suggestions

there should be: 

- composable recorder node.
- recorder service
- recorder client
- recorder cli

## Testing Notes / Suggestions

the distributed recording service should be tested in the AV domain project like autoware auto, https://github.com/autowarefoundation/autoware.universe

fujitatomoya:
This feature sounds interesting, but we surely need to consider more details for implementation note and suggestions. i believe that [REP](https://github.com/ros-infrastructure/rep) would be the good way to start the more discussion in detail. as 1st step, i would recommend that we can have discussion in [ROS 2 Tolling WG](https://docs.ros.org/en/rolling/The-ROS2-Project/Governance/Working-Groups.html#rosbag2-and-tooling).

btw, as user-experience,

[...]

PR #2248, opened 2025-11-26
Add Record, Stop, StartDiscovery, StopDiscovery and IsDiscoveryRunning services for Recorder

### LINKED ISSUE #1634: Add record, stop, start_discovery and stop_discovery services for rosbag2_transport::Recorder

MichaelOrlov:
## Description

This PR is a replacement for the stale PR #1840 and adds the following services to the `rosbag2_transport::Recorder`
- Record
- Stop
- StartDiscovery
- StopDiscovery
- IsDiscoveryRunning

- Fixes #1634
- Relates #https://github.com/ros2/rosbag2/issues/1548

Yes. The user will have the ability to manage the Rosbag2 recorder instance remotely via the service calls.

Yes. I am using GitHub Copilot, GPT-5.0, mostly to help me with tests and Doxygen comments.

### Additional Information
No APi/ABI breaking changes -> can be backported


MichaelOrlov:

MichaelOrlov:
@christophebedard @fujitatomoya Friendly ping here for review.  Or maybe @ahcorde or @Barry-Xu-2018 some of you can review this PR?

Barry-Xu-2018 (approved):
LGTM.

I have a small question.

In RecorderImpl::stop(), `in_recording_` is set to false only after the stop processing is completed.  
https://github.com/ros2/rosbag2/blob/a8b9c82d8665ffe387b4dc8397e5d6f469eed9f9/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L321-L334

In RecorderImpl::record(), `in_recording_` is set to true before record processing is complete.
https://github.com/ros2/rosbag2/blob/a8b9c82d8665ffe387b4dc8397e5d6f469eed9f9/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L353-L356

In RecorderImpl::stop(), is it possible to set `in_recording_` to false before stop processing is started ?  (I'm not sure if there were any specific considerations for current implementation.)

Considering that there is some processing time between when RecorderImpl::stop() is called and when `in_recording_ is` set to false, if a service callback is invoked during this period, the first check of `in_recording_` would be meaningless.
```c++
  srv_stop_ = node->create_service<rosbag2_interfaces::srv::Stop>(
    "~/stop",
    [this](
      const std::shared_ptr<rmw_request_id_t>/* request_header */,
      const std::shared_ptr<rosbag2_interfaces::srv::Stop::Request>/* request */,
      const std::shared_ptr<rosbag2_interfaces::srv::Stop::Response>/* response */)
    {
      if (!in_recording_) {
        RCLCPP_WARN(node->get_logger(),
          "Received Stop request while not in recording. Ignoring request.");
      } else {
        try {
          this->stop();
          RCLCPP_ERROR(node->get_logger(), "Error during Stop request: %s", e.what());
        }
      }
    }
  );
```

MichaelOrlov:
@Barry-Xu-2018, good question and catch-up.  
We use `if (in_recording_.exchange(true))` in the record(uri) method as a leftover from the times when we were using it without a mutex lock.
https://github.com/ros2/rosbag2/blob/a8b9c82d8665ffe387b4dc8397e5d6f469eed9f9/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L353-L356

You are right, we shall set `in_recording_ = true;` at the end of the `record(uri)` method when the recorder and all its underlying data types are fully initialized and ready for recording to avoid race conditions and undefined behavior.

I made a fix in my last https://github.com/ros2/rosbag2/pull/2248/commits/0b2e4789647c3f62ff7ee081a3dfd08ca3cecf9a commit

MichaelOrlov:
@christophebedard @fujitatomoya @ahcorde This PR was reviewed, and findings were addressed. Can some of you formally approve this PR or review it additionally?

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -438,17 +517,135 @@ void RecorderImpl::record(const std::string & uri)
...
     [this](
       const std::shared_ptr<rmw_request_id_t>/* request_header */,
       const std::shared_ptr<rosbag2_interfaces::srv::Pause::Request>/* request */,
       const std::shared_ptr<rosbag2_interfaces::srv::Pause::Response>/* response */)
     {
-      pause();
+      std::lock_guard<std::mutex> state_lock(start_stop_transition_mutex_);
+      // Note: We don't check if we are in recording here, as pausing when not recording is a valid
+      // operation and can be used to set the initial state before starting recording.
+      this->pause();
```
Tiryoh, line 647: Thanks for documenting pre-record usage. One concern: `record()` currently resets `paused_` to `record_options_.start_paused` (line 402),
so a pre-record `/pause` or `/resume` would be overwritten when `record()` is called.
Should we preserve the existing `paused_` state instead (init once in constructor) so this pre-start use case works?

MichaelOrlov: @Tiryoh Thank you for your review. That is a good catch and a good point.
As far as I recall the original reasoning why we were using `record_options_.start_paused` in `record()` method is because we have a getter method 
```cpp
rosbag2_transport::RecordOptions &
Recorder::get_record_options()
{
  return pimpl_->record_options_;
}
```
which is returning a **non-const** reference, and someone decided that it is a good idea to change record_options via reference after start, and expect that the recorder will take into account changes.  As far as I recall, the main reasoning for that expectation was a sort of workaround needed for a composable node. Back in those times, we didn't have a default Recorder constructor and options to pass parameters to the recorder after construction. If one were to need to use a composable recorder, they would need to create a workaround with a derived recorder class with a constructor accepting a default record options and set the actual options later via a getter. 

Since now we have a fully pledged implementation for composable nodes, it makes sense to remove this redumental behaviour and use `record_options_.start_paused` only in the constructor.  
I've made changes for that in the https://github.com/ros2/rosbag2/pull/2248/commits/2978286f9dc3a8935bb22968e5f04db91ed8c7da commit.

MichaelOrlov:
Pulls: ros2/rosbag2#2248
Gist: https://gist.githubusercontent.com/MichaelOrlov/a037975e5fc0e47f16e9b4531842f99f/raw/4f6447b9d365c81b919c094079a426d215d6c412/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17856
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26723/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20355/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6524/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26314/)

rosbag2_transport/src/rosbag2_transport/recorder.cpp
Tiryoh: @MichaelOrlov Thanks for the context and for adjusting it. Using `start_paused` only at construction time makes sense now that composable options are in place. LGTM!

Tiryoh (approved):

MichaelOrlov:
Made a rebase and resolved conflicts.

MichaelOrlov:
@ahcorde @fujitatomoya A friendly ping for a formal approval or additional review here.

MichaelOrlov:
Re-run CI after rebase and fixing conflicts
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17878
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26808/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20377/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6557/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26336/)

rosbag2_interfaces/srv/IsDiscoveryRunning.srv
```diff
@@ -0,0 +1,4 @@
+---
+# True if discovery is running, false otherwise
+bool running
+
```
fujitatomoya, line 4: (this comment does not block the PR)

although i understand there is already `~/is_paused` as a service (this is consistent for the current implementation), i would use a read only parameter instead here. this is purely a state query with no side effects, and i believe exactly what parameters are designed for. in this way, we can reduce service endpoint to take advantage of internal parameter service.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -438,17 +518,135 @@ void RecorderImpl::record(const std::string & uri)
...
     });
 
+  srv_record_ = node->create_service<rosbag2_interfaces::srv::Record>(
+    "~/record",
+    [this](
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Record::Request> request,
+      const std::shared_ptr<rosbag2_interfaces::srv::Record::Response> response)
+    {
+      if (in_recording_) {
```
fujitatomoya, line 601: are we missing mutex here?

```suggestion
      std::lock_guard<std::mutex> state_lock(start_stop_transition_mutex_);
      if (in_recording_) {
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -438,17 +518,135 @@ void RecorderImpl::record(const std::string & uri)
...
+  );
+
+  srv_stop_ = node->create_service<rosbag2_interfaces::srv::Stop>(
+    "~/stop",
+    [this](
+      const std::shared_ptr<rmw_request_id_t>/* request_header */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Stop::Request>/* request */,
+      const std::shared_ptr<rosbag2_interfaces::srv::Stop::Response>/* response */)
+    {
+      if (!in_recording_) {
```
fujitatomoya, line 624: same here?

```suggestion
      std::lock_guard<std::mutex> state_lock(start_stop_transition_mutex_);
      if (!in_recording_) {
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -438,17 +518,135 @@ void RecorderImpl::record(const std::string & uri)
...
+        RCLCPP_WARN(node->get_logger(),
+          "Received Stop request while not in recording. Ignoring request.");
+      } else {
+        try {
+          this->stop();
+        } catch (const std::exception & e) {
+          RCLCPP_ERROR(node->get_logger(), "Error during Stop request: %s", e.what());
```
fujitatomoya, line 631: probably we can consider adding return_code/error_string to the Stop response for consistency with other services? errors are only logged, not returned to the caller.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "rosbag2_transport/recorder.hpp"
-
 #include <algorithm>
+#include <chrono>
+#include <cstdint>
+#include <condition_variable>
```
fujitatomoya: is this required to add? doesn't appear to be used.

fujitatomoya (approved):
just a few minor comments. overall lgtm with all comments are resolved.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: It doesn't make sense to lock a mutex here just to check a flag.  
The flag `in_recording_` itself is `atomic_bool`, and it is safe to check it from multiple threads. 
The mutex will be locked inside `record/stop` calls anyway, where the actual state transition occurs. 

Since there are no potential race conditions nor determinism or UB issues we decided to simplify and **not** lock mutex here.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: It doesn't make sense to lock a mutex here just to check a flag.  
The flag `in_recording_` itself is `atomic_bool`, and it is safe to check it from multiple threads. 
The mutex will be locked inside `record/stop` calls anyway, where the actual state transition occurs. 

Since there are no potential race conditions nor determinism or UB issues we decided to simplify and **not** lock mutex here.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Good catch. I've removed unused includes in the latest commit https://github.com/ros2/rosbag2/pull/2248/changes/d858e26461ec5478134f589c9782c345e126d368

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Good point. Originally, we were a bit conservative about changing already existing Stop service request which is also used in the Player. 
However, I see a rationale to be consistent with other service requests. I went ahead and added a `return_code` to the `~/stop` service request in the https://github.com/ros2/rosbag2/commit/d858e26461ec5478134f589c9782c345e126d368 commit.

MichaelOrlov: @fujitatomoya I recall why we didn't add it at the very beginning - because I want to be this PR with NO API/ABI breaking changes to be able to backport it.  
I am reverting my last commit. I will move it to a follow-up PR only for Rolling.

MichaelOrlov:
Pulls: ros2/rosbag2#2248
Gist: https://gist.githubusercontent.com/MichaelOrlov/af1ba364e5ae98d985cd6a4af59d41ec/raw/4f6447b9d365c81b919c094079a426d215d6c412/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_interfaces rosbag2_transport rosbag2_tests
TEST args:  --packages-above rosbag2_interfaces rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17904
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26845/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20406/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6590/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26364/)

rosbag2_interfaces/srv/IsDiscoveryRunning.srv
MichaelOrlov: Well, the situation is that we are trying to establish a remote interface for the Rosbag2 recorder and Player to be able to manage multiple Rosbag2 recorder/player instances remotely. Perhaps from UI in the future. 
Some interfaces really require services some of them are simple as you mentioned and could be managed as read only parameters. 
However, it would be odd if some of the interfacse would be done via services but some via the parameters API. 
Also, I think the parameters are for a different purpose. It would be logically correct to map parameter changes and incur to match the inner `RecordOptions`, `PlayerOptions` and `StorageOptions`, which really represent the parameters or config options. 
Although using a parameter to incur an inner state of the Player or Recorder is somewhat odd because the inner state is not a parameter or option. In my opinion, the services are more logically suited for that.

MichaelOrlov:

PR #1006, opened 2022-05-11
0.15.2

audrow:
Signed-off-by: Audrow Nash <audrow@hey.com>

PR #2249, opened 2025-11-29
0.15.2 (#1006)

tirth10x:
## Description

Fixes # (issue)


fujitatomoya (commented):
@tirth10x thanks for creating PR.

i think it would be helpful to explain the overview in the PR header what you are trying to address or fix for reviewers to share the context.

here is a couple of things before starting review this PR.

- DCO is missing: https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#change-control-process
- `rolling` is the development branch. basically all the fixes need to be merged into `rolling` 1st, and then we can backport the fix to the downstream branches. https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#change-control-process

PR #2265, opened 2025-12-17
Update Rosbag2 filename format to index+name+timestamp

### LINKED ISSUE #2198: Make timestamp-indexed naming the default for split bagfiles
## Description

~~This proposal introduces a new `--split-naming` option to improve the usability of split bagfiles by allowing timestamps to be included directly in the filenames.~~

> [!NOTE]
> Supplementary comment:
> 
> - This proposal changes course from adding a new CLI option to instead adopt timestamp-indexed naming (start-timestamp plus index) as the default behavior for split bagfiles. The goal is to improve human-readability and make filenames naturally sortable by time. To preserve backward compatibility during transition, an explicit flag to force the legacy serial naming (for example `--split-naming=serial`) may be provided, but the baseline behavior will be timestamp-indexed naming.

Currently, rosbag2 names split bagfiles using sequential numbering only (`record_0.mcap`, `record_1.mcap`, etc.). While this is functional, it makes it harder to quickly understand the time range each file covers when browsing files in a directory.

With the proposed `--split-naming` option, users could select between:

* **`serial` (default):** Current behavior with sequential numbering.
* **`timestamp-index`:** Each split bagfile name includes the start timestamp plus an index (e.g., `record_2025-02-21-16-21-00_0.mcap`).

This change would provide clearer, human-readable filenames, giving users immediate temporal context and helping those migrating from ROS1 workflows feel more at home. It is especially valuable in continuous operation scenarios, where stopping and restarting `ros2 bag record` frequently is not practical for systems with many topics or complex discovery.

## Related Issues

This proposal is related to #1917 (Timestamp split bagfiles instead of serial numbering), which in turn is part of #1915 (Always-on recording). While all of these proposals aim to make bagfile naming more informative, they serve slightly different needs:

* **Purpose:** #1917 is designed for always-on recording, where session boundaries do not apply. This proposal focuses on general split recording scenarios where timestamp context is important.
* **Format:** #1917 suggests timestamp-only names (`recording_20250221T162100Z.mcap`). This proposal keeps both timestamp and index (`recording_2025-02-21-16-21-00_0.mcap`), ensuring both clarity and ordering.
* **Compatibility:** These approaches can coexist under a unified `--split-naming` interface, giving users flexibility to choose the scheme that best fits their workflow.

## Completion Criteria

* Add `--split-naming` option to the `ros2 bag record` command.
* Implement `serial` mode (default, maintains current behavior).
* Implement `timestamp-index` mode with filesystem-safe timestamp format.
* Ensure compatibility with all splitting mechanisms (`--max-bag-duration`, `--max-bag-size`, manual splitting via `~/split_bagfile`).
* Preserve backward compatibility by defaulting to `serial`.
* Add appropriate unit and integration tests.
* Update documentation and CLI help text.

* Timestamps should be filesystem-safe and sortable (suggested: `YYYY-MM-DD-HH-mm-ss`).
* Implementation would involve extending the `SequentialWriter` to handle naming modes.
* Use `std::strftime` or similar for formatting.
* Recommend UTC timestamps for consistency across environments.
* Timestamp should reflect the **start time** of each bagfile.
* Index should always be included to ensure uniqueness, even if multiple splits occur in the same second.

* **Unit tests:** Verify filename generation for both modes.
* **Integration tests:** Confirm behavior with size, duration, and manual splits.
* **Edge cases:**

  * Rapid splits (sub-second).
  * Cross-platform filesystem compatibility.
  * Uniqueness guarantees under load.
* **Backward compatibility:** Verify unchanged behavior when `--split-naming` is not specified.
* **Playback compatibility:** Ensure generated bagfiles can be played back regardless of naming mode.

Tiryoh:
## Description

This PR changes the automatically generated **bagfile (split) filenames** produced by the sequential writer from the current `"{prefix}_{counter}"` pattern to a new pattern:

```
{counter}_{prefix}_{timestamp}.{extension}
```

* **`counter`**: split index (integer starting from 0, *not zero-padded*)
* **`prefix`**: derived from the bag directory name, with any default timestamp suffix removed
* **`timestamp`**: local time at file creation, formatted as `YYYY_MM_DD-HH_MM_SS`
* **`extension`**: bag file extension. e.g., `.mcap`, `.db3`

This change was discussed by @fujitatomoya, @MichaelOrlov, and me in [[rosbag2#2198](https://github.com/ros2/rosbag2/issues/2198)](https://github.com/ros2/rosbag2/issues/2198).
It makes each split file **self-descriptive** and **chronologically traceable**, improving usability for both human inspection and external tooling.

**Key points / possible concerns**

* **User-visible filename change** for new recordings only
* The **reader behavior** remains unchanged (it uses metadata)
* **Backward compatibility** is preserved: existing bags can still be replayed and inspected
* **Custom scripts** that assume the old pattern (e.g., `mybag_0.mcap`, `mybag_1.mcap`) may need to adjust globs or regexes
* Timestamp is generated in **local time** for human readability (UTC could be added later if desired)
* The **`reindexer`** was updated to recognize the new filename pattern (including optional double extensions), and corresponding tests/fixtures were adjusted

  * Note: The `reindexer` still does not decompress file-compressed bags; compressed files must be accessible or decompressed beforehand

- Fixes [https://github.com/ros2/rosbag2/issues/2198](https://github.com/ros2/rosbag2/issues/2198)

---

Yes.

**When a custom name is specified (e.g., `ros2 bag record -o mybag -a`):**

| Before               | After                                    |
| :------------------- | :--------------------------------------- |
| `mybag/mybag_0.mcap` | `mybag/0_mybag_2025_11_04-12_30_20.mcap` |
| `mybag/mybag_1.mcap` | `mybag/1_mybag_2025_11_04-12_45_20.mcap` |

This represents the case where the user explicitly specified the output name with `-o`.

**When no name is specified (default timestamped directory):**

| Before                                                           | After                                                            |
| :--------------------------------------------------------------- | :--------------------------------------------------------------- |
| `rosbag2_2025_09_30-09_11_30/rosbag2_2025_09_30-09_11_30_0.mcap` | `rosbag2_2025_09_30-09_11_30/0_rosbag2_2025_09_30-09_11_30.mcap` |
| `rosbag2_2025_09_30-09_11_30/rosbag2_2025_09_30-09_11_30_1.mcap` | `rosbag2_2025_09_30-09_11_30/1_rosbag2_2025_09_30-09_12_30.mcap` |

This represents the default behavior when no `-o` argument is provided, and the bag directory name is automatically timestamped.

**Impact:**

* Easier manual identification of file order and creation time
* External tooling and scripts relying on filename patterns may require minor adjustments
* Reading and playing existing recordings remains unaffected

---

Yes.

* **ChatGPT (GPT-5)** was used for drafting this PR description text
* **Cursor (Composer 1)** was used for writing/updating the test code

---

### Additional Information

* Timestamp format: `YYYY_MM_DD-HH_MM_SS` (local time)
* Implementation modifies `format_storage_uri` within the sequential writer, updates `reindexer` filename parsing, and refreshes related unit/integration tests
* No metadata or storage layout changes; behavior is limited to filename generation/recognition
* Optionally, UTC timestamps or configurable formats could be considered in a future PR
* Test resources in `rosbag2_tests/resources/*/cdr_test/metadata.yaml` were intentionally **not** updated

  * The test logic now uses `get_actual_cdr_test_filename()` to retrieve actual filenames from metadata at runtime, eliminating the need to regenerate test bags for the new naming convention

---

#### Record behavior with split service

While recording with `ros2 bag record -a`, the split service was called during recording and the generated filenames were verified.

**Before (legacy filename pattern):**

```bash
$ ls rosbag2_2025_12_17-20_37_09/
metadata.yaml  rosbag2_2025_12_17-20_37_09_0.mcap  rosbag2_2025_12_17-20_37_09_1.mcap  rosbag2_2025_12_17-20_37_09_2.mcap

$ ls mybag/
metadata.yaml  mybag_0.mcap  mybag_1.mcap  mybag_2.mcap  mybag_3.mcap
```

**After (new filename pattern):**

```bash
$ ls rosbag2_2025_12_17-20_55_50/
0_rosbag2_2025_12_17-20_55_50.mcap  1_rosbag2_2025_12_17-20_56_06.mcap  2_rosbag2_2025_12_17-20_56_16.mcap  metadata.yaml

$ ls mybag_new/
0_mybag_new_2025_12_17-20_57_22.mcap  1_mybag_new_2025_12_17-20_57_31.mcap  2_mybag_new_2025_12_17-20_57_40.mcap  metadata.yaml
```

This confirms that split files created during an active recording session follow the new naming convention for both default timestamped directories and explicitly named bags.

---

#### Reindex behavior

Reindexing was tested using the updated `ros2 bag` command on both legacy and new filename patterns.

**Legacy filename pattern:**

```bash
$ ls mybag
metadata.yaml  mybag_0.mcap  mybag_1.mcap  mybag_2.mcap  mybag_3.mcap
$ ros2 bag reindex mybag
[INFO] [1765972695.600521161] [rosbag2_cpp]: Beginning reindexing bag in directory: "mybag"

closing.

closing.

closing.

closing.
[INFO] [1765972695.620651394] [rosbag2_cpp]: Reindexing complete.
```

**New filename pattern:**

```bash
$ ls mybag_new/
0_mybag_new_2025_12_17-20_57_22.mcap  1_mybag_new_2025_12_17-20_57_31.mcap  2_mybag_new_2025_12_17-20_57_40.mcap  metadata.yaml
$ ros2 bag reindex mybag_new
[INFO] [1765972868.127763161] [rosbag2_cpp]: Beginning reindexing bag in directory: "mybag_new"

closing.

closing.

closing.
[INFO] [1765972868.152821525] [rosbag2_cpp]: Reindexing complete.
```

This confirms that `ros2 bag reindex` works correctly with both legacy and new split filename patterns.


rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
MichaelOrlov, line 1: Please declare the following helper functions as public methods of the `SequentialCompressionWriterTest` test fixture to check the path in multiple places, the same way, and avoid code duplication.
```cpp

  bool path_match_expected_regex(const std::string & path) const
  {
    // New filename format: {counter}_{bag_base_dir}_{timestamp}.{compressor}
    // Timestamp is generated at runtime, so validate using regex
    // Use static to avoid recompiling regex on each test run
    static const std::string file_pattern_str =
      R"(\d+_)" + bag_base_dir_ + R"(_)" +
      std::string(rosbag2_cpp::writers::TIMESTAMP_PATTERN) +
      R"(\.)" + std::string(DefaultTestCompressor);
    static const std::regex file_pattern(file_pattern_str);
    return std::regex_match(path, file_pattern);
    // "Path '" << path << "' does not match expected pattern for file " << counter;
  }

  ::testing::AssertionResult file_counter_at_start(const std::string & path, size_t counter) const
  {
    std::stringstream expected_prefix;
    expected_prefix << counter << "_" << bag_base_dir_ << "_";
    if (path.find(expected_prefix.str()) == 0) {
      return ::testing::AssertionSuccess();
    }
    return ::testing::AssertionFailure() << "Path '" << path <<
           "' does not start with expected prefix '" << expected_prefix.str() << "'";
  }
````

Expected usage:
```cpp
  size_t counter = 0;
  for (const auto & path : intercepted_write_metadata_.relative_file_paths) {
    // Verify that filename matches expected format
    EXPECT_TRUE(path_match_expected_regex(path)) <<
      "Path '" << path << "' does not match expected pattern for file " << counter;

    // Verify that counter is at the correct position
    EXPECT_TRUE(file_counter_at_start(path, counter));
    counter++;
  }
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -295,9 +298,45 @@ std::string SequentialWriter::format_storage_uri(
   // Right now `base_folder_` is always just the folder name for where to install the bagfile.
   // The name of the folder needs to be queried in case
   // SequentialWriter is opened with a relative path.
```
MichaelOrlov, line 303: I think we can delete the original comment it has become irrelevant and outdated 
```suggestion
```

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -295,9 +298,45 @@ std::string SequentialWriter::format_storage_uri(
...
+
+  std::stringstream timestamp_stream;
+  timestamp_stream << std::put_time(&tm_buf, "%Y_%m_%d-%H_%M_%S");
+  std::string timestamp = timestamp_stream.str();
+
+  // Generate filename in format {storage_count}_{prefix}_{timestamp}
+  // Note: Underscores are used as separators. If the prefix contains underscores,
+  // this creates theoretical ambiguity when parsing filenames. However, parsing is
+  // typically done by matching the timestamp pattern from the end, which avoids
+  // ambiguity in practice.
```
MichaelOrlov: Nitpick: to shorten the body and rephrase a bit comments.
```suggestion
 // Extract prefix from directory name by removing timestamp pattern if present
  // This handles the case when --output is not specified and default timestamped directory is used
  // Currently, the default timestamp format is `YYYY_MM_DD-HH_MM_SS`
  std::string dir_name = fs::path(base_folder).filename().generic_string();
  // Handle edge case where filename() returns empty (e.g., base_folder is "/" or ".")
  // This should not happen in practice since base_folder is validated in open(), but
  // we add this check for defensive programming.
  if (dir_name.empty()) {
    dir_name = "rosbag2";  // Use default prefix
  }
  static std::regex timestamp_pattern("_" + std::string(TIMESTAMP_PATTERN) + "$");
  std::string prefix = std::regex_replace(dir_name, timestamp_pattern, "");

  // Generate timestamp in local time.
  // Note: During DST switches the same string may occur twice. However, we're also adding the
  // sequence counter as part of the filename, so duplicates still remain distinguishable.
  auto time_t = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
  std::tm timestamp{};
#ifdef _WIN32
  localtime_s(&timestamp, &time_t);
#else
  localtime_r(&time_t, &timestamp);
#endif

  // Generate filename in format {storage_count}_{prefix}_{timestamp}
  // Note: Underscores are used as separators. If the prefix contains underscores, this creates
  // theoretical ambiguity when parsing filenames. However, parsing is typically done by matching
  // the timestamp pattern from the end, which avoids ambiguity in practice.
  std::stringstream storage_file_name;
  storage_file_name << storage_count << "_" << prefix << "_" <<
    std::put_time(&timestamp, "%Y_%m_%d-%H_%M_%S");

  return (fs::path(base_folder) / storage_file_name.str()).generic_string();
```

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
```diff
@@ -89,7 +89,8 @@ class ROSBAG2_CPP_PUBLIC Reindexer
   std::vector<rosbag2_storage::TopicMetadata> topics_metadata_{};
 
 private:
-  std::string regex_bag_pattern_;
+  std::string new_format_regex_;
+  std::string old_format_regex_;
```
MichaelOrlov: Nitpick: Please rename to the `new{old}_file_format_regex_str_`
I would also make them as a `const` and init right away here using `rosbag2_cpp::writers::TIMESTAMP_PATTERN`.
```cpp
  const std::string new_file_format_regex_str_ =
    R"((\d+)_(.*)_)" + std::string(rosbag2_cpp::writers::TIMESTAMP_PATTERN) +
    R"(\.[a-zA-Z0-9]+){1,2})";
  const std::string old_file_format_regex_str_ = R"((.*)_(\d+)(\.[a-zA-Z0-9]+){1,2})";
```

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -44,6 +49,90 @@ using namespace testing;  // NOLINT
 using rosbag2_test_common::ParametrizedTemporaryDirectoryFixture;
 namespace fs = std::filesystem;
 
+// Helper function to check if filename matches new naming pattern: {counter}_{prefix}_{timestamp}
+bool matches_filename_pattern(
+  const std::string & filename, const std::string & prefix,
+  uint64_t expected_counter)
+{
+  // Expected format: {counter}_{prefix}_{timestamp}
+  std::string expected_prefix = std::to_string(expected_counter) + "_" + prefix + "_";
```
MichaelOrlov: Please rewrite `matches_filename_pattern(const std::string & path, const std::string & prefix,, size_t expected_counter)` to include stripping file extension to reduce boilerplate code in tests.
i.e.
```cpp
   // Extract filename from path (remove directory part if present)
    std::string filename = fs::path(path).filename().generic_string();
    // Remove extension if present (e.g., .mcap)
    std::string filename_no_ext = fs::path(filename).stem().generic_string();
```
- Also suggest using the `size_t` type for the counter to avoid type casts.
- Also, please move helper functions inside the test fixture class. To have the ability to use `ASSERT_TRUE` inside helper functions if needed.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
MichaelOrlov, line 1: Please avoid checking the timestamp with tolerance in bag split tests. Checking the sequence file number is enough. Those tests are intended to verify that the bag split works correctly and that the file's sequence numbers are as expected. Checking the timestamp with tolerance is a source of flakiness in the tests on CI, and only overcomplicates the tests.
For timestamp correctness in the filename, we should have a separate unit test. e.g. `filename_extracts_prefix_from_timestamped_directory_name`

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -430,12 +524,22 @@ TEST_F(SequentialWriterTest, writer_splits_when_storage_bagfile_size_gt_max_bagf
...
+
+    EXPECT_TRUE(matches_filename_pattern(filename_no_ext, bag_base_dir_, counter))
+      << "Filename '" << filename_no_ext << "' does not match expected pattern";
+    EXPECT_EQ(extract_counter_from_filename(filename_no_ext), static_cast<uint64_t>(counter))
+      << "Counter in filename '" << filename_no_ext << "' does not match expected value";
+
+    // Verify timestamp matches recorded time (with tolerance)
+    std::string filename_timestamp = extract_timestamp_from_filename(filename_no_ext);
+    EXPECT_TRUE(timestamp_matches(filename_timestamp, recorded_time))
+      << "Timestamp in filename '" << filename_timestamp << "' does not match recorded time";
```
MichaelOrlov: Please simplify the checks. I would expect it to be as simple as
```cpp
  size_t counter = 0;
  for (const auto & path : fake_metadata_.relative_file_paths) {
    EXPECT_TRUE(matches_filename_pattern(path, bag_base_dir_, counter))
      << "Filename '" << path << "' does not match expected pattern";
    EXPECT_EQ(extract_counter_from_filename(path), counter)
      << "Counter in filename '" << path << "' does not match expected value";
    counter++;
  }
```
Also applicable for all other places with similar checks.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -1162,6 +1302,56 @@ TEST_P(ParametrizedTemporaryDirectoryFixture, split_bag_metadata_has_full_durati
...
+  // * --output is not specified
+  // * default timestamped directory is used
+  const std::string timestamped_dir = "rosbag2_2025_11_04-20_21_42";
+  storage_options_.uri = (tmp_dir_ / timestamped_dir).string();
+
+  // Set up metadata interception
+  ON_CALL(*metadata_io_, write_metadata).WillByDefault(
+    [this](const std::string &, const rosbag2_storage::BagMetadata & metadata) {
+      fake_metadata_ = metadata;
+    });
```
MichaelOrlov: We have already done the same `ON_CALL` in the test fixture, so we don't need to repeat it again.
```suggestion
```

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
```diff
@@ -1162,6 +1302,56 @@ TEST_P(ParametrizedTemporaryDirectoryFixture, split_bag_metadata_has_full_durati
...
+  EXPECT_CALL(*metadata_io_, write_metadata).Times(1);
+
+  auto sequential_writer = std::make_unique<rosbag2_cpp::writers::SequentialWriter>(
+    std::move(storage_factory_), converter_factory_, std::move(metadata_io_));
+  writer_ = std::make_unique<rosbag2_cpp::Writer>(std::move(sequential_writer));
+
+  std::string rmw_format = "rmw_format";
+  writer_->open(storage_options_, {rmw_format, rmw_format});
```
MichaelOrlov: Can be simplified to be as:
```suggestion
  writer_->open((tmp_dir_ / timestamped_dir).string());
```

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.cpp
MichaelOrlov, line 1: This file can be intact since no modifications were made in the test resource files. Please revert the unnecessary delta.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -372,10 +372,42 @@ TEST_P(RecordFixture, record_end_to_end_with_splitting_max_size_not_reached) {
...
+        filename.find(".mcap.zstd") != std::string::npos)))
+      {
+        actual_bag_file_count++;
+      }
+    }
+  }
+
+  EXPECT_EQ(metadata.files.size(), actual_bag_file_count) <<
+    "Expected " << metadata.files.size() << " bag files, but found " << actual_bag_file_count <<
+    " bag files in directory.";
```
MichaelOrlov: This part is over-engineered. Please replace with the simple version
```cpp
  // Count actual bag files in directory (excluding metadata files)
  std::vector<std::string> found_bag_files;
  std::string bag_files_list_str;
  for (const auto & entry : fs::directory_iterator(root_bag_path_)) {
    if (entry.is_regular_file()) {
      const auto & path = entry.path();
      const auto filename = path.filename().generic_string();
      // Skip metadata file
      if (filename == "metadata.yaml") {
        continue;
      }
      found_bag_files.push_back(filename);
      bag_files_list_str += filename + ", ";
    }
  }
  
  EXPECT_EQ(found_bag_files.size(), 1u) << "Expected 1 bag file, but found " <<
    found_bag_files.size() << " bag files in directory. {" << bag_files_list_str << "}";
```

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
MichaelOrlov: These helper functions are no longer needed and can be deleted to prevent incorrect usage in the future.
```suggestion
```

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
```diff
@@ -73,16 +76,68 @@ class ComposableRecorderIntegrationTests : public CompositionManagerTestFixture
...
+      "\" in directory: \"" << root_bag_path_.generic_string() << "\"";
+    if (!found_files.empty()) {
+      error_msg << ". Found files: [";
+      for (size_t i = 0; i < found_files.size(); ++i) {
+        if (i > 0) {error_msg << ", ";}
+        error_msg << "\"" << found_files[i] << "\"";
+      }
+      error_msg << "]";
+    }
+    ASSERT_TRUE(false) << error_msg.str();
```
MichaelOrlov: Nitpick: Suggested simplification and cleanup
```cpp
    // Filename format: {counter}_{bag_base_dir}_{timestamp}.{ext}
    // Get bag_base_dir from root_bag_path_ directory name
    std::string dir_name = root_bag_path_.filename().generic_string();
    using rosbag2_cpp::writers::TIMESTAMP_PATTERN;
    // Remove timestamp pattern if present (same logic as format_storage_uri() in SequentialWriter)
    static std::regex timestamp_pattern("_" + std::string(TIMESTAMP_PATTERN) + "$");
    const std::string bag_base_dir = std::regex_replace(dir_name, timestamp_pattern, "");

    const auto storage_id = GetParam();
    // Build regex pattern for first file: 0_{bag_base_dir}_{timestamp}.{ext}
    // Escape dot in extension for regex (replace . with \.)
    std::string escaped_extension = (storage_id == "sqlite3") ? "\\.db3" : "\\.mcap";

    std::stringstream pattern_ss;
    pattern_ss << R"(0_)" << bag_base_dir << R"(_)" << TIMESTAMP_PATTERN << escaped_extension;
    std::regex file_pattern(pattern_ss.str());

    const auto start_time = std::chrono::steady_clock::now();
    std::vector<std::string> found_files;
    while (std::chrono::steady_clock::now() - start_time < timeout && rclcpp::ok()) {
      // Search for matching file in directory
      if (fs::exists(root_bag_path_) && fs::is_directory(root_bag_path_)) {
        found_files.clear();
        for (const auto & entry : fs::directory_iterator(root_bag_path_)) {
          if (entry.is_regular_file()) {
            const std::string filename = entry.path().filename().generic_string();
            found_files.push_back(filename);
            if (std::regex_match(filename, file_pattern)) {
              return;  // Found matching file
            }
          }
        }
      }
      std::this_thread::sleep_for(50ms);  // wait a bit to not query constantly
    }

    // Build error message with debug info
    std::stringstream error_msg;
    error_msg << "Could not find storage file matching pattern: \"" << pattern_ss.str() <<
      "\" in directory: \"" << root_bag_path_.generic_string() << "\"";
    if (!found_files.empty()) {
      error_msg << ". Found files: [";
      for (size_t i = 0; i < found_files.size(); ++i) {
        if (i > 0) {error_msg << ", ";}
        error_msg << "\"" << found_files[i] << "\"";
      }
      error_msg << "]";
    }
    ASSERT_TRUE(false) << error_msg.str();
```

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
```diff
@@ -73,16 +76,68 @@ class ComposableRecorderIntegrationTests : public CompositionManagerTestFixture
 
   void wait_for_storage_file(std::chrono::duration<float> timeout = std::chrono::seconds(10))
   {
-    const auto storage_path = get_bag_file_path(0);
+    // Filename format: {counter}_{bag_base_dir}_{timestamp}.{ext}
+    // format_storage_uri extracts prefix from directory name
+    // by removing timestamp pattern if present
```
MichaelOrlov: I would remove this part of the comment. Seems AI stuff, and doesn't really help with code understanding.
```suggestion
```

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
```diff
@@ -73,16 +76,68 @@ class ComposableRecorderIntegrationTests : public CompositionManagerTestFixture
...
+        escaped_extension += c;
+      }
+    }
+
+    std::stringstream pattern_stream;
+    pattern_stream << R"(0_)" << bag_base_dir << R"(_)" <<
+      rosbag2_cpp::writers::TIMESTAMP_PATTERN << escaped_extension;
+    const std::string pattern_str = pattern_stream.str();
+    std::regex file_pattern(pattern_str);
+
```
MichaelOrlov: Suggested simplification and cleanup:
```suggestion
    using rosbag2_cpp::writers::TIMESTAMP_PATTERN;
    // Remove timestamp pattern if present (same logic as format_storage_uri() in SequentialWriter)
    static std::regex timestamp_pattern("_" + std::string(TIMESTAMP_PATTERN) + "$");
    const std::string bag_base_dir = std::regex_replace(dir_name, timestamp_pattern, "");

    const auto storage_id = GetParam();
    // Build regex pattern for first file: 0_{bag_base_dir}_{timestamp}.{ext}
    // Escape dot in extension for regex (replace . with \.)
    std::string escaped_extension = (storage_id == "sqlite3") ? "\\.db3" : "\\.mcap";

    std::stringstream pattern_ss;
    pattern_ss << R"(0_)" << bag_base_dir << R"(_)" << TIMESTAMP_PATTERN << escaped_extension;
    std::regex file_pattern(pattern_ss.str());

```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -546,14 +566,44 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_msg_comp
 
   ASSERT_GE(opened_files.size(), num_splits + 1);
   ASSERT_GE(closed_files.size(), num_splits + 1);
+
+  // New filename format: {counter}_{bag_base_dir}_{timestamp}
+  // Timestamp is generated at runtime, so validate using regex
+  static const std::string file_pattern_str =
+    R"(\d+_)" + std::string("test_bag") + R"(_)" +
```
MichaelOrlov:
```suggestion
     R"(\d+_)" + bag_base_dir_ + R"(_)" +
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -546,14 +566,44 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_msg_comp
...
+    std::string closed_filename = closed_path.filename().generic_string();
+    EXPECT_TRUE(std::regex_match(closed_filename, file_pattern)) <<
+      "Closed file '" << closed_filename << "' does not match expected pattern for file " << i;
+
+    // Verify counter is at the correct position
+    std::stringstream expected_prefix;
+    expected_prefix << i << "_" << bag_base_dir_ << "_";
+    EXPECT_TRUE(closed_filename.find(expected_prefix.str()) == 0) <<
+      "Closed file '" << closed_filename << "' does not start with expected prefix '" <<
+      expected_prefix.str() << "'";
```
MichaelOrlov:
```suggestion
    EXPECT_TRUE(file_counter_at_start(closed_filename, i));
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -616,15 +666,49 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_file_com
 
   ASSERT_GE(opened_files.size(), num_splits + 1);
   ASSERT_GE(closed_files.size(), num_splits + 1);
+
+  // New filename format: {counter}_{bag_base_dir}_{timestamp}.{compressor}
+  // Timestamp is generated at runtime, so validate using regex
+  static const std::string closed_file_pattern_str =
+    R"(\d+_)" + std::string("test_bag") + R"(_)" +
```
MichaelOrlov:
```suggestion
    R"(\d+_)" + bag_base_dir_ + R"(_)" +
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -616,15 +666,49 @@ TEST_P(SequentialCompressionWriterTest, split_event_calls_callback_with_file_com
...
+    std::string(rosbag2_cpp::writers::TIMESTAMP_PATTERN) +
+    R"(\.)" + std::string(DefaultTestCompressor);
+  static const std::regex closed_file_pattern(closed_file_pattern_str);
+  static const std::string opened_file_pattern_str =
+    R"(\d+_)" + std::string("test_bag") + R"(_)" +
```
MichaelOrlov:
```suggestion
    R"(\d+_)" + bag_base_dir_ + R"(_)" +
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -690,15 +774,48 @@ TEST_F(SequentialCompressionWriterTest, snapshot_writes_to_new_file_with_file_co
   ASSERT_EQ(opened_files.size(), 2);
   ASSERT_EQ(closed_files.size(), 2);
 
+  // New filename format: {counter}_{bag_base_dir}_{timestamp}.{compressor}
+  // Timestamp is generated at runtime, so validate using regex
+  static const std::string closed_file_pattern_str =
+    R"(\d+_)" + std::string("test_bag") + R"(_)" +
```
MichaelOrlov:
```suggestion
    R"(\d+_)" + bag_base_dir_ + R"(_)" +
```

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
```diff
@@ -690,15 +774,48 @@ TEST_F(SequentialCompressionWriterTest, snapshot_writes_to_new_file_with_file_co
...
+    std::string(rosbag2_cpp::writers::TIMESTAMP_PATTERN) +
+    R"(\.)" + std::string(DefaultTestCompressor);
+  static const std::regex closed_file_pattern(closed_file_pattern_str);
+  static const std::string opened_file_pattern_str =
+    R"(\d+_)" + std::string("test_bag") + R"(_)" +
```
MichaelOrlov:
```suggestion
    R"(\d+_)" + bag_base_dir_ + R"(_)" +
```

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -129,18 +138,62 @@ class RecordFixture : public ParametrizedTemporaryDirectoryFixture
...
+    const auto bag_path = root_bag_path_.generic_string();
+
+    // Wait for metadata file to exist
+    const auto start_time = std::chrono::steady_clock::now();
+    while (std::chrono::steady_clock::now() - start_time < std::chrono::seconds(10)) {
+      if (metadata_io.metadata_file_exists(bag_path)) {
+        break;
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(50));
+    }
```
MichaelOrlov:
```suggestion
    wait_for_metadata(std::chrono::seconds(10));
```

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
MichaelOrlov: These helper functions are no longer needed and can be deleted to prevent incorrect usage in the future.

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
```diff
@@ -129,18 +138,62 @@ class RecordFixture : public ParametrizedTemporaryDirectoryFixture
       << "Could not find metadata file.";
   }
 
+  std::filesystem::path get_actual_bag_file_path(int split_index = 0) const
```
MichaelOrlov: Please rename and redesign this method to be `get_bag_file_path_from_metadata(BagMetadata metadata, int split_index = 0)`. 
Also, please move it to the `rosbag2_test_common` in a new `bag_files_helpers.hpp` file as well as other bag file helper functions such as 
- `wait_for_metadata(std::chrono::duration<float> timeout = std::chrono::seconds(5)) const`
- `std::filesystem::path get_relative_bag_file_path(int split_index)`
- `std::filesystem::path get_compressed_bag_file_path(int split_index)`
- `void wait_for_storage_file(const std::filesystem::path & root_bag_path, std::chrono::duration<float> timeout)`
- etc...

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
```diff
@@ -136,6 +138,34 @@ class Rosbag2CPPGetServiceInfoTest
     return rosbag2_test_common::bag_filename_for_storage_id(bag_file_name.str(), storage_id);
   }
 
+  std::filesystem::path get_actual_bag_file_path(int split_index = 0) const
```
MichaelOrlov: Please use previously suggested `get_bag_file_path_from_metadata(BagMetadata metadata, int split_index = 0)` instead of this duplicating function, which is going to be in the new `bag_files_helpers.hpp` in the `rosbag2_test_common` package.

MichaelOrlov (changes requested):
@Tiryoh Thank you for your contribution.
While the changes in the code itself are pretty much minor, I would say. However, the changes in the tests are trickier and needs to be adjusted and cleaned up.

Can you please rename and redesign the `std::filesystem::path RecordFixture::get_actual_bag_file_path(int split_index = 0) const` method to be `get_bag_file_path_from_metadata(BagMetadata metadata, int split_index = 0)`.  i.e. we can wait for metadata explicitly with another helper functions in tests and this API will be more clear to understand how are we getting a filename.
Also, please move it to the `rosbag2_test_common` in a new `bag_files_helpers.hpp` file as well as other bag file helper functions such as 
- `wait_for_metadata(std::chrono::duration<float> timeout = std::chrono::seconds(5)) const`
- `std::filesystem::path get_relative_bag_file_path(int split_index)`
- `std::filesystem::path get_compressed_bag_file_path(int split_index)`
- `void wait_for_storage_file(const std::filesystem::path & root_bag_path, std::chrono::duration<float> timeout)`
- etc...

Tiryoh:
@MichaelOrlov Thank you for the code review. I will go through the comments one by one and address them accordingly.

fujitatomoya (commented):
lgtm with green CI and @MichaelOrlov 's comment resolved.

most of the changes are adjustment for tests, implementation looks good to me.
the only concern is that, breaking change for scripts. any external tooling assuming the old pattern (mybag_0.mcap) will need updates as acknowledged in the PR.

MichaelOrlov:
@Tiryoh, friendly ping here to reiterate or update about the status and plans for this PR.

Tiryoh:
Thanks for the kind reminder. I'll update the pull request today!

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied, thanks.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Tiryoh: I have removed the comments

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Tiryoh: Applied

rosbag2_cpp/include/rosbag2_cpp/reindexer.hpp
Tiryoh: I have renamed and made as a `const`, also updated `reindexer.cpp`

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
Tiryoh: I have updated

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
Tiryoh: I updated the split tests so they no longer validate timestamp tolerance. They now focus on what those tests are meant to verify: that splitting happens correctly and the sequence numbers in filenames are correct.

Timestamp correctness is now covered separately in filename_extracts_prefix_from_timestamped_directory_name, so we still validate the filename timestamp format/behavior without introducing CI flakiness into split tests.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
Tiryoh: I updated the similar checks to follow that simplified pattern across the relevant tests: iterate with `size_t counter`, validate with `matches_filename_pattern(...)`, and verify with `extract_counter_from_filename(...)`. This removes extra filename handling logic and keeps the assertions focused and consistent.

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
Tiryoh: Thank you, removed

rosbag2_cpp/test/rosbag2_cpp/test_sequential_writer.cpp
Tiryoh: Applied

rosbag2_tests/test/rosbag2_tests/test_rosbag2_info_end_to_end.cpp
Tiryoh: reverted the changes

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
Tiryoh: Applied

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
Tiryoh: Sorry, I went ahead and checked the other comments too.

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_compression/test/rosbag2_compression/test_sequential_compression_writer.cpp
Tiryoh: Applied

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
Tiryoh: Applied

rosbag2_transport/test/rosbag2_transport/test_composable_recorder.cpp
Tiryoh: I have removed these lines

rosbag2_tests/test/rosbag2_tests/record_fixture.hpp
Tiryoh: I have added `bag_files_helpers.hpp`

rosbag2_tests/test/rosbag2_tests/test_rosbag2_cpp_get_service_info.cpp
Tiryoh: I have updated

fujitatomoya:
Pulls: ros2/rosbag2#2265
Gist: https://gist.githubusercontent.com/fujitatomoya/1fc7a12f0d94fdaa4f2d8d3749d4d54f/raw/b20294925cbc4740470823768cec59b0653601be/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_compression rosbag2_cpp rosbag2_test_common rosbag2_test rosbag2_transport
TEST args:  --packages-above rosbag2_compression rosbag2_cpp rosbag2_test_common rosbag2_test rosbag2_transport
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18270
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27753/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20788/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7619/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26752/)

Tiryoh:
The build and tests are passing on my end, but rclcpp might be a bit old. Let me double-check.

MichaelOrlov:
I am in process of doing a second round of review.

MichaelOrlov:
I am getting a test failure on my local run
```bash
 [ RUN      ] TestRecordEndToEnd/RecordFixture.record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size/0
1: stdin is not a terminal device. Keyboard handling disabled.[INFO] [1771556220.390328864] [rosbag2_recorder]: Press SPACE for pausing/resuming
1: [INFO] [1771556220.390351957] [rosbag2_recorder]: Event publisher thread: Started
1: [INFO] [1771556220.391551560] [rosbag2_recorder]: Starting recording to '/tmpfs/tmp_test_dir_50579f/record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size_0_sqlite3'
1: [WARN] [1771556220.391570434] [rosbag2_recorder]: No input serialization format specified, using default rmw serialization format: 'cdr'.
1: [INFO] [1771556220.393370453] [rosbag2_storage]: Opened database '/tmpfs/tmp_test_dir_50579f/record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size_0_sqlite3/0_record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size_0_sqlite3_2026_02_19-18_57_00.db3' for READ_WRITE.
1: [INFO] [1771556220.394739176] [rosbag2_recorder]: Listening for topics...
1: [INFO] [1771556220.394832891] [rosbag2_recorder]: Recording...
1: [INFO] [1771556220.394871815] [rosbag2_recorder]: Topics discovery started.
1: [INFO] [1771556220.498601748] [rosbag2_recorder]: Subscribed to topic '/test_topic'
1: [INFO] [1771556220.498638374] [rosbag2_recorder]: All requested topics are subscribed. Stopping discovery...
1: [INFO] [1771556220.498643486] [rosbag2_recorder]: Topics discovery stopped.
1: [INFO] [1771556221.709183576] [rosbag2_recorder]: Pausing recording.
1: [INFO] [1771556221.709270752] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
1: [INFO] [1771556221.710287927] [rosbag2_recorder]: Recording stopped
1: [INFO] [1771556221.711846045] [rosbag2_recorder]: Event publisher thread: Exited
1: /home/morlov/ros2_rolling/src/ros2/rosbag2/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp:318: Failure
1: Expected equality of these values:
1:   actual_splits
1:     Which is: 1
1:   expected_splits
1:     Which is: 3
1: 
1: [  FAILED  ] TestRecordEndToEnd/RecordFixture.record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size/0, where GetParam() = "sqlite3" (1833 ms)
```

Tiryoh:
@MichaelOrlov Thanks for the logs. I still can't reproduce it on my end, but I'll try to fix it while building with GitHub Actions based on the logs you shared.

Tiryoh:
I found the root cause of the CI failures. There were two separate issues:

1. **Flaky test**: One of the split tests was flaky. I couldn't reproduce it locally, but I've added a fix to wait for the expected number of storage files.

2. **Build error from deprecated `ament_index_cpp` APIs**: I had not been building with `-DCMAKE_CXX_FLAGS="-Werror"` locally, so I missed that `ament_index_cpp` 1.13.1 deprecated `get_resource()` and `get_resources()` (introduced in ament/ament_index#104). This causes a build failure with `-Werror=deprecated-declarations`. I've migrated to the new `PathWithResource get_resource()` and `get_resources_by_name()` APIs.


@MichaelOrlov Both issues have been fixed. Could you please review again when you get a chance?
@fujitatomoya Could you re-trigger the CI on ci.ros2.org?

MichaelOrlov:
Hi @Tiryoh I've cleaned up your branch a bit.  
I made a rebase on top of the latest Rolling branch and dropped your commit with the ament_index API changes 
> 2. **Build error from deprecated ament_index_cpp APIs:** I had not been building with -DCMAKE_CXX_FLAGS="-Werror" locally, so I missed that ament_index_cpp 1.13.1 deprecated get_resource() and get_resources() (introduced in https://github.com/ament/ament_index/pull/104). This causes a build failure with -Werror=deprecated-declarations. I've migrated to the new PathWithResource get_resource() and get_resources_by_name() APIs.

- I dropped your commit because on Rolling we have already migrated to the new `ament_index_cpp` API in the scope of the #2268, and your changes were overlapping with the original changes.

Tiryoh:
@MichaelOrlov Thanks for the rebase. I think it's reasonable.

MichaelOrlov (approved):
@Tiryoh, I pushed https://github.com/ros2/rosbag2/pull/2265/commits/e7ca3b6a025082666a4b836c8c6e944e5de52263 commit with fixes from the second round of my review on your branch.
There are mostly minor changes in tests and helper functions to improve code readability and maintainability.
Please review it.

If CI passes green, I think we are good to go.

Thank you for your contribution!

Tiryoh:
@MichaelOrlov Thank you for your careful and detailed review. I appreciate the time you took to provide clear feedback on both code and tests.

MichaelOrlov:
Pulls: ros2/rosbag2#2265
Gist: https://gist.githubusercontent.com/MichaelOrlov/c93f5cd04b06969d0424c3c17b3a6619/raw/b20294925cbc4740470823768cec59b0653601be/ros2.repos
BUILD args:  --packages-above-and-dependencies rosbag2_cpp rosbag2_transport rosbag2_compression rosbag2_tests rosbag2_test_common
TEST args:  --packages-above rosbag2_cpp rosbag2_transport rosbag2_compression rosbag2_tests rosbag2_test_common
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18282
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27773/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20801/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7637/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26765/)

PR #2281, opened 2025-12-23
Rolling 0.34.0 release

ahcorde:
Rolling 0.34.0 release.

As soon as this PR is reviewed, I will squash commits, merge on the Rolling branch, fix the tag, and then do a bloom release for the Rosbag2 repo.


MichaelOrlov (changes requested):
@ahcorde Thank you for trying to do a releases for the Rosbag2 repo.  
I was expected to do a release for the Rolling, Kilted and Jazzy sometime in January. (first or second week). So many things need to be merged.  
At least I need to merge https://github.com/ros2/rosbag2/pull/2286, because those PR has an API breaking changes in the `RecorderEventNotifier` class. However, if we will merge it before the next Rosbag2 release on Rolling, Jazzy and Kilted. We will not break API because the `RecorderEventNotifier` class itself was introduced recently after the latest current release. The changes will be considered as non-breaking in this case.

PR #2289, opened 2026-01-04
Add "--max-cache-duration" option to the Recorder to facilitate the time bounded snapshot feature

### LINKED ISSUE #900: Snapshot based on duration

MichaelOrlov:
## Description
This PR adds the ability to limit the inner Recorder cache not only by size, but also by duration in seconds.
This is a more generic alternative to the  "--snapshot-duration" CLI option implemented in the https://github.com/ros2/rosbag2/pull/1844

Originally, the implementation was planed to be made only for the `snapshot mode` to be used as a time bounded snapshot feature. However, it was generalized and the same "--max-cache-duration" CLI option can be used as a cache limiter for the regular recording.
Note: The max-cache-duration it is set to 0 by default for backward compatability behavior. i.e. only `--max-cache-size` cache limiter will be applied by default.

- Fixes https://github.com/ros2/rosbag2/issues/900
- Replaces https://github.com/ros2/rosbag2/pull/1844

Yes, a new "--max-cache-duration" CLI option was added

Yes. GitHub Copilot, GPT-5 was used to help with unit tests and Doxygen comments.

### Additional Information
This PR is not backportable, because it has API/ABI breaking changes.


MichaelOrlov:
@christophebedard @fujitatomoya @emersonknapp A friendly ping here for review.

christophebedard:
I'll start reviewing this today.

rosbag2_storage/src/rosbag2_storage/storage_options.cpp
```diff
@@ -15,6 +15,7 @@
 #include <string>
 #include <unordered_map>
 
+#include "rclcpp/duration.hpp"
```
christophebedard: This include doesn't seem to be necessary/used.

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
 This entire buffer can be dumped to disk on request, saving data only in specified circumstances such as a detected error condition or point of interest, capturing the "last N bytes" of incoming data, therefore making sure that you can trigger snapshot after the fact of the event.
 
-The snapshot is taken by calling the `~/snapshot` service on the recorder, described previously.
+The snapshot is taken by calling the `~/snapshot` service on the recorder, described previously.\
+Triggering a snapshot via CLI:
+```
+$ ros2 service call /rosbag2_recorder/snapshot rosbag2_interfaces/srv/Snapshot
+```
+
+#### Time-limited buffering with --max-cache-duration
```
christophebedard: "time-limited buffering" is more of a low-level feature that supports high-level time-based snapshots, as mentioned below. I think it would be better for users if the (sub)section title mentioned this directly instead of "buffering." Users are probably looking for "time-based snapshot" instead of "time-based buffering."

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
+
+- `--max-cache-duration <seconds>`: Maximum cache duration window, in seconds.
+  - Default: `0` — buffer is limited only by `--max-cache-size`.
+  - If `> 0`: buffer is limited by both time and size:
+    - Time bound: retains only the most recent messages within the duration window.
+    - Size bound: respects `--max-cache-size` (bytes).
+
+Configuring bounds options:
+- Time-only buffer: `--max-cache-size 0 --max-cache-duration > 0`
+- Size-only buffer: `--max-cache-duration 0 --max-cache-size > 0`
```
christophebedard: Minor feedback, but I feel like users will see this and simply copy `--max-cache-size 0 --max-cache-duration > 0` into their terminal, which will lead to an error. Is there a different way to express this that doesn't look like a copy-pasteable CLI option?

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
 The Recorder provides a "snapshot mode", enabled via `--snapshot-mode` or `StorageOptions.snapshot_mode`, which does not write messages to disk as they come in, but instead keeps an in-memory circular buffer of size `--max-cache-size`.
```
christophebedard: Should this be updated to mention `--max-cache-duration`?

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
+- When time bounding is active:
+  - Snapshot mode: the recorded timespan corresponds to the current buffer window at snapshot 
+    trigger time. i.e., up to `--max-cache-duration` seconds.
+  - Non-snapshot (regular) mode: due to producer/consumer buffer swap and write cadence, the
+    effective observed timespan can be up to approximately `2 × --max-cache-duration` seconds.
+
+Examples:
+- Regular recording, time-only buffer (keep last 30 seconds in cache, regardless of size):
+  ```
+  ros2 bag record -a --max-cache-size 0 --max-cache-duration 30
```
christophebedard: for consistency:


```suggestion
  $ ros2 bag record -a --max-cache-size 0 --max-cache-duration 30
```

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
+  ```
+- Snapshot mode, size-only buffer (keep last ~100 MB of recent messages):
+  ```
+  ros2 bag record -a --snapshot-mode --max-cache-size 100000000 --max-cache-duration 0
```
christophebedard: for consistency:


```suggestion
  $ ros2 bag record -a --snapshot-mode --max-cache-size 100000000 --max-cache-duration 0
```

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
+  ```
+- Snapshot mode, time-and-size bounded (keep last 10 seconds, max 50 MB):
+  ```
+  ros2 bag record -a --snapshot-mode --max-cache-size 50000000 --max-cache-duration 10
```
christophebedard: for consistency:


```suggestion
  $ ros2 bag record -a --snapshot-mode --max-cache-size 50000000 --max-cache-duration 10
```

README.md
```diff
@@ -149,7 +149,57 @@ These services enable full remote control of the recording process, allowing you
...
+  ```
+- Snapshot mode, time-only buffer (keep last 7 seconds regardless of size):
+  ```
+  ros2 bag record -a --snapshot-mode --max-cache-size 0 --max-cache-duration 7
```
christophebedard: for consistency:


```suggestion
  $ ros2 bag record -a --snapshot-mode --max-cache-size 0 --max-cache-duration 7
```

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -39,11 +39,20 @@ struct StorageOptions
   // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_duration = 0;
 
-  // The cache size indiciates how many messages can maximally be hold in cache
+  // The cache size. Indiciates how many messages can maximally be hold in cache
```
christophebedard:
```suggestion
  // The cache size. Indicates how many messages can maximally be held in cache
```

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -39,11 +39,20 @@ struct StorageOptions
...
   // before these being written to disk.
-  // A value of 0 disables caching and every write happens directly to disk.
+  // Works together with max_cache_duration bound if set.
+  // A value of 0 disables caching and every write happens directly to disk if
+  // max_cache_duration also set to 0.
```
christophebedard:
```suggestion
  // max_cache_duration is also set to 0.
```

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
```diff
@@ -39,11 +39,20 @@ struct StorageOptions
   // A value of 0 indicates that bagfile splitting will not be used.
   uint64_t max_bagfile_duration = 0;
 
-  // The cache size indiciates how many messages can maximally be hold in cache
+  // The cache size. Indiciates how many messages can maximally be hold in cache
   // before these being written to disk.
-  // A value of 0 disables caching and every write happens directly to disk.
+  // Works together with max_cache_duration bound if set.
+  // A value of 0 disables caching and every write happens directly to disk if
```
christophebedard:
```suggestion
  // A value of 0 disables size-based caching and every write happens directly to disk if
```

(double-check line length)

ros2bag/ros2bag/verb/record.py
```diff
@@ -313,6 +322,17 @@ def validate_parsed_arguments(args, uri) -> str:
     if args.stats_max_publishing_rate < 0 or args.stats_max_publishing_rate > 1000.0:
         return print_error('stats_max_publishing_rate must be between 0 and 1000.')
 
+    if args.max_cache_duration < 0:
+        return print_error('max_cache_duration must be a non-negative integer.')
```
christophebedard, line 333: Why are we checking that `max_cache_duration` isn't negative, but not that `max_cache_cache` isn't negative?

rosbag2_cpp/CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   find_package(test_msgs REQUIRED)
   find_package(std_msgs REQUIRED)
+  find_package(rcutils REQUIRED)
```
christophebedard: do we need this, since we already have it above?

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,62 @@
...
 
 namespace rosbag2_cpp
 {
 namespace cache
 {
-
-MessageCacheBuffer::MessageCacheBuffer(size_t max_cache_size)
-: max_bytes_size_(max_cache_size)
+MessageCacheBuffer::MessageCacheBuffer(size_t max_cache_size, uint32_t max_cache_duration)
+: max_bytes_size_(max_cache_size), max_cache_duration_ns_(max_cache_duration * 1'000'000'000ULL)
```
christophebedard: You should use `RCUTILS_S_TO_NS()` instead for the conversion to ns. `rosbag2_cpp` already depends on `rcutils`

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,62 @@
 #include <memory>
 #include <vector>
 
+#include "rosbag2_cpp/logging.hpp"
 #include "rosbag2_cpp/cache/cache_buffer_interface.hpp"
 #include "rosbag2_cpp/cache/message_cache_buffer.hpp"
```
christophebedard:
```suggestion
#include "rosbag2_cpp/cache/cache_buffer_interface.hpp"
#include "rosbag2_cpp/cache/message_cache_buffer.hpp"
#include "rosbag2_cpp/logging.hpp"
```

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,62 @@
...
+    // Check if adding this message would exceed the time limit
+    if (max_cache_duration_ns_ > 0 && buffer_.size() > 1) {  // If we have at least 2 messages
+      // Note: the newest messages are at the back of the vector
+      auto current_buffer_duration =
+        buffer_.back()->recv_timestamp - buffer_.front()->recv_timestamp;
+      if (current_buffer_duration < 0) {
+        ROSBAG2_CPP_LOG_ERROR_STREAM("Inconsistent timestamps in cache buffer: "
+          << "oldest message timestamp " << buffer_.front()->recv_timestamp
+          << " is earlier than newest message timestamp " << buffer_.back()->recv_timestamp
+          << "Dropping message!");
```
christophebedard:
```suggestion
          << " is earlier than newest message timestamp " << buffer_.back()->recv_timestamp
          << ". Dropping message!");
```

Otherwise this will look like "...timestamp 1234Dropping message!"

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,62 @@
...
+        drop_messages_ = true;
```
christophebedard: Why/when would this happen? Should we really drop all future messages if this happens? Since we use `recv_timestamp`, I think the answer is "yes," but I just wanted to double-check.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,62 @@
...
+      }
+      if (static_cast<uint64_t>(current_buffer_duration) >= max_cache_duration_ns_) {
+        drop_messages_ = true;
```
christophebedard: It's unclear to me if this should be `>=`. This means that choosing a max cache duration of 1 second means we would cache a message with timestamp 1.0 s but drop a message with timestamp 2.0 s. At the very least, this isn't very clear or intuitive in the `max_cache_duration`/`--max-cache-duration` documentation.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -25,9 +25,15 @@ namespace rosbag2_cpp
 namespace cache
 {
 
-MessageCacheCircularBuffer::MessageCacheCircularBuffer(size_t max_cache_size)
-: max_bytes_size_(max_cache_size)
+MessageCacheCircularBuffer::MessageCacheCircularBuffer(
+  size_t max_cache_size,
+  uint32_t max_cache_duration)
+: max_bytes_size_(max_cache_size), max_cache_duration_ns_(max_cache_duration * 1'000'000'000ULL)
```
christophebedard: You should use `RCUTILS_S_TO_NS()` instead for the conversion to ns. `rosbag2_cpp` already depends on `rcutils`

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -38,17 +44,43 @@ bool MessageCacheCircularBuffer::push(CacheBufferInterface::buffer_element_t msg
...
+  // Remove any old items until there is room for a new message
+  while (max_bytes_size_ > 0 &&
+    buffer_bytes_size_ > (max_bytes_size_ - msg->serialized_data->buffer_length))
+  {
     buffer_bytes_size_ -= buffer_.front()->serialized_data->buffer_length;
     buffer_.pop_front();
   }
-  // Add new message to end of buffer
+  // Remove old messages until the time span between the oldest and newest message
+  // is less than or equal to max_cache_duration_ns_.
```
christophebedard, line 61: This comment is not correct. We remove old messages until the time difference between the oldest and newest is less than (not less than or equal) to `max_cache_duration_ns_`, because a time difference equal to `max_cache_duration_ns_` means we drop the oldest message.

However, like I wrote in `rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp`, I think we may want to implement it this way, i.e., not drop a message

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -38,17 +44,43 @@ bool MessageCacheCircularBuffer::push(CacheBufferInterface::buffer_element_t msg
...
+    while (buffer_.size() > 1 && current_buffer_duration_exceed_limit()) {
+      buffer_bytes_size_ -= buffer_.front()->serialized_data->buffer_length;
+      buffer_.pop_front();
+      // Note: the newest messages are at the back of the deque
+      current_buffer_duration = buffer_.back()->recv_timestamp - buffer_.front()->recv_timestamp;
+    }
+  }
+  // Add a new message to the end of the buffer
   buffer_bytes_size_ += msg->serialized_data->buffer_length;
   buffer_.push_back(msg);
```
christophebedard: Since we never check the duration of the buffer _with_ the new message, this means that pushing the new message might mean that we exceed the duration. Right? We should drop old messages until we don't exceed the max duration _with_ the new message:

https://github.com/ros2/rosbag2/blob/8e0a12d52a13f6546d89050296628e8a04b05b04/rosbag2_cpp/include/rosbag2_cpp/cache/message_cache_circular_buffer.hpp#L64-L65

christophebedard (commented):
First pass, without looking at tests for now.

README.md
MichaelOrlov: Well, the time-limited buffering goes above and beyond the "time-limited snapshot" case and is applicable for the generic non-snapshot use cases. What I can do is to add time-limited snapshot at the beginnig of the title to have a somewhat compromise.
```md
### Time-based snapshot and time-limited buffering with --max-cache-duration
```

README.md
MichaelOrlov: Good point I will address it.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: Good point. We should not even add a message with the ill-formed timestamp to the buffer at first hand. And you are right, we shouldn't drop all the next messages until the buffer is swapped by the cache consumer.
I've went ahead and redesigned this part. Now I am using a `prospected_buffer_duration = msg->recv_timestamp - buffer_.front()->recv_timestamp;` instead of the `current_buffer_duration` and changed logic accordingly to address aformentioned issues.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: I've redesigned the logic. Now it is 
```cpp
if (static_cast<uint64_t>(prospected_buffer_duration) > max_cache_duration_ns_) {
        drop_messages_ = true;
}
```
Perhaps it is cleaner now. The logic is simple and pretty obvious, in my opinion: if the `max-cache-duration` is 1 second, we should keep messages in the buffer so they don't exceed 1 second. A timespan of **1 second** between messages is a valid case.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
MichaelOrlov: Your concern is correct. I've redesigned the logic to use 
```cpp
auto prospected_buffer_duration = msg->recv_timestamp - buffer_.front()->recv_timestamp;
```
instead of the `current_buffer_duration`.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
MichaelOrlov: Fixed.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: Fixed

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: Fixed

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: Fixed

rosbag2_cpp/CMakeLists.txt
MichaelOrlov: Removed

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Added checks for the `max_cache_size`

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
MichaelOrlov: Fixed.

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
MichaelOrlov: Fixed

rosbag2_storage/include/rosbag2_storage/storage_options.hpp
MichaelOrlov: Fixed

README.md
MichaelOrlov: Fixed.

MichaelOrlov:
@christophebedard Thank you for the thorough code review and findings. 
I've addressed all your comments. Please verify.

README.md
christophebedard: > Well, the time-limited buffering goes above and beyond the "time-limited snapshot" case and is applicable for the generic non-snapshot use cases.

Absolutely, but I think most users will be looking for "time-limited snapshot" and may overlook this section.

> What I can do is to add time-limited snapshot at the beginnig of the title to have a somewhat compromise.

Yep, that's great, thank you!

README.md
christophebedard: Looks good!

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
```diff
@@ -15,31 +15,74 @@
...
+    // Note: the oldest messages are at the front of the vector
+    prospected_buffer_duration = msg->recv_timestamp - buffer_.front()->recv_timestamp;
+    if (prospected_buffer_duration < 0) {
+      ROSBAG2_CPP_LOG_ERROR_STREAM("Can't calculate prospected cache buffer duration: "
+        << "oldest message timestamp " << buffer_.front()->recv_timestamp
+        << " is earlier than new message timestamp " << msg->recv_timestamp
+        << ". Dropping new message!");
+      // Note: We drop the newly adding message in this case, as we can't determine
+      // whether adding it would exceed the time limit. However, the next message may have
+      // a correct timestamp and be added to the buffer.
```
christophebedard:
```suggestion
      // Note: We drop the newly added message in this case, as we can't determine
      // whether adding it would exceed the time limit. However, the next message may have
      // a correct timestamp and be added to the buffer.
```

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
```diff
@@ -38,17 +45,53 @@ bool MessageCacheCircularBuffer::push(CacheBufferInterface::buffer_element_t msg
...
+    // Note: the oldest messages are at the front of the deque
+    auto prospected_buffer_duration = msg->recv_timestamp - buffer_.front()->recv_timestamp;
+    if (prospected_buffer_duration < 0) {
+      ROSBAG2_CPP_LOG_ERROR_STREAM("Can't calculate prospected circular cache buffer duration: "
+        << "oldest message timestamp " << buffer_.front()->recv_timestamp
+        << " is earlier than new message timestamp " << msg->recv_timestamp
+        << ". Dropping new message!");
+      // Note: We drop the newly adding message in this case, as we can't determine
+      // whether adding it would exceed the time limit. However, the next message may have
+      // a correct timestamp and be added to the buffer.
```
christophebedard:
```suggestion
      // Note: We drop the newly added message in this case, as we can't determine
      // whether adding it would exceed the time limit. However, the next message may have
      // a correct timestamp and be added to the buffer.
```

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
christophebedard: Looks good!

rosbag2_cpp/test/rosbag2_cpp/test_message_cache.cpp
```diff
@@ -23,25 +23,33 @@
 #include <vector>
 #include <thread>
 
+#include "rcutils/time.h"
 #include "rosbag2_storage/ros_helper.hpp"
 #include "rosbag2_storage/serialized_bag_message.hpp"
+#include "rosbag2_cpp/cache/message_cache_buffer.hpp"
```
christophebedard: nitpick: order includes

```suggestion
#include "rcutils/time.h"
#include "rosbag2_cpp/cache/message_cache_buffer.hpp"
#include "rosbag2_storage/ros_helper.hpp"
#include "rosbag2_storage/serialized_bag_message.hpp"
```

rosbag2_cpp/test/rosbag2_cpp/test_circular_message_cache.cpp
```diff
@@ -135,10 +143,174 @@ TEST_F(CircularMessageCacheTest, circular_message_cache_ensure_empty) {
...
+  auto consumer_buffer = circular_message_cache->get_consumer_buffer();
+  auto messages = consumer_buffer->data();
+  circular_message_cache->release_consumer_buffer();
+
+  ASSERT_EQ(messages.size(), expected_count);
+  const auto oldest = messages.front()->recv_timestamp;
+  const auto newest = messages.back()->recv_timestamp;
+  const auto span_ns = newest - oldest;
+  EXPECT_GT(span_ns, 0);
+  EXPECT_LE(span_ns, RCUTILS_S_TO_NS(max_buffer_duration));
```
christophebedard, line 192: This only checks that the messages in the buffer fit inside the maximum time window (1 s). Shouldn't we also verify that the buffer contains the last 1 second of all synthetic messages? We should specifically check that we drop old messages and not new messages. Same for other tests below that check time/duration constraints.

Unlike the (non-circular) buffer, where it drops the newest messages and returns `false` (which you can check), you would have no way of knowing here, since `push()` always returns `true`.

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
```diff
@@ -681,6 +678,80 @@ TEST_P(RecordFixture, rosbag2_record_and_play_multiple_topics_with_filter) {
...
+  cleanup.cancel();
+
+  finalize_metadata_kludge();
+  wait_for_metadata();
+
+  auto msgs = get_messages_for_topic<test_msgs::msg::Strings>(topic_name);
+  EXPECT_THAT(msgs, SizeIs(Eq(expected_count)));
+  for (const auto & m : msgs) {
+    EXPECT_EQ(m->string_value, "time_only");
+  }
```
christophebedard, line 715: Do we check the time window for messages here? Same for the test below with duration & size limits. The tests only seem to publish N messages and then check that it recorded the same number (N) of messages.

MichaelOrlov: Nop. Because we don't have information about timestamps here.

rosbag2_cpp/test/rosbag2_cpp/test_circular_message_cache.cpp
MichaelOrlov: Fixed. I've added a check for the time window here.

rosbag2_cpp/test/rosbag2_cpp/test_message_cache.cpp
MichaelOrlov: Fixed.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: With these changes, I would disagree. We haven't **added** a new message yet, but are in the process of considering adding it. This is why it is in continuous form.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
MichaelOrlov: With these changes, I would disagree. We haven't **added** a new message yet, but are in the process of considering adding it. This is why it is in continuous form.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
christophebedard: it's not valid English, though. Simply changing "the newly adding message" to "the new message" would work.

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
christophebedard: same as https://github.com/ros2/rosbag2/pull/2289#discussion_r2719058535

rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp
christophebedard: Could we play a (known) test bag, record it with an appropriate duration value (N seconds), and then assert the number of messages we recorded (knowing how many messages are in the last N seconds of the bag)?

Anyway, feel free to ignore. We're already testing this lower down the stack by testing the buffer itself.

christophebedard (approved):
There are 2 minor comments about wording, but otherwise LGTM

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_buffer.cpp
MichaelOrlov: Agree. Fixed in the https://github.com/ros2/rosbag2/pull/2289/changes/79cc1034ab46a04eb8c7bf1a807f3c1642ab8627

rosbag2_cpp/src/rosbag2_cpp/cache/message_cache_circular_buffer.cpp
MichaelOrlov: Fixed in the https://github.com/ros2/rosbag2/pull/2289/changes/79cc1034ab46a04eb8c7bf1a807f3c1642ab8627

MichaelOrlov:
Pulls: ros2/rosbag2#2289
Gist: https://gist.githubusercontent.com/MichaelOrlov/1a95cb8c86d1f6ac0222a0a190c41390/raw/a1907149afbc72c122672cfb9ef548364112270e/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
TEST args:  --packages-above ros2bag rosbag2_cpp rosbag2_py rosbag2_storage rosbag2_storage_sqlite3 rosbag2_transport rosbag2_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18027
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27193/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20535/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6923/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26490/)

tonynajjar:
Cool stuff thanks!

mcp-ats:
this is awesome, is there a plan to back-port this to Jazzy?

MichaelOrlov:
> This PR is not backportable, because it has API/ABI breaking changes.

Sorry, but this PR is not backportable because it has API/ABI-breaking changes.

PR #2300, opened 2026-01-07
Fix Kilted CI to use the right ROS distro

ahcorde:
## Description

I noticed here https://github.com/ros2/rosbag2/pull/2294 that `kilted` CI is using `rolling`

I also updated some other links to documentation


MichaelOrlov (approved):
@ahcorde Thank you for the updates.  
I missed it since a Kilted release, and it was working somehow for some time.

PR #2319, opened 2026-01-23
Static topic list for recorder

### LINKED ISSUE #2318: Static topic list for recorder
## Description

We need more determinism when we are starting recording in a way that we would like to avoid some uncertainty related to the topics discovery and following subscription to the specified topics.

In the current implementation, the Rosbag2 recorder relies on the topics discovery from the ROS graph for the topics to be subscribed. While this is Okay for most scenarios, it doesn't give a guarantee that the messages will be recorded as soon as they are published.  

In the current workflow, users usually start publishing nodes, i.e., all system and processing nodes expected to be running beforehand, then the user is expected to start the Rosbag2 recorder, and it will use ROS graph discovery to explore available topics and subscribe to them, even if the topics are specified explicitly via a list.

Such a workflow has a somewhat fussy start when nobody knows at what exact moment the Rosbag2 will start recording on a particular topic, and most likely, it will miss some messages at the beginning. 

As a workaround  for the current workflow to achieve determinism on start, it could be used `wait_for_matched` idiom. However, it means that every publishing node shall be adjusted to wait while the Rosbag2 will discover its topics and create subscriptions before the node starts publishing. This workaround works well for the unit tests; however, in practice, when we have a working system and need to record everything from the beginning. Going ahead and changing each node could be a very tedious task. The system may have more than 100 nodes.

To mitigate the aforementioned limitations, it would be good to have a more deterministic workflow when the Rosbag2 recorder starts, first with the explicitly specified topic list and creates subscriptions regardless of whether they are discoverable via the ROS graph or not. i.e., subscribe to the topics even if they don't yet have publishers.
This way, the user can start the Rosbag2 recorder and then run the entire system without changes and will get all recordings from the very beginning.

One of the reasons why the Rosbag2 recorder uses discovery via ROS graph is that to create a subscription, we need to know both the topic name and the topic type. However, the Rosbag2 CLI option either relies on some regex or relax option like `--all-topics` or, in the best case, we can specify a topic list explicitly via the `--topics` CLI option. But the missing part in this case is a topic type.

## Related Issues

N/A

## Completion Criteria

The user shall be able to specify the "static" topic list with topic types that can be used to create subscriptions to the topics at the start of the recorder, even if the topics are not available/discoverable via the ROS graph.

- Add an extra CLI option "--static-topics-list", with an argument as a path to the yaml config file with a list of pairs "topic_name - topic_type".
- The '--static-topics-path' CLI option will expect yaml file in the following format:
```
static_topics_and_types_list:
  - [topic1_name, topic1_type]
  - [topic2_name, topic2_type]
  - [topic3_name, topic3_type]
```

The relevant test shall be added on the `rosbag2_transport` layer as well as integration tests in the `rosbag2_tests` package.

MichaelOrlov:
## Description

This PR introduces a new CLI option "--static-topics-list", with an argument as a path to the yaml config file with a list of pairs "topic_name - topic_type".
The '--static-topics-path' CLI option will expect yaml file in the following format:
```
static_topics_and_types_list:
  - [topic1_name, topic1_type]
  - [topic2_name, topic2_type]
  - [topic3_name, topic3_type]
```
The static topic list can be used to create subscriptions to the topics at the start of the recording, even if the topics are not available/discoverable via the ROS graph and even if publishers do not yet exist on those topics.

- Fixes #https://github.com/ros2/rosbag2/issues/2318

Yes, a new "--static-topics-list" CLI option will be added to the `ros2 bag record` command.

Yes. The GitHub Copilot, GPT-5 was used to help with writing some tests and doxygen comments.

### Additional Information
This PR is not backportable since it introduces API/ABI-breaking changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ static_topics_and_types_list:
 ```
 
 For topics from the static topics list, subscriptions will be created even if they are not
-discoverable via ROS graph. 
+discoverable via the ROS graph and even if publishers do not yet exist on those topics.
 
 If not further specified, `ros2 bag record` will create a new folder named to the current time stamp and stores all data within this folder.
 A user defined name can be given with `-o, --output`.
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ To record a set of predefined topics, one can specify them on the command line e
 $ ros2 bag record <topic1> <topic2> … <topicN>
 ```
 
-Press ++ctrl+c++ to stop the recording.
+Press `Ctrl+C` to stop the recording.
 
 The specified topics don't necessarily have to be present at the start time.
 The discovery function will automatically recognize if one of the specified topics appears.
--- a/README.md
+++ b/README.md
@@ -70,9 +70,23 @@ To record a set of predefined topics, one can specify them on the command line e
 $ ros2 bag record <topic1> <topic2> … <topicN>
 ```
 
-The specified topics don't necessarily have to be present at start time.
-The discovery function will automatically recognize if one of the specified topics appeared.
-In the same fashion, this auto discovery can be disabled with `--no-discovery`.
+Press ++ctrl+c++ to stop the recording.
+
+The specified topics don't necessarily have to be present at the start time.
+The discovery function will automatically recognize if one of the specified topics appears.
+In the same fashion, this auto-discovery can be disabled with `--no-discovery`.\
+The topics can also be specified via a static topic list with the `--static-topics-path` option.
+Recorder will expect yaml file in the following format:
+
+```yaml
+static_topics_and_types_list:
+ - [/topic_name1, topic_type1]
+ - [/topic_name2, topic_type2]
+ - [/topic_name3, topic_type3]
+```
+
+For topics from the static topics list, subscriptions will be created even if they are not
+discoverable via ROS graph. 
 
 If not further specified, `ros2 bag record` will create a new folder named to the current time stamp and stores all data within this folder.
 A user defined name can be given with `-o, --output`.
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -71,6 +71,14 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
         '--topics', type=str, default=[], metavar='Topic', nargs='+',
         help='Space-delimited list of topics to record.')
     parser.add_argument(
+        '--static-topics-path', type=FileType('r'),
+        help='Path to a YAML file with statically defining topic names and types.'
+             'Recorder will expect a YAML file in the following format:\n'
+             'static_topics_and_types_list:\n'
+             ' - [/topic_name1, topic_type1]\n'
+             ' - [/topic_name2, topic_type2]\n'
+             ' - [/topic_name3, topic_type3]')
+    parser.add_argument(
         '--services', type=str, metavar='ServiceName', nargs='+',
         help='Space-delimited list of services to record.')
     parser.add_argument(
@@ -247,12 +255,13 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
 
 def check_necessary_argument(args):
     # At least one options out of --all, --all-topics, --all-services, --all-actions, --services,
-    # --actions --topics, --topic-types or --regex must be used
+    # --actions --topics, --topic-types, --static-topics-path or --regex must be used
     if not (args.all or args.all_topics or args.all_services or args.all_actions or
             (args.services and len(args.services) > 0) or
             (args.actions and len(args.actions) > 0) or
             (args.topics and len(args.topics) > 0) or
-            (args.topic_types and len(args.topic_types) > 0) or args.regex):
+            (args.topic_types and len(args.topic_types) > 0) or args.regex or
+            args.static_topics_path):
         return False
     return True
 
@@ -265,7 +274,8 @@ def validate_parsed_arguments(args, uri) -> str:
 
     if not check_necessary_argument(args):
         return print_error('Need to specify at least one option out of --all, --all-topics, '
-                           '--all-services, --services, --topics, --topic-types or --regex')
+                           '--all-services, --services, --topics, --topic-types,'
+                           '--static-topics-path or --regex')
 
     if args.exclude_regex and not \
             (args.all or args.all_topics or args.topic_types or args.all_services or
@@ -381,6 +391,10 @@ class RecordVerb(VerbExtension):
             key_value_pairs = [pair.split('=') for pair in args.custom_data]
             custom_data = {pair[0]: pair[1] for pair in key_value_pairs}
 
+        static_topics_uri = ''
+        if args.static_topics_path:
+            static_topics_uri = args.static_topics_path.name
+
         storage_config_file = ''
         if args.storage_config_file:
             storage_config_file = args.storage_config_file.name
@@ -398,6 +412,7 @@ class RecordVerb(VerbExtension):
             custom_data=custom_data
         )
         record_options = RecordOptions()
+        record_options.static_topics_uri = static_topics_uri
         record_options.all_topics = args.all_topics or args.all
         record_options.all_services = args.all_services or args.all
         record_options.all_actions = args.all_actions or args.all
--- a/ros2bag/ros2bag/verb/record.py
+++ b/ros2bag/ros2bag/verb/record.py
@@ -72,7 +72,7 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
         help='Space-delimited list of topics to record.')
     parser.add_argument(
         '--static-topics-path', type=FileType('r'),
-        help='Path to a YAML file with statically defining topic names and types.'
+        help='Path to a YAML file with statically defined topic names and types.'
              'Recorder will expect a YAML file in the following format:\n'
              'static_topics_and_types_list:\n'
              ' - [/topic_name1, topic_type1]\n'
--- a/rosbag2_transport/CMakeLists.txt
+++ b/rosbag2_transport/CMakeLists.txt
@@ -598,6 +598,16 @@ if(BUILD_TESTING)
     rosbag2_test_common::rosbag2_test_common
     rcutils::rcutils
   )
+
+  ament_add_gmock(test_record_static_topics
+    test/rosbag2_transport/test_record_static_topics.cpp)
+  target_link_libraries(test_record_static_topics
+    ${PROJECT_NAME}
+    ${test_msgs_TARGTES}
+    rosbag2_test_common::rosbag2_test_common
+    test_msgs::test_msgs_includes
+  )
+
 endif()
 
 ament_package()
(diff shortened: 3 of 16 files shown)


MichaelOrlov:
@christophebedard @fujitatomoya @emersonknapp A  friendly ping for review.

ros2bag/ros2bag/verb/record.py
```diff
@@ -70,6 +70,14 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
     topics_args_group.add_argument(
         '--topics', type=str, default=[], metavar='Topic', nargs='+',
         help='Space-delimited list of topics to record.')
+    parser.add_argument(
+        '--static-topics-path', type=FileType('r'),
+        help='Path to a YAML file with statically defining topic names and types.'
```
christophebedard:
```suggestion
        help='Path to a YAML file with statically-defined topic names and types.'
```

ros2bag/ros2bag/verb/record.py
```diff
@@ -70,6 +70,14 @@ def add_recorder_arguments(parser: ArgumentParser) -> None:
     topics_args_group.add_argument(
         '--topics', type=str, default=[], metavar='Topic', nargs='+',
         help='Space-delimited list of topics to record.')
+    parser.add_argument(
+        '--static-topics-path', type=FileType('r'),
```
christophebedard, line 74: it seems annoying to me to have to write a YAML file and then pass its path through `--static-topics-path`. What about a list of two strings: `--static-topic /topic_name1 topic_type1 --static-topic /topic_name2 topic_type2 --static-topic /topic_name3 topic_type3`. This could be stored as a list/vector of pairs of strings.

Just something to consider.

rosbag2_transport/include/rosbag2_transport/topic_filter.hpp
```diff
@@ -20,6 +20,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
+#include <utility>
```
christophebedard:
```suggestion
#include <utility>
#include <vector>
```

README.md
```diff
@@ -70,9 +70,23 @@ To record a set of predefined topics, one can specify them on the command line e
 $ ros2 bag record <topic1> <topic2> … <topicN>
 ```
 
-The specified topics don't necessarily have to be present at start time.
-The discovery function will automatically recognize if one of the specified topics appeared.
-In the same fashion, this auto discovery can be disabled with `--no-discovery`.
+Press ++ctrl+c++ to stop the recording.
```
christophebedard: `++ctrl+c++` doesn't get formatted in Markdown like it does in RST

Image

Maybe this instead?


```suggestion
Press `Ctrl+C` to stop the recording.
```

rosbag2_transport/test/rosbag2_transport/test_record_static_topics.cpp
```diff
@@ -0,0 +1,77 @@
+// Copyright 2024, Apex.AI
```
christophebedard: update year

rosbag2_transport/CMakeLists.txt
```diff
@@ -598,6 +598,16 @@ if(BUILD_TESTING)
     rosbag2_test_common::rosbag2_test_common
     rcutils::rcutils
   )
+
+  ament_add_gmock(test_record_static_topics
+    test/rosbag2_transport/test_record_static_topics.cpp)
+  target_link_libraries(test_record_static_topics
+    ${PROJECT_NAME}
+    ${test_msgs_TARGTES}
```
Barry-Xu-2018: typo

```suggestion
    ${test_msgs_TARGETS}
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -1150,6 +1188,39 @@ void RecorderImpl::warn_if_new_qos_for_subscribed_topic(const std::string & topi
...
+        throw std::runtime_error(
+                "Top-level key 'static_topics_and_types_list' must contain a list of "
+                "'topic_name, topic_type' pairs");
+      }
+      for (const auto & bag_node : bag_nodes) {
+        const auto topic_name = bag_node[0].as<std::string>();
+        const auto topic_type = bag_node[1].as<std::string>();
+        if (topic_type.empty()) {
+          RCLCPP_ERROR_STREAM(node->get_logger(),
+            "Static topic " << topic_name.c_str() << " has no corresponding type");
```
Barry-Xu-2018:
```suggestion
            "Static topic " << topic_name << " has no corresponding type");
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -470,6 +483,16 @@ void RecorderImpl::record(const std::string & uri)
   if (!record_options_.use_sim_time) {
     subscribe_topics(get_requested_or_available_topics());
   }
+
+  // Disable discovery if only static topics defined
+  if (!static_topics_.empty() &&
+    !(record_options_.all_topics || !record_options_.topics.empty() ||
+    record_options_.all_services || !record_options_.services.empty() ||
```
Barry-Xu-2018, line 490:
```suggestion
    record_options_.all_services || !record_options_.services.empty() ||
    record_options_.all_actions || !record_options_.actions.empty() ||
```

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -1150,6 +1188,39 @@ void RecorderImpl::warn_if_new_qos_for_subscribed_topic(const std::string & topi
...
+        const auto topic_name = bag_node[0].as<std::string>();
+        const auto topic_type = bag_node[1].as<std::string>();
+        if (topic_type.empty()) {
+          RCLCPP_ERROR_STREAM(node->get_logger(),
+            "Static topic " << topic_name.c_str() << " has no corresponding type");
+        }
+        static_topics_.emplace_back(topic_name, topic_type);
+      }
+    } catch (std::exception & e) {
+      RCLCPP_ERROR_STREAM(node->get_logger(), "Failed to read static topics list: " << e.what());
```
Barry-Xu-2018, line 1219: This error log indicates that the static topics list cannot be read, but some data may have already been inserted into static_topics_.

Should we clear static_topics_?

Barry-Xu-2018 (commented):
When writing a static topic YAML file, if I want to configure a static service or a static action, does it mean I need to manually specify the service event topic for the service or the five internal topics required for the action?

ros2bag/ros2bag/verb/record.py
MichaelOrlov: Fixed.

ros2bag/ros2bag/verb/record.py
MichaelOrlov: We have a case where the system is statically defined and has around **300** topics. Using the proposed option with `--static-topic /topic_name1 topic_type1 --static-topic /topic_name2 topic_type2 --static-topic /topic_name3 topic_type3` is not scalable and will be a nightmare in this case.

rosbag2_transport/include/rosbag2_transport/topic_filter.hpp
MichaelOrlov: Fixed

README.md
MichaelOrlov: Good catch. Fixed

rosbag2_transport/test/rosbag2_transport/test_record_static_topics.cpp
MichaelOrlov: Fixed

rosbag2_transport/CMakeLists.txt
MichaelOrlov: Good catch! Fixed

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Good catch! Fixed.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: That is a good finding. I agree that it would be more consistent to clear the `static_topics_` list in case of partial readout.  
I've added extra logging and cleanup for the `static_topics_` list in the catch statemenent.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
MichaelOrlov: Fixed

MichaelOrlov:
@Barry-Xu-2018 As regards:
> When writing a static topic YAML file, if I want to configure a static service or a static action, does it mean I need to manually specify the service event topic for the service or the five internal topics required for the action?

The answer is yes, need to manually specify the service event topic for the service or the five internal topics required for the action. We don't have special markups or separate lists for services and actions.  
A statically defined system is rarely changed and usually only needs to be written once. Usually, it is not a big hurdle to add or remove topics for services or actions explicitly using a copy-paste technique in the text editor.

MichaelOrlov:
@christophebedard @Barry-Xu-2018 Thank you for your review. I've addressed your review comments.
Please reiterate on review.

Barry-Xu-2018:
>> When writing a static topic YAML file, if I want to configure a static service or a static action, does it mean I need to manually specify the service event topic for the service or the five internal topics required for the action?
>
> The answer is yes, need to manually specify the service event topic for the service or the five internal topics required for the action. We don't have special markups or separate lists for services and actions.
> A statically defined system is rarely changed and usually only needs to be written once. Usually, it is not a big hurdle to add or remove topics for services or actions explicitly using a copy-paste technique in the text editor.

Thanks for explaining that.  
Now user using the command line don’t really know how services and actions get recorded, so the command line just asks them to put in the service name and action name. If users want to set up a static topic for a service or action, they need to know a bit about how these are actually recorded—basically, they need to configure the right topics and types themselves. But I think in the future it’d be simpler if we could just have the service name and type, or the action name and type, written straight into the config.

I don’t have any other comments.

christophebedard:
> Now user using the command line don’t really know how services and actions get recorded, so the command line just asks them to put in the service name and action name. If users want to set up a static topic for a service or action, they need to know a bit about how these are actually recorded—basically, they need to configure the right topics and types themselves. But I think in the future it’d be simpler if we could just have the service name and type, or the action name and type, written straight into the config.

I agree. This could be improved later on

christophebedard (approved):
LGTM with green CI

MichaelOrlov:
Pulls: ros2/rosbag2#2319
Gist: https://gist.githubusercontent.com/MichaelOrlov/2bd5840553690b6d7c6e73b3a47c3055/raw/d709427538997af0f602282af2dca72a273ea923/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18104
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20620/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7384/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26575/)

PR #2320, opened 2026-01-25
[jazzy] Add accidentally deleted `typesupport_helper.{cpp}hpp` files

MichaelOrlov:
## Description

- This PR adds back accidentally deleted `typesupport_helper.{cpp}hpp` files in the https://github.com/ros2/rosbag2/pull/2130

Fixes # N/A
- Relates https://github.com/ros2/rosbag2/pull/2316#issuecomment-3797109708

### Is this user-facing behavior change?
No.

### Did you use Generative AI?
No.

### Additional Information
N/A

PR #2316, opened 2026-01-21
Jazzy 0.26.10 release

marcoag:
## Description

Jazzy release 0.26.10

Will tag the `jazzy` branch and release to rosdistro once this is merged.

Fixes # (issue)

### Is this user-facing behavior change?

No.

### Did you use Generative AI?

No.

### Additional Information

N/A

PR #2322, opened 2026-01-25
Revert "Jazzy 0.26.10 release (#2316)"

MichaelOrlov:
## Description

- This PR reverts the Jazzy 0.26.10 release due to an API breakage introduced in the https://github.com/ros2/rosbag2/pull/2130 and described in the https://github.com/ros2/rosbag2/pull/2316#issuecomment-3797109708.
- The API breakage will be addressed in the #2320. After that, we can make the Jazzy 0.26.10 release again.

Fixes # N/A

No.

No.

### Additional Information
N/A


christophebedard:
0.26.10 was already tagged (and tags are not supposed to be deleted), so I think the way to solve this is to simply create another release after the fix was made, i.e., 0.26.11

MichaelOrlov:
> 0.26.10 was already tagged (and tags are not supposed to be deleted), so I think the way to solve this is to simply create another release after the fix was made, i.e., 0.26.11

Well, I think we can move tag to a new advanced 0.26.10 release.

christophebedard:
The point is that, while you can technically move the tag, you shouldn't. A 0.26.10 release was already created. It's probably not the case in practice, but what if someone is already using version 0.26.10, somehow? We'd end up with two different versions out there with the same version number.

Anyway, I'll approve and let you go ahead if you still want to.

christophebedard (approved):

MichaelOrlov:
- One important correction to note the `0.26.10` release was reverted in the https://github.com/ros/rosdistro/pull/49579.

It turns out the release `0.26.10` doesn't exist anymore, and nobody got it. Therefore, it is safe to advance the `0.26.10` tag to a new base with a fix.

marcoag:
My preference would be to not move the tag and create a new one too, not because of it being released or not but mostly because afaik it's a general good practice to not delete/move tags. With that said I'll leave it up to you @MichaelOrlov.

marcoag (approved):

MichaelOrlov:
I prefer a clean and concise sequence of tags and releases. 
IMO, a tag without a release will cause more confusion and questions.

Merging this PR then.

michaelcarlstrom-civrobotics:
Now that this all seems resolved could we get a new non-breaking release into rosdistro? I'm really excited to try out the new pause/resume api.

MichaelOrlov:
@michaelcarlstrom-civrobotics I can do a new release.  
But I am a bit confused by your statement
> I'm really excited to try out the new pause/resume api.

Can you please clarify which PR you are referring to?   
As far as I recall, we made changes to `pause/resume` only for Rolling, and Jazzy doesn't seem to have them.

PR #2317, opened 2026-01-22
mcap_vendor: Update to the latest release (v2.1.1)

MichaelOrlov:
## Description
- This PR updates `mcap_vendor` package to refer to the latest [mcap/cpp  v2.1.1 release](https://github.com/foxglove/mcap/releases/tag/releases%2Fcpp%2Fv2.1.1)

- Full changelog can be found here https://github.com/foxglove/mcap/compare/releases/cpp/v1.4.0...releases/cpp/v2.1.1

Fixes # N/A

### Is this user-facing behavior change?
No.

### Did you use Generative AI?
No.

### Additional Information
N/A

PR #2323, opened 2026-02-03
Update mcap vendor to version 1.4.2

mosfet80:
## Description

Changelog:
PR 1291 added an optimization, which caused the behavior reported in issue 1355. The intention of that change was to close a non-empty chunk before writing a message that would overflow the chunk size. It checks whether the chunk is non-empty _after_ writing `Channel` & `Schema` records to the chunk. Therefore, it's possible to write a chunk that contains no messages.

The writer has a small bug in the metadata that it writes into the chunk index record. The `currentChunkStart_` register is initialized as `MaxTime`, and only updated when a message is added to the chunk. But when writing a message-less chunk, we need to write the message start time as 0.

The indexed message reader has a small bug in the way it detects the presence of a message index. It only looks for a message index on the first chunk index record, instead of scanning all chunk index records. This is why the test case provided in 1355 fails: the first chunk contains only `Channel` & `Schema` records without messages, and therefore the first chunk index has no message index.

Remove dead members in McapReader
cpp: Fix for undefined behavior in mcap::ParseByteArray(..)

### DIFF
--- a/mcap_vendor/CMakeLists.txt
+++ b/mcap_vendor/CMakeLists.txt
@@ -34,8 +34,8 @@ endif()
 macro(build_mcap_vendor)
   include(FetchContent)
   fetchcontent_declare(mcap
-    URL https://github.com/foxglove/mcap/archive/refs/tags/releases/cpp/v1.4.0.tar.gz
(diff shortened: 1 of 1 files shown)


MichaelOrlov:
- Closing in favor of the https://github.com/ros2/rosbag2/pull/2317 we need to bump mcap to the latest version.

PR #2331, opened 2026-02-11
Fix YAML decode/encode for ignore_leaf_topics, start_paused and use_sim_time in RecordOptions

Baranbologur:
## Description

This PR adds missing YAML encode/decode support for the following rosbag2_transport::RecordOptions fields:

- ignore_leaf_topics
- start_paused
- use_sim_time

These fields are defined in RecordOptions but were not handled in the YAML::convert<RecordOptions> specialization. As a result, when using YAML-based configuration, these options were silently ignored and always retained their default values.

This change updates both:

- encode() to properly serialize these fields.
- decode() to correctly parse them from YAML.

The behavior remains unchanged unless these parameters are explivitly provided in a YAML configuration file.

Yes, but only in the case where these parameters are provided in YAML configuration.

Previously, the following YAML fields had no effect:

- ignore_leaf_topics
- start_paused
- use_sim_time

After this change, these parameters are correctly applied when specified in YAML profiles.

There is no change for users who do not use these fields in YAML.

No.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/record_options.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/record_options.cpp
@@ -60,6 +60,9 @@ Node convert<rosbag2_transport::RecordOptions>::encode(
   node["topic_qos_profile_overrides"] = qos_overrides;
   node["include_hidden_topics"] = record_options.include_hidden_topics;
   node["include_unpublished_topics"] = record_options.include_unpublished_topics;
+  node["ignore_leaf_topics"] = record_options.ignore_leaf_topics;
+  node["start_paused"] = record_options.start_paused;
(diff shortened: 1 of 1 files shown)


MichaelOrlov (changes requested):
@Baranbologur Thank you for the PR. And sorry that we overlooked it at the first place.

Can you please retarget your branch against the Rolling branch?  
We need to first fix the issue on Rolling, then backport to other distros.

Baranbologur:
As requested, I have opened a new PR targeting the `rolling` branch: #2334

Please let me know if this is aligned with what you had in mind.  
This Humble PR can either be closed or kept for backport after the Rolling PR is merged.

MichaelOrlov:
- Closing this PR for now. 

Will create another one automatically by the mergify agent

PR #2362, opened 2026-03-09
Fix for a possible race condition in compression writer on close

MichaelOrlov:
## Description

It was possible a race condition in the close() method. If all compressor threads are busy compressing files, we may end up exiting before scheduling to compress the last file, which would lead to the issue when the last file appears uncompressed.

The race condition was resolved by sequentially closing the current file, waiting until all compressors finish and then compressing the last file in the current thread instead of pushing the last file into the compressors' queue and stopping all compressors.

Fixes # (issue) N/A

### Is this user-facing behavior change?
No.

### Did you use Generative AI?
Yes. Codex GPT-5.4 to help with comments and code cleanup

### Additional Information
Can be backported

Issue #2108, opened 2025-07-31
Bag recorder sometimes loses/skips data on bag split


RMichaelSwan:
## Description
The bag recorder seems to skip recording some data when large bag file splits (~4-5 GB) are saved. This behavior does not seem to occur for smaller split sizes. It is important to note that this occurred when using an external hard drive for storage.

### Expected Behavior
Bag recorder would not lose data it is able to capture.

### Actual Behavior
Data is missing for 1 minute every ~10 minutes (time it takes for us to record 4.7GB on our robot). In previous recording tests with smaller splits (~50MB) this does not occur. The bag recorder node does not give any error in the logs.

## To Reproduce
1. Record bag with large split (e.g. 5GB)
2. After 1 or more splits occur, end recording
3. Review data timestamps, look for gaps

## System (please complete the following information)
 - OS: Ubuntu Noble
 - ROS 2 Distro: Jazzy
 - Install Method: APT
 - Version: 0.26.6 (latest on Jazzy apt as of 2025-04-02)

## Additional context
We were directing the recorder to save to an external hard drive [like this one](https://www.seagate.com/products/creative-pro/lacie-rugged-ssd/) which has a USB 3.1 Gen 2 link rate. My guess is that the file writer took long enough that it held up the recorder from continuing to capture data. I suppose that this kind of issue could be partially alleviated by the proposal from #2007 though a more robust solution would be preferred if possible. Additional error messaging if this is not logically resolveable without changing settings/environment would be helpful .

MichaelOrlov:
Hi @RMichaelSwan are you using MCAP storage plugin for recording?
If a messages loss happened on the Rosbag2 recorder side, you should see the warning message lke 
```bash
Cache buffers lost messages per topic: 
Total lost: 
```
At the end of the recording or after the each bag split.
Form issue description, it looks like the disk too slow when dumping messages from the Rosbag2 inner cache.
Very litle we can do in this case. We already have double buffering scheme in the recorder. That allows us to decouple transport layer from the storage. I can recommend playing around with the inner Rosbag2 recorder cache size. 
This is a `--max-cache-size` CLI option which is in bytes. Also I would recommend increasing DDS QoS queue depth via QoS override CLI option. Please refer to the https://github.com/ros2/rosbag2/blob/rolling/README.md#overriding-qos-profiles for details.

RMichaelSwan:
This is with the MCAP storage plugin, yes. I do not see any warnings at all actually.

[...]

PR #2382, opened 2026-03-16
[WIP] Add non-blocking asynchronous bag splitting to Rosbag2 recorder/writer

MichaelOrlov:
## Description

Add non-blocking asynchronous bag splitting to rosbag2 recorder/writer.
At a high level, the PR changes bag splitting from a blocking storage switch into an asynchronous operation that can run without stalling ongoing message writes. The concurrency work ensures the writer can continue accepting messages while the split is being processed, and the recorder now uses that async path directly.
Change also includes test updates for the async path and removes one timing-sensitive assertion that was no longer reliable with asynchronous execution.

- Fixes #https://github.com/ros2/rosbag2/issues/2108

Yes.
From a user perspective, bag splitting during recording becomes more robust under load.
User-visible impact:
- recorder-triggered splits no longer block recording in the same way as before
- incoming messages can continue to be processed while a split is happening
- the risk of stalls and message loss during long split operations is reduced
- service- or recorder-triggered splits may complete slightly after the exact request point because execution is asynchronous
This is most relevant for users:
- recording high-rate topics
- splitting by size or time during active recording
- triggering bag splits while recording is ongoing
- using recorder services that depend on split behavior

### Backward Compatibility
This change is largely backward compatible at the API and workflow level, but there is an observable behavioral nuance.
Compatible aspects:
- existing recording workflows continue to work
- bag splitting remains available through the same recorder flows
- no user-facing CLI split options are removed or renamed
- the new async capability is additive at the writer API level

Behavioral difference to call out:
- split timing is no longer strictly synchronous from the caller’s perspective
- callers should not assume the new bagfile is opened immediately at the exact instruction boundary
- if any downstream code relied on split completion being instantaneous/blocking, that assumption may no longer hold

So the practical compatibility summary is:
- no breaking interface change for normal users
- minor semantic change for code or tests that depended on synchronous split completion timing

Yes. I used Codex gpt-5.4 to help with some tasks and analysis of the problems.

### Additional Information
Not backportable due to the API/ABI breaking changes.

This PR depends on the:
1. [Add "--max-cache-duration" option to the Recorder to facilitate the time bounded snapshot feature (#2289)](https://github.com/ros2/rosbag2/pull/2289)
2. [Feature: Circular logging by split count (--max-bag-files) (#2218)](https://github.com/ros2/rosbag2/pull/2218)
3. [Update Rosbag2 filename format to index+name+timestamp (#2265)](https://github.com/ros2/rosbag2/pull/2265)
4. [Fix for a possible race condition in compression writer on close (#2362)](https://github.com/ros2/rosbag2/pull/2362)


carlos-apex:
@MichaelOrlov I continued working on top of your commits. I pushed our changes to the following branch. 

https://github.com/ros2/rosbag2/compare/rolling...carlos-apex:rosbag2:csv/async-split-nonblocking-upstream

Here is a short description and justification for the new commits.
_____

https://github.com/ros2/rosbag2/commit/3e4ac61805417408775771ed2240484242c46cad

This commit changes the order in which pending split timestamps are handled. It could be potentially moved to a follow-up PR with some re-work.

____

https://github.com/ros2/rosbag2/commit/3328e9ed310dc6fd382f2eb4115425f77a43846c

This commit addresses a problem where split_bagfile_shared_future_ was not clean-up properly. The future was expected to be cleaned-up as part of `wait_for_pending_split`. However, this might not be the case if the function is not called. 

The commit does a lazy clean-up at the `start_split_bagfile_async` function beginning. As a consequence, simultaneous async splits are not allowed. If there is an ongoing async split the new one is rejected. 


_____

https://github.com/ros2/rosbag2/commit/857ded738bda33faa97ee252535905524c4d9ab0

This commit addresses this TODO by adding a specific mutex.

// TODO(morlov): Protect messages_dropped_per_topic_ with mutex since we can call write(msg) concurrently

_____

https://github.com/ros2/rosbag2/commit/0888531d6cc4e291c45cdf68cd69f50ca7e91ed7

Added new tests to exercise split concurrency 

_____

https://github.com/ros2/rosbag2/commit/290d6bf12edf9eed856a30412cbeabf85a73d862

This commit addresses a TSAN violation caught by the tests in the commit above

MichaelOrlov:
@carlos-apex Thank you for following up and your contribution. 
Unfortunately, I currently don't have much capacity to finish this PR, while I am searching for a new job. I will get back to it when things settle down.

carlos-apex:
Hi @MichaelOrlov I would like to include here some additional commits for completeness.

Don't worry if you don't have capacity now, I just wanted to include some potential optimizations so they are available when reviewing this PR.

__________

With the current implementation we are still observing message drops for high throughput recordings.

The problem is we are still doing some tasks while the cache is paused. The main idea is to keep the cache-consumer pause as short as possible.

We are considering different optimizations by deferring some tasks during the pause. 

Here is a list of the optimizations we implemented:

https://github.com/carlos-apex/rosbag2/commits/csv/async-split-nonblocking-upstream-optimization/

- [Defer old storage close after cache restart](https://github.com/carlos-apex/rosbag2/commit/0e86ac98ea356390b6edca7e381bca1e69be751c)
- [Defer old metadata update after cache restart](https://github.com/carlos-apex/rosbag2/commit/986b16269ec5bc694f7542af6f918a282ac35d98)
- [Skip intermediate MCAP metadata update on split](https://github.com/carlos-apex/rosbag2/commit/646f5753a6a806efc59c574e5fb9ca874f2d5e20)
- [Snapshot split topics before cache stop](https://github.com/carlos-apex/rosbag2/commit/288fb8c9967dc7084db238cf0938bb932fcd36a9)

We measured the time contribution and the impact of message drops with a benchmark and the 3 first commits are the ones which are most relevant. The last one is a minor change.

After those changes the main contributor to the pause duration is the topic creation. I considered create the topics before the pause. However, this looks like a bigger behavior change and I'm not sure about the side effects.

Let me know if the these optimizations make sense to you.

Issue #2409, opened 2026-04-22
Recorder produces empty bags on second record() after stop()


brow1633:
## Description
In Jazzy and Kilted, recorders will not resume after stop.  This was ~quietly~ fixed in #2313 for rolling, which was not back ported due to new `disable_callbacks` API. ~however the removal of `pause()` from `stop()` probably should be.~

Seems like this was first introduced in 4da63dbf by removing the `paused_` reset in `record()` with the rationale:
```
    * Make record_options_.start_paused to be used only in constructor
     
    Signed-off-by: Michael Orlov <morlovmr@gmail.com>
```

### Expected Behavior
The sequence `record()`, `stop()`, `record()` should result in a recorder writing data.

### Actual Behavior
The above sequence results in a recorder in the paused state.

## To Reproduce
```
ros2 run demo_nodes_cpp talker
ros2 bag record -o /tmp/bag1 /chatter
ros2 service call /rosbag2_recorder/stop rosbag2_interfaces/srv/Stop {}
ros2 service call /rosbag2_recorder/record rosbag2_interfaces/srv/Record "{uri: "/tmp/bag2"}"
ros2 service call /rosbag2_recorder/stop rosbag2_interfaces/srv/Stop {}
```

Results in these bags:
```
Files:             bag1_0.mcap
Bag size:          8.3 KiB
Storage id:        mcap
ROS Distro:        jazzy
Duration:          47.000014960s
Start:             Apr 21 2026 20:42:58.811797179 (1776829378.811797179)
End:               Apr 21 2026 20:43:45.811812139 (1776829425.811812139)
Messages:          48
Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 48 | Serialization Format: cdr
Service:           0
Service information: 
```
```
Files:             bag2_0.mcap
Bag size:          4.6 KiB
Storage id:        mcap
ROS Distro:        jazzy
Duration:          0.000000000s
Start:             Apr 11 2262 16:47:16.854775807 (9223372036.854775807)
End:               Apr 11 2262 16:47:16.854775807 (9223372036.854775807)
Messages:          0
Topic information: Topic: /chatter | Type: std_msgs/msg/String | Count: 0 | Serialization Format: cdr
Service:           0
Service information: 
```

## System (please complete the following information)
 - OS: Ubuntu
 - ROS 2 Distro: Jazzy/Kilted
 - Install Method: apt
 - Version: release

MichaelOrlov:
[...]

PR #2313, opened 2026-01-12
Fix for the flaky "can_record_again_after_stop" test

MichaelOrlov:
## Description

This PR addresses undefined behaviour and possible call for the callbacks of the already deleted subscriptions after calling `Recorder::Stop()` API.

Note: The subscription's callbacks propagated to the executor and may still be in the executor's queue, but they will no longer be called after disabling callbacks.

- This PR depends on the non-backportable https://github.com/ros2/rclcpp/pull/2985

- Fixes #https://github.com/ros2/rosbag2/issues/1914

### Is this user-facing behavior change?
No.

### Did you use Generative AI?
No.

### Additional Information
- This PR is not backportable because it depends on the non-backportable https://github.com/ros2/rclcpp/pull/2985

PR #2410, opened 2026-04-22
[Kilted] Don't pause recorder after stop

brow1633:
## Description

Removes pause from stop, as in #2313 from rolling.

Fixes [# 2409](https://github.com/ros2/rosbag2/issues/2409)

Yes, but should fix a regression.

Claude Opus 4.6 assisted with triage.

### Additional Information

Can be back ported to Jazzy as well.

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -273,7 +273,6 @@ void RecorderImpl::stop()
   }
 
   stop_discovery();
-  pause();
   subscriptions_.clear();
   writer_->close();  // Call writer->close() to finalize current bag file and write metadata
 


MichaelOrlov (changes requested):
This pause was introduced on purpose to mitigate undefined behaviour.

The problem is that the subscription's callbacks are propagated to the executor and may still be in the executor's queue after finishing the `stop()` operation.
Using a pause was to ensure that even if callbacks from deleted subscriptions would be called, they wouldn't propagate and call the deleted writer.
Of course pause is not completely fixing the undefined behaviour, but removing it will make it worse and more often will cause unexpected crashes.

The proper fix is to use a new API `disable_callbacks` as in the https://github.com/ros2/rosbag2/pull/2313,  but unfortunately, we can't backport it.

brow1633:
> This pause was introduced on purpose to mitigate undefined behaviour.
> 
> The problem is that the subscription's callbacks are propagated to the executor and may still be in the executor's queue after finishing the `stop()` operation. Using a pause was to ensure that even if callbacks from deleted subscriptions would be called, they wouldn't propagate and call the deleted writer. Of course pause is not completely fixing the undefined behaviour, but removing it will make it worse and more often will cause unexpected crashes.
> 
> The proper fix is to use a new API `disable_callbacks` as in the #2313, but unfortunately, we can't backport it.

Ah, understood.  In that case, can we reset the `paused_` state when calling `record()`?  Having pause here (and a lack of a reset in `record()`) changed behavior.  More info: https://github.com/ros2/rosbag2/issues/2409

MichaelOrlov:
As regards 
> In that case, can we reset the paused_ state when calling record()?

Unfortunately, not because in this case we will break another feature `start-paused`. We do have a dedicated CLI option `--start-paused` initialized here
https://github.com/ros2/rosbag2/blob/9687a1111b22ecd0866d807072a79bad7275dbad/rosbag2_transport/src/rosbag2_transport/recorder.cpp#L204

The workaround would be to call `resume()` API or service call explicitly after the `stop()` and before the next `record()` API calls.

**Update:** or even better to call `resume()` API after the second `record()` call to avoid undefined behavior.

brow1633:
The current behavior seems to break the most common/obvious use case, but
yes we have implemented the workaround.  With start paused set to false, I
would never expect to need to resume when calling record again or after
calling stop.

As an anecdote, due this change, we lost several terabytes of data in a single day.

MichaelOrlov:
@brow1633 
> As an anecdote, due this change, we lost several terabytes of data in a single day.

Sorry about this. To make a proper fix, I had to make significant changes on a RCLCPP layer. It was a fundamental problem in the ROS 2 executor design.

MichaelOrlov:
@brow1633 I created a follow-up PR https://github.com/ros2/rosbag2/pull/2412 to outline this behavior as a known limitation in the README file.

MichaelOrlov:
- Closing as wouldn't be merged. Since pause on stop was introduced as a workaround to mitigate undefined behavior when callbacks from subscriptions are called after the subscription's destruction. 
- The current deviated behavior was documented in the README.md file and headers in the follow-up https://github.com/ros2/rosbag2/pull/2412 PR.

PR #2423, opened 2026-05-06
feat: Add support for writing custom metadata records to the bag

tonynajjar:
## Description

Add `write_custom_metadata()` API to allow users to write arbitrary MCAP Metadata records through the rosbag2 writer stack.

Currently, there is no way to write custom metadata records to an MCAP bag through rosbag2. The only metadata written is the internal `"rosbag2"` summary record via `update_metadata()`. Users who need to annotate bags with custom key-value metadata (e.g., robot name, event markers, provenance info) must resort to binary surgery on the MCAP file after it is closed — reopening the file, truncating the summary section, appending metadata records, and rewriting the footer/index.

This PR adds a clean API path through the entire stack:

1. **`rosbag2_storage::BaseWriteInterface`** — new virtual `write_custom_metadata(name, kv)` with a default no-op (non-breaking for existing storage plugins like sqlite3)
2. **`rosbag2_storage_mcap::MCAPStorage`** — overrides it to write a native `mcap::Metadata` record
3. **`rosbag2_cpp::writer_interfaces::BaseWriterInterface`** — new pure virtual
4. **`rosbag2_cpp::writers::SequentialWriter`** — forwards to `storage_->write_custom_metadata()`
5. **`rosbag2_cpp::Writer`** — public thread-safe method

### Usage

```cpp
auto writer = std::make_shared<rosbag2_cpp::Writer>();
writer->open(storage_options);

// ... write messages ...

writer->write_custom_metadata("my_application", {
  {"robot_name", "arri-160"},
  {"created_at", "2026-05-06T08:00:00Z"},
  {"map_name", "warehouse_A"},
});

writer->close();
```

The metadata records appear in the MCAP summary section and are indexed for fast retrieval without scanning messages.

Fixes # (issue)

Yes. Users can now write custom MCAP Metadata records through `rosbag2_cpp::Writer::write_custom_metadata()`. This is additive — no existing behavior is changed. Storage plugins that don't support custom metadata (e.g., sqlite3) inherit the default no-op.

Yes — GitHub Copilot (Claude Opus 4.6) was used to implement and iterate on this change.

### Additional Information

- The `BaseWriteInterface` method has a default empty implementation (`{}`) so existing storage plugins (sqlite3, test mocks) are not broken.
- The `BaseWriterInterface` method is pure virtual since all writer implementations (SequentialWriter) must explicitly handle it.
- The MockSequentialWriter in rosbag2_transport tests has been updated with a no-op override.
- The MCAP implementation uses `metadata.metadata.insert(kv.begin(), kv.end())` to convert from `std::map` to the `std::unordered_map` used by `mcap::Metadata`.

FYI @doisyg

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer.hpp
@@ -16,6 +16,7 @@
 #define ROSBAG2_CPP__WRITER_HPP_
 
 #include <cstddef>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -262,6 +263,15 @@ public:
    * \return True if there is any callback registered for the event, false otherwise.
    */
   [[nodiscard]] bool has_callback_for_event(bag_events::BagEvent event) const;
+
+  /**
+   * \brief Write a custom metadata record to the bag.
+   * \param name Identifier for this metadata record.
+   * \param kv Key-value pairs to store.
+   */
+  void write_custom_metadata(
+    const std::string & name,
+    const std::map<std::string, std::string> & kv);
 
 private:
   std::mutex writer_mutex_;
--- a/rosbag2_cpp/include/rosbag2_cpp/writer_interfaces/base_writer_interface.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writer_interfaces/base_writer_interface.hpp
@@ -16,7 +16,9 @@
 #define ROSBAG2_CPP__WRITER_INTERFACES__BASE_WRITER_INTERFACE_HPP_
 
 #include <cstddef>
+#include <map>
 #include <memory>
+#include <string>
 
 #include "rosbag2_cpp/bag_events.hpp"
 #include "rosbag2_cpp/converter_options.hpp"
@@ -80,6 +82,15 @@ public:
    * \return True if there is any callback registered for the event, false otherwise.
    */
   [[nodiscard]] virtual bool has_callback_for_event(bag_events::BagEvent event) const = 0;
+
+  /**
+   * \brief Write a custom metadata record to the bag.
+   * \param name Identifier for this metadata record.
+   * \param kv Key-value pairs to store.
+   */
+  virtual void write_custom_metadata(
+    const std::string & name,
+    const std::map<std::string, std::string> & kv) = 0;
 };
 
 }  // namespace writer_interfaces
--- a/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
@@ -16,6 +16,7 @@
 #define ROSBAG2_CPP__WRITERS__SEQUENTIAL_WRITER_HPP_
 
 #include <deque>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -161,6 +162,15 @@ public:
    * \return True if there is any callback registered for the event, false otherwise.
    */
   [[nodiscard]] bool has_callback_for_event(bag_events::BagEvent event) const override;
+
+  /**
+   * \brief Write a custom metadata record to the bag.
+   * \param name Identifier for this metadata record.
+   * \param kv Key-value pairs to store.
+   */
+  void write_custom_metadata(
+    const std::string & name,
+    const std::map<std::string, std::string> & kv) override;
 
 protected:
   std::string base_folder_;
--- a/rosbag2_cpp/src/rosbag2_cpp/writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writer.cpp
@@ -195,4 +195,12 @@ bool Writer::has_callback_for_event(bag_events::BagEvent event) const
   return writer_impl_->has_callback_for_event(event);
 }
 
+void Writer::write_custom_metadata(
+  const std::string & name,
+  const std::map<std::string, std::string> & kv)
+{
+  std::lock_guard<std::mutex> writer_lock(writer_mutex_);
+  writer_impl_->write_custom_metadata(name, kv);
+}
+
 }  // namespace rosbag2_cpp
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -19,6 +19,7 @@
 #include <ctime>
 #include <filesystem>
 #include <iomanip>
+#include <map>
 #include <memory>
 #include <regex>
 #include <stdexcept>
@@ -889,6 +890,15 @@ void SequentialWriter::add_event_callbacks(const bag_events::WriterEventCallback
 bool SequentialWriter::has_callback_for_event(bag_events::BagEvent event) const
 {
   return callback_manager_.has_callback_for_event(event);
+}
+
+void SequentialWriter::write_custom_metadata(
+  const std::string & name,
+  const std::map<std::string, std::string> & kv)
+{
+  if (storage_) {
+    storage_->write_custom_metadata(name, kv);
+  }
 }
 
 }  // namespace writers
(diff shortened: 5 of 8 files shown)


MichaelOrlov (changes requested):
Hi @tonynajjar Thank you for your effort, but we already have this feature for storing custom key-value metadata in the Rosbag2 from a very long time ago. 
- Originally, it was implemented in the https://github.com/ros2/rosbag2/pull/976

Now it exists on all levels, starting from the `struct BagMetadata`
https://github.com/ros2/rosbag2/blob/9c0f7ba876cddfc3314b3de720b0a02c12310cb8/rosbag2_storage/include/rosbag2_storage/bag_metadata.hpp#L43-L56
and all the way up to the CLI option
https://github.com/ros2/rosbag2/blob/9c0f7ba876cddfc3314b3de720b0a02c12310cb8/ros2bag/ros2bag/verb/record.py#L241-L245

Therefore, I am proposing to close this PR as it will not be merged.

In the future, I would also encourage having discussions or at least creating an issue if something is not working as expected or needs some new functionality before initiating development for a new feature or fix.

tonynajjar:
> In the future, I would also encourage having discussions or at least creating an issue if something is not working as expected or needs some new functionality before initiating development for a new feature or fix.

Yes I usually do it if it's a big contribution but when it's relatively small, then I just use the PR as grounds for discussion - I wasn't expecting an auto-approve :smile: 

Secondly, regarding this feature already existing: I believe currently what exists is the ability to **statically** write custom metadata to `metadata.yaml`. 

What I'd like to enable is writing native MCAP Metadata records (not metadata.yaml) dynamically during recording. These are different from custom_data in two ways: (1) they're embedded in the .mcap binary (self-contained, indexed, visible in Foxglove), and (2) they can be written at any point during recording as events occur, not just at open time.

FYI, in my application I'm using the `rosbag2::Writer` separately  so this PR would be enough for me but we could also think of exposing some service for `ros2 bag record` users to also be able to dynamically inject into mcap's metadata

tonynajjar:
@MichaelOrlov what do you think of my arguments for this PR?

MichaelOrlov:
@tonynajjar As regards
> Secondly, regarding this feature already existing: I believe currently what exists is the ability to statically write custom metadata to metadata.yaml.

> What I'd like to enable is writing native MCAP Metadata records (not metadata.yaml) dynamically during recording. These are different from custom_data in two ways: (1) they're embedded in the .mcap binary (self-contained, indexed, visible in Foxglove), and (2) they can be written at any point during recording as events occur, not just at open time.

Let me clarify a bit.
1. We are writing metadata in the metadata.yaml and mcap file directly. The MCAP file is self-contained. The `metadata.yaml` file exists for usability and overrides metadata from the MCAP file. It was made on purpose to be able to quickly change metadata fields in the text editor for debug purposes.
2. We are writing metadata into the mcap file twice, once at the beginning, right after opening the file, and the second time right before closing the file.
3. The metadata written into the `mcap` file is in the specific Metadata `mcap` section according to the mcap spec and should be visible in tooling like Foxglove Studio.   https://github.com/ros2/rosbag2/blob/64e82abbc0307ef2ba72a728714d4beb0bdf1b3f/rosbag2_storage_mcap/src/mcap_storage.cpp#L1005-L1013
5. There is available public API `MCAPStorage::update_metadata(const rosbag2_storage::BagMetadata & bag_metadata)`. However, we are not exposing it on the upper Writer level. In theory the one may write metadata to the mcap file as many times as needed; the Rosbag2 reader will take the last one saved in the mcap file. We do have a special logic for that during readout.

**The key point is that all metadata shall be consistent and in the specified Rosbag2 format.**

I think the only missing thing from your list is the ability to write metadata at will by the user. 
But that is a sort of abstraction which is inherently hidden in the `rosbag2_cpp::sequential_writer` class. 

Can you please clarify what would be a real world usecase when you would really need to update metadata in real time, rather than just at the beginning during the `SequentialWriter::open(..)` call?

tonynajjar:
Thanks for these clarifications.

> Can you please clarify what would be a real world usecase when you would really need to update metadata in real time, rather than just at the beginning during the SequentialWriter::open(..) call?

My usecase for writing metadata while recording is to record unpredictable events, not known at the beginning, like collisions, start of motion, delocalizations etc... in the metadata. You could argue that we can save them and write them at file closure but this has the drawback of risking the event to not be written in case of process sudden crash (and implementation is more complex)

MichaelOrlov:
In this case, I would propose to add the following public API to the Writer classes
```cpp
void SequentialWriter::update_custom_metadata(
  const std::unordered_map<std::string, std::string> & custom_data)
{
  std::unique_lock<std::mutex> lock(metadata_mutex_);
  metadata_.custom_data = custom_data;
}
  
void SequentialWriter::write_metadata_to_storage()
{
  std::unique_lock<std::mutex> lock(metadata_mutex_);
  storage_->update_metadata(metadata_);
}
```

tonynajjar:
The problems I have with this API:

1-  I would be forced to write the metadata under `rosbag2` and `serialized_metadata` fields whereas I would like to have my own independent ones (one reason being that they show better in visualization tools like Foxglove)

2- It's a disadvantage to keep adding to `metadata_.custom_data` as we will be unnecessarily serializing the same growing data over and over again: each `update_custom_metadata` call serializes the full BagMetadata (topics, files, durations, message counts) just to persist a 100-byte event annotation. The payload-to-overhead ratio gets worse with every event.


FYI this is the state with this PR:
image

MichaelOrlov:
@tonynajjar I need some time (a couple of days) to consider more thoroughly what we can do to make it optimal and in a consistent way.

MichaelOrlov:
Hi @tonynajjar, sorry for the delayed response.

I took some time to think through the proposal more carefully. I still have strong concerns about merging this PR and the proposed `write_custom_metadata(name, kv)` API as-is.

The main reason is that this would introduce a second, independent concept of user-defined metadata in rosbag2, separate from the existing `rosbag2_storage::BagMetadata::custom_data` field. That can easily become confusing and inconsistent: users would have two different places to put arbitrary key-value metadata, with different storage behavior, different visibility, and different read paths.

At the same time, I understand the use case better now. From your clarification, the data you want to write is not really bag-level metadata known at the start of recording. It is closer to a stream of runtime events or state transitions that happen during recording, for example, collisions, start of motion, delocalization events, etc.

For that kind of data, my preferred recommendation would be to record it as messages on a dedicated topic. That gives the events a well-defined schema, timestamps, type information, and natural support in introspection tools such as Foxglove Studio. It also keeps rosbag2 metadata focused on describing the bag itself, rather than mixing bag metadata with runtime application events.

However, I also understand that this may not always be practical, especially when the event is triggered inside a custom recorder that uses the `rosbag2_cpp` writer layer directly and does not naturally go through a ROS topic.

As a compromise, I think we could consider a more constrained API that explicitly models this use case as **event metadata**, rather than as arbitrary custom metadata. For example, instead of allowing users to write any MCAP metadata record with any name, we could define a rosbag2-supported metadata record name, such as:
```cpp
mcap::Metadata metadata;
metadata.name = "rosbag2_event_metadata";
```
or similar.

Then we could define a small set of required/common fields, for example:
```text
event_name
event_type
event_timestamp
schema_version
```
and allow additional user-provided key-value fields after that.

This would make the intent much clearer:

* `BagMetadata::custom_data` remains bag-level custom metadata.
* `rosbag2_event_metadata` metadata represents runtime events that occurred during recording.
* Tooling and readers can distinguish between Rosbag2 metadata and event annotations.
* We avoid opening a fully generic “write arbitrary MCAP metadata records” API at the rosbag2 writer level.

If we go in this direction, I think the feature should also include a corresponding read API. Otherwise, we would provide a way to write these records, but no rosbag2-level way to consume them. At minimum, we should think about APIs to read all event metadata records, and possibly filter by `event_name`, `event_type`, or timestamp.

Another important point is that rosbag2 metadata is currently meant to describe the bag and the recorded files themselves: storage identifier, duration, topics, message counts, compression, custom bag-level data, etc. Runtime application events are a different concept, so I would prefer to keep that separation explicit in the API and storage model.

Regarding the Foxglove visualization issue: I agree that the current rendering of the Rosbag2 metadata record is not very user-friendly. The serialized metadata appears as one long string, for example:
```text
serialized_metadata version: 9 storage_identifier: mcap duration: nanoseconds: 3000000000 starting_time: nanoseconds_since_epoch: 0 message_count: 3...
```

However, the serialized YAML metadata does contain newline delimiters. It looks like Foxglove Studio is either stripping those delimiters or not rendering them properly. That part seems like a tooling/rendering issue, and I think it would be reasonable to ask the Foxglove Studio folks whether they can render the existing `serialized_metadata` field in a more readable way.

So my current position is:

* I am still against merging the current generic `write_custom_metadata(name, kv)` API as-is.
* I agree there is a valid use case for recording runtime event annotations during recording.
* I would prefer to solve that with a constrained, rosbag2-defined event metadata API, or by encouraging users to record such data as typed messages on a topic where possible.
* If we add a write API, we should also design the corresponding read/query API at the same time.

MichaelOrlov:
Closing as a stale PR with no response from the author. 
Please feel free to reopen if you want to follow up on this PR and continue development/discussion.

PR #2438, opened 2026-05-30
feat(rosbag2_transport): add progress reporting to ros2 bag convert

### LINKED ISSUE #2435: ros2 bag convert should print progress
## Description

I would be very nice if 'ros2 bag convert' would print the progress of the conversion (in percentage).
We have some very large bags, and it is difficult to gauge how long time is remaining in the conversion process.

## Related Issues

** Is this feature dependent on any other features? Is it part of a larger project? Note here. **

## Completion Criteria

ros2 bag convert prints a progress bag with the overall progress in percentage.

Count the current processed message number and compare it to the total.

** All features in this project need tests. Please give some input on cases that will need to be tested - and how the testing might be implemented. **

Sourav0607:
## Description
Add progress reporting to `ros2 bag convert` so users can gauge how long
large bag conversions will take.

Previously, the convert command gave no feedback during conversion, making
it impossible to estimate remaining time for large bags.

Changes:
- Sum total message count from input bag metadata before conversion starts
- Print progress at each 1% increment with message count
- Handle edge case where total message count is unknown
- Print final done message with total messages written

Example output:
```
[INFO] [rosbag2_transport]: Converting bag: 0% (0 / 30 messages)
[INFO] [rosbag2_transport]: Converting bag: 50% (15 / 30 messages)
[INFO] [rosbag2_transport]: Converting bag: 100% (30 / 30 messages)
[INFO] [rosbag2_transport]: Converting bag: done (30 messages written)
```

Fixes #2435

Yes. Users will now see progress output when running `ros2 bag convert`.

No.

### Additional Information
Tested manually with a recorded bag on Rolling (Lyrical Luth tarball).

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
@@ -21,11 +21,14 @@
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <cstdio>
+#include <cinttypes>
 
 #include "rosbag2_cpp/reader.hpp"
 #include "rosbag2_cpp/writer.hpp"
 #include "rosbag2_transport/reader_writer_factory.hpp"
 #include "rosbag2_transport/topic_filter.hpp"
+
 
 namespace
 {
@@ -177,8 +180,30 @@ void perform_rewrite(
 
   auto topic_outputs = setup_topic_filtering(input_bags, output_bags);
 
+  // Sum total message count across all input bags for progress reporting
+  // Sum total message count across all input bags for progress reporting
+  uint64_t total_messages = 0;
+  for (const auto & [reader, storage_options] : input_bags) {
+    if (reader) {
+      const auto & metadata = reader->get_metadata();
+      for (const auto & topic_info : metadata.topics_with_message_count) {
+        total_messages += topic_info.message_count;
+      }
+    }
+  }
+  if (total_messages > 0) {
+    fprintf(stdout, "Converting bag:   0%% (0 / %" PRIu64 " messages)", total_messages);
+    fflush(stdout);
+  } else {
+    fprintf(stdout, "Converting bag: starting (total unknown)...");
+    fflush(stdout);
+  }
+
   std::vector<std::shared_ptr<rosbag2_storage::SerializedBagMessage>> next_messages;
   next_messages.resize(input_bags.size(), nullptr);
+
+  uint64_t processed = 0;
+  int last_printed_pct = 0;
 
   std::shared_ptr<rosbag2_storage::SerializedBagMessage> next_msg;
   while ((next_msg = get_next(input_bags, next_messages))) {
@@ -188,7 +213,21 @@ void perform_rewrite(
         writer->write(next_msg);
       }
     }
+    ++processed;
+
+    if (total_messages > 0) {
+      int pct = static_cast<int>((processed * 100) / total_messages);
+      if (pct > last_printed_pct) {
+        fprintf(
+          stdout, "\rConverting bag: %3d%% (%" PRIu64 " / %" PRIu64 " messages)   ",
+          pct, processed, total_messages);
+        fflush(stdout);
+        last_printed_pct = pct;
+      }
+    }
   }
+  fprintf(
+    stdout, "\nConverting bag: done (%" PRIu64 " messages written)\n", processed);
 }
 
 }  // namespace


rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -177,9 +179,30 @@ void perform_rewrite(
...
+      for (const auto & topic_info : metadata.topics_with_message_count) {
+        total_messages += topic_info.message_count;
+      }
+    }
+  }
+  if (total_messages > 0) {
+    RCUTILS_LOG_INFO_NAMED(
+      "rosbag2_transport",
+      "Converting bag: 0%% (0 / %" PRIu64 " messages)", total_messages);
+  } else {
```
cwit-vcas, line 197: I think for other progress commands this is printed inplace. i.e. on the same line each time.
I think you can just insert \r in the beginning of the print statement.

Sourav0607: Thank you for the suggestion! I tested inserting \r directly in 
RCUTILS_LOG_INFO_NAMED but it doesn't work cleanly because the logger 
prepends [INFO] [timestamp] [name]: and appends \n, causing garbled output:

Progress: 0%310942.030369353] [test]:
Progress: 10%10942.330651408] [test]:

Instead I used fprintf(stdout, "\r...") with fflush(stdout) which gives 
a true single-line in-place update — the percentage overwrites itself on 
the same line from 0% to 100%, then \n moves to the next line for the 
done message.

The conversion appears instant for small bags, but for large bags with 
millions of messages the single-line progress updates are clearly visible.

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -21,11 +21,14 @@
...
 #include <utility>
 #include <vector>
+#include <cinttypes>
 
 #include "rosbag2_cpp/reader.hpp"
 #include "rosbag2_cpp/writer.hpp"
 #include "rosbag2_transport/reader_writer_factory.hpp"
 #include "rosbag2_transport/topic_filter.hpp"
+#include <cstdio>
+#include <cinttypes>
```
cwit-vcas: You include this header twice

Sourav0607: Fixed! Removed the duplicate <cinttypes> include and reorganized the 
includes section.

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -188,7 +213,21 @@ void perform_rewrite(
...
+        fprintf(
+          stdout, "\rConverting bag: %3d%% (%" PRIu64 " / %" PRIu64 " messages)   ",
+          pct, processed, total_messages);
+        fflush(stdout);
+        last_printed_pct = pct;
+      }
+    }
   }
+  fprintf(
+    stdout, "\nConverting bag: done (%" PRIu64 " messages written)\n", processed);
```
cwit-vcas, line 230: Should this be processed instead of written?

```suggestion
    stdout, "\nConverting bag: done (%" PRIu64 " messages processes)\n", processed);
```

cwit-vcas (commented):
LGTM, but I do not have any power to merge.

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
Sourav0607: I think "written" is more accurate here — the count reflects exactly how many 
messages were written to the output bag. If topic filters are applied, some 
messages may be read but not written, so "written" correctly describes what 
the user gets in the output. The function itself is also called perform_rewrite, 
so "written" is consistent with the naming convention.

Happy to change it if you feel strongly, but wanted to share the reasoning.

Sourav0607:
@MichaelOrlov  could you please review this PR when you get a chance? 
The implementation adds progress reporting to `ros2 bag convert` as requested 
in issue #2435. All CI checks should be passing now.

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
cwit-vcas: But you increment the "processed" variable no matter if the message is actually written or not?
So the processed variable reflects the total processed messages, not the actual written messages?

rosbag2_transport/src/rosbag2_transport/bag_rewrite.cpp
```diff
@@ -177,9 +180,31 @@ void perform_rewrite(
 
   auto topic_outputs = setup_topic_filtering(input_bags, output_bags);
 
+  // Sum total message count across all input bags for progress reporting
+  // Sum total message count across all input bags for progress reporting
```
cwit-vcas, line 184: Double comment

MichaelOrlov (commented):
@Sourav0607 Thank you for you interest to contributing and making Rosbag2 better.
I didn't have time to run through a thorough review.
However, for a high level design approach I would expect an implementation of the progress bar in a separate class similar to how it was made in the #1836 PR.

PR #2440, opened 2026-06-02
Add is_recording and get_uri services to the recorder

### LINKED ISSUE #2439: Add is_recording and get_uri services to the recorder
## Description

The recorder exposes services for controlling state (`record`, `stop`, `pause`, `resume`) and querying some state (`is_paused`, `is_discovery_running`), but there's no way to query whether the recorder is actively recording or what URI it's recording to.

When integrating the recorder into a larger system via its service interface, the common workaround for checking recording state is to call `~/record` and check if it fails with "Recorder is already recording", which is fragile and creates throwaway recordings that need cleanup.

Similarly, the recording URI is only available to the caller that initiated the `~/record` request. Other nodes have no way to discover what path the recorder is writing to, which is needed for post-processing workflows (e.g. adding metadata to completed mcap files).

Proposed services:
- `~/is_recording` (`IsRecording.srv`): empty request, response: `bool recording`
- `~/get_uri` (`GetUri.srv`): empty request, response: `string uri` (empty if not recording)

## Related Issues

Extends the service interface established in #1634 / #2248.
Complements #2167 (runtime configurable URI in snapshot mode).

## Completion Criteria

- `IsRecording.srv` and `GetUri.srv` definitions added to `rosbag2_interfaces`
- Services registered in `RecorderImpl::create_control_services()`
- `~/is_recording` returns `true` between `record` and `stop` calls, `false` otherwise
- `~/get_uri` returns the current storage URI while recording, empty string otherwise
- Integration tests added
- README updated

Both values already exist internally and just need to be exposed:
- `is_recording`: reads `in_recording_.load()` (same pattern as `is_paused` / `is_discovery_running`)
- `get_uri`: reads `storage_options_.uri`, guarded by `start_stop_transition_mutex_`

Follows the pattern established in PR #2248 for adding recorder services.

Add test cases to `test_record_services.cpp` following the existing pattern:
- `is_recording` returns `false` before `record`, `true` after `record`, `false` after `stop`
- `get_uri` returns empty before `record`, the expected URI after `record`, empty after `stop`
- `get_uri` returns consistent value across `pause`/`resume` cycles

shalev-balas:
## Description

Add two new state query services to the recorder:
- `~/is_recording`: returns whether the recorder is currently in a recording session, reading the existing `in_recording_` atomic boolean.
- `~/get_uri`: returns the current recording URI/path, or empty string if not recording, reading `storage_options_.uri` under the state transition mutex.

Both follow the same pattern as `~/is_paused` and `~/is_discovery_running`.

Fixes #2439

Yes. Two new services are available on the recorder node: `~/is_recording` and `~/get_uri`. No existing behavior is changed.

Yes — Claude (Anthropic, Opus 4.6) was used to assist with the implementation. All code was reviewed and validated manually.

### Additional Information

This is a non-breaking addition — safe to backport. Integration tests added in `test_record_services.cpp`.


shalev-balas:
We'd appreciate this being backported to Jazzy as well. The .srv and CMakeLists.txt changes should apply cleanly. The recorder.cpp changes may need manual adaptation due to divergence between rolling and jazzy, but the pattern is the same — happy to submit a separate Jazzy PR if needed.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
```diff
@@ -1500,6 +1528,17 @@ void RecorderImpl::stop_discovery()
   }
 }
 
+bool RecorderImpl::is_recording() const
+{
+  return in_recording_.load();
+}
+
+std::string RecorderImpl::get_uri()
```
fujitatomoya: idiomatic fix is to mark start_stop_transition_mutex_ as mutable and make get_uri() const on both RecorderImpl and Recorder. this is just a minor comment which should not block this PR, but it keeps the getter API uniform?

rosbag2_interfaces/CMakeLists.txt
```diff
@@ -21,7 +21,9 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/Burst.srv"
   "srv/GetRate.srv"
   "srv/GetSubscribedTopics.srv"
+  "srv/GetUri.srv"
   "srv/IsDiscoveryRunning.srv"
+  "srv/IsRecording.srv"
```
fujitatomoya: this should be after `srv/IsPaused.srv` in alphabet order?

fujitatomoya (approved):
1st of all, the fix itself looks good to me, so it is approved. it just aligns with current other endpoints like `is_paused` and so on. extending those design to get more status would be useful for user application and command line interfaces.

IMO this current design is not well considered. having multiple service endpoint via RPC could generate some problems to get the recorder status. the major problem for me is that snapshot consistency. with separate services, a client that wants "am I recording, and if so, where?" has to call is_recording then get_uri as two RPCs, and the state can transition in between already.
i would introduce a `~/get_status` returning a `RecorderStatus` message (recording, uri, and mirror paused / discovery_running for completeness). that gives a single forward-extensible endpoint, and future fields like bag size, duration, or message count become message-field additions rather than new services. it also stops the proliferation at the new surface rather than adding to it.

i will leave this decision to @MichaelOrlov

rosbag2_interfaces/CMakeLists.txt
shalev-balas: Good catch — moved IsRecording.srv after IsPaused.srv. Done in 3e5cb714.

rosbag2_transport/src/rosbag2_transport/recorder.cpp
shalev-balas: Marked start_stop_transition_mutex_ as mutable and made get_uri() const on both RecorderImpl and Recorder. Done in 3e5cb714.

shalev-balas:
@fujitatomoya @MichaelOrlov
Both comments were applied.
As for having one unified "~/get_status" service, i agree it would be cleaner and safer, however one of my main concerns here was to not break the current interface and allow for easy and fast back-porting to jazzy, which is what my company currently runs on.
But i'd love to hear from you guys about what are your considerations and what you value more in these kind of dilemmas.

MichaelOrlov:
I am second to @fujitatomoya comment
> i would introduce a ~/get_status returning a RecorderStatus message (recording, uri, and mirror paused / discovery_running for completeness). that gives a single forward-extensible endpoint, and future fields like bag size, duration, or message count become message-field additions rather than new services. it also stops the proliferation at the new surface rather than adding to it.

I don't envision any problems backporting it into Jazzy, since it will be a similar kind of change by adding a new API; we don't need to change an existing API.

PR #2443, opened 2026-06-11
Surface recorder spin thread failures in rosbag2_py and CLI

### LINKED ISSUE #2442: ros2 bag record should surface asynchronous storage write failures to Python/CLI
## Description

`ros2 bag record` should reliably report unrecoverable recording failures from the underlying C++ recorder/writer back to the Python caller and CLI.

Currently, some failures that occur inside the C++ recording implementation may not propagate through the Python wrapper in a user-visible way. A concrete example is a disk-full / `ENOSPC` condition while recording. Storage plugins such as sqlite3 or MCAP may throw while writing a message, but message writes can occur from subscription callbacks, cache consumer threads, or other recorder worker paths rather than directly inside the top-level Python `Recorder.record()` call.

As a result, the Python caller may not receive a normal exception, and the CLI may fail unclearly, terminate unexpectedly.

From a user perspective, if recording can no longer continue, `ros2 bag record` should stop, exit non-zero, and print a clear error message containing the underlying cause, such as `No space left on device`.

## Related Issues

I could not find an existing issue for this specific failure propagation problem.

This appears related to:

* the Python wrapper for `rosbag2_transport::Recorder`
* the `rosbag2_transport` recorder execution model
* storage write errors from rosbag2 storage plugins
* cached writing / cache consumer error handling
* metadata writing error handling

## Expected Behavior

* Unrecoverable recording failures from C++ async paths are surfaced to the recorder’s top-level control flow.
* `rosbag2_py.Recorder.record()` raises a normal Python exception when recording fails irrecoverably.
* `ros2 bag record` exits with a non-zero status and prints a clear error message.
* Disk-full / `No space left on device` during message writing is treated as an unrecoverable recording failure.
* Existing expected non-fatal behavior, such as QoS incompatibility warnings, remains unchanged.
* Signal handling and normal user cancellation behavior remain unchanged.

## Possible Implementation Direction

One possible approach would be to capture exceptions from executor/spin paths, subscription callbacks, discovery, cache consumers, event publisher threads, and storage write paths into shared recorder state, for example using `std::exception_ptr`.

The recorder could then notify the outer control flow that recording failed and should stop. The Python wrapper could check for a stored recorder exception after the wait loop exits and rethrow it, allowing pybind11 to translate it into a Python exception.

It may also be useful to distinguish fatal recording errors from recoverable discovery errors. For example, failure to write a message due to a storage error should likely be fatal, while transient topic discovery failures may remain logged and retried if that is the current intended behavior.

`MetadataIo::write_metadata()` may also need review if it writes using `std::ofstream` without explicitly checking stream failure, since disk-full during metadata writing may otherwise not raise an exception.

## Suggested Tests

Tests should avoid actually filling the CI filesystem. Deterministic error injection would be preferable.

Possible tests:

* Add a `rosbag2_cpp::SequentialWriter` unit test where mocked storage `write()` throws `std::runtime_error("No space left on device")`, and verify that the writer propagates the exception.
* Add a `rosbag2_transport::Recorder` test using a mock writer whose `write()` throws during a subscription callback, and verify that the recorder surfaces the failure to the controlling thread.
* Add a Python wrapper or CLI-level test that simulates a recorder failure and verifies that `rosbag2_py.Recorder.record()` raises an exception and `ros2 bag record` exits non-zero with a useful message.
* Add a cached-writing test where the cache consumer write callback throws.
* Add a metadata write failure test if `MetadataIo::write_metadata()` is changed to detect stream errors.

dskkato:
## Description

This PR surfaces asynchronous recorder failures from the `rosbag2_py` recorder spin thread back to the Python/CLI caller.

Previously, storage write failures raised from subscription callbacks, for example disk-full / `"No space left on device"`, occurred on the recorder executor thread and had no propagation path to the main thread used by `ros2 bag record`. This could leave the CLI without a clear failure path.

### Changes:

- Capture exceptions escaping the recorder executor spin thread as `std::exception_ptr`.
- Add `rosbag2_py.Recorder.throw_if_spin_failed()` so non-blocking users can poll and rethrow async recorder failures on their caller thread.
- Wake blocking recorder wait paths when a spin-thread failure is captured.
- Update `ros2 bag record` to poll for async recorder failures, enter cleanup, and report the original recording failure.
- Preserve separate cleanup errors from `stop()` and `stop_spin()` so recording failures are not overwritten by cleanup failures.
- Add tests covering subscription callback write failure propagation and CLI error reporting.

Fixes #2442

Yes.

For `ros2 bag record`, asynchronous recorder failures that previously escaped the recorder spin thread and could terminate the process are now reported as CLI errors.

For non-blocking `rosbag2_py.Recorder` users that call `start_spin()` and `record()` directly, exceptions from the recorder executor spin thread are now captured instead of escaping the thread. Such users should poll `throw_if_recording_failed()` to surface those asynchronous recording failures in their caller thread. If they do not poll it, the spin thread may stop after storing the failure and the exception will not be raised in user code automatically.

Yes. OpenAI Codex was used to help implement the changes and tests.

### Additional Information

The non-blocking Python recorder API cannot automatically throw exceptions into the caller thread after `record()` has returned. Callers using `start_spin()` / `record()` directly should poll `throw_if_spin_failed()` to surface asynchronous recorder failures.

## Evaluation results

The behavior when an asynchronous storage write failure occurs was compared before and after this patch.

For the recording side, the following cli command was used.
```sh
ros2 bag record --storage mcap --max-bag-size 10485760 -o bag --topics /topic
```

To force a storage rollover failure, the bag output directory was made read-only after recording started:

```
chmod a-w bag
```

The following script was used to publish large messages to `/topic`.

<details><summary>publish script</summary>

```py
#!/usr/bin/env python3

import rclpy
from rclpy.node import Node
from std_msgs.msg import String

class BigPub(Node):
    def __init__(self):
        super().__init__('big_pub')
        self.pub = self.create_publisher(String, '/topic', 10)

        self.msg = String()

        # 10MB
        self.msg.data = "a" * (10 * 1024 * 1024)

        self.timer = self.create_timer(1, self.publish)

    def publish(self):
        self.pub.publish(self.msg)
        self.get_logger().info(
            f"published {len(self.msg.data)/1024/1024:.1f} MB")

rclpy.init()
node = BigPub()
try:
    rclpy.spin(node)
except KeyboardInterrupt:
    pass
finally:
    node.destroy_node()
    rclpy.shutdown()
```

</details>

### Before this patch

The exception escaped the recorder executor thread and terminated the process with an unhandled `std::runtime_error`, producing a core dump.

```sh
ros@17c0229bcae8:~/ros2_ws$ ros2 bag record --storage mcap --max-bag-size 10485760 -o bag --topics /topic
[INFO] [1781364121.186279687] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1781364121.186321382] [rosbag2_recorder]: Event publisher thread: Started
[INFO] [1781364121.189263604] [rosbag2_recorder]: Starting recording to 'bag'
[WARN] [1781364121.189277343] [rosbag2_recorder]: No input serialization format specified, using default rmw serialization format: 'cdr'.
[INFO] [1781364121.191424694] [rosbag2_recorder]: Listening for topics...
[INFO] [1781364121.191439635] [rosbag2_recorder]: Recording...
[INFO] [1781364121.191465173] [rosbag2_recorder]: Topics discovery started.
[INFO] [1781364137.542254728] [rosbag2_recorder]: Subscribed to topic '/topic'
[INFO] [1781364137.542286894] [rosbag2_recorder]: All requested topics are subscribed. Stopping discovery...
[INFO] [1781364137.542291414] [rosbag2_recorder]: Topics discovery stopped.
[INFO] [1781364139.144205011] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781364140.128502274] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781364141.132174042] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[ERROR] [1781364141.132652742] [rosbag2_storage]: Could not open 'bag/3_bag_2026_06_13-15_22_21' with 'mcap'. Error: failed to open file "bag/3_bag_2026_06_13-15_22_21.mcap" for writing
[ERROR] [1781364141.132683523] [rosbag2_storage]: Could not load/open plugin with storage id 'mcap'
[INFO] [1781364141.132688046] [rosbag2_storage]: Available read-write storage plugins: 'mcap', 'my_test_plugin', 'sqlite3', 
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to rollover bagfile to new file: "bag/3_bag_2026_06_13-15_22_21"!
Aborted                    (core dumped) ros2 bag record --storage mcap --max-bag-size 10485760 -o bag --topics /topic
```

### With this patch

The recorder failure was propagated back to the CLI, recording was stopped cleanly, and the original error was reported without a core dump.

```sh
ros@17c0229bcae8:~/ros2_ws$ ros2 bag record --storage mcap --max-bag-size 10485760 -o bag --topics /topic
[INFO] [1781363902.733629157] [rosbag2_recorder]: Press SPACE for pausing/resuming
[INFO] [1781363902.733793004] [rosbag2_recorder]: Event publisher thread: Started
[INFO] [1781363902.736761695] [rosbag2_recorder]: Starting recording to 'bag'
[WARN] [1781363902.736826237] [rosbag2_recorder]: No input serialization format specified, using default rmw serialization format: 'cdr'.
[INFO] [1781363902.739161369] [rosbag2_recorder]: Listening for topics...
[INFO] [1781363902.739176403] [rosbag2_recorder]: Recording...
[INFO] [1781363902.739190444] [rosbag2_recorder]: Topics discovery started.
[INFO] [1781363929.228746571] [rosbag2_recorder]: Subscribed to topic '/topic'
[INFO] [1781363929.228794139] [rosbag2_recorder]: All requested topics are subscribed. Stopping discovery...
[INFO] [1781363929.228802038] [rosbag2_recorder]: Topics discovery stopped.
[INFO] [1781363931.140027610] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363932.147199326] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363933.178353138] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363934.164901793] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363935.176594083] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363936.172421978] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363937.154529955] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363938.135169843] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363939.147004561] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363940.151331321] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363941.186313775] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363942.155911538] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363943.162030557] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[ERROR] [1781363943.162921184] [rosbag2_storage]: Could not open 'bag/13_bag_2026_06_13-15_19_03' with 'mcap'. Error: failed to open file "bag/13_bag_2026_06_13-15_19_03.mcap" for writing
[ERROR] [1781363943.162956166] [rosbag2_storage]: Could not load/open plugin with storage id 'mcap'
[INFO] [1781363943.162963337] [rosbag2_storage]: Available read-write storage plugins: 'mcap', 'my_test_plugin', 'sqlite3', 
[INFO] [1781363943.210902277] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
[INFO] [1781363943.211995761] [rosbag2_recorder]: Recording stopped
[ERROR] [ros2bag]: Failed to rollover bagfile to new file: "bag/13_bag_2026_06_13-15_19_03"!
[INFO] [1781363943.222777762] [rosbag2_recorder]: Event publisher thread: Exited
Error. Can't return old SIGINT handler, someone override our signal handler
```

Note that the final SIGINT handler warning is unrelated to this change and was also reproducible independently.


ros2bag/ros2bag/verb/record.py
```diff
@@ -512,20 +512,46 @@ def main(self, *, args):  # noqa: D102
...
             termination_requested.clear()
 
+        # Preserve each failure source separately. If recording itself failed, report that first;
+        # otherwise report cleanup failures in the order they were attempted.
+        if record_error is not None:
+            return print_error(str(record_error))
+        if stop_error is not None:
+            return print_error(str(stop_error))
+        if stop_spin_error is not None:
+            return print_error(str(stop_spin_error))
```
fujitatomoya: if it hits any of them, it returns here anyway. do we really need to cache the all errors in here? IMO we can only have recorder_exception and return it because record_error always prevails stop_error.

ros2bag/test/test_recorder_args_parser.py
```diff
@@ -33,6 +34,42 @@ def test_arguments_parser():
     return parser
 
 
+def test_record_main_reports_async_recorder_failure(test_arguments_parser, monkeypatch, tmp_path):
```
fujitatomoya, line 37: it looks like that test replaces the entire recorder with a fake object whose "did it fail?" method just raises on command, so the real exception handoff logic (the core of this PR) never runs. and the almost same thing goes to C++ test which only proves that a throwing writer makes spin() throw.

rosbag2_py/src/rosbag2_py/_transport.cpp
```diff
@@ -616,14 +617,24 @@ class Recorder
...
       [this]() {
-        exec_->spin();
+        try {
+          exec_->spin();
+        } catch (...) {
+          // Subscription callbacks run on this thread. Store failures so the caller thread can
+          // poll and rethrow them; C++ exceptions do not cross thread boundaries automatically.
+          store_spin_exception(std::current_exception());
+          exit_ = true;
+          wait_for_exit_cv_.notify_all();
```
fujitatomoya, line 635: it sets that flag without holding the lock the waiter uses. there's a racy timing window where the wake-up signal can be missed, and the waiter sleeps forever?

fujitatomoya (changes requested):
i think this PR is not matured enough, can you check my comments below. i would like to have another review on this before CI.

AFAIS, metadata is written with std::ofstream, which can silently fail when the disk is full without throwing. the PR doesn't add a check for that, so a disk-full during the final metadata write still goes unreported?

dskkato:
@fujitatomoya Thanks for the review and for pointing this out.

I agree that the problem described in the original issue can manifest under a variety of conditions, and the metadata write failure due to a full disk is one such case. To make sure we're addressing the right problem, I'd like to spend some time organizing the different reproduction scenarios and clarifying which ones are intended to be covered by this PR versus follow-up work.

I'll take another look at the failure modes and possible approaches, then come back with a proposal. In the meantime, I'll mark the PR as draft.

dskkato:
A quick update after investigating this further.

I reproduced the behavior with a full storage device and compared the SQLite3 and MCAP backends. One correction to my earlier assumption: McapWriter writes through `std::fwrite()`, but write failures are only checked via assertions. Write failures are effectively checked only via an assertion, which means they may go unnoticed in release builds.

```cpp
void FileWriter::handleWrite(const std::byte* data, uint64_t size) {
  assert(file_);
  const size_t written = std::fwrite(data, 1, size, file_);
  (void)written;
  assert(written == size);
  size_ += size;
}
```

https://github.com/foxglove/mcap/blob/8da6e1e4f27107458925e185d8dd305204b552c6/cpp/mcap/include/mcap/writer.inl#L56-L62

On the SQLite3 side, the failure is reported as an exception and ultimately results in `std::terminate()`, so the error is at least surfaced.

```sh
terminate called after throwing an instance of 'rosbag2_storage_plugins::SqliteException'
  what():  Error when processing SQL statement. SQLite error (13): database or disk is full
zsh: IOT instruction (core dumped)  ros2 bag record -a -o rosbag_full_test/mnt/bag_test
```

On the MCAP side, the underlying write failure can remain hidden within the writer implementation. 

What I originally wanted to address was this silent failure mode. However, after tracing the code path, it appears that the silent-failure mode originates in the MCAP library itself. Because rosbag2 sits downstream of the writer, the kind of detection implemented in this PR can only catch failures at limited synchronization points (for example, when splitting or finalizing a file). It cannot reliably detect arbitrary write failures occurring inside the MCAP writer.

So one outcome of this investigation is that the proposed rosbag2-side handling can only provide partial coverage. It may improve detection for some scenarios, but it does not address the general silent-failure case that motivated the original issue. I'm therefore reassessing whether the underlying problem should instead be addressed in the MCAP writer implementation and what the appropriate scope of this PR should be.

Given that this appears to be a broader discussion about failure semantics, error propagation, and the division of responsibility between MCAP and rosbag2, I'm also wondering what the best venue for that discussion would be.

My initial assumption was that this could be addressed with a relatively localized fix in rosbag2, but the investigation so far suggests the problem is more fundamental than that. Before narrowing the scope of this PR further, would it make sense to continue the discussion in a GitHub issue (either here or upstream in MCAP), or is there another channel that would be more appropriate for design-level discussions like this?

In any case, thank you for taking the time to review this PR and for raising these questions. They helped uncover aspects of the problem that I had not fully appreciated when I opened the issue, and I think the discussion has been valuable in clarifying the actual scope of the problem.

MichaelOrlov:
@dskkato Thank you for digging deeper into this problem.
I think the proper way to handle this particular issue with the missed handling of the `std::fwrite` return in the MCAP implementation would be to create a separate a follow up issue with the description in the Rosbag2 repo. Then create an issue/PR in the MCAP C++ repo with fix. Then we can continue with a fix on the Rosbag2 side.  
But all these motions with separate issues/PRs are to unblock the current PR and let it go separately.

rosbag2_py/src/rosbag2_py/_transport.cpp
dskkato: It seems that this is a pre-existing synchronization issue rather than something introduced in this PR. I’ll address it separately before proceeding this PR.

dskkato: It seems there may be places where this kind of lock is necessary, but I fixed this part with lock_guard in 31f24d05d936a4d1771b8e973477c2cc2831fb9a . Thank you.

dskkato:
> AFAIS, metadata is written with std::ofstream, which can silently fail when the disk is full without throwing. the PR doesn't add a check for that, so a disk-full during the final metadata write still goes unreported?

This point is true, but to be honest, I'm not sure the proper way to handle this issue in the current implementation. At least, if some I/O issue happens, I modified to throw an exception for the following two points. 

1. file open error
2. file write error

1057d1b0430269632683e2c450f31b6b1b70cbd1

For other places, zstd_compressor/zstd_decompressor checks file open failures, but do not check whether the contents are actually flushed to the disk (ios_base::badbit may happen). Others are for tests, so I didn't touch them.

ros2bag/ros2bag/verb/record.py
dskkato: Thanks, I agree that `stop_spin_error` is probably unnecessary here and I can simplify that part.

However, I think we still need to keep `record_error` and `stop_error` separately. With the metadata write failure now detected during `recorder.stop()`, the recording path and the finalization path can fail independently. For example, the recorder spin thread may report a write failure first, and then `stop()` may also fail while finalizing metadata.

If `stop()` is allowed to throw directly from the `finally` block, it could mask the original recording error. So the intention here is to preserve the original recording failure and report it first, while still allowing a stop/finalization-only failure to be reported when there was no earlier recording error.

So, `stop_spin_error` is now removed 09e0cf65eb389018dfdf1d6e46e468c4a214ffd8

ros2bag/test/test_recorder_args_parser.py
dskkato: I agree that the current Python test does not exercise the real C++ exception handoff path.
My intention is to split the coverage into two layers:

1. A C++ test that exercises the real recorder/spin-thread exception handoff: a writer failure occurs on the recorder spin path, the exception is captured, and throw_if_spin_failed() rethrows it on the caller thread.
2. A Python/CLI test that treats throw_if_spin_failed() as the rosbag2_py API boundary and verifies that ros2 bag record handles such an async recorder failure correctly, including cleanup and error reporting.

I think a full Python end-to-end test that reliably triggers real disk I/O failure is brittle,
and injecting a C++ mock writer from Python appears difficult with the current test structure.
So I would like to avoid making the CLI test depend on real filesystem failure and instead
cover the cross-thread handoff in C++ and the CLI handling in Python separately.

dskkato:
The abort still occurs:

```sh
[INFO] [1782499108.071593824] [rosbag2_cpp]: Writing remaining messages from cache to the bag. It may take a while
terminate called after throwing an instance of 'rosbag2_storage_plugins::SqliteException'
  what():  Error when processing SQL statement. SQLite error (1): SQL logic error
Aborted (core dumped)
```

This is because the cache writer may use a dedicated thread where the actual `storage_->write()` call occurs, I may need some additional work to achieve a graceful shutdown. Let me spend some more time investigating those code paths.

It may turn out to be more complicated than expected, and I may end up leaving it as is, since the current exception at least provides a helpful error message as shown above.

dskkato:
After working through the implementation and incorporating the review feedback, I've come to a different conclusion about this PR.

My original goal was to improve how `ros2 bag` reports internal errors instead of terminating unexpectedly. However, while implementing this, I realized that the current architecture makes exception propagation across thread boundaries much more complicated than I initially expected. Since the recorder owns its own worker thread, handling exceptions cleanly requires additional mechanisms that significantly increase the complexity of the implementation.

I believe this PR now handles the cases it originally set out to solve. However, I'm no longer convinced that the tradeoff is worthwhile. The implementation has become considerably more complicated, and I'm concerned about its long-term maintainability.

During this investigation, I also found that this is not only a `rosbag2` problem. The current MCAP write API has similar limitations, meaning that a complete solution would require changes beyond `rosbag2` itself rather than incremental fixes in the CLI.

Given all of this, my current opinion is that we probably should not pursue this direction unless we are willing to revisit the threading model itself. With the current architecture, most unexpected internal exceptions already end up terminating the process, so users are generally not left with silent failures. While that behavior is not ideal, it may be a more reasonable tradeoff than introducing substantially more complex exception-handling logic.

I apologize for changing direction after you've already spent time reviewing this PR. I'd really appreciate your thoughts on whether abandoning this approach is the right decision, or whether you think this is still worth pursuing despite the added complexity.

dskkato:
Since I haven't received further feedback, and after spending more time on this, I've concluded that this approach is probably not the right direction.

I'll go ahead and close this PR. Thank you all for taking the time to review it and for the valuable feedback—it helped me better understand the limitations of the current architecture.

The issues related to the MCAP write API appear to be independent of this PR, so I'll investigate those separately and see if they can be addressed in a more focused way.

PR #2444, opened 2026-06-17
[jazzy] Fix per-file message count for zero-cache writes

Hpeox:
## Description

Fix `files[].message_count` being incremented twice when `max_cache_size == 0`, which is the write path used by `ros2 bag convert`.

Per-file counts are now updated after messages are written to storage for both direct and buffered writes. Snapshot behavior is unchanged.

To reproduce, record 10 messages and convert the bag. Before this change, the converted metadata reports 20 messages for the output file while the top-level and topic counts remain 10.

Yes. `files[].message_count` in converted `metadata.yaml` files now matches the actual number of messages.

Yes. ChatGPT 5.5 was used to analyze the bug and draft the PR text. The code changes and tests were performed manually.

### Additional Information

Tested with:

- `colcon test --packages-select rosbag2_cpp`
- Recording and converting a 10-message MCAP bag

### DIFF
--- a/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
@@ -545,10 +545,6 @@ void SequentialWriter::write_messages(
   }
   storage_->write(messages);
 
-  if (!storage_options_.snapshot_mode) {
-    metadata_.files.back().message_count += messages.size();
-  }
-
   if (storage_options_.snapshot_mode) {
     // Update FileInformation about the last file in metadata in case of snapshot mode
     const auto first_msg_timestamp = std::chrono::time_point<std::chrono::high_resolution_clock>(
@@ -557,8 +553,9 @@ void SequentialWriter::write_messages(
       std::chrono::nanoseconds(messages.back()->recv_timestamp));
     metadata_.files.back().starting_time = first_msg_timestamp;
     metadata_.files.back().duration = last_msg_timestamp - first_msg_timestamp;
-    metadata_.files.back().message_count = messages.size();
   }
+
+  metadata_.files.back().message_count += messages.size();
   metadata_.message_count += messages.size();
   std::lock_guard<std::mutex> lock(topics_info_mutex_);
   for (const auto & msg : messages) {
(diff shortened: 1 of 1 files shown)


rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -545,6 +544,11 @@ void SequentialWriter::write_messages(
     return;
   }
   storage_->write(messages);
+
+  if (!storage_options_.snapshot_mode) {
```
MichaelOrlov: IMO We should increment the counter in the snapshot mode too, because this is a callback when we are dumping messages to the storage.
It would be better to place it near line 562 to have it together with the 
```cpp
 metadata_.message_count += messages.size();
```
update.

MichaelOrlov (changes requested):

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
Hpeox: Updated as suggested. Tested with both `rosbag2_cpp` tests and a 10-message record-and-convert test.

MichaelOrlov (approved):
LGTM

MichaelOrlov:
Pulls: ros2/rosbag2#2444
Gist: https://gist.githubusercontent.com/MichaelOrlov/4138f1d2e8d9e7ffe788c5db6df811ff/raw/0e4d96b2290aafddbf63f56bba13b39a77b4a019/ros2.repos
ROS Distro: jazzy
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19640
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29497/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22284/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28447/)

MichaelOrlov:
Windows CI failures are unrelated. Merging then.

Issue #2445, opened 2026-06-24
Read-only support for opening a whole-file zstd-compressed MCAP (`.mcap.zstd`) directly


Tiryoh:
## Description

We would like to open a **standalone, whole-file zstd-compressed MCAP file** (`recording.mcap.zstd`) read-only — directly through rosbag2 or the storage layer — without first having to:

- reconstruct a rosbag2 bag directory with a `metadata.yaml`, or
- decompress the file to a temporary file on disk.

Our recordings leave the robot as standalone `*.mcap.zstd` artifacts, produced and handled by external tooling and storage pipelines. On the consuming side, we often only have that single compressed file and want to use it for playback, inspection, or conversion.

The existing options do not quite cover this case:

- MCAP's built-in per-chunk zstd/lz4 compression remains seekable, but it does not exploit redundancy across chunks. For our structured telemetry-heavy bags, whole-file zstd compression can produce significantly smaller files.
- `rosbag2_compression` file compression mode requires a `metadata.yaml` and decompresses the file to a temporary file on disk before the storage plugin opens it.

The gap we would like to address is:

> Open a standalone `.mcap.zstd` file read-only, with no `metadata.yaml` and no on-disk temporary file.

We understand that this trades away MCAP's random-access and seek benefits and, in our current prototype, requires buffering the decompressed MCAP in memory. We would document this limitation clearly and keep the feature read-only.

Before opening a PR, we would appreciate maintainers' guidance on the preferred shape of this feature:

- (a) a dedicated read-only storage plugin, for example with storage id `mcap_zstd`, auto-detected by the `.mcap.zstd` extension;
- (b) extending `rosbag2_compression` to support metadata-less, streaming or in-memory file decompression for MCAP; or
- (c) another approach that would better fit the project.

We have a working prototype of approach (a) and would be happy to upstream it in whatever form best fits the project.

## Motivation / Measurements

Whole-file zstd compression is useful for our use case because many of our diagnostic bags are structured and telemetry-heavy. In those cases, compressing the entire MCAP file can exploit cross-chunk redundancy that MCAP's built-in per-chunk compression cannot.

Measured on our own recordings using the same uncompressed MCAP source (`mcap compress --compression zstd` vs. whole-file `zstd`):

| data profile | MCAP chunk-zstd | whole-file zstd -3 | whole-file zstd -19 |
|---|---:|---:|---:|
[...]

PR #2448, opened 2026-07-06
Fallback to infile metadata for compressed bag files

MichaelOrlov:
## Description
This PR introduces a fallback behavior for per-file compressed bag files by trying to open metadata from storage if a separate metadata.yaml file is not found. 
Rationale: 
1. To have the ability to open directly just one compressed file.
2. For consistency with the uncompressed files

- Fixes # N/A
- Relates #2445

Yes. Added the ability to open compressed bag files directly without metadata.yaml file.

No.

### Additional Information
Can be backported.

### DIFF
--- a/rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
+++ b/rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp
@@ -95,12 +95,6 @@ void SequentialCompressionReader::open(
   const rosbag2_storage::StorageOptions & storage_options,
   const rosbag2_cpp::ConverterOptions & converter_options)
 {
-  if (!metadata_io_->metadata_file_exists(storage_options.uri)) {
-    std::stringstream errmsg;
-    errmsg << "Could not find metadata for bag: \"" << storage_options.uri <<
-      "\". Bags without metadata (such as from ROS 1) not supported by rosbag2 decompression.";
-    throw std::runtime_error{errmsg.str()};
-  }
   SequentialReader::open(storage_options, converter_options);
 }
 
--- a/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
+++ b/rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
@@ -110,25 +110,20 @@ void SequentialReader::open(
     current_file_iterator_ = file_paths_.begin();
     load_current_file();
   } else {
-    storage_ = storage_factory_->open_read_only(storage_options_);
-    if (!storage_) {
-      throw std::runtime_error("No storage could be initialized for the input URI: " +
-            storage_options.uri);
-    }
-    if (!set_read_order(read_order_)) {
-      ROSBAG2_CPP_LOG_WARN(
-        "Could not set read order on open(), falling back to file order");
-      if (!set_read_order(kFallbackOrder)) {
-        throw std::runtime_error("Could not set read order on open()");
-      }
-    }
+    ROSBAG2_CPP_LOG_WARN("No metadata.yaml file found. Assuming a single storage file and trying "
+                         "to read metadata from storage file.");
+    file_paths_ = {storage_options_.uri};
+    current_file_iterator_ = file_paths_.begin();
+    load_current_file();
     metadata_ = storage_->get_metadata();
     if (metadata_.relative_file_paths.empty()) {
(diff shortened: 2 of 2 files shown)


fujitatomoya (approved):
i think this is better for https://github.com/ros2/rosbag2/issues/2445, lgtm.

@Tiryoh can you also take a look ?

fujitatomoya:
Pulls: ros2/rosbag2#2448
Gist: https://gist.githubusercontent.com/fujitatomoya/ede735c58c77bd8bdcf8ec169937a47c/raw/1034b67b3f83c78d97b2687618fb66a5140ae6d6/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19783
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22431/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28622/)

MichaelOrlov:
Please hold off from merging. I forgot to remove exception in compression reader.

MichaelOrlov:
ob: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19797
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/29733/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22445/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9636/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28643/)

Tiryoh (commented):
Thanks for the fast follow-up on this! Confirmed `4e18cadb` removes the `metadata_file_exists` guard in `SequentialCompressionReader::open()`.

One thing I wasn't 100% sure about, so wanted to check rather than assume: `ReaderWriterFactory::make_reader()` in rosbag2_transport still only picks `SequentialCompressionReader` when `metadata.yaml` exists and reports a non-empty `compression_format`:

```cpp
if (metadata_io.metadata_file_exists(storage_options.uri)) {
  auto metadata = metadata_io.read_metadata(storage_options.uri);
  if (!metadata.compression_format.empty()) {
    reader_impl = std::make_unique<rosbag2_compression::SequentialCompressionReader>();
  }
}
if (!reader_impl) {
  reader_impl = std::make_unique<rosbag2_cpp::readers::SequentialReader>();
}
```

So with a `metadata.yaml`-less per-file-compressed bag, `ros2 bag play`/`ros2 bag rewrite` would still end up on the plain `SequentialReader` (whose `preprocess_current_file()` is a no-op) rather than `SequentialCompressionReader`. I might be missing something about how this is meant to be used, or maybe that's just out of scope for this PR and planned separately — just flagging in case it's useful.

MichaelOrlov:
@Tiryoh That is the way how we determine whether we need to create a simple SequentialReader or a SequentialCompressionReader. i.e., by checking the `!metadata.compression_format.empty()`, otherwise the upper application layer has no idea what to use.

Therefore, unfortunately, it seems there is no way to make Player or bag rewriter to work with compressed bag files without `metadata.yaml` file.

Even with current changes and using the `SequentialCompressionReader` directly, it is going to fail and throw an exception at the moment we would try to create a decompressor in the `preprocess_file()`. Because `metadata_` field hasn't been read and initilized to the moment when we are trying to read a `metadata_.compression_mode` from it.
https://github.com/ros2/rosbag2/blob/802ecdc6ce116fade30a2ffbaade287087d6cf51/rosbag2_compression/src/rosbag2_compression/sequential_compression_reader.cpp#L52-L58

I think we need to close this PR without merging it, since it will not be helpful anyway.

Tiryoh:
@MichaelOrlov Thanks for digging into this and for the clear explanation. It makes sense that the reader type has to be selected before the compression metadata is available, and that `SequentialCompressionReader` cannot obtain that metadata from the storage file until it already knows how to decompress it.

I understand closing this PR, since the current approach cannot provide a usable path for the original use case.

One small thought: would it be worth leaving a short comment near `ReaderWriterFactory::make_reader()` and/or the `metadata_file_exists` guard in `SequentialCompressionReader::open()` explaining this dependency? It might help someone who considers the same approach in the future.

No strong opinion either way, since the reasoning is now captured clearly in this PR thread.

PR #2454, opened 2026-07-14
Add playback progress output to rosbag2 player

umiklos:
## Description

This PR adds ROS1-style playback progress output to `ros2 bag play`.

During playback, the player now prints the current bag time and the elapsed playback duration relative to the total bag duration. The progress output is also refreshed smoothly during playback so that the status line updates even when messages are sparse.

In addition to the player-side output, the playback wait behavior was adjusted so that the progress line can refresh periodically instead of only updating on message publication boundaries.

Yes.

Users running `ros2 bag play` will now see playback progress in the terminal, including the current bag time and elapsed duration versus total duration, similar to the feedback available in ROS1. This makes it easier to understand where playback currently is during debugging, validation, and dataset inspection.

Yes.

GitHub Copilot using GPT-5.4 was used to help draft and implement parts of the code changes, validation steps, and PR text.

### Additional Information

This change is implemented in the common playback path, so it is storage-backend agnostic and is expected to apply to playback using both MCAP and SQLite3/DB3 bags.

Validated locally with:

```bash
source setup.bash
colcon build --packages-select rosbag2_cpp rosbag2_transport rosbag2_py --cmake-args -DBUILD_TESTING=OFF

### DIFF
--- a/rosbag2_cpp/include/rosbag2_cpp/clocks/time_controller_clock.hpp
+++ b/rosbag2_cpp/include/rosbag2_cpp/clocks/time_controller_clock.hpp
@@ -39,8 +39,8 @@ public:
    * \param now_fn: Function used to get the current steady time
    *   defaults to std::chrono::steady_clock::now
    *   Used to control for unit testing, or for specialized needs
-   * \param sleep_time_while_paused: Amount of time to sleep in `sleep_until` when the clock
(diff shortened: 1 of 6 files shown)


MichaelOrlov (commented):
This PR has ABI breaking changes in the `rosbag2_transport/include/rosbag2_transport/player.hpp`. Unfortunately we can't merge it to Humble.

PR #2468, opened 2026-08-05
Make the topic filter match the first message.

clalancette:
## Description

Previously the ReadersManager pre-read one message from each reader in its constructor.  There are 2 problems with that:

1.  This happens before any filter could be set via set_filter(), so the first message from each bag bypassed the topic filter.  The problem was latent because the player always did a seek back to the beginning before playback, but a naive user of the API could run into it.

2.  The pre-read could misalign the next-message cache with the reader indices when a reader had no messages, silently dropping messages from other bags.

Fix both of these issues by filling the cache lazily on first access instead, with one index-aligned slot per reader.

(this fix is a prerequisite to a follow-up patch that I'm going to open for mixed serialization formats)

No.

Yes, Claude Fable 5.

### Additional Information

N/A

### DIFF
--- a/rosbag2_transport/include/rosbag2_transport/readers_manager.hpp
+++ b/rosbag2_transport/include/rosbag2_transport/readers_manager.hpp
@@ -60,8 +60,10 @@ public:
 
   /// \brief Constructor which initializes the ReadersManager with multiple readers and their
   /// associated storage options.
-  /// \note The readers will be opened during construction and cache will be populated with the
-  /// first message from each reader (if available).
+  /// \note The readers will be opened during construction. The internal cache of next messages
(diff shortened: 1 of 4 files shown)


fujitatomoya (approved):
lgtm with green CI.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30065/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22832/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9950/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28984/)

PR #2469, opened 2026-08-05
Cleanup the tests

clalancette:
## Description

There are two changes in here to make the test situation a bit nicer:

1.  Remove printouts from the test plugin.
    
    The main reason to do this is that, during development,
    the test plugin gets installed as a "regular" plugin.
    Thus when you run any "ros2 bag" commands, the plugin
    gets loaded and prints out spurious messages.
    There is really no need for the plugin to print out, so just make it silent.
    
2.  Don't install test plugins into the "production" location.
    
    Instead utilize pluginlib_enable_plugin_testing() to install
    to the build directory, and then inject the path into the
    tests so they get found properly.

No.

Yes, Claude Fable 5

### Additional Information

Strictly speaking, patch 1 is not required if we just take patch 2.  But it still seems like unnecessary printouts to me, so I left it in.

### DIFF
--- a/rosbag2_compression/CMakeLists.txt
+++ b/rosbag2_compression/CMakeLists.txt
@@ -70,33 +70,45 @@ if(BUILD_TESTING)
     test/rosbag2_compression/fake_compressor.cpp
     test/rosbag2_compression/fake_decompressor.cpp)
   target_link_libraries(fake_plugin ament_cmake_ros_core::ament_ros_defaults ${PROJECT_NAME})
-  install(
-    TARGETS fake_plugin
-    ARCHIVE DESTINATION lib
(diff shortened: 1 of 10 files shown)


fujitatomoya (approved):
lgtm with green CI.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30066/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22833/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28985/)

PR #2470, opened 2026-08-11
Enhance action status topic handling and filtering logic

tonynajjar:
## Description

Enables recording a **single action interface topic** (most notably an action's `status` topic) by name, without pulling in the entire action, and ensures such recordings get a proper, self-contained message/service definition in the bag.

Previously, selecting just one action sub-topic (e.g. `.../_action/status`) was not possible via the `--topics` list, and even when recorded the message definition came out as an "unknown" encoding with an empty definition because the owning action type could not be resolved.

**Changes:**

- **Topic filtering** (`rosbag2_transport/topic_filter.cpp`): an exact match in the `topics` list now selects an individual action interface topic on its own. When selected this way it keeps parity with regular topic handling — honoring hidden-topic gating (`--include-hidden-topics` required, since these topics are hidden) and `exclude_topics` precedence.
- **Message definitions** (`rosbag2_cpp/local_message_definition_source.cpp`): adds fallbacks in `get_full_text_ext` when the owning action type is unknown — `action_msgs/msg/GoalStatusArray` emits its plain message definition (`MSG`) and `action_msgs/srv/CancelGoal_Event` emits the plain service definition (`SRV`). Also fixes delimiter handling so the top-level `MSG` definition is emitted first with no delimiter.
- **Tests**: coverage for both the filter behavior and the definition fallbacks.

Yes. Users can now record an individual action interface topic by listing its exact name (e.g. `ros2 bag record /my_action/_action/status --include-hidden-topics`) instead of being forced to record the whole action. Such recordings now also contain a valid `ros2msg` message/service definition rather than an empty/unknown one. Existing behavior for whole-action, regex, and `--all-actions` selection is unchanged, and action interface topics remain hidden by default.

yes Opus 4.8

### Additional Information

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
@@ -279,17 +279,17 @@ bool TopicFilter::topic_selected_by_lists_or_regex(
       if (include_action_interface_names_.find(topic_name) ==
         include_action_interface_names_.end())
       {
-        // An exact match in the topics list selects an individual action
-        // interface topic (e.g. only the status topic of an action).
-        selected_as_topic = topic_in_list(topic_name, record_options_.topics);
-        if (!selected_as_topic) {
-          // Not match include regex
-          if (!record_options_.regex.empty()) {
-            std::regex include_regex(record_options_.regex);
-            if (!std::regex_search(action_name, include_regex)) {
-              return false;
-            }
-          } else {
+        bool selected_by_regex = false;
+        if (!record_options_.regex.empty()) {
+          std::regex include_regex(record_options_.regex);
+          selected_by_regex = std::regex_search(action_name, include_regex);
+        }
+        if (!selected_by_regex) {
+          // An exact match in the topics list selects an individual action
+          // interface topic (e.g. only the status topic of an action). Whole-action
+          // selection takes precedence over this fallback.
+          selected_as_topic = topic_in_list(topic_name, record_options_.topics);
+          if (!selected_as_topic) {
             return false;
           }
         }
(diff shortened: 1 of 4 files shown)


tonynajjar:
To backport to lyrical once merged :pray:

rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
```diff
@@ -262,6 +262,7 @@ bool TopicFilter::topic_selected_by_lists_or_regex(
     if (!record_options_.all_actions &&
       static_topics_and_services_.empty() &&
       record_options_.actions.empty() &&
+      record_options_.topics.empty() &&
```
fujitatomoya, line 265: exclude_regex isn't applied to the topic name? this seems like a parity gap?

fujitatomoya (commented):
i am okay with maintainers approval on this.
this could be one of the corner-cases, but usecase still stands.

one minor comment though.

rosbag2_transport/src/rosbag2_transport/topic_filter.cpp
tonynajjar: Thanks fixed

tonynajjar:
Friendly ping

fujitatomoya:
Pulls: ros2/rosbag2#2470
Gist: https://gist.githubusercontent.com/fujitatomoya/d4d95bae6d5c6e0bf316beeba1d8a680/raw/e3b7b378069aa9ae428cf7e1a236120ed594ed4f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20290
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30376/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23069/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10178/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29258/)

fujitatomoya:
@tonynajjar in the meantime, let's start the CI.

tonynajjar:
I'm fine with waiting e.g. a week if there is another review in the horizon, but if it's unknown I would say we can merge? This is the only change keeping me from using rosbag2 from binaries

fujitatomoya (approved):

fujitatomoya:
@tonynajjar yeah, i do not want you to wait on this for that. let's merge it, we can always roll back if anything comes after.

tonynajjar:
@fujitatomoya thanks a lot, can we backport to lyrical please :pray:

fujitatomoya:
@Mergifyio backport lyrical

PR #2476, opened 2026-08-24
Support playing bags with topics in mixed serialization formats.

clalancette:
## Description

Prior to this commit, "ros2 bag play" would refuse to open a bag where some topics had a different serialization format.  While it is completely reasonable for it to refuse to play *topics* it can't decode, if the user selects only topics that it can decode, it should be able to play the bag.

With this commit, that restriction is lifted.  If all topics selected can be decoded, then the bag is played.  If any topics are selected that cannot be decoded, it prints an error listing the offending topics and suggests excluding them via --topics, --exclude-topics, or --exclude-regex.

Note that this removes two protected virtual methods from SequentialReader: check_topics_serialization_formats(), which used to throw on bags with mixed serialization formats, and check_converter_serialization_format(), which despite its name only set up the converter and is now inlined into open().  This could be considered an API/ABI break, but I didn't find any in-tree users of these APIs so I think it is relatively safe.

Yes, after this change users should be able to open mcap/bag files that have multiple encodings in them.

Yes, Claude Opus 5

### Additional Information


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30269/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22967/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10079/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29145/)

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -137,11 +138,33 @@ void SequentialReader::open(
...
+  {
+    if (storage_serialization_formats.size() > 1u) {
+      throw std::runtime_error(
+              "The bag contains topics with mixed serialization formats, and none of them "
+              "matches the requested output serialization format '" +
+              output_serialization_format + "'. Conversion of a bag with mixed serialization "
+              "formats is not supported.");
+    }
+    converter_ = std::make_unique<Converter>(
+      *storage_serialization_formats.begin(), output_serialization_format, converter_factory_);
```
fujitatomoya: what if topics are empty? this coule be guraded already? i would add a cheap `if (storage_serialization_formats.empty()) return;` before here.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -137,11 +138,33 @@ void SequentialReader::open(
...
+    for (const auto & topic : topics) {
+      converter_->add_topic(topic.topic_metadata.name, topic.topic_metadata.type);
+    }
+  }
+  // Messages already stored in the requested output serialization format (or all messages,
+  // when no output serialization format is requested) are returned as stored, without
+  // conversion. Topics which the consumer cannot decode have to be filtered out by the
+  // consumer.
```
fujitatomoya: sounds like this is significant change for the API contract for user application.

before this PR, requesting output_serialization_format = "cdr" guaranteed that every message from read_next() came back CDR-encoded, either it was stored that way, or a converter was created.
but after this PR, if the requested format is present in the bag but the bag is mixed (e.g. CDR + protobuf topics, output format cdr), it takes neither branch here. that means no converter is created and no error is raised. and eventually open() succeeds, and protobuf messages flow out of read_next() in their stored encoding, right?

i think that is fine but probably we would want to add this behavior on docstring of `Reader::open()`?

fujitatomoya (approved):
minor 2 comments, lgtm with green CI.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
clalancette: It's not in this diff, but up above at line 135 there is a check for `if (topics.empty())`.  Since `storage_serialization_formats` can only be empty if `topics` is empty, that *should* handle this case.  Let me know if you agree.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
clalancette: > before this PR, requesting output_serialization_format = "cdr" guaranteed that every message from read_next() came back CDR-encoded, either it was stored that way, or a converter was created. but after this PR, if the requested format is present in the bag but the bag is mixed (e.g. CDR + protobuf topics, output format cdr), it takes neither branch here. that means no converter is created and no error is raised. and eventually open() succeeds, and protobuf messages flow out of read_next() in their stored encoding, right?

Yes, you are correct about that.  In a mixed bag, no converter is created, no error is raised, and protobuf messages are just returned and possibly mis-interpreted as CDR.  I agree that this isn't a great outcome.

I think the way to deal with this would be to add the serialization format to `rosbag2_storage::SerializedBagMessage`.  That way we could reject messages that have a serialization format we know is bogus before we try to publish/replay it.  It would add 50-100 lines to this PR and be an ABI break, but would make this kind of failure loud.  @fujitatomoya do you think this is worth doing?

rosbag2_cpp/include/rosbag2_cpp/readers/sequential_reader.hpp
```diff
@@ -156,29 +156,6 @@ class ROSBAG2_CPP_PUBLIC SequentialReader
...
-  * factory.
-  * If not, changes the serialization format of the converter factory to use the serialization
-  * format of the storage factory.
-  *
-  * \param converter_serialization_format
-  * \param storage_serialization_format
-  */
-  virtual void check_converter_serialization_format(
-    const std::string & converter_serialization_format,
-    const std::string & storage_serialization_format);
```
clalancette, line 180: @MichaelOrlov Just as an FYI, this is technically an API break.  But I looked into all of the open-source Rolling packages, and none of them override this, so I think this is safe to do.  I'd still like to have you take a look at it and make sure this is OK.

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
fujitatomoya: probably, i do not think of any other better solution here although we could expect performance overhead.
i think that is much safer behavior than a silent failure that could produce garbage data or crashes with no indication of why.

clalancette: Cool, sounds good.  I've implemented this in [739ca99](https://github.com/ros2/rosbag2/pull/2476/commits/739ca994521bb693d5c4aa0a809bf81001ca597f) now; please take another look when you get a chance.

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30401/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23100/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10203/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29282/)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30404/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23102/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10205/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29283/)

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
```diff
@@ -182,14 +209,38 @@ std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::read_ne
...
-      auto message = storage_->read_next();
-      return converter_ ? converter_->convert(message) : message;
+      return finalize_message(storage_->read_next());
     }
     throw std::runtime_error("Bag is at end. No next message.");
   }
   throw std::runtime_error("Bag is not open. Call open() before reading.");
 }
 
+std::shared_ptr<rosbag2_storage::SerializedBagMessage> SequentialReader::finalize_message(
```
fujitatomoya, line 219: this appears to hard-error whenever any selected topic cannot be decoded?

i would suggest taht we could hard-error only when the topic was explicitly requested (--topics)? auto-exclude with a warning when the topic is only implicitly included by the play-everything default. otherwise the common case, that is `ros2 bag play mixed_bag` with default options, that still fails outright, which partially defeats the PR's stated goal?

this is just a suggestion, so i will leave this up to you.

lgtm with green CI.

fujitatomoya (approved):

rosbag2_cpp/src/rosbag2_cpp/readers/sequential_reader.cpp
clalancette: Yeah, this is a good point.  My initial thought was to make this the problem of the user (fail very loudly), but I think it is a nicer user experience to just play what we can.  So In the latest push I've changed it to do exactly that.  Please take another look when you get a chance!

clalancette:
New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30479/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23172/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10277/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29344/)

Issue #2211, opened 2025-10-16
Throttle recording rate per topic


dw25628:
## Description

This feature  was previously proposed in #682, having the ability to throttle the rate of recording on a per topic basis. For example an image being published at a frame rate of 30hz but only wanting to log it at 1hz.

I would be interested in the communities feedback on this feature and my potential proposal for implementation. I would be happy to take an initial stab at implementing it.

## Completion Criteria

- [ ] Ability to enable throttle from cli
- [ ] Check whether or not a msg should be logged within the callback based on the last write time
- [ ] Update documentation

## Implementation Notes / Suggestions

- Implement a record option for enabling the topic throttle and specifying the rate
Potential example call, no rate entry or a rate entry less than 0 would not throttle the message
Rate would be hz
`ros2 bag record --topics_and_rates /tf /gps/fix 5 /image/rgb 0.5 /rosout -1`

- Store a vector or map of rates per topic within RecordOptions
- Store the last write time for each topic
- Add function within TopicFilter to compare the current receive time to the last write time for that topic. Return whether the message can be written
- For each subscription callback check if the message should be written

MichaelOrlov:
Long story short: You can throttle/decimate the topic first, then record the throttled one.

Here are two ways of how to do it in the ROS 2 ecosystem.

### 1. Throttle by rate (≈1 Hz)
 - Pass through ≤1 msg/sec
```bash
ros2 run topic_tools throttle messages /camera/image_raw 1.0 /camera/image_raw_1hz
```
- Then record only the throttled topic
```bash
ros2 bag record /camera/image_raw_1hz
```
Notes:
`topic_tools throttle` works for any message type, including images, and keeps the original message contents/stamps. You can set `--ros-args -p use_wall_clock:=true` if you want throttling against wall time (ignoring /clock). 
[GitHub](https://github.com/ros-tooling/topic_tools). Please see https://github.com/ros-tooling/topic_tools for details.

### 2. Decimate by dropping N of every M messages (exact 1/30)
Let exactly 1 of every 30 images through (≈1 Hz if source ≈30 Hz)
```bash
ros2 run topic_tools drop /camera/image_raw 29 30 /camera/image_raw_decimated
ros2 bag record /camera/image_raw_decimated
```

I wouldn't go the way of adding this feature inside the Rosbag2 recorder. It already has a pretty complicated logic and a lot of various features. Another one good to have feature which is could be accomplished by using a third-party package, will overcomplicate the logic even more.

[...]

Issue #682, opened 2021-03-17
Support record enhancements from rosbag_fancy - UI and Throttling


d-walsh:
## Description

rosbag_fancy, created by @xqms, is a great tool for recording rosbags in ROS1: https://github.com/xqms/rosbag_fancy

It adds two really useful features:
1. A live UI for monitoring the recording with useful metrics
2. The ability to throttle the rate of recording per topic

It would be great if these features could be integrated into rosbag2 record.

Here's the UI:
![](https://camo.githubusercontent.com/ee17e859f6eed713baeabb713e9a0f51d70740a54ee8195a8d248b5c79b7c737/68747470733a2f2f78716d732e6769746875622e696f2f726f736261675f66616e63792f616e696d2e737667)

emersonknapp:
This is a cool feature set - I would love to see a design proposal for it before an implementation, if anybody decides to pick this up.

xqms:
Hey, author of `rosbag_fancy´ here. I wouldn't want to lead development of a rosbag2 port, since I don't use ROS 2 myself (yet),
but I'm available for design discussion / help :)

PR #2478, opened 2026-08-26
Adds native per-topic frequency throttling to `ros2 bag record`

guilyx:
## Description

Adds native per-topic frequency throttling to `ros2 bag record` so high-rate topics (IMUs, cameras, point clouds) can be recorded at a lower Hz without paying writer cache / compression / disk I/O for dropped messages.

## Fixes

- https://github.com/ros2/rosbag2/issues/2211
- https://github.com/ros2/rosbag2/issues/682

```bash
ros2 bag record -a --topic-throttle-frequency /imu=10.0 /camera/image_raw=2.0
```

Plumbing mirrors the existing `--repeat-transient-local` pattern end-to-end:

- `RecordOptions::topic_throttle_frequencies` (+ YAML round-trip via `unordered_map<string, double>`)
- `record.topic_throttle_frequency` composable-recorder node parameter
- `rosbag2_py` binding + stub
- `ros2bag` CLI flag with validation
- Design doc under `docs/design/per_topic_frequency_throttling_during_recording.md`

No, everything happen in a new RecordOptions

### Behavior

- Drop happens in the recorder subscription callback (transport edge), before `Writer::write()`
- Throttle compares the same receive timestamp written into the bag (honors `use_sim_time`); a backwards time jump restarts the window instead of stalling
- First message on a throttled topic is always recorded; unlisted topics are untouched
- Direct `Recorder::write_message()` is intentionally not throttled
- Invalid frequencies (non-finite / ≤ 0) fail fast in CLI, node params, and `record()`

## Did you use Generative AI?

Yes

## Test plan

- [x] `colcon build --parallel-workers 2 --packages-up-to rosbag2_transport rosbag2_py ros2bag` (Rolling Docker `ros:rolling-ros-base-resolute`)
- [x] `colcon test --packages-select rosbag2_storage rosbag2_transport rosbag2_py ros2bag` — **1097 tests, 0 failures**
- [x] Integration: throttled topic records 1 of 5 burst messages; unthrottled records all 5
- [x] `record()` rejects 0 / negative / inf / NaN frequencies
- [x] YAML round-trip of `topic_throttle_frequencies`
- [x] Composable recorder parses `record.topic_throttle_frequency` from params file
- [x] CLI parser accepts valid entries; rejects missing / non-positive / non-finite frequencies
- [x] Manual smoke: `ros2 topic pub -r 100 /chatter std_msgs/msg/String "{data: hi}"` + `ros2 bag record /chatter --topic-throttle-frequency /chatter=5.0` → ~5 Hz in bag info

## Additional Information

Design discussions started in https://github.com/ros2/rosbag2/issues/2211

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,39 @@ $ ros2 bag record -a --repeat-all-transient-local --repeat-transient-local /map=
 listed in `--repeat-transient-local`, the QoS override takes precedence. Ensure the override
 includes `transient_local` durability to receive latched messages.
 
+#### Throttling per-topic recording frequency
+
+High-frequency topics (IMUs, cameras, point clouds) can dominate bag size even when the recorded
+data is only needed at a lower rate. The `--topic-throttle-frequency` option sets a maximum
+recording frequency in Hz for individual topics. Messages arriving faster than the requested
+frequency are dropped inside the recorder at the subscription edge — before serialization
+buffering, compression, and disk I/O — so throttled data costs no write bandwidth.
+
+Format: `--topic-throttle-frequency <topic>=<frequency> [...]`
+
+Examples:
+
+```bash
+# Record everything, but limit the IMU to 10 Hz and the camera to 2 Hz
+$ ros2 bag record -a --topic-throttle-frequency /imu=10.0 /camera/image_raw=2.0
+
+# Record two topics, throttling only one of them
+$ ros2 bag record /odom /scan --topic-throttle-frequency /scan=5.0
+```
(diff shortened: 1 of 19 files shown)


sylvesterkaczmarek (commented):
The design says timestamp-based split requests observe the full incoming stream, but handle_pending_bag_split_request() runs after the throttle early-return. A dropped message therefore cannot trigger a pending split. Should the split check run before throttling, like resume does?

PR #2479, opened 2026-08-28
Add minimum free disk space limit to the recorder (#2477)

### LINKED ISSUE #2477: Feature: stop recording when free disk space is low
## Description

`ros2 bag record` keeps writing until the disk is full. There is no option to stop before that happens.

The existing limits do not cover this. `--max-bag-size`, `--max-bag-files` and `--max-record-duration` are all bounds on the bag itself. They do not protect against anything else on the same disk growing: system logs, core dumps, a second recorder, a software update. The disk can still fill up with those options set.

A full disk on an embedded robot does more than stop the recording. System logs cannot be written, DDS shared memory files cannot be created, and services start to fail. Recovering usually needs someone on site.

ROS 1 `rosbag record` had `--min-space`, with a default of 1 GB. It stopped recording when the free space went below that value. rosbag2 has no equivalent.

We record continuously for many hours on machines with no operator nearby, so we added this check in our own wrapper node around `ros2 bag record`.

## Related Issues

* #1915 mentions a free space bound as one of its open points.

## Completion Criteria

* `ros2 bag record` accepts a minimum free space limit, in bytes and/or as a percentage of the filesystem.
* With no option given, behaviour is unchanged.
* When the limit is reached, recording stops, one error is logged, and an event is published so other nodes can react.
* When circular logging is enabled, there should probably be a way to choose "delete older splits and continue" instead of "stop". I do not have a strong opinion on which should be the default.
* The new options are documented.

I would be happy to implement this, but I wanted to propose the idea first. ROS 1 had this option and rosbag2 did not keep it, so there may be a reason for that which I have missed.

fujitatomoya:
## Description

Closes #2477

Adds a minimum free disk space limit to the recorder, the rosbag2 equivalent of ROS 1's `rosbag record --min-space`. The existing `--max-bag-*` options bound the bag but not the disk, so anything else growing on the same filesystem could still fill it up during long unattended recordings.

New `StorageOptions` fields `min_free_space_bytes`, `min_free_space_percent` and `low_free_space_action` (exposed as `--min-free-space`, `--min-free-space-percent`, `--low-free-space-action` in `ros2 bag record`, as `storage.*` node parameters and in `rosbag2_py`). When the available space of the bag directory's filesystem drops below the limit, the writer either stops writing (`stop`, default) or deletes the oldest splits of the current recording and continues (`delete_oldest_files`, falling back to stop). One error is logged and a new `rosbag2_interfaces/msg/LowDiskSpaceEvent` is published on `events/low_disk_space` so other nodes can react. The recorder stops asynchronously via the task runner, since the writer callback runs with the writer mutex held. With no option set, behaviour is unchanged.

Tests added per layer (storage options YAML, writer stop/delete paths, event publishing, recorder stop, CLI parsing) and the README documents the new options.

Fixes # (issue)

No, new options are just added.

Yes, Claude Fable 5


fujitatomoya (commented):
the implementation got some code quantity, but it is really straight-forward to add the new storage space class during the storage access.

ros2bag/ros2bag/verb/record.py
```diff
@@ -402,6 +427,19 @@ def validate_parsed_arguments(args, uri) -> str:
...
+        return print_error('--min-free-space must be a non-negative integer.')
+
+    if args.min_free_space_percent < 0.0 or args.min_free_space_percent > 100.0:
+        return print_error('--min-free-space-percent must be in the range [0, 100].')
+
+    if args.low_free_space_action == 'delete_oldest_files' and \
+            not (args.max_bag_size or args.max_bag_duration):
+        print(print_warn('--low-free-space-action delete_oldest_files has no old bag files to '
+                         'delete unless bag splitting is enabled via --max-bag-size or '
+                         '--max-bag-duration. Recording will stop when free space is low.'),
```
clalancette: I find this help text a little confusing.  Maybe something like:


```suggestion
        print(print_warn('--low-free-space-action delete_oldest_files can only be used if '
                         'bag splitting is enabled via --max-bag-size or '
                         '--max-bag-duration.'),
```

Also, rather than continue here, I think we should make this a hard error and exit.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -677,31 +708,127 @@ void SequentialWriter::delete_oldest_files_if_needed()
...
-        break;  // Keep file in tracking; retry on next split to avoid tight loop
+  const auto & oldest_file = metadata_.files.front();
+  const auto file_path = fs::path(base_folder_) / oldest_file.path;
+
+  // Delete file from filesystem
+  if (fs::exists(file_path)) {
+    const uint64_t file_size = fs::file_size(file_path);
+    const int64_t file_duration_ns = oldest_file.duration.count();
+    std::error_code ec;
+    bool file_removed = fs::remove(file_path, ec);
```
clalancette, line 719: This concern isn't completely new, but I'm worried about both TOCTTOU races between `fs::exists` and `fs::remove`, as well as symlink attacks with this code.

We can probably remove the TOCTTOU race by trying to just remove, and then gracefully handling the error.  For the symlink attack, we might want to consider guarding with `fs::is_regular_file` before trying to delete it.  There is still a TOCTTOU race, but it is harmless; in the worst case we delete the symlink itself rather than the file we were trying to remove.

rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -677,31 +708,127 @@ void SequentialWriter::delete_oldest_files_if_needed()
...
+  info->min_free_space_bytes = min_free_space_bytes;
+
+  if (storage_options_.low_free_space_action ==
+    rosbag2_storage::LowFreeSpaceAction::DELETE_OLDEST_FILES)
+  {
+    // Delete the oldest files, but never the one currently being written to, until there is
+    // enough free space.
+    while (space_info.available < min_free_space_bytes && metadata_.files.size() > 1) {
+      std::string deleted_file_path;
+      if (!delete_oldest_file(deleted_file_path)) {
```
clalancette, line 804: I *think* this is thread-unsafe with the cache thread.  `switch_to_next_storage`, which is the only other caller of `delete_oldest_file`, makes sure to stop the cache thread before doing deletions.  That isn't the case here, so I think the cache thread could still be writing to a file while we delete it.  Further, I think we could have unsynchronized access to `metadata_.files` and `per_file_topic_message_counts_` between the two threads.  I think we'll have to synchronize things somehow.

clalancette: I also think this has another failure mode when interacting with file mode compression.  This probably isn't a new failure mode, but a latent one.  There are 2 lists: `metadata_.files` and `metadata._relative_file_paths_`.  When creating a new file, they both get the uncompressed name.  But when a file eventually gets compressed, it rewrites itself in `relative_file_paths` only and deletes the original file.  Because `delete_oldest_path` only looks at `files` (which contains the *uncompressed* filename), it will never delete anything, and this loop will just spin forever.

The fix is probably to have `compress_file` rewrite `path` at the same time it rewrites `relative_file_paths`.

clalancette (changes requested):

ros2bag/ros2bag/verb/record.py
fujitatomoya: yeah agree.

retuning error is more predictable behavior here. i will take your suggestion.

sylvesterkaczmarek (commented):
Should a failure of fs::space() fail closed? This feature is specifically protecting the robot from a full filesystem, but the current error path logs a warning and keeps writing when free space cannot be determined.

Issue #2383, opened 2026-03-17
Jazzy: rosbag2_transport Player Node does not terminate when `playback_duration` or `playback_until_timestamp` is specified.


balthasarschuess:
## Description
- when running `ros2 bag play --playback-duration 5s /path/to/bag` the command terminates after playback is complete
- when using `rosbag2_transport player node`  with the playback duration parameter, message publishing terminates at the right time, but the /clock topic continues being published and the node does not shut down

### Expected Behavior
- I expect the node to publish the /clock topic and messages until the specified time
- I expect the node to terminate, once the playback duration is exceeded

### Actual Behavior
- messages are published until the specified time
- /clock is being published beyond the specified time until infinity (if termination is not intended, I'd expect at least this.)
- node does not terminate

## To Reproduce
```python3
from amr_bringup.logging_utils import add_rosout_support
from launch import LaunchDescription
from launch.actions import (
    DeclareLaunchArgument,
    RegisterEventHandler,
    Shutdown,
    LogInfo,
)
from launch.event_handlers import OnProcessExit
from launch.substitutions import LaunchConfiguration
from launch_ros.actions import Node, SetParameter

from lv.appconfig import resolve_launch_param_files


def generate_launch_description():  # noqa
    bag_path = LaunchConfiguration("bag_path")

    declare_bag_path = DeclareLaunchArgument(
        "bag_path",
        description="Path to the rosbag to replay",
    )

    player = Node(
        package="rosbag2_transport",
        executable="player",
        name="rosbag_replay",
        output="screen",
        parameters=[
             "path/to/rosbag_play.yaml",
            {"storage.uri": bag_path},
        ],
        respawn=False,
    )

    ld = LaunchDescription(
        [
            SetParameter(
                name="use_sim_time",
                value="true",
            ),
            declare_bag_path,
            player,
            RegisterEventHandler(
                OnProcessExit(
                    target_action=player,
                    on_exit=[
                      LogInfo(msg="Node exited"),
                      Shutdown(reason="Rosbag replay finished"),
                   ],
                )
            ),
        ]
    )
    add_rosout_support(ld)

    return ld
```
```yaml

/**:

  ros__parameters:
    play:
      rate: 1.0
      loop: false
      start_paused: false
      disable_keyboard_controls: true
      playback_duration:
        sec: 5
        nsec: 0
```

## System (please complete the following information)
 - OS: Ubuntu 22.04
 - ROS 2 Distro: jazzy
[...]

PR #2482, opened 2026-09-01
Stop fixed-frequency /clock publication after playback

ktyang512:
## Summary

A Player configured for fixed-frequency clock publication leaves its wall timer active after playback finishes. When the Player node remains alive, this causes `/clock` to continue advancing after message playback has stopped, including when playback reaches a configured duration or timestamp boundary.

This change cancels the fixed-frequency clock timer when the playback session finishes. The existing reset at the beginning of `play()` restarts the canceled timer for subsequent playback sessions.

The Player node and its executor remain alive. This is intentional because Player can be loaded as a component in a shared process, where shutting down the ROS context would also affect unrelated nodes.

Addresses the `/clock` behavior reported in #2383.

## Testing

- Added a regression test that keeps the executor running after playback finishes, verifies that fixed-frequency `/clock` publication stops, and verifies that a subsequent playback restarts clock publication.
- Built `rosbag2_transport` and its dependencies with `BUILD_TESTING=ON` and `-Werror` in a ROS Rolling container.
- Ran the `test_play_publish_clock` target together with `cpplint` and `uncrustify`: 3/3 passed.
- Ran the complete `rosbag2_transport` package test suite: 39/39 CTest targets passed; 612 tests, 0 errors, 0 failures, 79 skipped.
- Repeated the complete `test_play_publish_clock` target 10 times with no failures.

### DIFF
--- a/rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
+++ b/rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
@@ -23,6 +23,7 @@
 #include <vector>
 
 #include <rclcpp/executors.hpp>
+#include "rcpputils/scope_exit.hpp"
 #include "rosbag2_transport/player.hpp"
 #include "rosgraph_msgs/msg/clock.hpp"
 #include "test_msgs/message_fixtures.hpp"
@@ -178,6 +179,12 @@ TEST_F(ClockPublishFixture, clock_stops_after_playback_and_restarts_on_next_play
   exec.add_node(player);
   exec.add_node(clock_subscriber);
   auto spin_thread = std::thread([&exec]() {exec.spin();});
+  auto cleanup_spin_thread = rcpputils::make_scope_exit([&]() {
+        exec.cancel();
+        if (spin_thread.joinable()) {
+          spin_thread.join();
+        }
+    });
 
   const auto discovery_deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
   while ((*clock_publisher)->get_subscription_count() == 0 &&
@@ -186,10 +193,6 @@ TEST_F(ClockPublishFixture, clock_stops_after_playback_and_restarts_on_next_play
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
   }
   const bool clock_subscription_matched = (*clock_publisher)->get_subscription_count() > 0;
-  if (!clock_subscription_matched) {
-    exec.cancel();
-    spin_thread.join();
-  }
   ASSERT_TRUE(clock_subscription_matched);
 
   ASSERT_TRUE(player->play());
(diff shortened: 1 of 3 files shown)


rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
```diff
@@ -131,6 +135,91 @@ TEST_F(ClockPublishFixture, clock_respects_playback_rate)
...
+  }
+  const bool clock_subscription_matched = (*clock_publisher)->get_subscription_count() > 0;
+  if (!clock_subscription_matched) {
+    exec.cancel();
+    spin_thread.join();
+  }
+  ASSERT_TRUE(clock_subscription_matched);
+
+  ASSERT_TRUE(player->play());
+  ASSERT_TRUE(player->wait_for_playback_to_finish(std::chrono::seconds(30)));
```
fujitatomoya, line 199: not a blocker, but should be fixed to harden the test quality.

if asserted here, there is spin_thread still joinable.
that means, a joinable std::thread destructor calls std::terminate() if that goes 1st.
and taking the whole test binary down and masking the real failure.

fujitatomoya (approved):
lgtm with green CI.

one minor comment for the documentation: this fixes the current behavior that some user application would rely on it already. could we update or add the doc in the README? (this is just a suggestion, not a blocker.)

rosbag2_transport/test/rosbag2_transport/test_play_publish_clock.cpp
ktyang512: Thanks for catching this. I added an `rcpputils::make_scope_exit` guard in 519e4a1a so every exit path cancels the executor and joins the spin thread before destruction; a local forced fatal assertion now reports as a normal test failure instead of terminating the test binary. I also documented the fixed-frequency `/clock` lifecycle in the README as suggested.

fujitatomoya:
Pulls: ros2/rosbag2#2482
Gist: https://gist.githubusercontent.com/fujitatomoya/409138343a013a6f62261826e99d3b19/raw/3913f955688896b44be3db1e88397acfcce93d2c/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20326
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30420/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23116/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10220/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29297/)

fujitatomoya:
@Mergifyio backport lyrical

PR #2484, opened 2026-09-01
fix SQLite::SQLite3  deprecation

mosfet80:
## Description

The target name SQLite::SQLite3 is deprecated

### DIFF
--- a/rosbag2_storage_sqlite3/CMakeLists.txt
+++ b/rosbag2_storage_sqlite3/CMakeLists.txt
@@ -31,7 +31,7 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
   rosbag2_storage::rosbag2_storage
   rcpputils::rcpputils
   rcutils::rcutils
-  SQLite::SQLite3
+  SQLite3::SQLite3
 )
 target_link_libraries(${PROJECT_NAME} PRIVATE
   ament_cmake_ros_core::ament_ros_defaults


ahcorde (changes requested):
Check this error:

```
  CMake Error at CMakeLists.txt:30 (target_link_libraries):
    Target "rosbag2_storage_sqlite3" links to:
  
      SQLite3::SQLite3
  
    but the target was not found.  Possible reasons include:
  
      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.
  
  
  
  -- Generating done (0.1s)
  CMake Generate step failed.  Build files cannot be regenerated correctly.
  ---
  --- stderr: rosbag2_storage_sqlite3
  CMake Error at CMakeLists.txt:30 (target_link_libraries):
    Target "rosbag2_storage_sqlite3" links to:
  
      SQLite3::SQLite3
  
    but the target was not found.  Possible reasons include:
  
      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.
  
```

fujitatomoya (changes requested):
see https://cmake.org/cmake/help/v4.3/module/FindSQLite3.html

this will not work with current rolling branch.

either closing this PR as it is, or following alias can be added as suggested.

```cmake
if(NOT TARGET SQLite3::SQLite3) # CMake < 4.3
  add_library(SQLite3::SQLite3 ALIAS SQLite::SQLite3)
endif()
```

PR #2485, opened 2026-09-01
Keep recorder discovery alive for mixed dynamic selectors

### LINKED ISSUE #2417: Recording with both --regex and --topics stops discovery early
## Description

When recording with both `--topics` and `--regex` topics the discovery seem to stop when the number of subscribed topics matched the number of topics passed to `--topics`.

### Expected Behavior

When using `--regex` discovery should never stop. A new topic that match the regex could appear at any point.

### Actual Behavior

The discovery seem to stop when the number of subscribed topics matched the number of topics passed to `--topics`.

## To Reproduce

1. Run `ros2 bag record --regex '/namespace/.*' --topics /test /other_test /another_test`
2. Publish some topics
  - `ros2 topic pub /namespace/qwerty nav_msgs/Odometry`
  - `ros2 topic pub /namespace/qwertz nav_msgs/Odometry`
  - `ros2 topic pub /namespace/azerty nav_msgs/Odometry`
3. Discovery stops at the third topic. A fourth topic is not recorded even if it was explicitly listed with `--topics`

 - OS: Ubuntu 24.04
 - ROS 2 Distro: Jazzy/Rolling
 - Install Method: apt, docker

Initially observed on Jazzy and verified to still be present on rolling using the docker image `osrf/ros:rolling-desktop`

yurekami:
## Summary

- stop discovery early only for finite, explicit topic selection
- keep discovery active when regex, type, service, or action selectors can still expand
- add late topic, service, and action regressions for mixed explicit-topic recording

## Why

Combining `--topics` with `--regex` could stop graph discovery as soon as the explicit topics subscribed, so later regex matches were never recorded. The same boundary applies to service/action selectors. This fixes #2417.

## Verification

- three bounded late-discovery regression paths wired into the existing isolated gmock target
- selector set cross-checked against `get_requested_or_available_topics()`
- `git diff --check` and `git show --check`

A local ROS/colcon test run was not available and remains CI-bound.

Fixes #2417

### DIFF
--- a/rosbag2_transport/src/rosbag2_transport/recorder.cpp
+++ b/rosbag2_transport/src/rosbag2_transport/recorder.cpp
@@ -1531,7 +1531,16 @@ void RecorderImpl::topics_discovery() noexcept
         std::lock_guard<std::mutex> subscriptions_lock(subscriptions_mutex_);
         subscriptions_count = subscriptions_.size();
       }
-      if (!record_options_.topics.empty() &&
+      const bool has_additional_or_dynamic_selection =
+        record_options_.all_topics ||
+        record_options_.all_services ||
+        record_options_.all_actions ||
+        !record_options_.topic_types.empty() ||
+        !record_options_.services.empty() ||
+        !record_options_.actions.empty() ||
+        !record_options_.regex.empty();
+      if (!has_additional_or_dynamic_selection &&
+        !record_options_.topics.empty() &&
         subscriptions_count == record_options_.topics.size())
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
i will close the all PRs from @yurekami because this is categorized into AI spam based on our policy.

- the internal reasoning/audit trailer format of an autonomous coding agent that leaked straight into the commit.
- never ran the real tests.
- did not use our original template.

PR #2490, opened 2026-09-06
avoid O(N²) metadata writes (the CPU/RSS/disk growth).

### LINKED ISSUE #2481: High bag Splitting Causes High CPU Usage
## Description

When bags split often it seems to cause the CPU rate to grow. With a bag split every one second It did 14% growth in one hour. Now obviously in practice I'm not really using 1 bag / second, but my use case involves a process running a long time and over time it is growing.

### Expected Behavior
Ideally the CPU% remained bounded

### Actual Behavior
Currently it grows unbounded

## To Reproduce
1. Run ros2 bag2 recordd with 1 second turn over
2. I add a junk noise publisher also running, but probably not actually needed
3. Watch CPU% rise

 - OS: Ubuntu Noble
 - ROS 2 Distro: Jazzy
 - Install Method: Local build
 - Version Head

(Also notice on apt install)

## Additional context

Image

Raw data
[samples.csv](https://github.com/user-attachments/files/31608088/samples.csv)

Taking a quick look at the code I would guess it has something to do with writing the file list to the manifest.yaml.

fujitatomoya:
## Description

this writes a copy of the metadata with relative_file_paths and files trimmed to the currently opened file only, so the YAML the storage plugins serialize into each bag file stays O(topics) instead of O(all files ever recorded).
the full file history is still written to metadata.yaml on close, that's what readers actually use.
this should kill the per-split CPU growth and the redundant on-disk metadata your analysis flagged.

Fixes https://github.com/ros2/rosbag2/issues/2481

No

Yes, Claude Fable 5. the affected pacakges are tested in local environment for sure.


rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp
```diff
@@ -727,13 +725,21 @@ void SequentialWriter::finalize_metadata()
...
-    const auto bag_path = fs::path{path};
+    // Note: The relative_file_paths contains file names relative to the bag folder
+    const auto bag_path = fs::path(base_folder_) / path;
 
-    if (fs::exists(bag_path)) {
-      metadata_.bag_size += fs::file_size(bag_path);
+    std::error_code ec;
+    const auto file_size = fs::file_size(bag_path, ec);
+    if (!ec) {
+      metadata_.bag_size += file_size;
```
fujitatomoya, line 734: this is a latent bug, not directly related to the issue.

rosbag2_cpp/include/rosbag2_cpp/writers/sequential_writer.hpp
```diff
@@ -265,6 +265,22 @@ class ROSBAG2_CPP_PUBLIC SequentialWriter
...
+  /// \brief Write the metadata about the currently opened bagfile into the storage file.
+  /// \details Storage plugins serialize the given metadata into the bag file on every call.
+  /// Therefore, this method writes a copy of metadata_ with the relative_file_paths and files
+  /// lists trimmed down to the currently opened file. These lists grow by one entry with every
+  /// bagfile split, and re-serializing the whole file history into the bag file on every split
+  /// caused unbounded CPU growth and redundant on-disk metadata for long recordings with many
+  /// splits. See https://github.com/ros2/rosbag2/issues/2481 for details.
+  /// The full file history is still written to the metadata.yaml file when the writer is closed.
+  /// \note Storage plugins ignore or overwrite the embedded file lists on read and use the
+  /// actually opened file instead.
```
clalancette, line 281: This side-note actually points to an interesting behavior in rosbag2.

By default, when using a rosbag2 reader with a metadata.yaml, we take the information from the metadata.yaml as canonical and use that.  However, if we point a rosbag2 reader at a raw `.mcap`/`.db3` file, it has a fallback path where it takes the data from within the file itself.  On `main`, this lists *every* file in the series up to this point, and is presumably why the N^2 behavior was implemented in the first place.

Now, this is inconsistently implemented.  The MCAP storage backend does not yield this at all presently, so `ros2 bag play` only plays the exact file you pointed out.  The SQLite3 storage backend yields a list of files, but it currently yields the wrong list of files (the one you pointed at, then 1, 2, 3...).

The question we need to answer is whether we want playing of sequential bags to work in the absence of a metadata.yaml.  If we *do* want it to work, then we need to continue to store all of this information, and we need to fix MCAP and SQLite3 storage backends so they actually work.  If we don't want it, we can take this PR and remove that path completely.  I'm not sure how to answer that question, though.  @MichaelOrlov @fujitatomoya any thoughts here?

clalancette (commented):
I've left one important comment inline about the change in behavior here.

PR #2494, opened 2026-09-09
Build rosbag2_performance_benchmarking with /bigobj on MSVC

mjcarroll:
`benchmark_publishers.cpp` instantiates `rclcpp::Publisher<T>`, and the whole type support stack behind it, for every message type handled by `msg_utils::create()` — 25 of them — in a single translation unit. That object needs more COMDAT sections than the default object file format allows once debug info is enabled.

Measured section counts for `benchmark_publishers.cpp.obj` (MSVC 19.51, x64), against the 65,535 limit:

| build type | sections | |
| --- | --- | --- |
| `Release` | 51,494 | 79% of the limit |
| `RelWithDebInfo` | 70,162 | over |
| `Debug` | 91,482 | over |

So the TU has been sitting just under the ceiling in `Release`, and goes over as soon as debug info is turned on.

`packaging_windows` builds `RelWithDebInfo`, and has failed on every run since [build 4098](https://ci.ros2.org/job/packaging_windows/4098/) (2026-08-31) with:

```
C:\ci\ws\src\ros2\rosbag2\rosbag2_performance\rosbag2_performance_benchmarking\src\benchmark_publishers.cpp(1,1): error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

`nightly_win_rel` builds `Release`, which is why the nightly never caught it.

`/bigobj` raises the ceiling to 2^31 sections, is the documented remedy for exactly this error, and costs nothing measurable.

Assisted by: Claude Code


mjcarroll:
Windows RelWithDebInfo: [Build Status](https://ci.ros2.org/job/ci_windows/29358/)

mjcarroll:
@miguelgonrod this should fix the nightly Windows debug build issues you were seeing.

PR #2311, opened 2026-01-11
Workaround for flaky "record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size" test

MichaelOrlov:
## Description

This PR addresses flakiness in the `RecordFixture::record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size` test by the following changes:
1. Use QoS settings for the publisher manager explicitly as `rclcpp::QoS qos = rclcpp::SystemDefaultsQoS().keep_last(message_count). reliable();`
2. Temporarily add `--max-cache-size 0` to the CLI "command" string. To run a test with the disabled writer's cache.

- Relates #https://github.com/ros2/rosbag2/issues/2310

### Is this user-facing behavior change?
No.

### Did you use Generative AI?
No

### Additional Information
Can be backported.

PR #2496, opened 2026-09-13
Account for cached messages size in bagfile split-by-size decision

### LINKED ISSUE #2310: Bag file size may be significantly larger than `--max-bag-size` when recording with cache
## Description
Bag file size may be significantly larger than `--max-bag-size` when recording with cache

### Expected Behavior
The bag file size shall not be "significantly" larger (more than one message size) than the value specified in the  `--max-bag-size` when recording with the enabled cache, i.e., with `--max-cache-size` > 0

### Actual Behavior
Sometimes, the bag file size can be significantly larger than the set limit in the `--max-bag-size` CLI option.

## To Reproduce
Run existing `RecordFixture::record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size` test locally:
https://github.com/ros2/rosbag2/blob/4da63dbf04219d67f410cbdf466d9e82b52e7130/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp#L332-L379

 - OS: Ubuntu 22.04
 - ROS 2 Distro: Rolling, Kilted, Jazzy
 - Install Method: any
 - Version: 4da63dbf04219d67f410cbdf466d9e82b52e7130

## Additional context
For some unknown reason, the test `record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size` is not failing on CI.

## RCA
1. The root cause is that we are not taking into account messages currently residing in the cache when doing an estimate for the file size when checking if we need to make a split 
https://github.com/ros2/rosbag2/blob/4da63dbf04219d67f410cbdf466d9e82b52e7130/rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp#L468-L479
2. The test also fails because we are using a default `rclcpp::QoS{rclcpp::KeepAll()` for the publishers, and not all messages may be delivered. To make the test more robust need to specify QoS settings for the publisher manager explicitly as `rclcpp::QoS qos = rclcpp::SystemDefaultsQoS().keep_last(message_count).reliable();`

KR-Ravindra:
## Description

Fixes #2310 (RCA item 1). Related: #2311 (the `--max-cache-size 0` test workaround).

### Problem
With the writer cache enabled (`--max-cache-size > 0`, the default for `ros2 bag record`), bag files can be significantly larger than `--max-bag-size`. The e2e test `record_end_to_end_with_splitting_bagsize_split_is_at_least_specified_size` currently works around it with `--max-cache-size 0` (#2311).

### Root cause
`rosbag2_cpp/src/rosbag2_cpp/writers/sequential_writer.cpp:637` (`SequentialWriter::should_split_bagfile`) compares only `storage_->get_bagfile_size()` with `max_bagfile_size`. Messages still residing in the cache are flushed to the current bagfile on split (`switch_to_next_storage()` stops the cache consumer first), so a file can overshoot the limit by up to the cache size. This is the RCA from #2310 and the `TODO(morlov)` at that line.

### Fix
- Add `MessageCacheInterface::get_current_size()` (virtual, default `0`), as named in the TODO.
- Implement it in `MessageCache` as the byte size of the producer buffer, read under `producer_buffer_mutex_` (new `MessageCacheBuffer::get_bytes_size()`). Messages already handed over to the consumer are not counted: they are being written and will show up in the storage size, which avoids double counting and keeps the "split is at least the specified size" guarantee.
- `should_split_bagfile()` now checks `storage size + cached size >= max_bagfile_size` when the cache is enabled. Snapshot mode never splits by size, so `CircularMessageCache` keeps the default.

The `--max-cache-size 0` workaround in the e2e test is left untouched, since I could not run `rosbag2_tests` here; it can be dropped in a follow-up once CI confirms.

### How tested
New `test_sequential_writer` case `writer_with_cache_splits_when_storage_and_cache_size_gt_max_bagfile_size` (storage 60 + cached 40 vs limit 100). Before the fix:
```
test_sequential_writer.cpp:721: Failure
Expected equality of these values:
  num_splits
    Which is: 0
  1u
Writer should split bagfile when storage size plus cached messages size reaches the max_bagfile_size
test_sequential_writer.cpp:726: Failure
  fake_metadata_.relative_file_paths.size()
    Which is: 1
  2u
[  FAILED  ] SequentialWriterTest.writer_with_cache_splits_when_storage_and_cache_size_gt_max_bagfile_size
```
After the fix: `test_sequential_writer` 36/36, `test_message_cache` 8/8 (new `get_current_size_returns_size_of_messages_in_producer_buffer`), full `rosbag2_cpp` ctest 16/16, and the ament linters (cpplint, uncrustify, cppcheck, copyright, lint_cmake, xmllint) pass on Rolling. The two cache/split tests were repeated 30x without failures.

The existing `writer_with_cache_splits_when_storage_bagfile_size_gt_max_bagfile_size` test now writes 1-byte messages: its fake storage size counts messages while the cached size is counted in bytes, so both are kept in the same units.

Yes. With the writer cache enabled, bag files now split when the on-disk size plus the size of the cached messages reaches `--max-bag-size`, so files land closer to the requested limit instead of overshooting by up to the cache size. Files are never smaller than before the change.

Yes. This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### Additional Information

- ABI: this adds a virtual method to the public `MessageCacheInterface`, so it is intended for Rolling only and is not backportable as-is to Jazzy/Kilted, where #2311's workaround was backported.
- Residual overshoot: messages already swapped into the consumer buffer but not yet reflected in `get_bagfile_size()` are counted by neither side, so the overshoot is bounded by one in-flight consumer batch rather than the whole cache.
- #2382 (async split, draft) touches the same files; happy to rebase whichever lands second.


KR-Ravindra:
Round 1 self-review.

Verified against `rolling`:
1. `should_split_bagfile()` (`sequential_writer.cpp:637`) only looks at `storage_->get_bagfile_size()`, and `switch_to_next_storage()` stops the cache consumer first, flushing the remaining cache into the old file; the overshoot described in #2310 follows from that. The fix is the `get_current_size()` API named in the `TODO(morlov)` on that line.
2. `MessageCache::get_current_size()` reads `buffer_bytes_size_` under `producer_buffer_mutex_`; the writer thread does not hold that mutex when it calls `should_split_bagfile()`, and the consumer only takes it in `swap_buffers()` / `wait_for_data()`, so there is no lock-order issue. `CircularMessageCache` keeps the default 0 because snapshot mode skips the size check in `write()`.
3. The new writer test is deterministic: storage 60 + cached 40 against a 100 limit gives 0 splits without the fix and 1 with it. The 1-byte-message change in the existing split test is needed because the fixture's fake storage size counts messages while the cache counts bytes.
4. Restructured the description to the ros2 PR template and added the ABI note (new virtual on `MessageCacheInterface`, Rolling only), the residual-overshoot bound and the overlap with #2382.

No duplicate: #2310 links only #2311 and its backports; #2382 touches the same files but not the split decision. DCO passes; the Lint/Test workflows need a maintainer to approve the run. Marking ready.

fujitatomoya (approved):
lgtm with green CI.

IMO overshoot by up to one consumer batch should be just fine.

fujitatomoya:
Pulls: ros2/rosbag2#2496
Gist: https://gist.githubusercontent.com/fujitatomoya/53b737bdef4de003ad2089eb564b4229/raw/cdcc53b8820fb219d117beebc605d7feb38aed52/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20463
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30567/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/23253/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10360/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29420/)

PR #2501, opened 2026-09-18
[rosbag2_py] Re-run Stubgen

InvincibleRMC:
## Description

Re ran the stub generation following the readme for some updates with new mypy/pybind.

Yes it lowers some warns around the use of `Set` and if users wanted to use int-like objects np.int32 for example whereas before since they would not technically be a subclass of int. This has been true at runtime for a long but, the stubs are always behind.

No

### Additional Information

Might be nice to run this on all builds, but would require adding mypy as a build dependency.


christophebedard:
> Re ran the stub generation following the readme for some updates with new mypy/pybind.

the version bump/update following the switch to Ubuntu Resolute? Just to make sure I understand where this update is coming from

InvincibleRMC:
> > Re ran the stub generation following the readme for some updates with new mypy/pybind.
> 
> the version bump/update following the switch to Ubuntu Resolute? Just to make sure I understand where this update is coming from

Yeah. This was just from running on Ubuntu 26 vs on Ubuntu 24. The typing changes have to do with some upstream pybind changes I and like 2 other guys did.

christophebedard (approved):
LGTM with green (GitHub) CI.

GitHub CI needs to be bumped to 26.04: #2502

InvincibleRMC:
Am I good to merge this, or should this happen after the CI bump PR?

christophebedard:
might as well wait I think

christophebedard:
alright, #2502 was merged, can you rebase on `rolling` or merge `rolling` into this?

christophebedard:
@mergifyio update

christophebedard:
@mergifyio backport lyrical

PR #2502, opened 2026-09-19
Switch to Ubuntu Resolute base images in GitHub CI

christophebedard:
## Description

And use ghcr.io instead of Docker Hub.

It was still on Ubuntu 24.04 Noble, which still works fine now, but eventually it's bound to break. For instance: #2501.

### Additional Information

Backport to `lyrical`.


christophebedard:
The "Is regeneration of Python stubs required?" check should fail because of a new version of mypy/stubgen . It will be fixed by #2501

InvincibleRMC (approved):

christophebedard:
@mergifyio backport lyrical

